PR #349, opened 2026-05-01
feat: native multi-arch Docker images (Apple Silicon, ARM64 Linux)

Luca-Timo:
## Summary

Pre-built backend and frontend images now ship as **multi-arch manifests
(`linux/amd64` + `linux/arm64`) on every published tag**. Apple Silicon
Macs (M1/M2/M3/M4) and ARM64 Linux hosts (Raspberry Pi, AWS Graviton,
Ampere, etc.) pull the native variant automatically. No Rosetta, no
`platform:` flags, no compose edits.

### What changes for users

Today on GHCR, `ghcr.io/the-luap/picpeak/{backend,frontend}` publish
`amd64`-only manifests for `:stable`, `:latest`, and `:beta`. Apple
Silicon and ARM64 Linux users running any of those channels are silently
falling back to Rosetta or qemu-user emulation. After this PR, every
channel produces a true multi-arch manifest.

| Channel | Before (upstream today) | After |
|---|---|---|
| `:stable` / `:latest` | amd64-only | multi-arch (amd64 + arm64) |
| `:beta` | amd64-only | multi-arch (amd64 + arm64) |
| `:vX.Y.Z` / `:X.Y.Z` / `:X.Y` / `:X` | not consistently published | multi-arch (amd64 + arm64) |

`postgres:15-alpine` and `redis:7-alpine` are already multi-arch upstream,
so the whole stack is native on every supported architecture.

The `PICPEAK_CHANNEL=stable|beta` switch in `.env` works exactly as
before, just resolves to multi-arch manifests now.

### What changes for CI

Multi-arch builds switch from a single QEMU-emulated job to the standard
**per-arch build + manifest merge** pattern documented at
<https://docs.docker.com/build/ci/github-actions/multi-platform/>:

- `build-backend` and `build-frontend` are now matrix jobs:
  `linux/amd64` on `ubuntu-latest`, `linux/arm64` on `ubuntu-24.04-arm`.
  Each leg pushes by digest. **No QEMU.**
- `merge-backend` and `merge-frontend` run after their matrix completes,
  use `docker buildx imagetools create` to assemble the digests into a
  multi-arch manifest, and apply the human-readable tags computed by
  `docker/metadata-action` (preserves all the existing `stable`/`beta`/
  `latest`/semver/sha tagging rules).
- Trivy scanning stays in the merge job (scans the resulting manifest by
  tag, identical to today).
- Repository-name lowercasing (`${GITHUB_REPOSITORY,,}`) so the workflow
  works on forks regardless of owner-name casing — GHCR rejects mixed
  case.

The old "skip arm64 on branch pushes" trade-off existed because QEMU
emulation was too slow/unreliable for `npm install`. Native runners
remove the trade-off entirely, so `:stable`, `:beta`, and `:latest` no
longer ship as amd64-only.

### Transitional fallback

For users still on a pre-multi-arch image tag (e.g. an old `:latest`
cached locally), the new `docker-compose.amd64.override.yml` forces
`platform: linux/amd64` on the backend and frontend services as a
documented fallback. It's opt-in, Linux ARM64 users would never load it
because forcing emulation on a native host is the wrong choice.

## Files changed

- `.github/workflows/docker-build.yml` — refactor to per-arch matrix +
  manifest merge; lowercase image names for fork compatibility
- `docker-compose.amd64.override.yml` — new; transitional override for
  ARM64 hosts on amd64-only image tags
- `README.md` — Apple Silicon / ARM64 support note + override pointer
- `DEPLOYMENT_GUIDE.md` — multi-arch architectures section + override
  usage block

No changes to `docker-compose.production.yml` or `docker-compose.yml`.
No changes to image build (Dockerfile-level), the existing Dockerfiles
already build cleanly on both architectures.

## Test plan

Validated end-to-end on a fork (`Luca-Timo/picpeak`):

- [x] Workflow runs to completion on a feature branch via `workflow_dispatch`
- [x] Both per-arch jobs succeed; merge job assembles the multi-arch manifest
- [x] `docker buildx imagetools inspect` confirms `linux/amd64` +
      `linux/arm64` on the resulting tag
- [x] Mac Studio (Apple Silicon, Docker Desktop) `docker compose pull`
      retrieves native `arm64/linux` images for backend, frontend,
      postgres, redis (4/4)
- [x] Stack starts, postgres/redis/frontend report `healthy`
- [x] All channel tags (`stable`, `beta`, `latest`, semver, sha) still
      apply correctly per the existing rules

Pre-existing migration ordering issue in `core/001_init.js` is unrelated
to this PR (would reproduce identically on a Linux amd64 host running
the same image) and will be filed separately.

## Backwards compatibility

- Existing users on `:stable` / `:latest` / `:beta` get a multi-arch
  image on next `docker compose pull`. Apple Silicon and ARM64 Linux
  users transition from Rosetta/qemu-user to native arm64 transparently.
  No `.env` changes, no compose edits required.
- Existing AMD64 deployments see no change beyond a fresh image rebuild
  — the amd64 manifest in the new multi-arch index is built from the
  same Dockerfile on the same `ubuntu-latest` runner as before.
- The new override file is opt-in; nothing references it unless the
  user passes `-f docker-compose.amd64.override.yml`.


the-luap:
Thanks @Luca-Timo, this is well-scoped and the per-arch matrix + manifest merge is the right pattern. Happy to take this on `beta`. Two small things before merging:

**1. Retarget to `beta`**
Please change the base branch to `beta` (the merge button has a "Compare & pull request" option, or we can do it from our side).

**2. Drop the `DEPLOYMENT_GUIDE.md` hunk**
That file was intentionally removed in 0faf9b3 ("docs: move documentation to docs.picpeak.app, drop in-repo copies"). On beta the file no longer exists, so the hunk won't apply and we can't keep it in this PR. I've already added the equivalent "Supported architectures" section to the docs site as a follow-up — it'll publish at docs.picpeak.app/deployment/docker once it lands.

**3. README hunk needs a small re-anchor against `beta`**
Beta's README has been restructured around docs.picpeak.app links — the `> Permissions:` quote block your patch sits below doesn't exist anymore. There's a `Note on Docker file permissions (PUID/PGID)` section now (around line 92). Easiest fix is to drop the ARM64 callout as a new bullet/note immediately after that section, or as part of the Quick Start lines higher up. Anywhere it'll be seen by someone evaluating the install is fine.

**On amd64 safety:** confirmed your trade-offs. Existing amd64 deployments pull the multi-arch manifest, Docker picks the amd64 variant transparently — no behaviour change. The override file is opt-in (only loaded with `-f docker-compose.amd64.override.yml`) and won't affect anything unless the user explicitly references it. ✅

Once the three items above are addressed I'll get this in. Thanks again!

Luca-Timo:
Thanks for the thorough review! All three items are addressed:

- Retargeted to beta — rebased the 5 commits onto origin/beta. Confirmed only my commits are in the diff, no main-only commits leaking through.
- Dropped the DEPLOYMENT_GUIDE.md hunk
- Re-anchored README hunk — placed the ARM64 callout immediately after the Note on Docker file permissions (PUID/PGID) section, as suggested. Also reworded it to be ARM64-generic (aarch64) rather than Apple Silicon–specific, since the multi-arch images benefit any ARM64 host.

Final PR diff: docker-build.yml, README.md, docker-compose.amd64.override.yml, nothing else.

Ready for another look!
And thank you very much for the great Project

PR #372, opened 2026-05-03
feat(cms): add external URL toggle for imprint and privacy pages

Luca-Timo:
## Description

Adds a per-CMS-page toggle "Use external URL" plus a URL input for the imprint and privacy-policy pages. When enabled, the gallery footer link points at the external URL and direct navigation to `/impressum` or `/datenschutz` redirects there. When disabled, behavior is unchanged.

Internal title/content/logo stay saved in the database when the toggle is on, so flipping it back restores the previous CMS page intact.

**Why:** operators (myself included) often maintain authoritative legal pages on a separate domain (e.g. a personal/agency site that already has a lawyer-reviewed Impressum and Datenschutzerklärung). Forcing PicPeak to host its own copy means duplicating content and risking drift between the two. The toggle lets each gallery deployment point at the canonical source per legal page, while keeping the internal CMS page as a fallback.

**Link behaviour:**
- **External URL configured →** rendered as `<a target="_blank" rel="noopener noreferrer">`, opens in a **new tab** so the gallery stays open behind it. Direct navigation to `/impressum` or `/datenschutz` performs `window.location.replace()` to the external URL (back button returns to the gallery, not the redirect).
- **External URL not configured →** rendered as the existing internal `<Link>`, opens in the **same tab** (current behavior, unchanged).

**Security:**
- Backend strictly validates `external_url` parses as a URL with protocol `https:` when the toggle is on. Rejects `http:`, `javascript:`, `data:`, missing schemes, and malformed input. The explicit protocol check is the security-relevant gate  `express-validator`'s `isURL()` is too permissive.
- The endpoint already requires the `cms.edit` permission, so no new auth surface is introduced.
- External link uses `rel="noopener noreferrer"` to prevent tab-nabbing and referrer leaks.

Fixes # (n/a — feature request, no issue filed)

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

- [x] Manual testing completed
- [x] Tested on Docker deployment

**Manual test steps:**
1. Online migration adds `use_external_url` (boolean, default false) and `external_url` (varchar) to `cms_pages` on backend startup, verified via `\d cms_pages` on a fresh PostgreSQL volume.
2. `PUT /api/admin/cms/pages/impressum` with `use_external_url: true` and a non-https URL → 400. With `https://...` → 200, persisted.
[...]

PR #378, opened 2026-05-04
feat(events): tree view for external media folder picker

Luca-Timo:
## Summary

- Replaces the per-level grid in `ExternalFolderPicker` (admin Event Details: form field + import modal) with a recursive folder tree using lazy-loaded children, expand/collapse chevrons, and a highlighted selected node.
- Tree auto-expands the ancestors of the currently-saved `external_path` on open, so a previously selected folder is visible immediately. A new "Clear" button replaces the old "Up"/"Select" buttons.
- No backend changes — children are fetched per-folder via the existing `GET /api/admin/external-media/list?path=...` endpoint.

## Why

Browsing deeper external-media mounts was tedious: users had to click into each folder and back up to compare siblings, with no overview of the hierarchy. The tree shows context at a glance.

## Design choices

- **Lazy loading on expand.** Each folder fetches its children only when the user opens it, and React Query caches the result (`['external-folder-children', path]`, 30 s stale time) so collapse/re-expand is instant. Keeps the picker snappy on deep mounts and avoids any backend changes.
- **Folders only — no images shown in the tree.** Tested on the dev instance with real data: showing image files alongside folders was cluttered and unintuitive (the tree's job is folder selection; files were noise), and listing them eagerly hurt performance on large directories. The picker's only purpose is selecting a folder, so files stay hidden.

## Behavior

- Empty expanded folder → "No subfolders" hint.
- Selecting the synthetic root sets `external_path = ''` (i.e. `/external-media` itself).
- Both call sites (Reference-folder field and Import-from-External modal) consume the same `{ value, onChange }` props, no API change.

## i18n

Two new keys under `events.*`: `externalFolderEmpty`, `clearSelection`.

- en + de: hand-written.

## Test plan

- [x] Open an event with a saved `external_path` → tree auto-expands to that folder, folder is highlighted.
- [x] Expand/collapse a folder → exactly one network request on first expand, none on subsequent expand/collapse (lazy load + cache).
- [x] Click a folder name → "Selected: /external-media/…" updates; save the event and reopen to confirm persistence.
- [x] Open "Import from External Folder" modal → same tree, selecting a folder enables import.
- [x] Empty folder → shows "No subfolders".
- [ ] Backend error on a path → inline error text, no crash.
- [x] Dark mode contrast on chevron, folder icon, hover, and selected states.

### DIFF
--- a/frontend/src/i18n/locales/de.json
+++ b/frontend/src/i18n/locales/de.json
@@ -865,6 +865,8 @@
     "externalFolder": "Externer Ordner",
     "externalFolderHint": "Diese Ordner stammen aus dem /external-media Mount innerhalb des Containers oder Hosts.",
     "externalFolderRequired": "Bitte wählen Sie vor dem Speichern einen externen Ordner aus.",
+    "externalFolderEmpty": "Keine Unterordner",
+    "clearSelection": "Löschen",
     "welcomeMessage": "Willkommensnachricht",
     "noWelcomeMessage": "Keine Willkommensnachricht festgelegt",
     "noWelcomeMessageSet": "Keine Willkommensnachricht festgelegt",
--- a/frontend/src/i18n/locales/en.json
+++ b/frontend/src/i18n/locales/en.json
@@ -416,6 +416,8 @@
     "externalFolder": "External Folder",
     "externalFolderHint": "These folders are read from the /external-media mount inside your container or host.",
     "externalFolderRequired": "Please select an external folder before saving.",
+    "externalFolderEmpty": "No subfolders",
+    "clearSelection": "Clear",
     "welcomeMessage": "Welcome Message",
     "noWelcomeMessage": "No welcome message set",
     "created": "Created",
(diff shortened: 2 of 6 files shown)


the-luap:
Really nice work — the lazy-loaded tree with auto-expanded ancestors is much better than the per-level grid, and great call on the scrollbar `color-scheme` fix. Especially appreciated that you self-flagged the nl/pt/ru as machine-suggested with English fallbacks; we'll get a native-speaker pass on those after merge along with the existing CMS translations from #372.

Approving as-is. Two minor follow-ups we'll handle post-merge so this can land:

- Type the `externalMediaService.list()` entries properly so the `(e: any)` / `(d: any)` casts can go away (pre-existing — not a regression in this PR).
- Native-speaker review of the nl/pt/ru strings (`externalFolderEmpty`, `clearSelection`).

Thanks for the thoughtful design notes too — "folders only, files would be noise" is exactly the right call for a folder picker. Merging.

PR #379, opened 2026-05-04
feat(events): prefill admin email + admin picker on event creation

Luca-Timo:
## Summary

When creating a new event, the **Admin Email** field now:

- **Auto-prefills** with the currently logged-in admin's email on first load (only if the field is empty, never clobbers typed input or a returned-to draft).
- **Shows a small "Pick from admins" dropdown** under the input when there are 2+ active admins, so you can switch to another admin's email in one click. The field stays a normal email input, typing a non-admin address still works and the dropdown reverts to "Custom email" to reflect that.

## Why

I shouldn't have to type my own email every time I create an event, it's the same value 99% of the time. And when there are multiple admins on the team, picking who should be the contact is one click instead of remembering and typing the address.

## Behavior

- **1 admin (you alone):** field auto-fills with your email. No dropdown.
- **2+ admins:** field auto-fills with your email; dropdown appears below the input listing `username (email)` for each active admin.
- **Low-permission admin** (no `users.view` permission): the admin-list API call is silently caught and returns `[]` — auto-fill from the auth context still works, the dropdown just doesn't render. No errors surface.

## Implementation notes

- Current admin pulled from `useAdminAuth()`.
- Admin list fetched via `userManagementService.getUsers()` with React Query (`['admin-users-list']`, 5 min stale, `retry: false`). Errors caught in the `queryFn` so missing `users.view` permission doesn't bubble up.
- Auto-prefill effect uses a `useRef` guard so it runs at most once, and only when `admin_email` is empty.

## i18n

Two new keys under `events.*`: `adminEmailPickFromAdmins`, `adminEmailCustom`.

- en + de: hand-written.

## Test plan

Verified on local dev instance:

- [x] Single-admin account: opening the create form shows my email pre-filled, no dropdown.
- [x] Multi-admin account: opening the create form shows my email pre-filled with the picker dropdown below; selecting another admin updates the field.
- [x] Typing a non-admin address keeps the input value and the dropdown shows "Custom email".
- [x] Auto-prefill does not overwrite an already-populated field.

### DIFF
--- a/frontend/src/i18n/locales/de.json
+++ b/frontend/src/i18n/locales/de.json
@@ -1017,6 +1017,8 @@
     "welcomeMessagePlaceholder": "Willkommen zu unserem besonderen Tag! Laden Sie diese Erinnerungen gerne herunter und teilen Sie sie...",
     "hostEmailPlaceholder": "kunde@beispiel.de",
     "adminEmailPlaceholder": "admin@beispiel.de",
+    "adminEmailPickFromAdmins": "Aus Admins wählen:",
+    "adminEmailCustom": "Eigene E-Mail",
     "securityAndAccess": "Sicherheit & Zugriff",
     "accessAndSecurity": "Zugriff & Sicherheit",
     "enterPassword": "Passwort eingeben",
--- a/frontend/src/i18n/locales/en.json
+++ b/frontend/src/i18n/locales/en.json
@@ -356,6 +356,8 @@
     "welcomeMessagePlaceholder": "Welcome to our special day! Feel free to download and share these memories...",
     "hostEmailPlaceholder": "customer@example.com",
     "adminEmailPlaceholder": "admin@example.com",
+    "adminEmailPickFromAdmins": "Pick from admins:",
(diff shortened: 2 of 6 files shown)


the-luap (approved):
Approving — same craftsmanship level as #378. Particularly nice:

- All 5 locales in the same PR with `t(key, 'fallback')` defaults
- Defensive `getUsers()` query so low-permission admins don't see errors
- `useRef` one-shot prefill matches the existing `requirePasswordDefaultApplied` / `brandingThemeApplied` pattern in this file
- Race-safe prefill check (`prev.admin_email ? prev : ...`) so typing-before-auth-loads can't get clobbered
- TypeScript clean, no `any` casts

Two minor observations (not blockers):
- The "Custom email" option in the select is a no-op once you've picked an admin (can't unset via the picker — would need to manually edit the input). Defensible UX since the input is the source of truth.
- `getUsers()` returns all users on first load; 5 min cache makes this fine. If ever expensive, an `activeOnly=true` server-side filter would help — but premature for now.

Translations look reasonable for the contexts. Thanks!

PR #385, opened 2026-05-04
feat(gallery): decouple header style from layout, add banner option

Luca-Timo:
## Summary

Today, `headerStyle` and `galleryLayout` are typed as independent fields, but `GalleryLayout.tsx` couples them: `standard` rendered as a compact inline header on grid, but as a sparse menu-only bar plus a colored hero banner on every non-grid layout. Switching photo layout silently flipped the entire header chrome, contradicting the data model and surprising users.

This PR makes them actually independent.

- **`standard` is now one consistent compact inline header** on every layout (logo, event name, dates, Download All, Logout).
- **New `banner` header style** = the `standard` header plus a colored banner block above. Five header options total: `hero`, `standard`, `banner`, `minimal`, `none`.
- **`controlsStyle` is also decoupled.** Previously, when unset it fell back to `(galleryLayout !== 'grid' || isHeroHeader)`, which silently moved the filters into the sidebar drawer for non-grid layouts — same root cause as the header coupling. Now `showSidebar` is true only when `theme.controlsStyle === 'sidebar'` is explicitly chosen.
- **Theme customizer** picker for header style now uses the same 3-column grid as the layout picker (was 5-col, looked inconsistent).

## Migration

`086_migrate_non_grid_standard_to_banner.js`: any event with `header_style='standard'` AND a non-grid layout (masonry/carousel/timeline/mosaic) is flipped to `header_style='banner'`. Their visual look is preserved (the colored banner stays). Reversible via the migration's `down`.

Events that explicitly set `controlsStyle === 'sidebar'` are unchanged. Events that relied on the implicit non-grid → sidebar default will now get the inline filter bar; they can re-enable the sidebar by selecting it in the theme customizer.

## Files

- `frontend/src/types/theme.types.ts` — add `'banner'` to `HeaderStyleType`; update `birthdayFun` and `corporateTimeline` presets.
- `frontend/src/components/gallery/GalleryLayout.tsx` — collapse three coupled blocks into one consistent `standard`/`banner` header for every layout; gate the colored banner on `isBannerHeader`; drop unused `isNonGridLayout`.
- `frontend/src/components/gallery/GalleryView.tsx` — `showSidebar` now keys on the explicit `controlsStyle === 'sidebar'` only.
- `frontend/src/components/admin/ThemeCustomizerEnhanced.tsx` — `banner` icon (`LayoutTemplate`); 3-col grid for the picker.
- `frontend/src/i18n/locales/{en,de,nl,pt,ru}.json` — `branding.headerStyleOptions.banner`, `branding.headerStyleDescriptions.banner`; tightened the now-misleading `standard` description.
- `backend/src/routes/adminEvents.js` — extend `header_style` validators to accept `'banner'`.
- `backend/migrations/core/086_migrate_non_grid_standard_to_banner.js` — preserve current visual look for existing data.

## i18n

- en + de: hand-written.

## Test plan

Verified live on a beta deployment with the dev instance:

- [x] `standard` renders the same compact inline header on `grid`, `masonry`, `carousel`, `timeline`, and `mosaic`.
- [x] `banner` renders the standard header plus the colored banner on every layout.
- [x] `hero`, `minimal`, `none` unchanged.
- [x] Migration round-trip: existing non-grid + standard events flip to `banner` (visually identical), `down` reverts cleanly.
- [x] Theme customizer shows 5 options in a 3-col grid; selecting `banner` persists.
- [x] `controlsStyle = 'sidebar'` still renders the drawer as before; unset/`'classic'` renders the inline filter bar regardless of layout.


the-luap:
Really nice architectural fix — collapsing those three coupled conditional blocks into one consistent header per style, plus the migration that preserves the visual appearance of existing events, is exactly the right shape. The TypeScript exhaustiveness on `headerStyleIcons` is a nice touch too.

One thing I'd like before we merge:

**Extend the migration to also preserve `controlsStyle` for affected events.** The current migration flips `header_style: standard → banner` for non-grid events (great — header looks identical). But the same events were previously getting the sidebar filter via the implicit `(galleryLayout !== 'grid' || isHeroHeader)` default in `GalleryView.tsx`. With that default removed, those events will now render **inline filters** instead of the sidebar drawer they had before. Header looks the same, filter bar moves.

Suggestion: in the same migration, for any event where you flipped to `banner` (i.e. non-grid + previously-standard) AND `theme.controlsStyle` is unset, set `theme.controlsStyle = 'sidebar'`. That way existing galleries are pixel-identical post-upgrade, and the cleaner decoupled defaults only apply to newly-created events. Same migration, same scope, same `down`-reversibility.

You explicitly noted this trade-off in the PR description (users can re-enable sidebar manually), so this is a judgment call — but my read is "preserve visuals on upgrade" beats "force users to opt back in" for an existing-deployment fix. Let me know if you'd rather keep the current scope and document the change in release notes instead; either is defensible.

Two non-blocking i18n nits we can polish either way:
- nl `"Compacte inline-header met evenementdetails"` — "inline-header" is tech jargon. More natural Dutch: `"Compacte koptekst met evenementdetails"`.
- ru `"Стандартный"` alone reads slightly incomplete; could be `"Стандартный заголовок"`. Probably fine in the picker-grid context though.

Otherwise this is great work. Ready to approve once we've decided on the migration scope.

Luca-Timo:
You are absolutely right, didn't notice the flip.

the-luap (approved):
All three asks addressed cleanly:

- **Migration `controlsStyle` preservation** — the `if (!theme.controlsStyle)` guard correctly preserves explicit user intent while pinning the implicit-default cases to `'sidebar'`. Symmetric down-migration with the honest "can't perfectly distinguish" caveat is exactly the right thing to write.
- **Dutch:** "Compacte koptekst" reads naturally.
- **Russian:** "Стандартный заголовок" — completes the label.

Nice defensive coding throughout (the `controlsPinnedCount` log counter is a nice touch for spotting how many events the upgrade actually affected). Approving.

PR #390, opened 2026-05-04
feat(branding): self-hosted webfonts with filesystem scanner

Luca-Timo:
## Summary

Replaces the Google Fonts CDN with a filesystem-driven font scanner. PicPeak now serves all webfonts from its own origin, no requests to `fonts.googleapis.com`, no visitor IPs leaked to a third party (relevant for GDPR / LG München 2022).

The font picker becomes data-driven: whatever the backend finds on disk, the picker offers. Drop a folder into the storage volume, restart the backend, and the new family appears in the picker — no code changes required.

## How it works

- **Bundled defaults** at `backend/assets/fonts/` (8 OFL families, baked into the image): Comic Neue, IBM Plex Sans, Inter, Jost, Montserrat, Noto Sans, Playfair Display, Poppins.
- **Runtime additions** at `STORAGE_PATH/fonts/` (optional, gitignored). Drop a folder there to add custom families. Same folder convention. User additions override bundled families of the same name.
- **Scanner** at `backend/src/services/fontsService.js` reads both locations, merges with user-wins, 30s TTL cache.
- **Listing API**: `GET /api/public/fonts` → `{ fonts: [{ family, weights }] }`.
- **Static serving**: `/fonts/<Family-Name>/<weight>.woff2` with `Cache-Control: max-age=7d, immutable`. Path-traversal protected. Frontend nginx proxies `/fonts` to the backend.
- **Lazy injection**: `ThemeContext.applyTheme` injects exactly one `@font-face` block per family the page actually uses. Other families never load for that visitor.
- **First-paint bootstrap**: `index.css` ships static `@font-face` blocks for Inter so the very first frame already has the default body font.

## Folder convention

backend/assets/fonts/
├── <Family-Name>/
  │ ├── 400.woff2
  │ ├── 600.woff2
  │ └── 700.woff2
└── LICENSE-OFL.txt

Folder name = family name with hyphens (`Playfair-Display/` → `Playfair Display`). File name = numeric weight + `.woff2`. Other names are silently ignored. Documented in `docs/fonts.md`.

## Out of scope (deferred)

- **Variable fonts** — single-file multi-weight format. Plan extension is straightforward but defers cleanly.
- **Italics** — the picker doesn't expose them.
- **Admin upload UI** — v1 workflow is "drop a folder via Docker volume / SFTP".
- **Per-option dropdown previews** — attempted in `b4f9b65`, reverted in `f410207`. Browser support for styling `<option>` elements is inconsistent: Safari ignores `font-family` on options entirely, Chrome/Firefox were unreliable in testing. Documented as a future improvement in `docs/fonts.md`. The follow-up would replace the native `<select>` with a custom dropdown component or render a separate "preview text" box below the picker.

## Migration

None. Existing event themes keep their saved `fontFamily` strings. The customizer's new dropdown matches both legacy unquoted (`Inter, sans-serif`) and quoted (`'Inter', sans-serif`) formats so events saved before this change still highlight the right option.

## Commits

- `bac51fe` feat: scanner + listing API + lazy injection + bootstrap CSS + 8 bundled fonts + docs
- `b4f9b65` feat: per-option dropdown preview (later reverted — see above)
- `e6c03e4` fix: nginx `/fonts` location proxying to backend (was missing in the initial commit, caused 404s when accessed via the production URL)
- `f410207` revert: per-option preview, deferred to a follow-up

## Smoke checks (verified live on dev instance)

- [x] Backend boots; logs show 8 families discovered.
- [x] `curl /api/public/fonts` returns the 8 families with their weights.
- [x] `curl /fonts/Inter/400.woff2` returns 200 with `Cache-Control: max-age=604800, immutable`.
- [x] `curl /fonts/../../../etc/passwd` returns 403 (path traversal blocked).
- [x] Theme customizer body + heading dropdowns show the 8 families plus "System UI". Selecting Jost saves and applies the font on the live gallery.
- [x] Network panel shows zero requests to `fonts.googleapis.com` or `fonts.gstatic.com` on a fresh visit.
- [x] OFL license served at `/fonts/LICENSE-OFL.txt`.


the-luap (approved):
Approving — substantial PR done with care. Real GDPR motivation, well-grounded in existing patterns, scope discipline is exemplary (the per-option preview attempt → revert with a documented browser-support reason is exactly the right way to handle that kind of thing).

Particularly nice:

- **Path traversal protection** via the existing `secureStatic` middleware, with the explicit `curl /fonts/../../../etc/passwd → 403` smoke check.
- **Two-tier user-wins logic** stays consistent across both the listing API (Map merge in `fontsService.js`) and the static file serving (storage path mounted before bundled in `server.js`) — overrides work end-to-end.
- **Backward compat** in `extractFamilyName` handles both legacy unquoted and new quoted formats so themes saved before this PR still highlight the correct dropdown option.
- **First-paint bootstrap** in `index.css` so the very first frame already has Inter — avoids FOUT on initial load.
- **OFL license** shipped at `backend/assets/fonts/LICENSE-OFL.txt` and served at `/fonts/LICENSE-OFL.txt` (required for SIL OFL compliance).
- Honest deferrals in the PR description — variable fonts, italics, admin upload UI, per-option previews — each with reasoning.

A few non-blocking follow-ups worth tracking after merge:

1. **Hardcoded `SERIF_FAMILIES` / `CURSIVE_FAMILIES` sets in `ThemeCustomizerEnhanced.tsx`** undermine the "drop a folder, no code changes" promise the moment someone adds a font that isn't Playfair Display or Comic Neue. A new `Roboto-Slab/` would get `'Roboto Slab', sans-serif` (wrong fallback) → during the ~200ms `.woff2` download a serif theme briefly renders in Helvetica, and if the font 404s the visual style is wrong permanently. Cosmetic, not functional, but it's the loose end in an otherwise data-driven design. A small v2: support an optional per-folder `meta.json` like `{ "generic": "serif" }` (or a marker file like `serif.txt`); scanner reads it, includes it in the API, frontend uses it instead of the hardcoded set.

2. **`immutable: true` + 7d cache + static `<weight>.woff2` filenames** means an admin replacing a font in storage won't see the change in browsers that already cached it for up to 7 days. Not catastrophic — fonts rarely change — but worth a note in `docs/fonts.md` explaining "to roll out a font replacement, also bump the file content via a different filename (e.g. add a v2 weight) or accept the 7-day rollover".

3. **5-min frontend cache vs 30s backend cache** mismatch — admin UI may show stale fonts list for up to 5min after a folder drop. Hard refresh fixes it. One-liner in `docs/fonts.md` would close the loop.

4. **No automated tests** for the scanner edge cases (weight parsing, dedup, malformed folders). Consistent with how features land in this repo, but the scanner has clear unit-testable logic that would be cheap to cover.

None of these block merging. Thanks for the thoughtful work — and the candid revert commit is appreciated.

the-luap (approved):
All four follow-ups addressed cleanly:

- **Generic fallback via `meta.json`** — explicit schema, allowlist validation, warning-on-invalid, bundled `serif` for Playfair and `cursive` for Comic Neue. Frontend now `font.generic ?? 'sans-serif'`. The "drop a folder, no code" promise holds end-to-end.
- **Cache rollout** — dropping `immutable` + relying on `If-Modified-Since` + the documented two-strategy rollout (wait-or-rename) is the right call.
- **Cache-mismatch docs** — clear note with the rationale for keeping the two caches independent.
- **24 unit tests** with the `PICPEAK_BUNDLED_FONTS_ROOT` env-var test seam — nice.

One small thing on the test file: 5 of the 24 assertions silently no-op because `beforeEach` calls `jest.resetModules()` followed by a fresh `require('fontsService')`, which means the `jest.mock` factory at the top no longer applies and `fontsService` picks up the real logger while the test's `logger` variable still points at the mocked one. The behaviour assertions all pass (so the scanner logic is verified correct) but the "warning logged" assertions don't actually run. We'll send a small follow-up that drops the unnecessary `resetModules()` — `clearFontsCache()` already covers the only stateful thing in the module.

Approving — substance is all there. Thanks for the thorough work, especially the docs additions.

PR #400, opened 2026-05-06
feat(branding): 8-token CI palette + force color mode + dark-mode consistency

Luca-Timo:
Addresses items 1–3 from the-luap/picpeak#397 (excluding header, separate effort) and follow-ups discovered during review.

## Summary

- Expands theme customisation from 4 colours to an **8-token CI palette** (Background, Surface, Elevated, Border, Text, Secondary text, Accent, Accent (filled)) drivable across the entire admin and public site.
- Adds an instance-wide **Force color mode** lock (Dark / Light / off) that hides the user toggle and overrides per-event themes.
- Sweeps **dark-mode consistency** across the admin (CMS editor, modals, tables, cards, badges, tooltips, sidebar, tab underlines) and the public CMS rendering.
- Adds **Sync from Branding** to the email and gallery theme customizers so admins can copy the site palette into emails and galleries with one click.
- Replaces ~80 hardcoded `primary-*` Tailwind classes with theme-token aliases so admin chrome follows the user's CI colours instead of the legacy green.

## Migration safety

- All 11 existing presets pinned the previously-implicit values onto the new fields, visual rendering of existing themes is byte-identical.
- `migrateThemeConfig` derives missing tokens from legacy 4-colour configs (`accentDarkColor` ← `primaryColor`, surfaces ← mode-aware defaults) so saved event themes upgrade silently.
- All 6 new email tokens default to the previously-hardcoded literals, emails sent by upgraded installs render byte-identical until an admin changes a colour.
- `--color-primary` kept as a legacy alias and synced from `accentDarkColor` in the customizer, so any consumer still reading it (CSS variable or older event configs) doesn't drift.
- Force color mode and Sync buttons are explicit opt-in, never auto-run.

### Foundation — 8-token CI palette
- `ThemeConfig` extended with `accentDarkColor`, `elevatedColor` (`frontend/src/types/theme.types.ts`)
- `applyTheme` writes all 8 tokens as CSS variables (`frontend/src/contexts/ThemeContext.tsx`)
- `:root.dark { }` block in `index.css` so admin dark mode flips variables without per-component overrides
- Tailwind aliases (`bg-surface`, `bg-elevated`, `bg-accent-dark`, `text-accent`, `border-token`, …) read CSS variables (`frontend/tailwind.config.js`)
[...]

PR #401, opened 2026-05-06
feat(gallery): icon-only menu, accent Download CTA (#386)

Luca-Timo:
Addresses [the-luap/picpeak#386](https://github.com/the-luap/picpeak/issues/386) — gallery header layout cleanup.

## Summary

- Menu icon is now icon-only (the redundant "Menu" text label is gone).
- Menu icon is absolute-positioned at the very left of the header.
- New accent-coloured **Download** button placed immediately left of Logout — always visible when downloads are allowed, replacing the previous primary-coloured "Download All" header button.
- Same Download CTA is rendered in standard, banner, hero, and minimal headers. Intentionally **not** rendered in the no-header variant (chromeless by design).

## Why

The previous header pushed the logo to the right of the menu icon (so it never lined up with the gallery image grid below) and tucked the download action behind either the sidebar or a secondary "Download All" button. Single-image downloads happen in the lightbox per-image; the natural batch action sits next to Logout — one click before leaving.

### Menu button
- `frontend/src/components/gallery/GalleryView.tsx` — drops `<span>{t('common.menu')}</span>`; the button now renders just `<Menu />` with `p-2` for tight icon padding.
- `frontend/src/components/gallery/GalleryLayout.tsx` — wraps the menu button in `position: absolute; left: 3/6/8 (responsive); top: 50%; -translate-y-1/2` inside a `relative` container. Logo gets `pl-12 sm:pl-14` *only when a menu button is rendered*, so the logo's left edge sits at the same x-coordinate as the leftmost image (both anchored at `.container` left padding). When no menu button (e.g. classic controls style), the logo is flush with the container — no extra padding.

### Download CTA (new)
- New props on `GalleryLayout`: `showHeaderDownload?: boolean` and `onHeaderDownload?: () => void`. When set, render a button with:
  - `style={{ backgroundColor: 'var(--color-accent)', color: '#ffffff' }}` — accent colour pulled from CSS variable, so the button automatically tracks whatever palette the admin has chosen (legacy `--color-accent: #22c55e` today, the user's CI accent once [#400](https://github.com/the-luap/picpeak/pull/400) lands).
  - Icon-only on mobile (`<span className="hidden sm:inline">Download</span>`).
  - `aria-label="Download"`, focus ring, disabled+loading states.
- Rendered in the standard, hero, and minimal header variants. **Not** in the no-header variant.
- Old `showDownloadAll` is kept on `GalleryLayout` for back-compat, but `GalleryView` now passes `showDownloadAll={false}` so only the new accent button renders in the header. The sidebar's own Download All is untouched.

## Migration safety

- All visible labels and visible behaviour change in the gallery header are the explicit design intent of #386. No backend schema changes.
- `var(--color-accent)` already exists on `upstream/beta` (default `#22c55e`), so the new Download button renders without depending on PR #400.
- No new settings, no migrations, no API surface changes.

## Files changed

```
frontend/src/components/gallery/GalleryView.tsx        (icon-only menu, drop old DownloadAll, wire new button)
frontend/src/components/gallery/GalleryLayout.tsx      (menu absolute-left, new Download button in std/hero/minimal headers, props)
```

## Test plan

- [x] Standard header — menu icon at far left, logo flush with leftmost image, Download button left of Logout in accent colour
- [x] Hero header — menu + headerExtra on left, accent Download + Logout on right
- [x] Minimal header — menu + title on left, accent Download + Logout on right
- [x] No-header — no Download button (intentional)
- [x] `controlsStyle: 'classic'` — no menu icon, logo flush with container left
- [x] Mobile (`<sm`) — Download button shrinks to icon-only
- [x] Disabled / loading state on Download button (during ZIP build) — opacity + no double-click
- [x] `npm run build` (frontend) passes

### DIFF
--- a/frontend/src/components/gallery/GalleryLayout.tsx
+++ b/frontend/src/components/gallery/GalleryLayout.tsx
@@ -40,6 +40,16 @@ interface GalleryLayoutProps {
   showDownloadAll?: boolean;
   onDownloadAll?: () => void;
   isDownloading?: boolean;
+  /**
+   * "Download" CTA shown immediately to the left of the Logout button in the
+   * standard / banner header. Same handler as Download All; the label and
+   * placement are intentionally simpler — single primary action right before
+   * Logout, the natural step at the end of a gallery visit (#386). Always
+   * visible when allowed (independent of sidebar state) so guests aren't
+   * forced to discover the download in the menu.
+   */
+  showHeaderDownload?: boolean;
+  onHeaderDownload?: () => void;
   headerExtra?: React.ReactNode;
   menuButton?: React.ReactNode;
   headerStyle?: HeaderStyleType;
@@ -54,6 +64,8 @@ export const GalleryLayout: React.FC<GalleryLayoutProps> = ({
   showDownloadAll = false,
   onDownloadAll,
   isDownloading = false,
+  showHeaderDownload = false,
+  onHeaderDownload,
   headerExtra,
   menuButton,
   headerStyle: headerStyleProp,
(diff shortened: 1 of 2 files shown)


the-luap (approved):
Approving — focused, well-scoped implementation of every item in #386's design proposal. i18n already covered (`gallery.download` in all 5 locales), `var(--color-accent)` keeps it palette-agnostic regardless of PR-merge order, accessibility done right (aria-label, focus-ring, disabled state), TypeScript clean. Verified locally.

A few small follow-ups worth tracking after merge — none blocking:

1. **Download button JSX is duplicated three times** (standard/banner, minimal, hero — ~15 lines each). Pragmatic for a focused PR, but a small extracted component (or const within the file) would deduplicate ~30 lines and make future tweaks one-place changes. Easy follow-up.

2. **Hardcoded `color: '#ffffff'` on the Download button.** If an admin picks a very pale accent colour (especially with PR #400 giving them full palette control), white-on-pale could fail WCAG contrast. Realistic risk is low — most accent colours are saturated mid-tones — but not enforced. Worth a follow-up that either derives the text colour from accent luminance, or exposes an `accentTextColor` token alongside the existing accent palette.

3. **Discussion #386 isn't formally linked in the body** — only referenced in the title. Discussions don't auto-close, but adding a one-line "Implements the design proposed in #386" in the description would help future archaeology when someone wonders why the menu went icon-only.

Thanks — clean execution, every change has an inline comment with the why and a discussion reference.

Issue #350, opened 2026-05-01
[BUG] login page loop


Rekoo-PS:
https://github.com/user-attachments/assets/e757dd86-edae-442b-9b1a-01c01348c8d5

**To Reproduce*
Idk if this is correct but this is how i notice its happening..

 Open any removed event link
Make sure your session is timed out (open link after one day)
Go back to home from that screen

The page will loop like example .

"It even shows the original logo not my logo in this loop "

the-luap:
Thanks for the report. This is already fixed on beta (commit 038e84c, in 3.30.0-beta.0+) — multiple admin queries were each firing window.location.href = '/admin/login' on 401 in parallel and aborting each other, which is the flicker storm in your video.

Stable (2.6.5) is missing the fix; it'll roll in with the next stable cut from beta. If you can switch the channel to beta in the meantime (PICPEAK_CHANNEL=beta in .env), it should be gone.

Closing — let me know if it still reproduces on beta.

Rekoo-PS:
<img width="1080" height="2340" alt="Image" src="https://github.com/user-attachments/assets/5d20b479-acbb-43c5-9504-5ac60b796e9d" />

Im already using this beta

the-luap:
@Rekoo-PS — apologies for closing prematurely. Pulled your video frame-by-frame and the loop is between `/admin/login` (default PicPeak logo) → `/admin/dashboard` (your "Arkan" logo, "Loading dashboard…") → back to `/admin/login`. The `038e84c` fix I cited was for parallel-401 dedupe, but your loop has a different root cause.

**The bug:** asymmetric JWT verification.

- `/auth/session` (the endpoint the frontend uses to ask "am I authenticated?") calls `jwt.verify(token, JWT_SECRET)` with no `issuer` check — it accepts old tokens.
- `adminAuth` middleware (which gates every admin endpoint) calls `jwt.verify(token, JWT_SECRET, { issuer: 'picpeak-auth' })` — it rejects them.

So if you have an admin cookie from before commit `23cd9cb` (when the `iss: 'picpeak-auth'` claim was added — Shannon security PR), the loop happens like this:

1. `/admin/login` asks `/auth/session` → "yes, you're admin" (lax check passes).
2. `<Navigate to="/admin/dashboard" />`.
3. Dashboard fires `/admin/dashboard/stats` etc. → `adminAuth` strict check **fails** → `401 Token expired/invalid`.
[...]

PR #403, opened 2026-05-06
feat: customer accounts (#354) — recurring logins, profile, password reset, branded customer surface

Luca-Timo:
Implements the **customer** tier from the-luap/picpeak#354 — a third user
type alongside admins and per-event guests. Photographers with recurring
clients can now invite a customer once; the customer logs in at
`/customer/login` and sees every gallery the admin has linked them to,
without re-entering a per-event password.

Wedding-planner role and public-selection tier from the original
discussion are deferred to follow-ups.

### Auth & accounts
- New token type `'customer'` (alongside `admin`/`gallery`/`guest`) and
  matching cookie `customer_token`. Strictly separate auth surface at
  `/api/customer/auth/*` — admin and customer can be logged in at the
  same time in the same browser without one clobbering the other.
- Admin invite flow mirroring `admin_invitations`: 64-char hex token,
  7-day expiry, email with set-password link.
- Admin password-reset flow — sends a 7-day single-use email link; the
  current password keeps working until the customer clicks it.

### Customer profile
- Salutation, first/last name, display name, phone, company name, VAT id,
  full billing address, preferred language. Admin can pre-fill any
  subset on invite; customer confirms or edits everything on the accept
  page; both customer and admin can edit later from their respective
  detail pages.
- Self-service `/customer/profile` page with password change. Admin can
  view and edit the same data from `/admin/customers/:id` plus trigger a
  reset.

### Dashboards & layout
- `/customer/dashboard` — inline list of assigned galleries, sortable
  by name / newest first / oldest first, per-row Open + Download
  buttons. Click → access-token exchange → gallery loads as if the
  per-event password was entered (no changes to gallery middleware).
- Admin-style sidebar layout for `/customer/*` (Outlet pattern, theme-
  aware, mirrors `AdminLayout`). Five entries: **Galleries**,
  **Calendar**, **Quotes**, **Bills**, **Profile**. Calendar/Quotes/Bills
  are intentionally placeholder stubs with a "Soon" pill — see toggles
  below.

### Admin controls
- New `/admin/customers` and `/admin/customers/:id` pages with invite
  modal (with optional prefill section), customer roster, profile
  editor, feature toggles, password-reset trigger. **BETA** badge in
  the admin sidebar so the not-yet-finished tabs are flagged honestly.
- New `Settings → Customer surface` tab:
  - Hide logo and/or company name in the customer header.
  - Global enable/disable for Calendar / Quotes / Bills. Visibility is
    AND-combined with per-customer overrides — global is the master
    switch (when off nobody sees the tab); per-customer is opt-out only
    (defaults to visible when global is on).
- New RBAC permissions seeded: `customers.view`, `customers.create`,
  `customers.delete` — granted to `super_admin` and `admin` system
  roles; editor/viewer remain locked out.

### Theming
- Customer surface uses theme-token CSS variables consistently; new
  `.customer-surface` scope retheme on `<Input>` so the existing
  `.input` class (hard-coded white in the gallery branding system)
  picks up `var(--color-surface)` etc. Profile-page tiles and password
  prompt headings on the gallery login were also reworked for
  contrast.

### Reliability fixes en route (worth highlighting)
- `/auth/session?slug=…` now prefers the gallery token over admin
  precedence so an admin dogfooding the customer flow doesn't get the
  per-event password prompt.
- `getGalleryTokenFromRequest` / `getAdminTokenFromRequest` only honor
  `Authorization: Bearer` when the JWT's `type` claim matches — admin
  Bearer headers leaking onto gallery requests no longer flip the
  reported identity.
- `/api/customer/events/:slug/access-token` writes the gallery cookie
  alongside the JSON response, so the dashboard → gallery handoff
  survives the cleanup helper that wipes sessionStorage on gallery
  mount.
- `revoked_tokens` insert uses `onConflict('token_id').ignore()` —
  double-logout is a no-op, not a 23505.
- `logActivity` coerces non-integer `actor_id` to null (a legacy
  guest-fingerprint caller was poisoning the audit writer).
- Migration 089's email template insert is schema-tolerant for
  installs with localized `subject_*`/`body_*_<locale>` columns.

## Schema changes

Three migrations:
- **087** — `customer_accounts` (full profile + auth columns),
  `customer_invitations`, `event_customer_assignments`. Seeds the three
  RBAC permissions and the `customer_invitation` email template.
- **088** — `prefill_data` JSON column on `customer_invitations`.
- **089** — `feature_calendar/quotes/bills` columns on
  `customer_accounts`, `customer_password_resets` table, five
  `customer_surface` settings rows, `customer_password_reset` email
  template.
- **090** — flips per-customer feature flag defaults to `true` and
  bulk-updates existing rows so global toggle is the master and
  per-customer is opt-out.

The pre-existing `customer_name` / `customer_email` / `customer_phone`
columns on `events` are freeform contact text — left untouched. The new
tables are independent.

## Verification

1. `npm run migrate` (backend) — all four migrations should land cleanly.
2. As `super_admin`, visit `/admin/customers` → invite a customer
   (try the prefill section), confirm the queued email in the inbox /
   `email_queue`.
3. Open the invite link in a private window → accept → log in →
   `/customer/dashboard` shows assigned galleries.
4. Assign the customer to an event from the event form's
   `<CustomerAccountPicker>` → refresh dashboard → event appears.
5. Click the event → no password prompt → gallery loads.
6. `/customer/profile` → edit fields → save → password change works.
7. From admin: `/admin/customers/:id` → toggle a feature off → switch
   to the customer tab → entry disappears (focus-refetch propagates the
   change without a manual reload).
8. `Settings → Customer surface` → hide logo + name → customer header
   renders empty brand area.
9. `/admin/customers/:id` → "Send password reset email" → customer's
   email gets a link → opens `/customer/reset-password/:token` →
   submitting a new password lets the customer log in with it.
10. Verify per-event password still works for guests in a private
    window with no customer auth.
11. `npm run test` (backend Jest) and frontend Vitest — both green.
    `npx tsc --noEmit` clean.

## i18n

EN + DE hand-translated. **NL / PT / RU keys are English fallbacks and
need native review** — same flagging convention used in earlier PRs:

- `customer.*` — login, dashboard, accept-invite, profile, calendar /
  quotes / bills coming-soon stubs, reset-password
- `customers.*` — admin management, detail page, invite modal,
  feature/reset sections
- `settings.customerSurface.*` — admin Settings tab
- `navigation.customers`, `navigation.betaTag`, `customer.nav.*`

## Out of scope (follow-ups)

- Wedding-planner role from #354
- Public-selection tier from #354
- Calendar / Quotes / Bills feature implementations (the tabs are
  stubs gated by the toggles, so they're easy to ship one at a time)
- Customer self-service email change (admin-only for now)
- Gallery-side `cleanupOldGalleryAuth` is overzealous — it nukes all
  per-slug session tokens on every gallery mount under the guise of
  clearing legacy state. The slug-specific keys aren't legacy; they're
  the current format. Worth scoping that helper to truly old global
  keys (`gallery_token` / `gallery_event` without suffix). For now the
  cookie path keeps things working.

### DIFF
--- a/backend/migrations/core/090_add_customer_accounts.js
+++ b/backend/migrations/core/090_add_customer_accounts.js
@@ -1,0 +1,313 @@
+/**
+ * Migration: Add Customer Accounts (recurring user logins)
+ *
+ * Implements the customer tier from discussion the-luap/picpeak#354.
+ *
+ * Three new tables:
+ *  - customer_accounts        : the user record (email + bcrypt password)
+ *  - customer_invitations     : admin → customer invite handshake (mirrors admin_invitations)
+ *  - event_customer_assignments: many-to-many junction with events
+ *
+ * Three new RBAC permissions seeded so super_admin and admin roles can
+ * manage customers immediately after migrate. Editor / viewer remain
+ * locked out by design (matches the existing users.* permissions pattern).
+ *
+ * Migration is idempotent — every step checks for existing state so a
+ * partial install can be resumed.
+ */
+
+exports.up = async function(knex) {
+  // ---- customer_accounts -----------------------------------------------
+  if (!(await knex.schema.hasTable('customer_accounts'))) {
+    await knex.schema.createTable('customer_accounts', (table) => {
+      table.increments('id').primary();
+      table.string('email', 255).unique().notNullable();
+
+      // --- auth ---------------------------------------------------------
+      // password_hash is nullable until the invitation is accepted —
+      // an unaccepted account row exists only after acceptInvitation, so
+      // in practice this is always set, but the column is nullable to
+      // allow for future "admin creates pre-loaded account" flows.
+      table.string('password_hash', 255);
+      table.boolean('must_change_password').notNullable().defaultTo(false);
+      table.boolean('is_active').notNullable().defaultTo(true);
+      // Tracks password-change time so JWTs issued before a password
+      // change are rejected by customerAuth middleware. Mirrors the
+      // admin_users.password_changed_at column.
+      table.timestamp('password_changed_at');
+      table.timestamp('last_login');
+      table.string('last_login_ip', 45);
+      table.string('preferred_language', 8).defaultTo('en');
+
+      // --- contact ------------------------------------------------------
+      // Salutation honorific (Herr / Frau / Mx / Dr / Other). Stored as
+      // free text rather than an enum so future locales (German "Frau",
+      // French "Mme", legal titles "Dr.", etc.) don't need a migration.
+      table.string('salutation', 32);
+      table.string('first_name', 80);
+      table.string('last_name', 80);
+      // Convenience display name kept separately so the dashboard can
+      // greet customers without joining first/last (e.g. "Welcome, Luca").
+      table.string('display_name', 120);
+      table.string('phone', 40);
+      table.string('company_name', 120);
+
+      // --- billing / address (for future quotes & invoicing) -----------
+      table.string('billing_email', 255);
+      table.string('vat_id', 40);
+      table.string('address_line1', 255);
+      table.string('address_line2', 255);
+      table.string('postal_code', 20);
+      table.string('city', 120);
+      table.string('state', 120);
+      table.string('country_code', 2); // ISO 3166-1 alpha-2
+
+      // --- audit --------------------------------------------------------
+      table.text('notes'); // free-text admin notes, never shown to the customer
+      table.integer('created_by_admin_id').unsigned()
+        .references('id').inTable('admin_users').onDelete('SET NULL');
+      table.timestamp('created_at').defaultTo(knex.fn.now());
+      table.timestamp('updated_at').defaultTo(knex.fn.now());
+
+      table.index(['email']);
+      table.index(['is_active']);
+      table.index(['last_name']);
+      table.index(['company_name']);
+    });
+  }
+
+  // ---- customer_invitations --------------------------------------------
+  if (!(await knex.schema.hasTable('customer_invitations'))) {
+    await knex.schema.createTable('customer_invitations', (table) => {
+      table.increments('id').primary();
+      table.string('email', 255).notNullable();
+      // 64 chars = 32 bytes hex = 256 bits — same entropy as admin invites.
+      table.string('token', 64).unique().notNullable();
+      table.integer('invited_by').unsigned()
+        .references('id').inTable('admin_users').onDelete('CASCADE').notNullable();
+      table.timestamp('expires_at').notNullable();
+      table.timestamp('accepted_at');
+      table.integer('accepted_customer_id').unsigned()
+        .references('id').inTable('customer_accounts').onDelete('SET NULL');
+      table.timestamp('created_at').defaultTo(knex.fn.now());
+
+      table.index(['token']);
+      table.index(['email']);
+      table.index(['expires_at']);
+      table.index(['accepted_at']);
+    });
+  }
+
+  // ---- event_customer_assignments --------------------------------------
+  if (!(await knex.schema.hasTable('event_customer_assignments'))) {
+    await knex.schema.createTable('event_customer_assignments', (table) => {
+      table.increments('id').primary();
+      table.integer('event_id').unsigned().notNullable()
+        .references('id').inTable('events').onDelete('CASCADE');
+      table.integer('customer_account_id').unsigned().notNullable()
+        .references('id').inTable('customer_accounts').onDelete('CASCADE');
+      table.integer('assigned_by_admin_id').unsigned()
+        .references('id').inTable('admin_users').onDelete('SET NULL');
+      table.timestamp('assigned_at').defaultTo(knex.fn.now());
+
+      table.unique(['event_id', 'customer_account_id']);
+      table.index(['customer_account_id']);
+      table.index(['event_id']);
+    });
+  }
+
+  // ---- RBAC permissions -------------------------------------------------
+  // Insert the three customers.* permissions if they're not already present
+  // (guards against re-running the migration in dev). The same idempotency
+  // pattern as 055_add_permissions_table.js.
+  const existingPermissions = await knex('permissions').select('name');
+  const existingNames = new Set(existingPermissions.map((p) => p.name));
+  const newPermissions = [
+    {
+      name: 'customers.view',
+      display_name: 'View Customers',
+      category: 'customers',
+      description: 'View customer accounts and their event assignments',
+    },
+    {
+      name: 'customers.create',
+      display_name: 'Invite Customers',
+      category: 'customers',
+      description: 'Issue customer invitations and assign customers to events',
+    },
+    {
+      name: 'customers.delete',
+      display_name: 'Deactivate Customers',
+      category: 'customers',
+      description: 'Deactivate customer accounts and revoke their access',
+    },
+  ].filter((p) => !existingNames.has(p.name));
+
+  if (newPermissions.length > 0) {
+    await knex('permissions').insert(newPermissions);
+  }
+
+  // Grant the three permissions to super_admin and admin so the feature
+  // is usable immediately. We look up role / permission ids fresh because
+  // the inserts above just landed.
+  const roles = await knex('roles').select('id', 'name')
+    .whereIn('name', ['super_admin', 'admin']);
+  const perms = await knex('permissions').select('id', 'name')
+    .whereIn('name', ['customers.view', 'customers.create', 'customers.delete']);
+
+  if (roles.length > 0 && perms.length > 0) {
+    const existing = await knex('role_permissions').select('role_id', 'permission_id');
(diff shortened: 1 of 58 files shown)


Luca-Timo:
@the-luap basic features of customer Dashboard are now working

the-luap:
Massive piece of work — this is a tier-defining feature and the engineering discipline shows. A few things stood out reading through:

**What's done really well**

- **Customer auth mirrors admin auth precisely.** `customerAuth` does the same issuer-claim verify, `isTokenRevoked` check, `password_changed_at` invalidation, and `is_active` gate. Same security surface, no shortcuts. The `getCustomerTokenFromRequest`/`getAdminTokenFromRequest` JWT-type matching in commit `ef45e045` is exactly the same shape as the symmetry fixes from the #350 redirect-loop family — nice consistency.
- **Idempotent migrations.** Every step in 087 checks `hasTable`/`hasColumn`. 090 flipping per-customer defaults to `true` so global is the master switch is a clean design call.
- **41-commit history shows real dogfooding.** 14 feat + 23 fix commits is a strong signal you actually used the feature end-to-end. The cookie scope, dashboard 500, audit-writer poisoning, and GDPR-erase fixes look like real-world finds rather than lab cases.
- **RBAC discipline** — three new permissions seeded for super_admin + admin only, editor/viewer locked out by design. Matches the existing `users.*` pattern.
- **240 i18n keys per locale** with EN/DE hand-translated and nl/pt/ru explicitly flagged — same convention as your prior PRs, appreciated.

**Two things to address before this can merge**

1. **Conflicting with current beta.** 14 files touched by both this PR and the recent beta merges (PRs #400, #401, #405, plus the #407 beta→main promotion that came back). The conflict surface includes AdminLoginPage, BrandingPage, EventDetailsPage, AdminSidebar, index.css, CHANGELOG, all 5 locale JSONs. Needs a rebase or beta-merge with manual conflict resolution — substantial but not impossible. Has to happen first.

2. **Diagnostic logging is too loud for prod.** `customerAuth.js` has `logger.info('[customerAuth] no token on request', { url, hasCookieHeader, cookieKeys })` that fires on every unauthenticated `/api/customer/*` request — page loads while not logged in, polling, etc. Commit `77369afd` ("add diagnostic logging to auth + events endpoints") looks like it was meant to be temporary debugging. Suggest dropping to `logger.debug` or removing entirely. There's at least one in `routes/customer.js` and a few in `customerAuth.js` worth a sweep.

**Two non-blocking nice-to-haves for after merge / follow-up**

- **No automated test for the critical auth flow.** `customerAccountsService.test.js` covers the service layer well (invitation creation, assignment management, access checks). The login → access-token exchange → gallery handoff is the core promise of this feature though — a Playwright spec ("customer can open a gallery without re-entering the password") would protect against regressions on a path that's now fairly load-bearing.
- **No unit tests for `customerAuth` middleware itself.** The symmetry mirrors `adminAuth` — same checks deserve the same test treatment as `__tests__/routes/authSession.symmetry.test.js` did for /auth/session. Easy follow-up.

**Suggested order**

1. Rebase onto current beta + resolve conflicts (currently the only hard blocker).
2. Quick `logger.info → logger.debug` sweep on the diagnostic lines.
3. Re-request review.

Once 1 + 2 land I'm ready to approve. The test additions can come as a follow-up PR.

Then before any stable promotion, this should soak on beta for a bit — tier-defining changes deserve real-world exposure before they cross to main. Thanks for the thoroughness.

Luca-Timo:
@the-luap should be ready for re-review

Thank you so much!

the-luap:
Thanks @Luca-Timo, this is a huge response — you addressed everything in the review and then some. The Playwright login→gallery spec, the `customerAuth` middleware unit tests, the `logger.info → debug` cleanup, the rebase, AND a customer-portal master toggle + GDPR erase on top. Going to do a fresh pass shortly.

One small thing from the rebase to clean up first: your branch now includes commit `c2b1854` ("fix(admin): test email always sends, regardless of update availability (#418)") which is from my open PR #420 — it doesn't belong on this branch. Could you `git rebase -i origin/beta` and drop that one commit?

Side effect: it also creates a duplicate migration-087 filename (`087_add_update_notification_test_template.js` alongside your `087_add_customer_accounts.js`). Dropping the commit removes the dup automatically.

Once that's clean I'll do the full re-review. Thanks again for the responsiveness here.

Luca-Timo:
@the-luap — ready for re-review when you have a moment.

Branch rebased onto the latest beta (f2f48f3) and the customer-portal work has been replumbed on top of your feature-flag reorg ([#443](https://github.com/Luca-Timo/picpeak/issues/443)) instead of the parallel toggle the original PR shipped. Key changes:

New customerPortal feature flag in your feature_flags table. Migration 095 seeds it TRUE on existing installs (events > 0), FALSE on fresh installs.
The Advanced features tab / CustomerSurfaceTab from the earlier draft are gone. The Customer portal card lives in Settings → Features; the customer-dashboard branding toggles (Show logo / Show company name) sit in Settings → Branding, between Company Information and Gallery Theme, feature-flag gated.
Migrations renumbered to 090–095 to avoid colliding with your 087 / 088 / 089.
RequireFeature route guard wraps the admin-side /admin/customers surface and the sidebar entry; the /customer/* tree itself is intentionally not gated (the guard reads from FeatureFlagsProvider, which only wraps admin chrome).
tokenUtils.js is byte-identical to your beta plus four customer-cookie helpers — COOKIE_SECURE='auto' default from [#427](https://github.com/Luca-Timo/picpeak/issues/427) is preserved.
server.js adds three mounts and nothing else.
Translations: en + de hand-translated. nl / pt / ru / fr are machine-translated — flagging as needing native review before merge.

the-luap:
Re-review pass on the rebased branch. Top-level: the cleanups from the previous round all landed cleanly and the replumb onto the feature-flags table (088) is the right architectural call — much tidier than the parallel toggle the earlier draft shipped.

**Verified from the previous review**

- ✅ Diagnostic noise sweep — `customerAuth.js:26` is now `logger.debug`; the remaining `logger.warn`/`logger.info` lines (token-revoked, IP-drift, password-changed) are appropriate signal, not chatter.
- ✅ Stray `c2b1854` from #420 dropped, no duplicate migration filename.
- ✅ Branch rebased onto `f2f48f3` (current beta), `mergeable=clean`.
- ✅ Migrations renumbered to 090–095 — no collision with 087/088/089 from beta.
- ✅ `customerPortal` feature flag seeded TRUE on installs with events, FALSE on fresh — consistent with 088's seeding rule.
- ✅ Customer login lockout key prefixed `customer:` so admin and customer attempt counters don't share a bucket. Generic error string (no enumeration) and reCAPTCHA gate are both there.
- ✅ Login response carries `features` + `branding` inline so first-render dashboard doesn't flash defaults — nice catch on the UX side.
- ✅ New tests pass locally (23/23 across `customerAuth.middleware.test.js` + `customerAccountsService.test.js`) plus the Playwright login→gallery spec exists.

**One blocker that just appeared (not your fault)**

- **Migration 090 collision incoming.** I just opened #455 (photo-dimensions backfill) which also numbers its migration 090. Whichever lands second has to renumber the entire 090–095 chain to 091–096. Easier path: I can hold #455, you merge first, then I rebase #455 to 091. Or vice versa. Either works — flag your preference.

**Three small things, all non-blocking**

1. **`must_change_password` is set on creation but never read on login.** `customerAuth` middleware doesn't gate on it, and the login route doesn't surface a "must change password" code. In practice the column is decorative because accept-invite always sets a customer-chosen password — but if you ever ship an "admin pre-loads a temporary password" flow, you'll need to wire this up like `adminAuth` does. A `// TODO when temp-password flow lands` next to the column would document the gap.

2. **Backend deliberately doesn't gate `/api/customer/auth/login` on the `customerPortal` flag** — already covered by the comment in `server.js:571`, but `customerAuth.js:51` (the login route itself) is silent on the choice. A one-line comment there ("portal toggle off ≠ access revoked; deactivate accounts individually to lock out existing customers") would save the next reader from wondering whether it was forgotten.

3. **`migrations/core/095_add_customer_portal_flag.js:1`** comment header says "Migration 094" — copy-paste from 094. Trivial, but easy to fix while you're in there.

**i18n** — your nl/pt/ru/fr "machine-translated, needs native review" flag is fair. I'm comfortable letting that soak on beta and inviting native-speaker corrections via the usual channel; not a blocker.

**Approval pending the 090 collision.** Once we settle the migration-number ordering with #455, this is good to go in. Plan still stands: merge to beta, soak, then a deliberate beta→main promotion since this is tier-defining.

Outstanding work — appreciate how thoroughly you closed the previous loop.

the-luap:
Quick correction on the previous comment — the migration 090 collision **isn't** a blocker for this PR.

Plan:
1. **Merge #403 first** when you're happy with it.
2. I'll rebase #455 onto the resulting beta tip, renumber its migration from 090 → 096, and **fold the three small notes from this review into the same PR** (the `must_change_password` TODO comment, the customer-login route note about the portal toggle, and the migration 095 header typo). That way they land together in one tidy follow-up rather than as drive-by changes here.

So this PR is good to merge from my side. Approval ✅ — go ahead whenever you're ready.

PR #429, opened 2026-05-09
fix(auth): default COOKIE_SECURE to 'auto' in production + first-install UX (#427)

### LINKED ISSUE #427: [BUG] Login fails silently over HTTP due to Secure cookie flag
Describe the bug
When setting up PicPeak for the first time without a reverse proxy (direct HTTP access via http://IP:3000), the login appears to work but immediately redirects back to the login page without any error message. The backend returns 200 OK on the login request, but the session cookie is silently discarded by the browser because it is flagged as Secure over a plain HTTP connection.
To Reproduce

Fresh install via picpeak-setup.sh with Docker on a local server
Access the admin panel via http://192.168.x.x:3000/admin (no reverse proxy, no HTTPS)
Enter valid credentials and click login
Page blinks briefly and returns to the login page — no error message shown

Expected behavior
Either the login works over HTTP, or a clear error message is displayed explaining that HTTPS is required.
Screenshots
Browser DevTools → Network tab shows:

POST /api/auth/admin/login → 200 OK with Set-Cookie: admin_token=...; Secure; HttpOnly
Immediately followed by GET /api/auth/session → 401 Unauthorized
Cookie is never stored because browser discards Secure cookies over HTTP

Environment:

OS: Debian 13 (Trixie) in Proxmox LXC
Browser: Chrome 147
PicPeak Version: 3.43.1
Deployment Method: Docker Compose (via picpeak-setup.sh)
Database: PostgreSQL 15

Logs
Backend logs show nothing during the failed login — the request never appears, which makes debugging extremely difficult without DevTools knowledge.
Browser console errors:
GET http://192.168.x.x:3000/api/auth/session 401 (Unauthorized)
Additional context

The COOKIE_SECURE environment variable exists and setting it to auto fixes the issue completely
However this variable is not documented in .env.example with a clear explanation
The backend already supports COOKIE_SECURE=auto which correctly detects HTTP vs HTTPS per request
This will affect every new user who tries PicPeak without a reverse proxy

Possible Solution

Change the default value of COOKIE_SECURE from true to auto in production — the backend already supports this and handles it correctly
Or add a visible warning in the UI when the session cookie is not stored after a successful login
Or clearly document in the README and setup script that HTTPS via reverse proxy is required before the first login attempt

the-luap:
Fixes #427 (reported by @iSchumi6210, also confirmed by @Rekoo-PS in the comments).

## What was broken

Two intertwined bugs that made first-install UX painful:

### 1. Login silently fails over HTTP

- Backend defaulted `COOKIE_SECURE` to `true` whenever `NODE_ENV=production`
- `picpeak-setup.sh` writes `NODE_ENV=production` and never writes `COOKIE_SECURE`
- Over plain HTTP the browser drops the `Secure` cookie → next `/auth/session` request returns 401 → redirect back to `/admin/login` → no error shown
- Backend log shows nothing because the `/auth/session` request never hits the auth path

DevTools is the only way to diagnose this. New users without DevTools experience just see the login page bouncing.

### 2. Admin credentials are generated but invisible

The `001_init.js` migration writes the generated random password to `data/ADMIN_CREDENTIALS.txt` inside the backend container. But `picpeak-setup.sh` only copies that file out to the host when `--reset-admin-password` is passed. Default-path users never see the file → resort to manually updating the bcrypt hash in psql to log in.

iSchumi noted both in the issue + comment thread.

### `backend/src/utils/tokenUtils.js`

Production default goes from `true` → `'auto'`.

- On real HTTPS via reverse proxy: `req.secure` is `true` → `Secure` flag still emitted. **No security regression** for HTTPS deployments.
- On plain HTTP: `req.secure` is `false` → `Secure` flag omitted → login works.
- Users who explicitly want strict HTTPS-only behavior can still set `COOKIE_SECURE=true`.

### `backend/.env.example`

Rewrite the `COOKIE_SECURE` block to make the new default obvious and explain *when* you'd override:
- Leave unset → `auto` in production. Right for almost everyone.
- `=true` → strict HTTPS-only (defense in depth on a deployment where you know HTTPS is the only access path)
- `=false` → opt-out of the per-request check entirely (rare)

### `scripts/picpeak-setup.sh` (both Docker and native install paths)

- Write `COOKIE_SECURE=auto` explicitly to the generated `.env`. Defense in depth — even if the backend default ever flips back, the user's `.env` says what they want.
- After migrations, **always** copy `ADMIN_CREDENTIALS.txt` from the backend data dir to the host, `chmod 600`, and print email + password to the install output. The credentials file stays as a backup record that the operator should delete after noting the password.

## Why 'auto' as the default isn't a security regression

`auto` mode reads `req.secure`, which Express derives from `X-Forwarded-Proto` when the proxy is in the trust list. The trust list (`127.0.0.1`, `10.x`, `172.16-31.x`, `192.168.x`, link-local) is the same one that's been in production for a long time.

- HTTPS via trusted proxy → `req.secure=true` → `Secure` flag → identical behaviour to old `true` default
- HTTP → `req.secure=false` → no `Secure` flag → cookie works (was previously a silent loop)
- HTTPS via untrusted proxy that doesn't forward `X-Forwarded-Proto` → `req.secure=false` → no `Secure` flag → degraded but still works (and this is the same misconfiguration you'd have to fix anyway for correct downstream `https` URL generation, etc.)

The risk vector is "MITM between proxy and backend on a misconfigured deploy." That risk already exists because the proxy → backend traffic is plain HTTP regardless of cookie flag. The Secure flag only protects the *browser → proxy* leg, which `auto` correctly handles when the proxy is configured even minimally.

## Verification

Tested all 4 permutations of `NODE_ENV` × `COOKIE_SECURE` in the dev container:

| `NODE_ENV` | `COOKIE_SECURE` | HTTPS req | HTTP req |
|---|---|---|---|
| production | (unset)  | `secure: true` ✓ | `secure: false` ✓ ← **was broken before** |
| production | `=true`  | `secure: true` | `secure: true` ← strict opt-in preserved |
| production | `=auto`  | `secure: true` | `secure: false` ← already-correct case |
| development | (unset) | `secure: false` | `secure: false` ← dev unchanged |

Lint clean on touched file.

## Test plan

- [ ] Fresh `picpeak-setup.sh` install on an HTTP-only test box (e.g. LXC without reverse proxy)
- [ ] Verify `.env` contains `COOKIE_SECURE=auto`
- [ ] Verify install output prints the admin email + password and writes `data/ADMIN_CREDENTIALS.txt` with `chmod 600`
- [ ] Verify login works over `http://IP:3000/admin` without DevTools workaround
- [ ] Network tab: `Set-Cookie` should NOT have `Secure` flag on HTTP, SHOULD have it on HTTPS
- [ ] Existing HTTPS-via-reverse-proxy deployments: behaviour unchanged (Secure flag still emitted via `req.secure`)
- [ ] Strict mode opt-in: setting `COOKIE_SECURE=true` in `.env` still forces Secure on every cookie

### DIFF
--- a/backend/.env.example
+++ b/backend/.env.example
@@ -10,18 +10,31 @@ PORT=3001
 JWT_SECRET=your-very-secure-jwt-secret-at-least-32-characters-long-example123456
 
 # Auth cookie Secure flag
-#   unset  - default: follows NODE_ENV (production=true, dev=false)
-#   true   - always set Secure (HTTPS-only cookies; breaks plain-HTTP access)
+#   unset  - default: 'auto' in production, false in dev (#427)
+#   true   - always set Secure (HTTPS-only cookies; breaks plain-HTTP access —
+#            login appears to succeed but the browser silently drops the
+#            cookie, leaving you in a redirect loop. Only set this if you
+#            ALWAYS reach the site via HTTPS)
 #   false  - never set Secure (allows HTTP; cookies not protected on HTTPS)
-#   auto   - decide per request: Secure on HTTPS, not on HTTP
+#   auto   - decide per request: Secure on HTTPS, not on HTTP. Reads
+#            req.secure from Express which respects X-Forwarded-Proto from a
+#            trusted reverse proxy. This is the default and is the right
+#            choice for most deployments.
 #
-# Use COOKIE_SECURE=auto if your deployment is reachable over both HTTPS
-# (via a reverse proxy like Nginx Proxy Manager, Traefik, Caddy) AND plain
-# HTTP (e.g. LAN access at http://192.168.x.x:3001). The backend reads
-# req.secure from Express, which respects the X-Forwarded-Proto header
-# when the proxy is in the trust list.
+# Why 'auto' is the default in production:
+#   - On real HTTPS (reverse proxy with X-Forwarded-Proto), req.secure is
+#     true → Secure flag is still emitted. No security regression vs. true.
+#   - On plain HTTP (LAN access, first-time install before reverse proxy is
+#     wired up), req.secure is false → Secure flag is omitted → login works
+#     instead of silently looping back to /admin/login.
 #
-# Requirements for auto mode:
+# When you'd set this explicitly:
+#   - COOKIE_SECURE=true → strict HTTPS-only deployments where you want
+#     defense in depth against accidentally serving over HTTP.
+#   - COOKIE_SECURE=false → you intentionally only ever serve over HTTP and
+#     don't want the per-request check (rare).
+#
+# Requirements for 'auto' mode to detect HTTPS correctly:
 #   1. Your reverse proxy MUST send X-Forwarded-Proto: https on HTTPS
 #      requests. Standard configs for NPM/Traefik/Caddy do this by default.
 #   2. The proxy must be on a trusted IP range. By default PicPeak trusts
--- a/backend/src/utils/tokenUtils.js
+++ b/backend/src/utils/tokenUtils.js
@@ -7,15 +7,24 @@ const DEFAULT_MAX_AGE_MS = 24 * 60 * 60 * 1000; // 24 hours
 
 /**
  * Cookie "Secure" flag mode:
- *   - true   → always set Secure (HTTPS-only)
- *   - false  → never set Secure (allow plain HTTP)
+ *   - true   → always set Secure (HTTPS-only — cookie won't be sent over HTTP at all)
+ *   - false  → never set Secure (allow plain HTTP — cookie has no in-flight protection)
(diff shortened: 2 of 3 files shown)


iSchumi6210:
Tested on: Fresh install via picpeak-setup.sh (Docker) on Debian 13 LXC, HTTP-only, no reverse proxy. Branch: beta (v3.43.3-beta.0)

✅ .env contains COOKIE_SECURE=auto
✅ Admin credentials printed at end of setup + written to data/ADMIN_CREDENTIALS.txt
✅ Login works over http://192.168.x.x:3000/admin without any workaround
✅ Set-Cookie has NO Secure flag on HTTP — cookie persists correctly
✅ No 401 redirect loop, dashboard loads cleanly

LGTM! 🚀

Issue #454, opened 2026-05-11
[BUG] ReferenceError: Cannot access 'I' before initialization


derooijmnl:
Hey Paul, im on the newest beta version v3.45.1-beta0 and having troubles opening the event page. When troubleshooting with Claude he gives me this:

ReferenceError: Cannot access 'I' before initialization
Dit is een bug in de beta van PicPeak specifiek op de events pagina. De backend heeft er niets mee te maken.

Versie: 3.45.1-beta.0
Pagina: /admin/events
Error: ReferenceError: Cannot access 'I' before initialization at r4 (index-jvYTdYC2.js:247:28919)

Its been a while since i looked, but the app was massivly upgraded, greatly appreciated the hard work! Keep it up. :)

Gr Mark

the-luap:
@derooijmnl — confirmed bug, fix is in PR #455 (targeting `beta`).

**Root cause:** the pagination-clamp `useEffect` shipped in #448 was inserted at the top of the component body, **before** the `useQuery` that declares `data`. Because the effect's dependency array `[data?.pagination, page]` is evaluated immediately when that line executes, every render hit a temporal dead zone access on `data` and threw `ReferenceError: Cannot access 'data' before initialization` — minified to `'I'` in production. Crashed the whole `/admin/events` page.

TypeScript actually flagged this at build time (`Block-scoped variable 'data' used before its declaration`) but the build doesn't fail on TS errors so it shipped anyway. Lesson learned on my side — going to look at making TS errors fatal in CI.

**Fix:** moved the effect to immediately after `useQuery` so `data` is in scope. No behaviour change otherwise.

Folded into PR #455 since that PR is already going through the smoke gate. Will land on `beta` in the next merge — apologies for the disruption, and thanks for the kind words on the upgrade!

the-luap:
Thanks @derooijmnl — really helpful report, the minified `'I'` was a great clue. Fix is in #455 and will land on beta shortly. Appreciate the heads-up! 🙏

the-luap:
Fixed in `2f63188` (fix(events): TDZ ReferenceError on /admin/events from #442 fix) and released in v3.44.0 stable (via PR #568).

Please reopen if you still see the `Cannot access 'I' before initialization` error on /admin/events on v3.44.0+.

PR #455, opened 2026-05-11
fix(import): capture photo dimensions in fileWatcher + s3AutoImporter (#447)

### LINKED ISSUE #447: [QUESTION] can we achieve this style ?
can we achieve this style ?
landscape photos are landscape thumbnails - and portraits are portraits.. 

Not every photo is cropped weirdly or only 1:1 .

would love to see this theme in this amazing script <3 

Image

-
-
-
-
or smth like this live example :
https://link.arkan-studio.com/gallery/777f3c1a-f4d5-4ccf-abbb-fecdcb19ab24/

the-luap:
## Summary

Closes the masonry "everything looks cropped to 1:1" complaint from #447. Two import paths were silently dropping `width`/`height` on insert, which forced `MasonryGalleryLayout` (and mosaic / justified) to use a hard-coded 800×600 fallback for every photo — so the layout looked uniform regardless of the source aspect ratio, no matter which thumbnail fit mode the admin picked.

### What changed

**1. `fileWatcher.js`** — filesystem auto-import (the default install path) now extracts width/height with `sharp.metadata()` before inserting the photos row. Mirrors what `photoProcessor.js` (uploads API) has been doing all along.

**2. `s3AutoImporter.js`** — S3 auto-import does the same, materialising a temp local copy via the existing `withLocalCopy` helper so it works without holding the whole object in memory.

**3. Migration `090_backfill_photo_dimensions_v2.js`** — re-runs the dimension extraction for any pre-existing rows still NULL after migration 064 ran. Skips videos (would need ffprobe). Local-fs only — S3 deployments cannot reach the storage backend from inside a migration; they rely on the writer fix in (2) for new photos.

**4. Default thumbnail fit** — `DEFAULT_THUMBNAIL_FIT` in `imageProcessor.js` flipped from `'cover'` to `'inside'`. This only takes effect when the seed setting is missing (almost never in practice — migration 040 seeds `'cover'`). Existing installs keep their saved value. Frontend `ThumbnailsTab` gets a tooltip recommending `inside` for masonry / mosaic / justified, `cover` for uniform 1:1 grids.

### Why this is the right fix

`MasonryGalleryLayout.tsx` is already aspect-aware end-to-end (Pinterest-columns, justified-rows, Flickr, quilted modes — all read `photo.width / photo.height`). The component doesn't need any changes; it just needs the data. With dims captured at write time and the one-shot backfill, the layout immediately starts rendering portraits as portraits and landscapes as landscapes.

### Out of scope

- Generating a separate masonry-friendly thumbnail size (e.g. 600px long-edge with `fit: inside`). Would double thumb storage; useful but not required for the bug fix.
- Changing the default gallery layout (still `grid` for new events).
- Touching `MasonryGalleryLayout.tsx` — already correct.

## Test plan

- [x] All 13 local smoke tests green.
- [ ] Manual — local fs install:
  - [ ] Drop a portrait + a landscape image into `events/active/<slug>/individual/`.
  - [ ] Inspect `photos` row → `width` / `height` populated.
  - [ ] Open the gallery in masonry layout → portrait card is taller than landscape card.
- [ ] Manual — backfill:
  - [ ] On an install where some photos have NULL dims, run `npm run migrate` and confirm migration 090 logs `updated N`.
  - [ ] After migration, masonry layout reflects per-photo aspect ratios.
- [ ] Manual — S3 install:
  - [ ] With `STORAGE_BACKEND=s3` and `STORAGE_AUTO_IMPORT=true`, drop a new image into the bucket prefix.
  - [ ] On the next poll tick, the `photos` row has dims set.
- [ ] Settings → Thumbnails: tooltip appears under the Fit Mode dropdown explaining the recommendation.

Targets `beta`. Closes #447 on the masonry side; the layout itself was already there.


the-luap:
Rebased onto beta after #403 merged. Two changes from the post-merge cleanup:

- **Migration renumbered 090 → 096** (`096_backfill_photo_dimensions_v2.js`) so it follows #403's customer-portal chain (090–095).
- **Folded in the three small notes from my #403 review** since they were trivial:
  - `must_change_password` TODO comment in `customerAccountsService.js` documenting that the column is decorative until a temp-password flow ships.
  - `customerAuth.js` /login route now carries a doc-comment explaining why the `customerPortal` toggle deliberately doesn't gate it (toggle off hides UI, doesn't revoke existing-customer access).
  - `095_add_customer_portal_flag.js` header typo fixed ("Migration 094" → 095).

All 23 customer tests + all 13 local smoke tests still green. PR scope unchanged otherwise.

the-luap:
Folded in a hot fix for #454 — `/admin/events` was throwing `ReferenceError: Cannot access 'I' before initialization` (minified `data`) on every render. Root cause: the pagination-clamp `useEffect` shipped in #448 was placed at the top of the component body, before the `useQuery` that declares `data`, so the dep array `[data?.pagination, page]` hit a TDZ on every render and crashed the whole page.

TS actually flagged it at build time (`Block-scoped variable 'data' used before its declaration`) but the build doesn't fail on TS errors so it shipped anyway.

Fix: move the effect to immediately after `useQuery`. Behaviour unchanged. 13/13 smoke tests still green; pushed cleanly.

PR #458, opened 2026-05-11
fix(customer-portal): post-merge fixes for event save, theme fonts, and customer→gallery handoff

Luca-Timo:
Two themes in this PR: post-merge bug fixes and a small feature (the **Clients** section scaffold).

## Bug fixes (post-merge follow-ups to #403)

**`fix(events): strip customer_account_ids from update spread`** — `PUT /admin/events/:id` spread `req.body` straight into `db('events').update(...)`. `customer_account_ids` isn't a column on `events` (it's consumed below by `setAssignmentsForEvent`), so SQL threw "column does not exist" and the edit surfaced as a generic "Failed to update event" toast. Adding the customer to an event then editing it now works end-to-end.

**`fix(customer): preserve slug-scoped gallery tokens on auth provider mount`** — `cleanupOldGalleryAuth()` ran on every `GalleryAuthProvider` mount and swept every `gallery_token_*` sessionStorage entry — including the one `CustomerDashboardPage` had stored 200ms before navigating. The customer-→-gallery handoff was being undone before the gallery page could read it, so customers got the per-event password prompt despite having just authenticated. Restricted the cleanup helper to its actual job (the legacy single-gallery `gallery_token` / `gallery_event` keys and the bare cookie) and left slug-scoped storage to its rightful owner.

**`fix(theme): "Same as body" heading font no longer inherits stale value`** — `ThemeContext.applyTheme()` only wrote `--heading-font-family` when `themeConfig.headingFontFamily` was truthy. Picking "Same as body" stores an empty string, so the property was never updated and the CSS variable retained whatever previous value `:root` had — typically a serif heading font that leaked in from a gallery's per-event theme on a prior visit. Customer profile, admin login, customer login, and customer dashboard all rendered with stale serif headings despite the active theme having body + heading set to the same font. Empty `headingFontFamily` now mirrors `fontFamily` explicitly.

**`feat(branding): toggle login-page logo frame + size`** — Restored the "Show tinted frame behind login logo" toggle and added a size selector for the same surfaces. Both apply *only* to `/admin/login` and `/customer/login` — gallery and admin chrome keep their own `logo_size` knob untouched. Lives in Settings → Branding under "Login pages logo".

**`fix(features-tab): icon tiles + preview pills follow CI accent`** — The icon tiles on the Features tab and the feature-driven pills in the Sidebar Preview were hardcoded to Tailwind's `primary-50` / `primary-700` palette, so they stayed green even when the admin had set a teal/blue/whatever accent in Branding → Theme. Added `.bg-accent-soft` / `.text-on-accent-soft` / `.border-accent-soft` utility classes derived from `--color-accent-dark` via `color-mix()` and switched the consumers. Bumped the dark-mode mix weight after the first pass produced unreadable contrast.

**`fix(settings): neutralize sidebar icons for a consistent palette`** — Active sidebar pill paints `accent-dark` + `text-white`, but the icon was set to `text-accent` (the lighter token), producing low-contrast accent-on-accent. Switched the active icon to `text-white` so pill + label + icon are visually consistent. The section heading icon on the right-hand panel was the only other accent fleck on an otherwise neutral page; switched to neutral too.

**`fix(events): CustomerAccountPicker hooks order crashed /admin/events/new`** — Surfaced in re-review. The picker called `useFeatureEnabled('customerPortal')` and then early-returned `null` BEFORE the rest of its hooks (`useState×4`, `useRef`, `useEffect×2`, `useMemo`). When the React Query for `/admin/feature-flags` resolved between renders, the flag value flipped from `DEFAULT_FLAGS`' false to the real value, the early return toggled off, the hook count differed between renders, and React crashed the entire `/admin/events/new` page through the global error boundary. Moved the early return below every hook so the count is stable across renders.

## Feature: top-level "Clients" section + sub-nav scaffold

Restructures the admin sidebar to host the customer-portal management inside a new top-level **Clients** section with a Settings-style nested sub-nav. Today the only sub-nav entry is **Accounts** (the existing customer-portal management page); future Calendar / Quotes / Bills / Messaging features (#354 roadmap) drop in as additional sub-nav entries with zero structural refactor.

- **New migration `097_add_clients_feature_flag.js`.** The `clients` flag is **derived** from its children — admins only toggle the specific feature they want (Accounts today; siblings later) and the Clients sidebar section appears automatically. No "enable two toggles for one feature" UX wart.
- **New `requireCustomerPortal` middleware** mounted on `/api/customer/*` AND `/api/admin/customers/*`. Toggling the feature off now hard-kills it — 410 Gone everywhere, not just hidden in the UI. Deliberate departure from the previous "soft kill" pattern documented inline in `customerAuth.js` and `server.js`.
- **New `ClientsLayout` component** (220px sub-nav rail, accent-dark active pill — mirrors `SettingsPage`). No placeholder UI; absent sub-features simply don't render.
- **Backward-compat redirects:** `/admin/customers` → `/admin/clients/accounts`, `/admin/customers/:id` → `/admin/clients/accounts/:id`.
- **i18n:** `navigation.clients`, `clients.*`, `settings.features.sections.clients`, `settings.features.customerPortal.title` (relabelled "Accounts"). en + de hand-translated; nl/pt/ru/fr machine-translated — fine to soak on beta per project convention.

## Test plan

- [x] Admin login works (regression catch — the cookie `Secure='auto'` default from #427 stays in place)
- [x] `/admin/events/new` opens and renders the full form (regression: Rules of Hooks crash)
- [x] Customer invite → accept-invite → land on customer login
- [x] Customer login → dashboard lists only galleries the customer is assigned to
- [x] Click "Open" on an assigned gallery: lands in the gallery without password prompt, even with an admin session in the same browser
- [x] Type the URL of a gallery the customer is NOT assigned to: password prompt as before
- [x] Admin can assign customers to an event from both create and edit forms; edit-then-save no longer 500s
- [x] Toggle Accounts OFF in Settings → Features → `/customer/*` and `/api/admin/customers/*` return 410; sidebar Clients entry disappears
- [x] Toggle Accounts ON → both surfaces work, Clients sidebar entry reappears, `/admin/clients/accounts` renders
- [x] `/admin/customers` redirects to `/admin/clients/accounts`
- [x] Settings → Branding → "Login pages logo" toggles affect `/admin/login` and `/customer/login` only
- [x] All headings (customer profile, dashboard, admin login) render in the configured body font when "Same as body" is selected
- [x] Features-tab icon tiles + Sidebar Preview pills follow the CI accent in light AND dark mode
- [x] `npx tsc --noEmit` clean
- [x] Customer auth Jest suites green (`customerAccountsService`, `customerAuth.middleware`)

## Translation note

en + de hand-translated. **nl / pt / ru / fr are machine-translated** and should be reviewed by native speakers — flagging per project convention. Maintainer indicated comfort soaking these on `beta` while native-speaker corrections come through normal channels.


Luca-Timo:
Please wait with the merge for a minute, would like to change the customer backend to CRM and add submenus for different functions

Luca-Timo:
Now Done, @the-luap ready for review!

the-luap:
Quick re-review. Three of the six documented fixes are clean as written; flagging the ones that need work before merge plus a couple of structural notes.

**Verified clean**

- ✅ `dde72a1` — `delete updates.customer_account_ids` before the `db('events').update(...)` spread. Right fix for the right reason; matches how `regenerate_client_token` is handled the same way two lines above.
- ✅ `7ac1d14` — `cleanupOldGalleryAuth` now scoped to legacy keys only. The doc-comment is excellent; future-me will know exactly why slug-scoped storage is off-limits to this helper.
- ✅ `35f5b86` — `effectiveHeadingFont = themeConfig.headingFontFamily || themeConfig.fontFamily` falls through correctly. Subtle bug, clean fix.
- ✅ `75e41eb` — login-page logo toggles, scoped to `/admin/login` and `/customer/login` only.
- ✅ `15d01f3` + `bf7ef14` + `2f00bbd` — accent-soft color-mix utilities + sidebar icon contrast. Light AND dark mode looked right when I poked at them.

**Two blockers**

1. **Rules of Hooks violation in `CustomerAccountPicker.tsx:36-48`.** The early `return null` after `useFeatureEnabled('customerPortal')` runs **before** 8 subsequent hooks (4×`useState`, `useRef`, 2×`useEffect`, `useMemo`). When the flag value changes between renders — which it always does, because `serverFlags ?? DEFAULT_FLAGS` returns the default `false` on the first render and then the real value once the React Query resolves — React sees a different hook count and throws "Rendered more hooks than during the previous render." That kills the entire `/admin/events/new` page through the global error boundary, which is what's failing smoke spec 09 (`Timeout` filling Admin Email — the form is gone, replaced by "Oops! Something went wrong"). Fix: move the `if (!customerPortalEnabled) return null;` check to **after** all hooks (you can render the gate as part of the JSX with a conditional `return`, or keep the early return but call all the hooks above it).

2. **Two "Save changes" buttons on `/admin/branding`.** The new `CustomerDashboardBrandingCard` adds a button labelled `Save changes` (lowercase 'c'); the existing main save is `Save Changes`. Smoke specs 06 and 07 use `getByRole('button', { name: /Save Changes|Änderungen speichern/i })` which is case-insensitive and matches both → strict-mode violation, both tests fail. The duplication is also a UX smell — two visually identical "Save changes" buttons on the same page invite accidental clicks on the wrong one. Suggest renaming the new one to something distinctive (`Save customer header`, `Apply to dashboard`, etc.) which fixes the ambiguity AND the smoke tests in one stroke.

So the smoke gate is currently **3/13 red** on this PR — none of those failures are flakes, all reproducible. (1) is the load-bearing one; (2) is cosmetic-but-test-breaking.

**Two structural notes (non-blocking but worth a sweep)**

3. **PR description undersells the diff.** "Six independent fixes" + the `feat(clients)` scaffold (commit `9091ed4`, +562 lines: new migration 097, new `requireCustomerPortal` middleware, new `ClientsLayout`, route restructure with `/admin/customers/*` → `/admin/clients/accounts/*` redirects, `clients` parent feature flag derived from sub-children). That's a proper feature, not a follow-up — it deserves its own bullet in the description so reviewers know to look at it. The architecture itself is good: the parent flag derived from sub-children means the sidebar entry appears/disappears automatically as Calendar/Quotes/Bills land later. Backward-compat redirects on the old `/admin/customers` paths are nice.

4. **Direct contradiction with my comment from #455.** `customerAuth.js:51-56` says the `customerPortal` flag *deliberately* doesn't gate the login route ("toggle off hides UI, doesn't revoke existing-customer access"). PR #458 then adds `requireCustomerPortalEnabled` middleware on `app.use('/api/customer/auth', ...)` in `server.js:586`, which **does** gate the login route. The comment now lies about the actual behaviour. The new design is more conventional ("if a feature is off, it's off everywhere") — I'm fine with that pivot — but the inconsistency must go. Just delete the comment block in `customerAuth.js:51-56` and update the `server.js:571-577` comment to clarify that the gate is now enforced both via the frontend and via the backend middleware. Worth a moment's thought on the UX of "admin flips toggle off → all logged-in customers get 410 on next request." Probably fine, just be deliberate.

**Translation flag** — your nl/pt/ru/fr machine-translated note is fair; happy to soak on beta and let native-speaker corrections come through the usual channel.

Plan: fix (1) + (2), I'll re-run smoke and approve. (3) is a description edit, (4) is two file touches.

the-luap:
Correction on (2) above — fair pushback, the "two save buttons" framing was overreach. They serve different scopes (the main one saves the full branding form, the new one saves just the customer-dashboard sub-card with its own dirty state) so the duplication is actually intentional and clean. That pattern is common on long settings pages.

The real issue is on **my smoke test side** — selector `getByRole('button', { name: /Save Changes|.../i })` is case-insensitive and too loose. That's in the gitignored local smoke suite, my code to fix. Will tighten the selectors before next push.

So updated triage:

- **(1) `CustomerAccountPicker` hook violation** — still a real blocker; this is the only thing failing for non-test reasons. Crashes the entire `/admin/events/new` page at runtime through the error boundary.
- ~~**(2) Two Save Changes buttons**~~ — withdrawn. Smoke selectors on my side need tightening.
- **(3) PR description undersells the diff** — still worth a description edit.
- **(4) Comment contradiction in `customerAuth.js`** — still needs the comment removed since the new middleware reverses that design.

Apologies for the noise on (2).

Luca-Timo:
I am right on it, changes got a bit messy with the rebase / new enable logic

Luca-Timo:
Addressed re-review feedback:

(1) Rules of Hooks violation in CustomerAccountPicker: fixed in 2a7ae07
(4) Comment contradiction between customerAuth.js and the new middleware: fixed in c0c6b4c
(3) PR description rewritten to surface the feat(clients) scaffold as its own theme

Ready for another smoke pass when you have a moment.

the-luap:
Re-pass on `c0c6b4c`. Both fixes landed exactly right.

**(1) Hooks fix in `CustomerAccountPicker` (`2a7ae07`)** — verified. All hooks (`useTranslation`, `useFeatureEnabled`, 4×`useState`, `useRef`, 2×`useEffect`, `useMemo`) now run before the `if (!customerPortalEnabled) return null` at line 119. Hook count is stable across renders. Doc comment at the top of the component is excellent — explicitly calls out the failure mode + the React Query timing that triggered it, so the next maintainer who's tempted to "tidy up" by moving the early-return back to the top has the load-bearing reason staring them in the face. **Smoke spec 09 now passes** (verified locally — /admin/events/new renders the full form).

**(4) Comment alignment (`c0c6b4c`)** — verified. The new `customerAuth.js` block accurately reflects the dual-enforcement model and points to the middleware in `server.js`. The `server.js` block reads cleanly as a single source of truth: two-place enforcement, login deliberately included in the kill switch, "deliberate departure from the prior design that left login alive when the rest of the surface was off" — exactly the level of explicit-ness this kind of design pivot needs. The note about already-issued JWTs continuing to be technically valid but every API call returning 410 is a nice touch; it preempts the "but my token is still valid, why am I locked out?" question.

**(3) PR description** — much better. The two-themes split (bug fixes vs Clients-section feature scaffold) is now clear, the hooks-fix has its own bullet, and the Clients-section section properly surfaces the migration / middleware / layout / redirects breakdown. Reviewers can now see the actual scope at a glance.

**Smoke status: 11/13 green.** The two remaining failures (specs 06 and 07) are entirely on my smoke-suite side — the case-insensitive `getByRole('button', { name: /Save Changes|.../i })` selector matches both the main "Save Changes" and the `CustomerDashboardBrandingCard` "Save changes" (lowercase 'c'). The card itself is right; my selector is too loose. I'll tighten it in the local smoke suite (gitignored, my problem). **Not a blocker on this PR.**

**Approved ✅** — merge whenever you're ready. Once this lands I'll rebase #461 (the small socials read-mapping fix for #460) onto it.

Real solid responsiveness here — turnaround from review to fix was tight and the doc-comments on both fixes are above average.

Issue #441, opened 2026-05-10
[BUG] Edit Footer


Rekoo-PS:
<img width="866" height="258" alt="Image" src="https://github.com/user-attachments/assets/5f260822-225b-41e2-9583-0e5cf5095546" />

Adding an option to remove these pages, as they are not required in my country .

Adding an option to add our Socials in the footer .
Facebook | Instagram | Send Whatsapp | Blah blah  | 

OR 
A fully customisable Events Footer with HTML for example, will be great. 🙏🏻

the-luap:
Hey @Rekoo-PS — relabelling this as `enhancement` since it's a feature request, not a bug, but the underlying need is clear and reasonable.

Current state in beta:
- ✓ `branding.footer_text` already lets you set arbitrary footer text (Settings → Branding)
- ✓ `branding.hide_powered_by` hides the "Powered by PicPeak" tagline
- ✓ Impressum + Datenschutz CMS pages already support an external URL (so you can point them somewhere else if needed)
- ✗ **No way to hide the legal links** entirely — they always render in the footer if the CMS page exists
- ✗ **No socials in the footer** — would need new fields

What I'd propose, in priority order:

1. **Per-page "Show in footer" toggle on CMS pages.** Already 80% there given the CMS page table has a row per legal page; just need a `show_in_footer` boolean and the footer reads it. Lets you remove Impressum + Datenschutz from your gallery's footer if your jurisdiction doesn't require them. Small PR.

2. **Socials in branding.** New optional fields on Branding settings: Facebook URL, Instagram URL, WhatsApp URL/phone, X/Twitter URL. Footer renders icons + links for any that are set, hides if empty. Bigger PR — needs UI for the Branding tab too.

3. **(Maybe)** A small "Custom HTML footer" textarea — but I'd push back on this. Custom HTML in user-settable content is an XSS minefield, and once you have #1 + #2 above you don't need the escape hatch for the common cases. If we ever do this, it'd need careful sanitization (DOMPurify allowlist).

Want me to start with #1 (the per-page footer toggle)? That alone solves your "remove these pages" half. We can scope #2 (socials) as a follow-up — happy to take it but it's its own UI work and i18n round.

the-luap:
Implemented in #450 (targeting `beta`). Both parts of your request are covered:

[...]

PR #461, opened 2026-05-11
fix(branding): socials + promo round-trip from DB to form (#460)

### LINKED ISSUE #460: [BUG] Socials
socials update : 
They not visible in the admin panel if you save > change tab , go back to the same tab ,, they will be gone .
but can still see them in the Events pages .
Unless you press Save while they empty, they will be removed from footer.
Image

 _Originally posted by @Rekoo-PS in [#447](https://github.com/the-luap/picpeak/issues/447#issuecomment-4419470592)_

the-luap:
## Summary

Closes #460. Fix for the "socials disappear from BrandingPage and get wiped on next save" bug — root cause is a stale read mapper.

When the footer-overhaul fields landed in #441 / #440, the `BrandingSettings` interface was updated and the BrandingPage form now reads/writes them — but `formatBrandingSettings` (the function that maps the raw `app_settings` rows to the form shape on load) was not updated. So:

1. User saves socials → backend persists ✅
2. User reloads /admin/branding → form initialises from `formatBrandingSettings`, which doesn't read the new keys → socials appear empty in the form ❌
3. User saves any other field → form's empty socials get sent → backend writes empty strings → DB wiped, public footer goes blank on next render

The asymmetric "visible to galleries, gone from admin" behaviour comes from step 2: the public footer keeps rendering the saved values until step 3 wipes them.

## Fix

Add the seven missing read mappings to `formatBrandingSettings`:
- `branding_facebook_url`, `branding_instagram_url`, `branding_whatsapp_url`, `branding_twitter_url`, `branding_youtube_url`
- `branding_promo_markdown`, `branding_promo_position`

Same pattern as the existing fields — `|| ''` for strings, narrowed enum for `promo_position`.

## Test plan

- [ ] On an install with socials configured: visit /admin/branding → socials populated in the form (regression check).
- [ ] Save the form without changing socials → reload → socials still there.
- [ ] Clear a social URL → save → reload → field shows empty (intentional empty save).
- [ ] Promo banner content + position round-trip the same way.

## Smoke

Local smoke gate is currently 3/13 red on `origin/beta` itself (`CustomerDashboardBrandingCard` adds a competing `Save changes` button on /admin/branding that breaks two specs, and `CustomerAccountPicker` has a Rules of Hooks violation that crashes /admin/events/new — both flagged on PR #458). Pushed with `PUSH_SKIP_E2E=1`. The failures are not caused by this PR — verified by snapshot check, the relevant page state matches plain beta.

Targets `beta`. Closes #460.

### DIFF
--- a/frontend/src/services/settings.service.ts
+++ b/frontend/src/services/settings.service.ts
@@ -323,7 +323,20 @@ export const settingsService = {
       login_logo_frame_enabled: this._parseBoolean(rawSettings.branding_login_logo_frame_enabled, true),
       login_logo_size: ['small', 'medium', 'large', 'xlarge'].includes(rawSettings.branding_login_logo_size)
         ? rawSettings.branding_login_logo_size
-        : 'medium'
+        : 'medium',
+      // Footer overhaul (#441 + #440 / #460). These were added to the
+      // BrandingSettings interface but not to the read mapper, so the
+      // BrandingPage form initialised them as empty strings on every
+      // load. The next save then sent the empty form back to the
+      // backend and wiped the saved values from the DB — even though
+      // the gallery footer kept rendering them until the next save.
+      facebook_url: rawSettings.branding_facebook_url || '',
+      instagram_url: rawSettings.branding_instagram_url || '',
+      whatsapp_url: rawSettings.branding_whatsapp_url || '',
+      twitter_url: rawSettings.branding_twitter_url || '',
+      youtube_url: rawSettings.branding_youtube_url || '',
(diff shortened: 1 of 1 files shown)


the-luap:
Rebased onto post-#458 beta. Trivial conflict in `settings.service.ts` — both PRs added independent field blocks to the same place in `formatBrandingSettings` (Luca's login-logo knobs vs my socials/promo). Kept both, comma between them. TS clean. Force-pushed with `--force-with-lease`.

PR #464, opened 2026-05-11
feat(email-templates): categorise + sub-section + Feature off chip (#354 follow-up)

Luca-Timo:
Reorganises the admin Email Templates editor so it's ready to host the upcoming CRM-area templates (calendar, quotes, bills, messaging — all from #354) without further structural changes. Today's templates are sorted into the new shape; new templates just declare their category and they land in the right bucket automatically.

## What changes for admins

The Templates sidebar (Settings → Email → Templates) is now grouped:

```
CORE
  Galleries         gallery_created · expiration_warning · gallery_expired · archive_complete
  Admin accounts    admin_invitation · admin_password_reset
  Backup & restore  database_backup_* · restore_* · backup_*
  System updates    version_update_available · version_update_test

CUSTOMERS
  customer_invitation
  customer_password_reset
```

**Feature off tag.** Every template now carries an optional `feature_flag` link. When the linked flag is OFF in Settings → Features, the template stays fully visible and editable — admins can prep copy ahead of a feature launch — but it gets a small amber **"Feature off"** chip next to its row so the dormant state is obvious. Today this applies to both customer-portal templates when `customerPortal` is disabled. Future calendar/quotes/bills templates will get the same treatment automatically.

**All templates always visible.** The categorisation is presentational only. Disabled-feature templates aren't hidden — that's a deliberate UX choice so admins can edit copy before flipping a feature on.

### Schema (migration 098)

Three new columns on `email_templates`:
- `category` (default `'core'`) — top-level grouping
- `subcategory` (nullable) — second-level grouping inside `core` (which is busy enough — 14 templates — to deserve its own internal headers)
- `feature_flag` (nullable) — name of the flag whose `false` state should chip out the row

Backfilled for every existing template_key. Empty categories hide automatically (calendar / quotes / billing render the moment a template with that category gets added — no UI change needed).

### Translation auto-fill (migrations 099 + 100)

The translations table had en/de coverage for ~11 templates and en/de/nl/pt/ru for the four gallery templates. Migration 099 fills the gaps — every template now ships with en/de/nl/pt/ru/fr rows.

Two templates (`customer_password_reset` from migration 092, `version_update_test` from migration 087) were inserted AFTER migration 075 — they had their EN/DE content in legacy `subject_en` / `body_html_en` / `subject_de` / `body_html_de` columns but zero rows in `email_template_translations`, so the admin Templates editor showed them as empty (`0/6`) until manually saved. Migration 099 seeds the full six-locale set for both, mirroring the original curated EN/DE copy.

### Why migration 100 exists

Knex tracks migrations by filename. While iterating on 098/099 during this PR, the migrations went out to a beta deploy in an earlier shape, then got recorded in `knex_migrations`. Edits to those files don't re-run. Migration 100 is an append-only backfill that re-applies the canonical metadata + missing translation rows so already-deployed instances catch up. Safe to re-run; safe on fresh installs (no-ops when values already match).

## Files

- `backend/migrations/core/098_add_email_template_category.js` — new columns + backfill
- `backend/migrations/core/099_seed_missing_email_template_translations.js` — fills missing locale rows for 16 templates
- `backend/migrations/core/100_backfill_email_template_subcategory.js` — append-only re-backfill (see above)
- `backend/src/routes/adminEmail.js` — GET endpoints return the new fields
- `frontend/src/services/email.service.ts` — `EmailTemplate` type extended
- `frontend/src/pages/admin/EmailConfigPage.tsx` — sidebar grouping + "Feature off" chip rendering
- `frontend/src/i18n/locales/{en,de,fr,nl,pt,ru}.json` — category labels, sub-category labels, "Feature off" chip + tooltip

## Translation note

**en + de hand-translated.** **nl + pt + ru + fr machine-translated** (i18n strings + ~60 template translation rows). Flagging for native-speaker review before next stable release per project convention. Templates with hand-tuned EN/DE in the original migrations (gallery_*, admin_*, customer_invitation) keep those untouched — only the missing locales were generated.

## Test plan

- [x] `docker compose pull && up -d` runs migrations 098/099/100 in sequence on a beta with the original customer-portal data → `email_templates` rows show the new category/subcategory/feature_flag values; `email_template_translations` has 6 rows per template
- [x] Settings → Email → Templates renders **Core** with **Galleries / Admin accounts / Backup & restore / System updates** sub-headers, plus **Customers** with both customer templates
- [x] Toggle `customerPortal` off in Settings → Features → both customer templates show the "Feature off" chip but remain selectable + editable
- [x] `customer_password_reset` and `version_update_test` no longer show `0/6` — both have 6/6 with editable content in every language tab
- [x] `npx tsc --noEmit` clean, backend Jest unit tests green
- [x] Migration 100 is idempotent — re-running is a no-op

Forward-compatibility: when the calendar / quotes / bills / messaging features ship, their email templates just declare `category` + `feature_flag` at seed time and they slot into the new shape automatically.


the-luap:
Reviewed end-to-end. Clean PR — the categorise + sub-categorise + feature-flag-link shape is exactly the right ground-floor for the upcoming CRM templates, and you've kept the data-model strictly additive so a forward-compat row never disappears from the UI.

**Verified clean**

- ✅ **Migration 098** — schema additions are properly idempotent (`hasColumn` checks before each `alterTable`), the `TEMPLATE_METADATA` map keyed by `template_key` won't accidentally update a row that's been renamed, and the column comment block at the top is doc-grade. Default `category='core'` on the column itself means an install that doesn't even hit the backfill loop still has a sane value for new rows.
- ✅ **Migration 099** — idempotency by `(template_id, language)` lookup before insert is the right shape; importantly, an admin-edited translation isn't ever overwritten because the row exists. 60+ translation rows seeded with the variable placeholders preserved so `safeTemplateReplace` keeps working — appreciated.
- ✅ **Migration 100** — pragmatic. The note up top is honest about why it exists (knex-tracked migrations can't re-run after edits) and the implementation is correctly append-only + idempotent. Defensive `columnInfo()` check before writing covers the "didn't run 098 first" scenario cleanly.
- ✅ **`adminEmail.js`** — `||` fallbacks on `category`/`subcategory`/`feature_flag` mean older installs keep working without a hard dependency on the new columns. Symmetric on both list + detail endpoints.
- ✅ **`EmailConfigPage.tsx`** — forward-compat is on point: unknown categories fall into 'core', empty categories hidden so an admin doesn't see a section header with no body, templates with unrecognised subcategory go into a trailing 'other' bucket. The "Feature off" amber chip + tooltip render side-by-side with the existing `n/6` translation count without crowding. Backend tests still green (`emailProcessor.safeTemplateReplace` + `htmlToText`).
- ✅ **i18n keys** present in all six locales (`email.categories.*`, `email.subcategories.*`, `email.featureOff`, `email.featureOffTooltip`).

**One process note (non-blocker)**

The migration-100-to-fix-098/099 pattern is the correct repair when migrations have already been deployed — knex-tracked filenames can't re-run. But this is the second time in two weeks where in-flight migration edits have leaked to beta and needed a follow-up patch (cf. the `customer_password_reset` / `version_update_test` post-075 inserts you mention in the PR body). Worth a habit shift: **keep migrations local until they're truly final**, even if it means re-merging a beta release later. Saves the "compensation migration" overhead and keeps `knex_migrations` accurately reflecting a single canonical schema. Doesn't change anything about THIS PR — just flagging the trend.

**Translation flag**

`nl / pt / ru / fr` machine-translated for both ~60 template content rows AND the new UI strings. That's a substantial copy surface — fine to soak on beta and invite native-speaker corrections through the usual channel, just want to make sure we don't promote to main with unreviewed marketing-tone copy in customer-facing emails like `customer_invitation` and `customer_password_reset`. Worth a deliberate "review pending" tag on those two specifically before the next stable cut.

**Minor**

PR title is `Feat/email templates reorg` (autogenerated from branch name). Worth retitling to something like `feat(email-templates): categorise + sub-section + Feature off chip (#354 follow-up)` so the changelog reads cleanly when release-please picks it up.

**Approved ✅** — merge whenever you're ready. Solid forward-compat groundwork for the CRM-area templates landing later.

PR #470, opened 2026-05-12
feat(customers): "Manage galleries" dialog with immediate access revocation + section reorder + portal-flag revert

Luca-Timo:
Four related changes that all landed while iterating on the customer detail page.

1. **Reordered the section tiles** so the page reads in the order an admin actually consumes it.
2. **New "Manage galleries" dialog** with immediate per-gallery access revocation when a customer is unassigned.
3. **Session-refresh hardening** — customer no longer gets logged out on transient errors.
4. **Reverted the customer-portal kill-switch** — the feature flag is back to a UI-visibility toggle. Real-time access changes happen through per-record tools that don't risk locking every customer out on a misclick.

## 1. Section reorder

Cards on the customer detail page now flow as:

```
Account            who they log in as
Personal info      name + contact
Internal notes     admin context first (was buried below billing)
Assigned events    what they're tied to (also was below billing)
Address & billing  the paperwork section
Customer features  per-portal toggles
Account actions    destructive admin operations last
```

The previous order put Notes + Events below the much-longer billing card, which meant the two things admins most often skim when opening a customer record (their internal notes + which galleries they're tied to) required a scroll. Reorder is presentational — no logic, mutation or i18n touched.

## 2. "Manage galleries" dialog

Adds a way for admins to manage which galleries a single customer can see, replacing today's flow where you had to click into each event individually and tick the customer picker. Lives on the Assigned events card.

New **Manage galleries** button on the *Assigned events* card opens a modal:

- Currently-assigned galleries render as removable chips
- Search box below filters every event in the system (admin-side `/admin/events` search, debounced 200ms)
- Click a result to add it as a chip; click the X on a chip to remove
- Save replaces the customer's entire assignment list in one round-trip
- Toast surfaces the diff: *"Assignments updated: 2 added, 1 removed"*

UX detail: the search query is preserved after adding an event — common case is a search returning multiple matches (e.g. "Smith" → engagement + wedding) that you want to add in sequence. An inline X clear button on the search input wipes the query when you're ready to start a new search.

The button is disabled for deactivated customers — their login is off anyway, reactivate first before planning their access.

### Access revocation (the important bit)

The previous flow (event detail → customer picker) only removed the junction row; any gallery JWT the customer had already minted via `/api/customer/events/:slug/access-token` kept working until its 24h expiry. That's no longer the case here.

When the customer obtains a gallery JWT through the customer-portal handoff, the payload carries `via:'customer'` + `customerId`. `verifyGalleryAccess` (the middleware behind every gallery + secure-image endpoint) now re-checks `event_customer_assignments` on any request carrying those claims. Remove the row → customer's very next request 403s with `code: 'CUSTOMER_ASSIGNMENT_REVOKED'`. No token blacklist, no enumeration, no Redis invalidation — just a query against the same junction table the admin just wrote to.

This only affects customer-minted JWTs. Per-event-password JWTs (no `via` claim) are untouched.

## 3. Session-refresh hardening

`CustomerAuthContext.refreshSession` used to clear the customer's state on **any** error from `/api/customer/auth/session` — network blip, transient 5xx, request abort, anything. That made the customer's tab fragile to perfectly normal connectivity hiccups.

Now `customerService.session()` distinguishes 401 (definitive "you're not authenticated" → return null → clear state) from any other error (transient → rethrow → caller keeps existing state). `refreshSession` honours that contract and only clears state on null, never on throw. The next focus/visibility tick retries; if the customer really is unauthenticated, that next try will see the 401 and clear properly.

Concrete win: the customer no longer gets bounced to `/customer/login` if their network skips a beat while the admin is saving an unrelated change.

## 4. Portal-flag kill-switch revert

PR #458 wrapped `/api/customer/*` and `/api/admin/customers/*` in `requireCustomerPortalEnabled` middleware — toggling `customerPortal` OFF in Settings → Features returned 410 on every customer endpoint. The intent was "hard kill", the practice was "one misclick in Features tab logs every paying customer out". Reverting.

The feature flag is back to its original role:

- **Frontend:** `RequireFeature` guards + `AdminSidebar` visibility hide the admin Clients section when the flag is off.
- **Backend:** no route-level gate. Admin endpoints stay protected by `adminAuth` + `customers.*` permissions; customer endpoints stay protected by `customerAuth` + `is_active` checks.

For close-to-realtime access changes the codebase already has dedicated, surgical tools:

| Action | Mechanism | Real-time? |
|---|---|---|
| Block one customer's access to one gallery | "Manage galleries" dialog → removes `event_customer_assignments` row → `verifyGalleryAccess` re-checks on every request | Instant |
| Block one customer entirely | Deactivate → `customer_accounts.is_active=false` + bumps `password_changed_at` → `customerAuth` rejects all their JWTs | Instant |
| Toggle customer-side calendar/quotes/bills for one customer | Per-customer feature toggles on customer detail page | Instant (≤60s session refresh) |
| Force re-login for one customer | "Send password reset email" → bumps `password_changed_at` on apply | On apply |

The `requireCustomerPortal` middleware file is deleted in this PR (no other consumers).

### Backend

- `customerAccountsService.setAssignmentsForCustomer(customerId, eventIds, adminId)` — inverse of `setAssignmentsForEvent`. Diffs against current state, deletes removed rows, inserts added rows after validating against `events.is_archived = false`. Returns `{ added, removed }`.
- `PUT /api/admin/customers/:id/events` — body `{ event_ids: number[] }`, gated by `customers.create` permission (same tier as invite/deactivate).
- `verifyGalleryAccess` — adds the `via === 'customer'` assignment re-check. Wrapped in the same `withRetry` shape as the other reads in that file so transient DB blips don't bounce a legitimate session.
- `server.js` + `customerAuth.js` — restored the original "flag is UI-only" comments. `requireCustomerPortal` middleware deleted.

### Frontend

- `customerAdminService.setEvents(id, eventIds)` — wraps the PUT.
- `AssignedEventsDialog` — new modal component. Multi-select autocomplete pattern (mirrors `CustomerAccountPicker`'s shape on the inverse direction). React Query invalidates `['admin-customer', id]` + `['admin-customers']` on success so the parent + list view stay fresh.
- `CustomerDetailPage` — section reorder + Manage galleries button + dialog mount on the Assigned events card.
- `customer.service.ts` — `session()` returns `null` only on explicit 401; rethrows everything else.
- `CustomerAuthContext` — `refreshSession` keeps existing state when `session()` throws, clears only on null.

## Files

```
backend/src/services/customerAccountsService.js
backend/src/routes/adminCustomers.js
backend/src/middleware/gallery.js
backend/src/middleware/requireCustomerPortal.js             (deleted)
backend/src/routes/customerAuth.js                          (comment restoration)
backend/server.js                                           (middleware removal + comment)
frontend/src/contexts/CustomerAuthContext.tsx
frontend/src/services/customer.service.ts
frontend/src/services/customerAdmin.service.ts
frontend/src/components/admin/AssignedEventsDialog.tsx      (new)
frontend/src/pages/admin/CustomerDetailPage.tsx             (reorder + dialog mount)
frontend/src/i18n/locales/{en,de,fr,nl,pt,ru}.json
```

## Translation note

en + de hand-translated. **nl + pt + ru + fr machine-translated** for the ~15 dialog strings — flagging for native-speaker review before the next stable cut per project convention.

## Test plan

- [x] Customer detail page renders cards in the new order: Account → Personal → Notes → Events → Billing → Features → Actions
- [x] Open a customer's detail page → "Manage galleries" button visible on Assigned events card, disabled when customer is deactivated
- [x] Modal opens; currently-assigned events shown as chips, search returns matching events
- [x] Adding an event keeps the search query so additional matches can be picked from the same result list; the added event drops out of the dropdown automatically; the inline clear button wipes the query
- [x] Save shows the added/removed diff toast and refetches the detail page
- [x] Customer with an open `/gallery/<slug>` tab on a now-removed event: next API call returns 403 `CUSTOMER_ASSIGNMENT_REVOKED` (confirmed against secureImages + galleryPhotos endpoints)
- [x] Per-event password JWTs (`via` claim absent) are untouched — guest with a normal gallery cookie still works after the admin removes a customer assignment that happens to be on the same event
- [x] Re-adding the assignment + customer re-mints via `/api/customer/events/:slug/access-token` → works again
- [x] Archived event ids in the request body are silently ignored by the service (logged, not 4xx'd) so a stale tab in the dialog can't error the save
- [x] Toggling `customerPortal` OFF in Settings → Features hides the admin Clients section + sidebar entry but does NOT log existing customers out and does NOT 410 their `/customer/*` requests
- [x] Forced 5xx on `/customer/auth/session` (e.g. backend restart mid-poll) no longer logs the customer out — the next successful refresh restores state silently
- [x] `npx tsc --noEmit` clean, backend Jest unit tests green
- [x] Permission gate: a non-`customers.create` admin gets 403 on the new endpoint

## Out of scope

- No bulk "add this customer to N events at once" entry point on the *events* side — that's the existing `CustomerAccountPicker` on event create/edit, untouched.
- No invitation flow change. This UI is for existing accepted customers; the invite path stays where it is.
- No customer-side notification when assignments change. The customer just sees the dashboard list update on next refresh.
- No `Cache-Control: no-store` hardening on customer endpoints yet — flagged during this PR's review (browsers cached a 410 from the now-reverted kill-switch and kept serving it). Worth a small follow-up so any future error response from these endpoints can't get stuck in browser cache, but defer to keep this PR focused.


the-luap:
Reviewed end-to-end. Substantial PR, all four documented themes land cleanly, plus a fifth one I'll mention below. The "Manage galleries" dialog + the `verifyGalleryAccess` re-check together replace the global kill-switch with something genuinely better — surgical, real-time, no Redis or blacklist machinery.

**Verified clean**

- ✅ **Section reorder** (`592fa1e`) — purely presentational, no logic touched. Notes + Events promotion above the long billing card is the right call.
- ✅ **`setAssignmentsForCustomer`** (`5377b88`) — symmetric with `setAssignmentsForEvent`, properly diffs current vs target, filters archived events server-side, transactional, returns the diff for the toast. Per-customer email queue runs post-commit + swallow-and-warn so a notification failure can't roll back the assignment write.
- ✅ **`verifyGalleryAccess` customer-assignment re-check** (`55a5846`) — the load-bearing piece. Only fires for tokens with `via === 'customer' && customerId` so per-event-password JWTs are untouched (no perf cost for the dominant case). Wrapped in `withRetry`, returns a specific `CUSTOMER_ASSIGNMENT_REVOKED` code, and the inline comment honestly states "no token-blacklisting machinery required" — the simplest correct design.
- ✅ **`AssignedEventsDialog`** (`6d1af7a` + `d0ad987`) — hooks all run before the conditional return (no #458-style Rules of Hooks issue), the "keep query after add" UX detail + explicit clear button are nice touches, deactivated customers gated correctly.
- ✅ **Session-refresh hardening** (`9e418c7`) — clear three-way contract (object/null/throw) in `customer.service.session()`, honoured in `CustomerAuthContext.refreshSession`. Doc-comments on both sides spell out why the previous "swallow all errors → null" was the wrong default. Real fix for a real user-facing bug.
- ✅ **Portal-flag revert** (`3f44193`) — middleware file deleted cleanly (no stragglers found), `server.js` + `customerAuth.js` comments rewritten honestly ("a stray click in the Features tab locks every paying customer out at once"). The per-record alternatives table in the PR body is exactly the kind of design rationale future-me will be grateful for.
- ✅ **Customer assignment notification email** (`c02c947`) — migration 101 slots into the 098/099/100 category/subcategory/feature_flag scheme, idempotent, en + de hand-translated, digest-style "one email per save, not one per gallery" is the right pattern, archived/expired events filtered out before the customer sees them.
- ✅ Customer auth Jest suites still green (23/23) post-revert.

**Three notes**

1. **PR description undersells the diff (again).** Body lists four themes; the actual diff has five — the customer-notification email (`c02c947`, migration 101, the `notifyCustomerOfNewAssignments` function, ~270 lines including six locales of template content) is a meaningful feature, not a tiny add-on. Same gentle reminder as on #464 — surfacing it explicitly in the description helps future-me and the changelog pick it up cleanly.

2. **The two new server-side primitives have no test coverage.** Both `setAssignmentsForCustomer` (the writer) and the `verifyGalleryAccess` customer-assignment re-check (the reader) are security-relevant — if either has a subtle bug, customer assignments don't actually revoke access, which is the entire promise of the new dialog. Worth at minimum:
   - A unit test for `setAssignmentsForCustomer` covering the diff math + the archived-event filter.
   - A route test asserting that removing an assignment causes the customer's next `GET /api/gallery/<slug>/info` (with their gallery JWT) to 403 with the expected code.
   Same shape as `authSession.symmetry.test.js` did for the #350 token-type checks. Not a merge blocker — but a follow-up worth scoping before the next stable cut. This codepath is the entire promise of the new UI.

3. **`Cache-Control: no-store` is worth doing.** You called this out in the "Out of scope" section — agreed it's a follow-up, but worth a tracking issue. Browsers that cached a 410 from the now-reverted middleware (PR #458 era) might serve stale responses until a hard-refresh. Low blast radius today, but a documented gotcha for anyone upgrading directly through both PRs.

**Translation flag** — your nl/pt/ru/fr machine-translated note covers both the dialog strings and the email template body. Same convention as recent PRs, fine to soak on beta.

**Design history**

This is the third design pivot on the same flag-gate surface (my #455 comment → your #458 middleware → this revert). The current shape is the right one and matches the original intent. Worth being explicit so a future reader doesn't read it as thrashing: the kill-switch experiment was reasonable to try, the real-world UX showed it failed (one misclick locks everyone out), the per-record tools you added in this PR achieve the same goal more safely. We landed where we should have started — that's iteration working as designed.

**Approved ✅** — merge whenever you're ready. Strong PR.

Issue #484, opened 2026-05-14
[BUG] Installation fail and documentaiton mismatch


MrGabri:
**Describe the bug**
I found two related issues during clean docker installation

A) Backend restart loop on fresh install

Following the manual install documentation on a clean system, the backend immediately enters a restart loop. With the default `.env` (where DB_NAME=picpeak_prod), the backend fails because it's looking for a database named picpeak, not picpeak_prod. Changing DB_NAME to picpeak gets past that, but then migrations fail because of multiple migration errors.
Note that this is a clean install with no prior state, wiping volumes and starting over produces the exact same result.

B) Setup script and documentation does installation very differenty

The `picpeak-setup.sh` script and the manual install documentation are inconsistent, and it's unclear which one is the intended architecture.
In the setup script, the frontend is served directly through the backend (as much as I understand),
there's a separate workers container,
and migrations are explicitly called after startup.
The manual docs describe a separate frontend container, make no mention of a workers container, and state that migrations run automatically on startup. 

**To Reproduce**
1. Clone the repository and copy `.env.example` to `.env`, leaving defaults as-is.
2. Run `docker compose up -d`
3. Observe the backend restart loop — it cannot find the database
4. Change `DB_NAME=picpeak`, run `docker compose down -v`, then `docker compose up -d` again
5. Observe migration failure: `relation "photos" does not exist`

**Expected behavior**
A clean install following the documentation should result in a working instance. 

**Environment (please complete the following information):**
 - OS: Debian 6.1.170-3
 - PicPeak Version: v3.43.1
 - Deployment Method: Docker Compose or Setup script
 - Database: PostgreSQL 15

**Logs**
Please include relevant logs:
```
# DB not found logs:
FATAL:  database "picpeak" does not exist

# Migration failure logs:
Migration failed: error: alter table "events" add constraint "events_hero_photo_id_foreign" foreign key ("hero_photo_id") references "photos" ("id") on delete SET NULL - relation "photos" does not exist
    at Parser.parseErrorMessage (/app/node_modules/pg-protocol/dist/parser.js:285:98)
    at Parser.handlePacket (/app/node_modules/pg-protocol/dist/parser.js:122:29)
    at Parser.parse (/app/node_modules/pg-protocol/dist/parser.js:35:38)
    at Socket.<anonymous> (/app/node_modules/pg-protocol/dist/index.js:11:42)
    at Socket.emit (node:events:519:28)
    at addChunk (node:internal/streams/readable:561:12)
[...]

PR #488, opened 2026-05-14
fix(install): silence pg healthcheck noise + drop legacy workers container (#484)

the-luap:
## Summary

Closes most of #484 — three install-experience bugs that compounded into MrGabri's "fresh install fails" report. The install was actually working; log noise scared the reporter into changing settings and re-creating the stack, which then hit a tainted-state migration error.

### 1. postgres healthcheck noise (the actual root cause of #484)

Both compose files used:
\`\`\`yaml
test: ["CMD-SHELL", "pg_isready -U \${DB_USER:-picpeak}"]
\`\`\`

\`pg_isready -U <user>\` without \`-d\` defaults to probing a database whose name matches the user. With the default \`DB_NAME=picpeak_prod\`, every healthcheck interval logged \`FATAL: database "picpeak" does not exist\` into postgres logs even though the install was working correctly. MrGabri saw the FATAL → assumed broken → restarted with \`DB_NAME=picpeak\` → got into tainted state → second migration attempt hit \`relation "photos" does not exist\` → opened the issue.

Fix: pin \`-d\` to \`\${DB_NAME}\` so the probe hits the real database. Silences the noise. Applied to both \`docker-compose.production.yml\` and the inline compose generated by \`scripts/picpeak-setup.sh\`.

### 2. backend `unhealthy` in `docker ps` even when /health is fine

Both compose files used \`curl -f\` for the backend healthcheck, but \`backend/Dockerfile\` only installs \`dumb-init + postgresql-client + ffmpeg\` — no curl. \`docker ps\` showed the backend perpetually unhealthy. Switched to \`wget --no-verbose --tries=1 --spider\` to mirror what \`backend/Dockerfile\`'s own \`HEALTHCHECK\` already does. Compose, image-side healthcheck, and \`docker ps\` now agree.

### 3. Stale separate `workers` container in setup-script-generated compose

\`scripts/picpeak-setup.sh\` still generated a second container running \`npm run workers\` alongside the backend, but workers (fileWatcher, expirationChecker, emailQueueProcessor, backgroundProcessor, webhookWorker) are started by \`server.js\` in-process now — the same script's systemd path already noted this consolidation (line ~895). The duplicate container caused two file watchers and two expiration checkers to compete for the same DB rows.

Removed from the generated compose. Install + upgrade paths now stop and rm any pre-existing \`picpeak-workers\` container so legacy installs clean up automatically on next \`picpeak-setup.sh\` run.

## Out of scope (issue B)

The bigger doc/setup-script architecture mismatch — script generates build-from-source compose with no frontend container; \`docker-compose.production.yml\` uses prebuilt images with a separate frontend container — deserves its own design pass to pick a canonical install path and align. Worth a separate issue/PR; out of scope here.

## Test plan

- [ ] Fresh install with default \`.env\` (\`DB_NAME=picpeak_prod\`) → no \`FATAL: database "picpeak" does not exist\` in postgres logs.
- [ ] \`docker ps\` shows backend as \`(healthy)\` (was \`(unhealthy)\` before).
- [ ] Re-running \`picpeak-setup.sh\` on a host with a leftover \`picpeak-workers\` container removes it.
- [ ] Fresh setup-script install no longer creates \`picpeak-workers\`; only \`picpeak-postgres / picpeak-redis / picpeak-backend\` are running, and the in-process workers (fileWatcher, expirationChecker, etc.) start as before.

Targets \`beta\`. Closes the install-experience portion of #484.

### DIFF
--- a/docker-compose.production.yml
+++ b/docker-compose.production.yml
@@ -15,7 +15,13 @@ services:
       - picpeak-network
     restart: unless-stopped
     healthcheck:
-      test: ["CMD-SHELL", "pg_isready -U ${DB_USER:-picpeak}"]
+      # `pg_isready -U <user>` without -d defaults to probing a database
+      # whose name matches the user — postgres then logs constant
+      # `FATAL: database "picpeak" does not exist` even though the
+      # actual DB is `picpeak_prod`. Pinning -d to DB_NAME makes the
+      # probe hit the real database and silences the log noise that
+      # made #484's reporter think the install was broken.
+      test: ["CMD-SHELL", "pg_isready -U ${DB_USER:-picpeak} -d ${DB_NAME:-picpeak}"]
       interval: 10s
       timeout: 5s
       retries: 5
@@ -64,8 +70,13 @@ services:
         condition: service_healthy
     restart: unless-stopped
     healthcheck:
-      # Backend exposes /health on internal port 3000
-      test: ["CMD", "curl", "-f", "http://localhost:3000/health"]
+      # Backend exposes /health on internal port 3000.
+      # The backend image only ships wget (Alpine base) — using curl
+      # here makes `docker ps` show the container as `unhealthy`
+      # indefinitely even when /health responds. Mirrors the wget-based
+      # HEALTHCHECK already declared in backend/Dockerfile so docker
+      # compose, plain `docker run`, and `docker ps` all agree.
+      test: ["CMD", "wget", "--no-verbose", "--tries=1", "--spider", "http://localhost:3000/health"]
       interval: 30s
       timeout: 10s
       retries: 3
--- a/scripts/picpeak-setup.sh
+++ b/scripts/picpeak-setup.sh
@@ -518,6 +518,17 @@ EOF
         setup_ssl_docker "$app_dir"
     fi
 
+    # Clean up the legacy picpeak-workers container from prior installs
+    # (workers are now in-process — see create_docker_compose_file).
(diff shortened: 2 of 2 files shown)


the-luap:
Bundled in two more install-experience fixes from the same #484 surface (commit \`d4155c4\`):

**3. Dropped the racy `docker compose exec -T backend npm run migrate`.** The backend container's `wait-for-db.sh` already runs `npm run migrate:safe` on startup. The script was racing it with a separate (and non-safe) `npm run migrate`, which is the most likely actual mechanism behind #484's "relation 'photos' does not exist" — partial schema visible to one of the two parallel migrators. Replaced with a bounded wait for the backend container's Docker healthcheck to go green (which only happens after wait-for-db.sh finishes its migration pass). Applied to both `install_docker` and `update_docker_installation`.

**4. Added the missing frontend container** to the script-generated compose. The script previously generated a postgres + redis + backend stack with no frontend at all (backend on host port 3001 with the API directly exposed) while `docker-compose.production.yml` ships a postgres + redis + backend + frontend stack (nginx `/api` proxy on host port 3000). MrGabri's issue B was directly this. Both compose shapes are now equivalent.

Remaining piece of issue B — picking ONE canonical install path (build-from-source script vs. prebuilt-image production compose) and deprecating the other — is a deployment-strategy call worth its own design pass. Both paths now produce architecturally-equivalent stacks regardless.

PR scope grew but the cluster of fixes all trace to #484 root causes. Easier to review as one coherent install-experience repair than four small PRs.

PR #501, opened 2026-05-16
fix(i18n): settings page resets UI language to server default

filpgame:
## Description

Fixes a bug where navigating to the Settings page resets the admin UI language to the server-stored `general_default_language` value, overriding the language the user had selected via the Language Selector.

In `useSettingsState.ts`, the `useEffect` that initialises state from the API response called `i18n.changeLanguage(settings.general_default_language)` on every query resolution. Since `general_default_language` is stored as `"en"` by default, any UI language other than English was silently reset every time the Settings page mounted.

The `general_default_language` setting is intended as the default language for public galleries — it should not control the admin UI language. The admin UI language is already correctly persisted via `localStorage` through `i18next-browser-languagedetector`.

Fixes # (no existing issue)

## Type of change

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## How Has This Been Tested?

**Steps to reproduce (before fix):**
1. Open the admin panel
2. Change the UI language to any non-English language (e.g. Português)
3. Click "Configurações" / "Settings" in the sidebar
4. Observe the UI language resets to English

**Verification (after fix):**
1. Set UI language to Português via the Language Selector
2. Navigate to Settings — language stays Português
3. Reload the page — language is restored from `localStorage` (i18next-browser-languagedetector)
4. Navigate between tabs within Settings — language unchanged

- [ ] Unit tests pass (`npm test`)
- [ ] Tested on Docker deployment
- [x] Tested on production-like environment

**Test Configuration**:
* Node.js Version: 18+
* Database: SQLite
* Browser: Chrome

## Checklist:

- [x] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] Any dependent changes have been merged and published

N/A — language state is not visible in a screenshot in a meaningful way.

## Additional Notes:

The `i18n` instance was also removed from the `useTranslation()` destructure and from the `useEffect` dependency array since it is no longer used in that hook. This eliminates a potential source of infinite re-renders (i18n object identity can change on language switch, which would previously re-trigger the effect and call `changeLanguage` again in a loop if the stored language differed).

### DIFF
--- a/frontend/src/features/settings/hooks/useSettingsState.ts
+++ b/frontend/src/features/settings/hooks/useSettingsState.ts
@@ -70,7 +70,7 @@ export interface SeoSettings {
 
 export function useSettingsState() {
   const queryClient = useQueryClient();
-  const { t, i18n } = useTranslation();
+  const { t } = useTranslation();
   const { updateUserProfile } = useAdminAuth();
 
   // Fetch settings
@@ -164,10 +164,6 @@ export function useSettingsState() {
   // Initialize settings from API
   useEffect(() => {
     if (settings) {
-      if (settings.general_default_language && settings.general_default_language !== i18n.language) {
-        i18n.changeLanguage(settings.general_default_language);
-      }
-
       setGeneralSettings({
         site_url: settings.general_site_url || '',
         default_expiration_days: toNumber(settings.general_default_expiration_days, 30),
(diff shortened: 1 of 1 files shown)


the-luap (approved):
Thanks for the contribution and for the clear write-up — the repro steps and the explanation of why `general_default_language` shouldn't drive the admin UI made the review trivial.

Nice catch on dropping `i18n` from the dep array too; that was a latent re-render hazard quite separate from the language reset itself.

Approving — will merge to `beta` and it'll ride out on the next beta release, then get promoted to `main` on the next stable cut.

PR #502, opened 2026-05-16
fix(categories): strip diacritics from auto-generated slugs

filpgame:
## Description

When creating or renaming a category with accented characters (e.g. Portuguese, French, Spanish), the slug generation silently drops those characters instead of converting them to their ASCII equivalents.

**Before:**
- `Decoração` → `/decorao`
- `Recepção` → `/recepo`
- `Família` → `/famlia`

**After:**
- `Decoração` → `/decoracao`
- `Recepção` → `/recepcao`
- `Família` → `/familia`

**Root cause:** The slug regex `/[^\w\s-]/g` strips any non-ASCII character because `\w` in JavaScript only matches `[a-zA-Z0-9_]`. Accented letters are entirely removed rather than transliterated.

**Fix:** Apply Unicode NFD normalization before the regex. NFD decomposes characters like `ã` into the base letter `a` plus a combining diacritical mark (`̀–ͯ`). A second replace then strips those combining marks, leaving only the ASCII base letter. The existing slug pipeline then handles the rest (lowercase, spaces → hyphens, etc.).

The fix is applied to both the POST (create) and PUT (update) handlers in `adminCategories.js`.

## Type of change

## How Has This Been Tested?

Verified with Node.js REPL:

```js
const slug = (name) => name
  .normalize('NFD').replace(/[̀-ͯ]/g, '')
  .toLowerCase()
  .replace(/[^\w\s-]/g, '')
  .replace(/\s+/g, '-')
  .replace(/-+/g, '-')
  .trim();

slug('Decoração')  // 'decoracao'
slug('Recepção')   // 'recepcao'
slug('Família')    // 'familia'
```

- [x] Tested on production-like environment (self-hosted LXC, SQLite)

**Test Configuration**:
* Node.js Version: v20.20.2
* Database: SQLite
* Browser: Chrome

## Checklist:

## Additional Notes:

This affects any language with diacritical marks: Portuguese (ã, ç, õ), French (é, è, ê), Spanish (ñ, á), German (ü, ö, ä), etc.

### DIFF
--- a/backend/src/routes/adminCategories.js
+++ b/backend/src/routes/adminCategories.js
@@ -56,7 +56,9 @@ router.post('/', adminAuth, requirePermission('settings.edit'), [
     const { name, slug, is_global = true, event_id = null } = req.body;
     
     // Generate slug if not provided
-    const categorySlug = slug || name.toLowerCase()
+    const categorySlug = slug || name
+      .normalize('NFD').replace(/[̀-ͯ]/g, '')
+      .toLowerCase()
       .replace(/[^\w\s-]/g, '')
       .replace(/\s+/g, '-')
       .replace(/-+/g, '-')
@@ -128,7 +130,9 @@ router.put('/:id', adminAuth, requirePermission('settings.edit'), [
 
     const updateData = {
       name,
-      slug: name.toLowerCase()
+      slug: name
+        .normalize('NFD').replace(/[̀-ͯ]/g, '')
+        .toLowerCase()
         .replace(/[^\w\s-]/g, '')
         .replace(/\s+/g, '-')
         .replace(/-+/g, '-')


the-luap (approved):
Thanks for the fix — clean diagnosis (NFD + strip combining marks is the right tool) and nice that you applied it to both POST and PUT.

Worth flagging for context: the same bug exists in a handful of other slug sites (`adminEvents.js`, `events.js`, `v1/events.js`, `adminPhotos.js`, `adminArchives.js`, `xmpGenerator.js`). Not your problem to fix in this PR — but if you're up for a follow-up that extracts a shared `slugify` util and converts the callers, that'd be a nice next contribution.

Approving — will ride out on the next beta release.

PR #503, opened 2026-05-16
fix(email): parse JSON-encoded language setting before using as locale

filpgame:
## Description

Emails always send in English regardless of the language configured in **Settings → General → Default Language**.

**Root cause:** `app_settings` stores all values as JSON-encoded strings (e.g. `"pt"` is stored as `\"pt\"` including the quotes). `getRecipientLanguage()` returned the raw `setting_value` without parsing it, so the language passed to `processTemplate()` was `'"pt"'` (with surrounding quotes) instead of `'pt'`. The translation lookup `WHERE language = '"pt"'` matched nothing and fell through to the English fallback.

**Fix:** Apply `JSON.parse()` with a try/catch fallback before returning the setting value — the same pattern used by `readSetting()` in `wrapEmailHtml()` and `getFrontendBaseUrl()` elsewhere in the codebase.

**Before:**
```js
if (langSetting && langSetting.setting_value) {
  return langSetting.setting_value; // returns '"pt"' with quotes
}
```

**After:**
```js
if (langSetting && langSetting.setting_value) {
  let lang = langSetting.setting_value;
  try { lang = JSON.parse(lang); } catch (_) {}
  if (typeof lang === 'string' && lang.trim()) return lang.trim(); // returns 'pt'
}
```

## Type of change

## How Has This Been Tested?

Verified by inspecting the DB value (`"pt"` with quotes) and tracing the lookup through `processTemplate()`. Confirmed emails now use the correct language translation after the fix.

- [x] Tested on production-like environment (self-hosted LXC, SQLite)

**Test Configuration**:
* Node.js Version: v20.20.2
* Database: SQLite
* Browser: Chrome

## Checklist:

## Additional Notes:

The same `JSON.parse` pattern is already used in `wrapEmailHtml()` (`readSetting` helper) and `getFrontendBaseUrl()`. This fix makes `getRecipientLanguage()` consistent with the rest of the codebase.

### DIFF
--- a/backend/src/services/emailProcessor.js
+++ b/backend/src/services/emailProcessor.js
@@ -116,7 +116,9 @@ async function getRecipientLanguage(email, eventId = null) {
       .where('setting_key', 'general_default_language')
       .first();
     if (langSetting && langSetting.setting_value) {
-      return langSetting.setting_value;
+      let lang = langSetting.setting_value;
+      try { lang = JSON.parse(lang); } catch (_) {}
+      if (typeof lang === 'string' && lang.trim()) return lang.trim();
     }
   } catch (error) {
     logger.error('Error fetching app settings language:', error);


the-luap (approved):
Thanks — sharp diagnosis. The JSON-encoded `app_settings` storage trips up anyone reading those values directly, and `getRecipientLanguage` was indeed the odd one out vs. `getSupportEmail`/`readSetting`/the template-var loop in the same file. Nice side-effect that empty-string values now fall through correctly to the email_configs fallback too.

Approving — three solid bugfixes in this batch (501/502/503), much appreciated. Will ride out on the next beta release.

PR #522, opened 2026-05-18
French translation

rpintodasilva:
## Description

Adds French translation to the project.

- adds i18n french locale
- adds language choice to picklist component in front interface
- adds language choice to picklist component in admin interface


the-luap:
Thanks so much for this contribution — and for the quality of the translation specifically. I compared your `fr.json` against the version currently on `beta` (added separately a couple of weeks ago by another contributor) and yours is meaningfully better: proper grammar, correct gender agreement (`Forte` vs `Robuste` for password strength, etc.), correct French terminology (`tirets bas` / `traits d'union` instead of "underscores"), and full natural sentences instead of the truncated machine-translated style currently on `beta`. Of the ~1,700 keys both files share, about 50% have different translations and yours read clearly better. This is exactly the kind of native-speaker review the existing French was missing.

**Quick branch logistics — could you retarget this PR from `main` to `beta`?** Here's the situation:

- We land all new features and language work on `beta` first, then promote to `main` on the next stable release cut (typically every few days).
- `beta` already has a French locale (the machine-quality one I described above) plus four sections that don't exist in your file yet: `clients`, `customer`, `customers`, and `photos` — these are recent customer-portal features that landed after `main` was last cut.
- `main` doesn't have French at all yet, but it'll get it (along with everything else on `beta`) within a few days.

What I'd ask:

1. **Retarget to `beta`.** GitHub UI → Edit → change base branch.
2. **Rebase against beta.** You'll see a conflict on `fr.json` since the file already exists. Resolve by keeping your translations for everything you have, and adding `beta`'s `clients`, `customer`, `customers`, and `photos` sections at the end. If you're willing to translate those four sections in the same proper-French style you've used elsewhere, that'd be the cherry on top — but it's optional, the current machine-translated content for those sections is at least functional until someone polishes it.
3. **You can drop the `permissions` section** you added — `grep`ing the frontend shows neither `permissions.insufficient` nor `permissions.viewOnly` is referenced anywhere in the code today, so those two keys aren't doing anything.

Once it's merged to `beta`, it'll ride out on the next beta release and then promote to `main` on the next stable cut — so your work will be on `main` within a few days regardless.

Let me know if the rebase logistics are unfamiliar and I'm happy to help. And thanks again — this is exactly the kind of contribution that's hard to come by (and impossible to fake with machine translation).

Issue #1, opened 2025-07-24
[DOCS]


dhoitink:
**What documentation needs improvement?**
Please specify which document or section needs attention:

- [X] DEPLOYMENT.md

Is it possible to run this locally, in stead of needing a HTTPS with FQDN ? 

This might sound strange, but for testing purpose it would be much easier.

Please add this to the documentation, or even better: in the .ENV file

Who is this documentation for?
- [X ] New users setting up PicPeak

the-luap:
@dhoitink did you check my latest update?
I cleaned up all the documentation files and added a lot of new information to deployment documentation.

Issue #530, opened 2026-05-19
CI: schema-diff test to catch initializeDatabase() vs migration-chain drift


the-luap:
Refined scope after a dry-run uncovered the actual drift hazard (see [issue #484 thread](https://github.com/the-luap/picpeak/issues/484#issuecomment-4489129298) for the original framing with @Ilya0527).

## Refined background

The original framing — `pg_dump` `initializeDatabase()` vs the migration chain, diff with `migra` — turned out to produce mostly noise. ~200 lines of differences, all expected: migrations add 35 new tables and incrementally add columns to the original 13 base tables over time. A CI test that flags every "yes a migration added a column" diff would be ignored within a week.

The actual architecture also doesn't quite match the original framing: `initializeDatabase()` isn't a parallel "fresh-install" path that diverges from migrations. It's invoked **by migration 001 itself**. So in any normal production install or upgrade, both `initializeDatabase()` AND the subsequent migrations always run sequentially — there's no real "bootstrap-alone" code path.

## The genuine drift hazard

A DB can land in a state where the modern bootstrap tables exist (`events`, `photos`, `photo_categories`, `cms_pages`, etc. — the post-`initializeDatabase()` 13-table state) but the `migrations` tracking table is empty. This isn't a normal install state — it happens when:

- A backup is restored that captured tables but not the `migrations` table (manifest divergence),
- Someone manually invoked `initializeDatabase()` outside the migration runner (recovery / debugging),
- The DB was moved between systems and the `migrations` table was not copied along.

In that state, `migrate:safe` detects the bootstrap tables, treats the DB as an "existing deployment", and runs the **legacy chain first**. `legacy/008_add_language_support_to_email_templates.js` renames `email_templates.subject` → `subject_en`. Then `core/029_add_backup_service_tables.js` (later in the chain) tries to insert email templates referencing the pre-rename column name — and fails with:

```
ERROR: null value in column "subject_en" of relation "email_templates"
violates not-null constraint
```

This is the real drift hazard. Fresh installs avoid it (they only run `core/*` and `core/059` handles the rename **after** `core/029` inserts). Real legacy upgrades avoid it (their `migrations` table already records that `legacy/008–028` ran historically).

## What the test does

Boot a fresh postgres, run `initializeDatabase()` directly (no migration runner), then run `migrate:safe`. The chain should:

1. Detect the modern bootstrap fingerprint and mark legacy migrations as applied,
2. Run only the core migrations (matching what a real fresh install runs),
[...]

PR #551, opened 2026-05-24
fix(email): event emails always sent in English ignoring system language

filpgame:
## Description

Event notification emails sent to clients (gallery created, expiry warning, etc.) were always sent in English, ignoring the **Settings → General → Language** setting.

### Root Cause

`events.language` was added with a DB-level `defaultTo('en')`. Since no INSERT path set the column explicitly, every event stored `language = 'en'`. The email processor's `getRecipientLanguage()` checks `event.language` as its **first priority**, so it always returned `'en'` before reaching `app_settings.general_default_language`.

### Fix

- **Migration 108**: resets all existing events with `language = 'en'` (DB-defaulted) to `NULL`; changes the column default to `NULL`
- **All five event INSERT paths** now set `language: null` explicitly — ensures correct behaviour even if the SQLite driver ignores the `ALTER COLUMN` default change
- `NULL` means "use system default" — `getRecipientLanguage` already falls through to `general_default_language` when `event.language` is falsy (no logic change needed in the email processor)

## Type of change

## How Has This Been Tested?

- Reviewed `getRecipientLanguage` priority chain in `emailProcessor.js` (lines 99–155)
- Traced all five `db('events').insert(...)` call sites and confirmed none set `language` explicitly before this fix
- Confirmed `email_template_translations` has `pt`, `nl`, `ru`, `de`, `en` rows for `gallery_created` (migration 075)
- Migration 108 verified against the same `alterTable` / `.alter()` pattern used in migration 073

**Config used for testing:** SQLite, Node.js backend, `general_default_language = 'pt'`

## Checklist

- [x] New and existing tests pass locally (Jest suite; no new test cases needed — the fix is a DB default + explicit null assignment)

## Additional Notes

The fix is backward-compatible: events that already had an explicit non-English language set are unaffected (only `'en'` rows are reset by the migration, and only because every `'en'` value came from the DB default — there is no UI for per-event language overrides).


the-luap:
Thanks for the fix, Felipe — the root-cause analysis is spot on, and `language: null` in all five INSERT paths is exactly the right surgical change.

Two things before I can merge though:

**1. Is the WhatsApp work in this PR intentional?**

The PR description scopes this as a single bug fix, but the diff is 34 files / +3886 lines — most of it is the WhatsApp Business API integration from #1 (migration `107_create_whatsapp_tables.js`, services, admin route, settings page, design docs, version bump to `4.0.0-beta`, CHANGELOG). I assume that happened because this branch was cut from your fork's `beta`, which still carries the closed #1 work, rather than from upstream `beta`.

If that was unintentional, could you rebase onto upstream `beta` and force-push? That would drop ~28 files of collateral and leave just the language fix (~6 files, ~50 lines). If it *was* intentional and you want to re-open the WhatsApp discussion as part of this PR, no worries — just let me know and I'll review it on those terms, but please update the title/description to reflect the wider scope.

**2. Migration 108 fails on fresh install + bootstrap upgrade**

Both install smoke checks are red. The error is:

```
❌ Migration failed: update "events" set "language" = $1 where "language" = $2
   - column "language" does not exist
```

The `hasColumn` guard is present, but it's checked *after* the UPDATE. On the modern bootstrap path (post-#530, which skips the legacy migration chain when it detects a clean fingerprint), `events.language` doesn't exist when 108 runs, so the UPDATE blows up before reaching the guard.

Fix is just to move the guard above the UPDATE:

\`\`\`js
exports.up = async function(knex) {
  if (!(await knex.schema.hasColumn('events', 'language'))) return;
  await knex('events').update({ language: null }).where('language', 'en');
  await knex.schema.alterTable('events', (table) => {
    table.string('language', 5).defaultTo(null).alter();
  });
};
\`\`\`

Same change for \`exports.down\`, since that one has the same pattern.

With those two cleaned up the PR is good to go from my side. Thanks again for chasing this — the \`getRecipientLanguage\` priority chain is a subtle one.

filpgame:
I've mistakenly put more unrelated code into this PR, I'm gonna close this and open another PR with the fix only

PR #555, opened 2026-05-26
feat: CRM module — quotes, contracts, invoices, hours, calendar, tax

Luca-Timo:
## Summary

Adds a CRM module on top of the existing photo platform: quotes →
contracts → invoices end-to-end with `deal_uuid` lineage, hour
logging, unified calendar, customer accounts, tax report. **All
new surfaces are off-by-default feature flags** — existing installs
see no UI change until opted in.

- 204 files, +47k / −0.7k
- 42 in-flight CRM migrations consolidated into one (`107_crm_consolidated`)
- CI added: Jest + Vitest run on every PR (was previously unrun)

## Commits

Reviewable by layer. Land in order or squash as you prefer.

| Commit | What |
|---|---|
| `60abe8c` | Consolidate migrations 102–143 + extract email seeds to self-heal services |
| `d543949` | Backend — services + routes + utils + tests |
| `a7e16e7` | Frontend — pages + services + components |
| `409d414` | i18n EN + DE (FR/NL/PT/RU are machine — need native review) |
| `e7db0bb` | Legal/financial "examples only" disclaimers |
| `88a6b34` | Fix: defer cross-table FKs in 107 (Postgres FK-order bug) |
| `b9cadf0` | Update upstream test mocks for new behavior |
| `482043f` | CI workflow runs Jest + Vitest on every PR |
| `3240137` | Schema-shape integration test (36 assertions) |

## Migrations

One consolidated migration replaces the 42-migration in-flight chain.
The Schema drift (#530) workflow caught a forward-reference FK bug
(`quotes.converted_contract_id` → `contracts(id)` declared inline
before `contracts` exists) — fixed by deferring both back-pointer
FKs to a post-create `ALTER TABLE`, same pattern as
`events.hero_photo_id`.

Migration 137 adds calendar time columns + `business_profile.timezone`,
backfilling `is_full_day=true` on existing rows so no operator
action is needed.

## Tests

- 533 backend + 40 frontend tests pass on every PR (new CI workflow).
- 6 pre-existing upstream-broken suites excluded inline (tracked as
  test-infra debt).
- New `crmSchema.test.js` pins the CRM table/column layout after
  migration 107 — catches drift like the FK bug above at the
  schema-shape level.

**Not covered (deferred):** HTTP route tests for adminQuotes /
Contracts / Invoices / Calendar, full lineage walk, PDF + email
rendering, frontend CRM components.

## Test plan

- [ ] Apply migrations on fresh Postgres (Schema drift workflow passes)
- [ ] Enable `clients` + `quotes` + `contracts` + `bills` +
      `hoursLogging` + `calendar` in Settings → Features
- [ ] Walk quote → accept → contract → countersign → invoice →
      send → Storno; verify `deal_uuid` consistent in lineage card
- [ ] Drag-create hour entry on the calendar; verify lock visual
      when billed
- [ ] Toggle off `calendar`: sidebar entry disappears
- [ ] Render UI in DE — every CRM string in German

## Disclaimers

Seeded contract blocks and default IBAN/QR/tax settings are
**examples only**. Editor + detail surfaces flag this; see
`docs/crm-disclaimers.md`. Operators must verify with a lawyer /
bank / tax advisor before sending real documents.

### Architectural decision needed

- **Multi-admin data isolation — possible horizontal privilege
  escalation.** Every admin with `<resource>.view` reads every
  record across the install; no `created_by_admin_id` scoping. If
  picpeak is meant to be a single-tenant operator team, this is
  fine — document it. If admins should each see only their own
  customers, this is a bug. ~3 days to remediate. Decision needed
  before any code.

### Known service-code issue (surfaced during testing)

- **`createQuote` → `nextQuoteNumber` self-deadlock on SQLite.**
  `nextQuoteNumber()` opens an inner transaction from inside
  `createQuote`'s outer transaction. SQLite's default 1-connection
  pool deadlocks; Postgres tolerates it because of the larger
  pool. Either thread `trx` through the sequence claim or formally
  document that CRM paths require Postgres. Not fixed here to keep
  PR blast radius bounded.

### Scoped feature work (designed, not yet built)

- **`updateInstallmentPlan` — atomic post-spawn plan edit.** Initial
  installment plans work today (the `InstallmentsPanel` is wired
  into the quote editor and conversion spawns N siblings sharing
  one `deal_uuid`). What's missing: editing the PLAN as a whole
  once siblings exist. Lock-in: allow when all siblings are still
  `scheduled`, refuse once any has shipped. Reuses existing
  invoice_numbers in one transaction (no sequence waste).
- **Invoice & payment splitting via Storno.** §14 UStG / §11 UStG
  locks issued invoices as immutable, so any future "split a sent
  invoice" or "split a partial payment" feature MUST go through
  Storno + N new siblings sharing `deal_uuid` (never in-place
  mutation). No code yet — design intent captured for when admin
  actually asks for it.

### Test coverage gaps (foundation laid in `3240137`)

- HTTP route tests for adminQuotes / Contracts / Invoices / Calendar
- End-to-end lineage walk (quote → contract → invoice → storno) —
  blocked on the SQLite deadlock above; needs a Postgres-backed
  CI workflow mirroring `schema-drift.yml`
- PDF + email rendering tests
- Frontend CRM page/component tests (only utilities have Vitest
  coverage today)
- 6 pre-existing upstream-broken suites excluded from CI as
  test-infra debt

### Translations

EN + DE are written natively (DE is the operator-facing locale). FR /
NL / PT / RU / ES have **no CRM strings at all** — every CRM key
falls through to i18next's English fallback for those locales. The
non-CRM parts of the UI (galleries, public site, etc.) are still
translated in those files; only the new CRM namespaces are missing.

Effect: a user with `language=fr` sees French chrome + English CRM
strings. Safer than partial / machine translations, but obviously
incomplete for those audiences. Adding the missing locales is its
own work — best done by native speakers per `feedback_translation_flagging`,
not by an LLM.

### Intentional scope deferrals (calendar)

Single-day events only. Multi-day, recurring events, iCal/.ics
export, conflict warnings on overlapping hours, and customer self-
booking (`calendarBooking` — UI-locked "coming soon") are all out
of scope here.

### Postponed (lower priority, tracked separately)

- **Public token hardening.** Hash-on-store + OTP for high-value
  contract signings + one-time-use semantics on the signing token
  itself
- **Letterhead / Briefkopf fields.** Some columns exist on
  `business_profile` from mig 102 but the Settings UI doesn't
  expose all of them — audit + close the loop
- **Logo + branding-palette colors on invoice PDFs.** Consistent
  logo rendering + decide which color tokens map to which PDF
  surfaces (header band, accent line, footer)
- **Contract integrity check — upload-and-verify variant.** Server-
  side disk check already shipped; this adds the "drop the PDF
  customer emailed back, verify against stored hash" sub-action
- **Per-customer hourly-rate currency override.** A new
  `customer_accounts.hourly_rate_currency` column for installs
  where customer currencies diverge from the business default
- **Reminder emails beyond invoice dunning + pre-event.** Signing
  nudge for unsigned contracts past N days, quote-response nudge
  before `valid_until`, custom one-off reminders
- **Per-customer split on multi-customer events.** One event with N
  customers → N separate invoices instead of one
- **Two-way messaging.** Admin ↔ customer thread on the portal
- **"Hide every UI surface tied to a disabled feature" sweep.**
  Sidebar + sub-nav done; some deep links + dashboard cards still
  surface when their feature is off
- **`Split into installments` UI doesn't yet respect
  `general_date_format`** setting


Luca-Timo:
just noticed a issue with the email processor

Luca-Timo:
i had a feew issues after the change to the pr-branch, they are now fixed:

### 1. `billing_email` actually routes mail (`09c5110`)
The column existed on `customer_accounts` and the admin UI rendered an input for it, but **no send path read it** — every invoice / Storno / reminder went to the primary `email`. New `resolveBillingRecipients` helper: when `billing_email` is set, billing-class sends go To: there with the primary email on CC + any per-doc `cc_pdf_email` folded in. Quotes / contracts / event reminders / gallery still route to the primary `email` (decision-maker address). Payment-check stays admin-routed. 10 unit tests cover the dedupe + case-folding rules.

### 2. CRM email templates self-heal at boot (`83933ba`)
`crmEmailTemplates.ensureCrmEmailTemplatesSeeded` was defined but **had no caller anywhere** — every install that didn't pre-exist its templates failed every `quote_sent` / `invoice_sent` / `storno_issued` / `invoice_reminder_*` send with "Email template '<key>' not found", queue processor retried 3× and gave up silently. Wired all three seeders (CRM, contracts, event reminders) into `server.js startServer()`. Also added a targeted recovery: any `email_queue` row stuck at `retry_count=3` whose template we just seeded is reset back to 0 so the next tick picks it up. Already-deployed installs heal on first restart after this lands.

### 3. Quote-response "change within {{minutes}}" rendered literally (`3c6a903`)
The DE string referenced `{{minutes}}` but the call site only passed `{ at }` — public quote page showed the raw placeholder text. Call site now computes `minutes = ceil((responseLockedAt - now) / 60_000)` and passes both `{ at, minutes }` so EN (`until {{at}}`) and DE (`innerhalb von {{minutes}} Minuten`) each pick up the variable they need. Rounds up so we never advertise a deadline the customer can't actually hit.

the-luap:
Re-reviewed @the-luap & I against the new commits — the three post-close fixes (`09c5110` billing routing, `83933ba` template self-heal, `45f0606`+`5ce0b6e` DE quote-response variable) all look sound, and the negative-line-items + reminderEmails flag changes are clean. Two items from the original review that we'd like to clear before merging into beta:

**1. Multi-admin data isolation — is this by design?**

You set `created_by_admin_id` on writes (e.g. `quoteService.js:537`) but no read-path filter uses it — every admin with `quotes.view` sees every other admin's quotes / contracts / invoices / hours. Picpeak does support multiple admin users today (see `adminUsers.js` + RBAC), so we wanted to ask:

- **Single-tenant per install (intentional):** all admins on one install share the data, same as a small studio team where any admin can pick up any customer. If that's the model, perfect — we'd just want to document it explicitly (CRM data is shared across admins) so operators understand it.
- **Per-admin scoping (oversight):** each admin should see only their own customers / quotes / hours. If that's the intent, this is a horizontal privilege escalation across the entire CRM surface; ~3 days of work to add the `created_by_admin_id` filters everywhere + a "shared" flag for documents that should be visible to all (e.g. company-wide invoices).

What's your design intent here? Happy with either answer — we just want to be deliberate about it before production data accrues, since retrofitting scoping later is much harder than declaring tenancy up-front.

**2. SQLite deadlock fix — one-line ask**

You already prepped `nextQuoteNumber(trx)` to accept a transaction parameter, but the `createQuote` call site doesn't pass it:

\`\`\`js
return await db.transaction(async (trx) => {
  const quoteNumber = await nextQuoteNumber();  // ← needs (trx)
  ...
});
\`\`\`

Postgres tolerates the inner transaction because of pool size; SQLite deadlocks on every quote creation with the 1-connection default. Could you thread `trx` through this one call site (and the equivalent in `nextInvoiceNumber` / sequence-claim if the same pattern exists)? That'd unblock SQLite installs without needing to formally drop SQLite support.

**Deferred to follow-up (filed as #570):**

HTTP route tests for the new `adminQuotes` / `adminContracts` / `adminInvoices` / `adminCalendar` / `customer` admin routes + the three `public*` routes. Service layer is well-covered; HTTP layer is the gap. We're OK landing this PR without them on the condition the issue tracks it.

Everything else from the original review is either resolved by your new commits or accepted as deferred-by-design (translations, end-to-end lineage walk).

Luca-Timo:
@the-luap On #1 (multi-admin scoping) — to clarify: I forgot to surfaced it as an
upstream design call because picpeak's tenancy model is fundamentally
a product-direction question, not a CRM-implementation choice —
whichever way it goes affects the whole admin surface (galleries,
customers, events, archives), not just the new CRM tables.

To make it easy to commit either way, here's how we see the two paths:

**Single-tenant (shared CRM across admins, our current default):**
- Zero additional code; the per-record `created_by_admin_id` we
  already write becomes a useful audit field
- Two reasons we kept it on every write even though no read path
  filters by it today:
  1. **Future-proofing** — if we later decide to add per-admin
     scoping, the data is already there for every record from day
     one; no backfill needed, no "rows created before scoping
     went in" edge case to handle
  2. **Accountability / misuse trail** — if an admin acts outside
     their remit (deletes a customer, mass-edits prices, sends a
     quote they shouldn't have), `created_by_admin_id` plus the
     existing `activity_log` table lets the install owner trace
     who did what without having to enable scoping itself
- We add a sentence to `docs/permission-model.md` (or wherever you
  prefer) stating: "All admins on one install share CRM data.
  Per-admin isolation is not enforced — use separate installs for
  genuinely separate operators. `created_by_admin_id` is captured
  on every record for audit purposes."
- Matches the small-studio mental model (any admin can pick up any
  customer)

**Per-admin scoping:**
- Adds `whereCreatedByOrSharedFor(adminId)` filters across
  `quoteService`, `contractService`, `invoiceService`,
  `customerHoursService`, `customerAdminService` read paths
- New `customer_admin_assignments` table for the "this customer is
  mine + shared with these other admins" model
- New `is_shared` flag on documents that should remain company-wide
  (e.g. tax exports, archived invoices)
- Schema migration on `customer_accounts` to backfill an
  `owner_admin_id` for existing rows (cheap because
  `created_by_admin_id` is already populated)
- Realistic estimate: 3 days implementation + 1 day for the
  migration / backfill / edge-case audit

I'm happy to go either way — what's your preference as the upstream
owner? If you genuinely don't have a preference, my default would be
**single-tenant + documentation** (option 1), since picpeak's existing
surfaces (galleries, events, customer accounts) already work that way
today, and adding scoping to only the CRM would create an
inconsistent admin model.

the-luap:
**Follow-up testing notes (non-blocking for merge)**

Tested the full CRM flow end-to-end on a fresh PR 555 install (`d1aecaa`). The architecture and core flows work; the items below are polish and translation gaps worth tracking but none of them block landing.

### i18n gaps

1. **Features tab** (`/admin/settings?tab=features`) — `hoursLogging`, `contracts`, and `crmDevelopment` cards have no German translations (likely missing in other locales too).
4. **Quote editor payment-schedule dropdown** (`/admin/clients/quotes/...`) — dropdown options not translated.
5. **New contract page** (`/admin/clients/contracts/new`) — German translations missing.
9. **New invoice payment conditions** (`/admin/clients/bills/new`) — **English** translations missing on the payment-conditions dropdown.
10. **CRM Settings tab** (`/admin/settings?tab=crm`) and the CRM-related menu items on `/admin/settings` — German translations missing.

### Data defaults / inconsistencies

2. **Country (ISO-2) default mismatch** — the "New Customer" modal shows `CH` as the default, but the "Edit Customer" modal on an existing record shows `FL`. Should be consistent (and likely `CH` is the intended value since `FL` = Liechtenstein, probably stale test data leaking through).
3. **Customer minimum-fields requirement** — creating a customer with only an email shows the name as "Unknown" in lists/details. Request: make at least one of (first name / last name / company name) required at create time so we never land in the "Unknown" state.

### Feature gaps / UX

6. **Contract PDF preview before send** — currently the "Download PDF" action only appears after the contract has been sent to the customer. Add a pre-send download/preview so the admin can verify the PDF before committing to send.
8. **Hours overview when no customer is selected** — when landing on `/admin/clients/hours` without a customer picked, the page is empty. Suggestion: show an aggregate list of all customers with their open (unbilled) hours and open monetary amount. Reference: `sevdesk` does this and it's genuinely useful for "what work is still unbilled across all clients" at a glance — also saves the click-through to pick a customer when you already know which one has work pending. (Open question whether this matches photography workflow — happy to deprioritise if not.)
11. **Cross-links from CRM screens → relevant settings** — several CRM forms reference values that come from `admin/settings` but there's no link to jump there. Add a small "Configure in Settings" link (ideally opens in a new tab so in-progress form data isn't lost) next to those fields. I had to search for the settings page manually on a few occasions.

### Error handling

7. **Missing hourly-rate error on `/admin/clients/hours`** — when the selected customer has no hourly rate configured, the page errors out. Three sub-issues with that error:
   - The error message is not translated to German.
   - The UI shows the install-wide default hourly rate alongside the error, which is confusing — it looks like the rate IS set, but the save still fails.
   - Better: either (a) actually fall back to the install default rate on save when the customer has none, or (b) inline-validate the customer's rate field and offer a "Set hourly rate for {customer}" CTA instead of a save-time error.

Luca-Timo:
Thanks for the thorough sweep @the-luap. I'd already spotted a few
of these myself but held back during the open-PR window — happy to
fold your observations and mine into the same backlog now.

I just didn't want to make to much changes, while the pr is open.

Everything you flagged is captured below. None of it blocks merge
from my side either.

### Design / scope decisions (awaiting your input)

- **Invoice & payment splitting** — architecture locked in this PR;
  implementation deferred until admin asks.
- **Multi-admin data isolation** — your call (the open thread above).

### Feature work — open

- **"Trigger-only" mode for accumulator invoices** — covers two
  surfaces in one flag: monthly billing (today auto-flushes on
  cycle day) AND combined invoices (manually-grouped on-demand
  batches). One opt-out on `customer_accounts`, scheduler skips,
  "Trigger invoice now" is the only flush path.
- **Email settings + send verification** — "Send test email" in
  Settings + per-document delivery-status surface.
- **Customer creation: country field missing** — on both the
  primary customer-create form AND the passive-customer create
  flow. (Natural bundle with your country-default-mismatch bug.)
- **Per-customer Skonto opt-out** — mirror of the per-invoice
  `skonto_disabled` toggle at the customer level.
- **Light / dark mode logo variants** — second slot on
  `business_profile`, admin UI + public site auto-pick by theme.
  Open Q: keep the dedicated PDF logo as override or always use
  the light variant?
- **Imported invoices: `event_name` field on import form** — free-
  text input + optional autocomplete against existing events.
- **Admin "Sent emails" tab** — paginated `email_queue` browser
  with status / type / date-range filters.
- **Admin gesture to delete / decline a quote** — admin decline
  (records `declined_at + reason`, locks the public token) +
  hard-delete for `draft` quotes only.
- **Surface silent background failures (page + opt-in email)** —
  triggered by a 14h `email_template not found` outage on my beta
  box this week. System Health page aggregating failed queue rows
  / scheduler runs / PDF renders + opt-in admin digest.
- **`updateInstallmentPlan` — atomic post-spawn plan edit** —
  allow plan edits while all siblings still `scheduled`; reuses
  existing invoice_numbers.
- **Contract PDF preview before send** (your ask) — pre-send
  download so admin can verify layout before committing.
- **Hours overview when no customer selected — aggregate view**
  (your ask) — all customers with unbilled hours + monetary
  amounts as the landing state.
- **Cross-links from CRM screens → relevant settings** (your ask) —
  small "Configure in Settings" link next to fields that reference
  admin/settings, opening in new tab.

### Bugs — open

- **Import-invoice form: date input renders 2 fields per section.**
- **Tax-report page: date-range selector uses wrong format.**
- **CRM dashboard "last 30 days" counts imported historical
  invoices** — query uses `created_at` instead of `issue_date`.
- **Quote → invoice conversion drops the due date.**
- **i18n gaps across CRM admin surfaces** (your audit) — Features
  tab cards (hoursLogging / contracts / crmDevelopment) missing
  DE, quote-editor payment-schedule dropdown not translated, new
  contract page DE missing, new invoice payment-conditions
  dropdown EN missing, CRM Settings tab + CRM menu items DE
  missing.
- **Country default mismatch on customer modals (`CH` vs `FL`)**
  (your finding).
- **Customer create allows empty name → lands as "Unknown"**
  (your ask) — require at least one of (first_name / last_name /
  company_name) at create time.
- **Hours page: missing-hourly-rate error UX broken** (your ask
  with three sub-issues) — leaning toward your option (a), fall
  back to install default rate when customer has none.

### Postponed

- Public token model rework (post-PR1) — hash-on-store, OTP option,
  one-time-use semantics.
- More letterhead (Briefkopf) fields in business profile settings.
- Logo + branding colors on invoice PDFs.
- Contract integrity check — upload-and-verify variant.
- Per-customer hourly-rate currency override.
- CRM reminder emails — broader scope (signing nudges, quote
  nudges, custom one-offs).
- Per-customer split on multi-customer events.
- **Messaging with customer (multi-channel)** — also being
  discussed in [#560](https://github.com/the-luap/picpeak/discussions/560).
  I posted a channel + feature-set discovery question there to see
  what the community would actually use before locking direction.

### Outside the CRM scope (spotted in passing)

- **SVG not handled for favicon + collapsed sidebar icon** — not a
  CRM concern but worth tracking. Upload pipeline either rejects
  SVG or accepts but renders broken. Security note: SVG accepts
  `<script>` / `on*` / `<foreignObject>`, so the fix needs
  DOMPurify (SVG profile) or `` sandboxing before
  shipping the "accept SVG" change.

---

### One question on PR shape

My preference is **one big PR** rather than splitting — bundled
review of related items in a single diff is easier to merge and
I'd rather not flood you with a dozen tiny PRs for half-day items.
Easier to bisect if something regresses, granted, but the items are
small enough that I'd rather pay the rare-bisect cost than the
constant review-overhead cost.

Sound reasonable, or would you prefer me to split them up despite
the overhead?

the-luap:
@Luca-Timo yes one big PR is totally fine for me as I also like it more to see the full picture rather than smaller items.

PR #571, opened 2026-05-28
fix(analytics): Umami config check always falls back to env vars

filpgame:
## Description

`GET /admin/settings` returns a plain object (key → value), but `AnalyticsPage.tsx` was calling `.reduce()` on the response as if it were an array of `{ key, value }` objects.

This threw `TypeError: settings.reduce is not a function`, silently caught by the surrounding `try/catch`, which caused the code to fall back to `VITE_UMAMI_*` env vars. Those are unset in a typical install, so `umamiConfig.enabled` was always `false` — the "Umami is not configured" banner always showed regardless of what was saved in Settings → Analytics.

**Fix:** use `response.data` directly as the settings map (it's already the plain object the check expects).

## Type of change

## How Has This Been Tested?

- Configured Umami URL + Website ID in **Settings → Analytics**, saved successfully
- Navigated to **Analytics** page — "not configured" banner no longer appears
- Verified `umamiConfig.enabled` is `true` when all three required fields are set

**Config:**
- SQLite, local dev (`npm run dev`)
- No `VITE_UMAMI_*` env vars set (confirms fix doesn't depend on env fallback)

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my own code
- [x] My changes generate no new console errors

## Screenshots

N/A — banner disappears when config is present.

## Additional Notes

Root cause: the `try/catch` around the fetch silently swallowed the `TypeError`, making this look like an API or auth problem rather than a JS runtime error in the response handler.


filpgame:
Opened against wrong repo — re-submitting from personal fork.

PR #579, opened 2026-05-29
fix(email): preserve dots + subaddresses across all normalization sites

### LINKED ISSUE #574: [BUG] Gmail addresses are normalized by removing dots, causing login mismatch after invitation v3.43.1
**Description**

When inviting a user with a Gmail address that contains dots, PicPeak appears to normalize the email address and removes the dots from the local part.

Example:

john.doe@gmail.com

gets stored/used as:

johndoe@gmail.com

The invitation email is still delivered correctly, because Gmail treats both addresses as the same mailbox. However, this causes confusion and a login mismatch, because the user was invited with john.doe@gmail.com, but the account/login expects johndoe@gmail.com.

Expected behavior

The email address entered during invitation should be preserved exactly as entered, especially for login/account identity purposes.

Example:

john.doe@gmail.com

should remain:

john.doe@gmail.com
Actual behavior

The Gmail address is normalized and dots are removed:

john.doe@gmail.com → johndoe@gmail.com
Possible cause

It looks like the backend uses normalizeEmail() from express-validator / validator.js without custom options.

For Gmail addresses, normalizeEmail() removes dots by default because gmail_remove_dots defaults to true.

Example location:

body('email').isEmail().normalizeEmail()
Suggested fix

Please consider disabling Gmail dot removal when normalizing email addresses:

.normalizeEmail({
  gmail_remove_dots: false,
  gmail_remove_subaddress: false
})

Or avoid normalizing the email address in a way that changes the value used for login/account identity.

Why this matters

Even though Gmail delivers both addresses to the same mailbox, the visible/login email should match what the admin entered during invitation. Otherwise users may not be able to log in with the address they were invited with, and admins may think the wrong email was created.

the-luap:
Closes #574.

## Summary

@blazmaric reported that inviting a Gmail address with dots (e.g. `john.doe@gmail.com`) silently stored the address as `johndoe@gmail.com`, breaking login with the as-invited form. Their diagnosis pointed at express-validator's `normalizeEmail()` defaulting to Gmail dot-stripping.

The diagnosis was correct — and the bug was wider than the one site they mentioned. **All 17 `normalizeEmail()` call sites in the codebase are identity-bearing** (login, invitations, account create/update, feedback identity) — none had a legitimate reason to strip dots for deduplication.

## Change

- New `backend/src/utils/emailNormalization.js` exports a single shared options object `IDENTITY_PRESERVING_NORMALIZE_EMAIL` that disables every provider-specific canonicalization:
  - `gmail_remove_dots: false` — the actual #574 bug
  - `gmail_remove_subaddress: false` — preserves `+tag` on Gmail
  - `gmail_convert_googlemaildotcom: false` — keeps `googlemail.com` (not folded to `gmail.com`)
  - `outlookdotcom_remove_subaddress: false`
  - `yahoo_remove_subaddress: false`
  - `icloud_remove_subaddress: false`
  - The only default left enabled is `all_lowercase` (safe — local-parts are case-insensitive in practice on every major provider, and lowercasing keeps login lookup consistent).
- All 17 call sites updated to pass the shared options:
  - `adminAuth.js` (admin profile email update)
  - `adminUsers.js` × 2 (admin create + update)
  - `adminCustomers.js` × 2 (customer create + update)
  - `customerAuth.js` (customer login lookup)
  - `events.js` × 2 (event create + update — `customer_email`)
  - `adminEvents.js` × 3 (event create + update + dedicated `customer_email`/`admin_email`)
  - `adminEvents-enhanced.js` × 2 (reference snippet for the enhanced flow; patched for consistency in case it gets wired up)
  - `feedbackValidation.js` × 4 (rating, comment, like, favorite)

Note on the `Object.freeze` design — initially tried freezing the options object, but validator.js's underlying `merge()` mutates the options to add its own defaults (notably `all_lowercase: true`), which would crash with `Cannot add property all_lowercase, object is not extensible`. Comment in the helper file explains why it's left mutable.

## Test plan

- [x] **New regression suite** `backend/__tests__/utils/emailNormalization.test.js` — 7 unit tests pinning: Gmail dots preserved, Gmail/Outlook/Yahoo/iCloud subaddresses preserved, googlemail.com domain preserved (not folded), local-part lowercased (kept default).
- [x] All 7 pass on the new options object.
- [x] Lint clean for all my changed files. Pre-existing `no-unused-vars` in `customerAuth.js:20` + `events.js:142` and `no-control-regex` in `feedbackValidation.js:88` are unchanged from beta.
- [ ] Manual: invite a customer with `john.doe@gmail.com` → stored email matches input → login with `john.doe@gmail.com` succeeds.
- [ ] Manual: invite a customer with `jane+work@gmail.com` → stored email matches → login with that exact address succeeds.

## Migration note

Existing accounts whose emails were already stripped pre-fix remain with the stripped form in the DB — they'll keep logging in with the stripped form they were assigned. The fix takes effect for new invitations going forward. If an admin re-invites an existing user with the un-stripped address, that would create a duplicate account; if that becomes a real problem we can add a backward-compat login fallback (try lookup with dot-stripped form too) as a separate change. Not in scope here.


blazmaric:
One more small UX note for Admin → Users:

The deactivate user action works, but after an admin user is disabled there does not seem to be an option to reactivate or permanently remove/delete that user from the UI.

Currently the disabled user remains listed, but there is no follow-up action to manage it afterwards.

Would it make sense to add “Reactivate user” and/or “Delete user” actions for disabled users?

the-luap:
Good catch @blazmaric — that was a real UX gap, the row was genuinely stranded once deactivated. Built it out in #588: deactivated users now get both **Reactivate** (UserCheck icon, restores access) and **Delete** (Trash2, hard-deletes with the usual self / last-super-admin guards). Backend gets symmetric `POST /:id/activate` + `DELETE /:id` endpoints, FK cascade audited so the delete is safe. 8 new service tests pin the guards.

Thanks for keeping an eye on the follow-on UX — these "oh, what happens after?" gaps are easy for me to miss when shipping the primary action.

PR #596, opened 2026-05-31
Backup & Restore hardening — close the silent files-only data-loss class

Luca-Timo:
Triggered by a real data-loss incident on a beta install (2026-05-29): four consecutive "Run Backup Now" runs silently shipped manifests with `database.backup_file = null`. I then did `docker compose down -v`, confident the backups were complete. They weren't — all CRM data (customers, quotes, invoices, contracts) was lost. luckly only mock / Test Data.

This branch closes that data-loss class at the source AND every layer between it and the admin: capture, configuration, surfacing, restore, disaster recovery, and documentation.

End-to-end validated on a real Postgres install by destroying it and rebuilding from backup — login with original credentials worked first try.

## Scope summary

35 commits across 4 themes. Bundle is intentional per the maintainer's preference for "full picture" diffs:

- **Stages A/B/C** — original three-stage hardening plan (capture + config + diagnostic)
- **Latent PG-restore bug fixes** — 11 dormant bugs surfaced by end-to-end DR validation
- **Tier 1/2/3 polish** — silent-failure surfaces, operator-meta persistence, per-path stats
- **Install-from-backup feature** — one-boot disaster recovery via trigger file
- **Documentation + i18n** — `docs/backup-restore.md` rewrite, en + de locale updates

### Stage A — Inline DB dump + fail-loud guard
`backupService.runBackup` now triggers `databaseBackupService.backup()` inline at the start of every run. New `backup_database_inline_dump` setting (default ON) lets operators opt out for installs with a separately-orchestrated DB dump. With opt-out enabled, a post-dump validator still aborts the run if no dump exists within the staleness window — no more silent files-only manifests.

### Stage B — Config-driven walker via `backup_paths` table
Migration 109 (renumbered from 108 to avoid the upstream `108_seed_sl_email_template_translations` collision — see "Migration note" below) seeds 7 canonical paths. Boot-time self-heal (`_backupPathsBoot.js`) handles already-deployed installs. Admins add custom paths via SQL — no code change needed.

### Stage C — Coverage diagnostic
New `GET /api/admin/system-health/backup-coverage` returns per-path coverage status + drift detection (top-level `STORAGE_PATH` subdirs without a `backup_paths` row). New `BackupCoverageCard` + Coverage tab in the admin UI. Catches the "feature shipped without a backup row" footgun in advance.

### Restore-path bug fixes (11 latent bugs)
Each was dormant because no test exercised the real-PG restore flow. All now closed:

1. `psql` DROP/CREATE DATABASE invoked without `-d` (failed when DB_USER ≠ DB_NAME)
2. DROP DATABASE blocked by backend's own connection pool (`pg_terminate_backend` + `WITH (FORCE)` fix)
3. `options.source = 'local'` used verbatim as a path (`local/database/...` garbage)
4. DB dump path basename-reconstructed instead of honoring the manifest's absolute path
5. Rollback never fired except on post-restore verification failure
6. Sequences not re-synced after psql load (every CREATE/INSERT into SERIAL columns failed with PK violation)
7. Knex pool stuck in destroyed state after restore (required container restart to log in)
8. `db.migrate.latest()` choked on `migrations/helpers.js` after every restore
9. `pg_dump --single-transaction` (never valid for pg_dump; belongs to pg_restore / psql)
10. `safeExec.spawnToFile` passed unopened WriteStream as stdio[1] (Node 22 strict mode rejection)
11. `databaseBackup.js` insert destructure (`const [runId] = ...` without `.returning()`) — broken on Postgres

### Tier 1/2/3 polish
- **#1** Restore wizard surfaces failure status + error_message in red banner instead of silently rendering "completed" at 0%
- **#2** Dashboard "Last backup" widget filters by `status='completed'`; surfaces failed/running attempts separately + flags zombie rows >30min old
- **#3** Restore wizard shows red "No DB" badge on files-only backups + explicit warning callout on the selected card
- **#4** Knex pool re-init after DROP/CREATE via `reinitPool()` + Proxy export (no container restart needed)
- **Tier 2** Operator-meta settings (`restore_allow_force`) persist across restore (snapshot before DROP, replay after psql load)
- **Tier 3** `backup_runs.statistics.per_path` carries per-`backup_paths`-row counters; History detail pane renders one row per entry with count + size

### Install-from-backup feature
Drop a `RESTORE_ON_INSTALL` file into the existing `/backup` bind mount. On next `docker compose up`, picpeak runs the restore before the admin UI surfaces. Two flavors: empty file = auto-pick newest manifest; non-empty = specific manifest path. Three safety gates (trigger presence + empty-DB check + auto-rollback on failure) + `INSTALL_FROM_BACKUP_FORCE=true` override. 7 integration tests.

### Documentation + i18n
- `docs/backup-restore.md` — comprehensive rewrite covering configure → run → restore (live) → restore (DR) → troubleshoot. ~340 lines, with anchored sections. Suggest publishing on `docs.picpeak.app/guides/backup-restore`.
- README link blurb updated to reflect actual scope.
- en + de i18n strings for new UI surfaces (Coverage tab, error banners, files-only warnings, History per-path rows).
- ⚠️ **Other locales (es / fr / nl / pt / ru) fall back to en keys until a native speaker reviews.** Marked in keeping with [[feedback_translation_flagging]] / [[feedback_translation_flagging]].

## Migration note

`108_add_backup_paths.js` (in earlier commits on this branch) collided with upstream/beta's `108_seed_sl_email_template_translations.js`. Renamed to `109_add_backup_paths.js` in the last commit. Mechanical rename, fully reversible via `hasTable` + `onConflict.ignore()` guards in the migration body. Beta installs that already applied the 108-named version get a harmless no-op when 109 runs. The orphaned `108_add_backup_paths.js` row in `migrations` sits harmlessly alongside the new `109_add_backup_paths.js` row.

When this merges, Ralf's beta install will pick up `108_seed_sl_email_template_translations` (new) + `109_add_backup_paths` (no-op). Fresh installs run them in order.

## Verification

- `cd backend && npx jest --no-coverage --testPathPattern='integration/(adminBackup|backupService|installFromBackupBoot)'` — **31 / 31 pass**
- `cd frontend && npm run build` — **clean**
- End-to-end on a real Postgres install:
  - `docker compose down -v`
  - Restore from disk via the new install-from-backup trigger file
  - Login with original pre-disaster credentials worked first try
  - Customer / quote / invoice / contract pages all loaded with the restored data
  - Integrity tab verified PDFs resolve, Coverage tab green
- Pre-existing failing integration suites (`webhookDelivery`, `storageBackend`, `adminPhotos.reference`, `imageProcessor.storage`) confirmed unrelated via `git stash` baseline run

## Test plan

- [ ] Pull the branch, run `cd backend && npx jest --no-coverage` — backup-related suites pass
- [ ] Build frontend — `cd frontend && npm run build` — clean
- [ ] On a fresh PG install: configure backup destination → click Run Backup Now → confirm `database.backup_file` is non-null in manifest
- [ ] On a fresh PG install: ⚠️ DESTRUCTIVE ⚠️ `docker compose down -v`, drop `RESTORE_ON_INSTALL` in `./backup`, `docker compose up -d` — confirm restore completes and original credentials work
- [ ] Open Coverage tab — all 7 paths green if no drift
- [ ] Open Integrity tab → Run check → confirms PDF paths resolve
- [ ] Backup History → expand a row → per-path breakdown visible
- [ ] Restore wizard → pick a backup → Force ✓ → completes; if any step fails, the error banner shows the actual error_message instead of "completed at 0%"

## Follow-ups (NOT in this PR)

Captured as task chips / backlog entries. Each ships in its own PR:

- Real-PG integration test in CI (would have caught every restore bug above)
- Restore wizard "files-only" badge in the integrity tab (cross-link from there)
- Sibling-class fix for `userManagementService` + `customerAccountsService` if they hit the same destructure pattern under load (audited tonight — current code is safe, but watch on diff)
- "Business-hours floor" for scheduled emails (separate backlog item, mentioned in [project_crm_backlog.md])
- Per-Stage-B-path backup_paths ordering / drift admin UI (the Coverage tab covers this; this would let admins edit `backup_paths` from the UI rather than SQL)
- `was_successful` flag set correctly on successful restores (cosmetic; data is back but the flag stays `false`)

## Notes for the reviewer

- This is NOT a stacked / split PR. The 35 commits are atomic (one concern per commit) but interrelated enough that splitting would create a review-order dependency. Per the bundled-PR preference, presented whole.
- Each "fix" commit has an inline comment block at the patch site explaining the root cause + why the fix is shaped the way it is. Reading the diff top-to-bottom should reconstruct the debugging story.
- The migration renumber is the last commit so it's easy to revert if you prefer a different number / coordination story.


the-luap:
Ran the full e2e DR cycle locally on a real Postgres install before merging — the headline Stage A fix is solid (`database.backup_file` non-null, dump on disk with checksums and per-table row_counts, exactly what the PR set out to close). Stage B/C and the safety gate all work too. One blocking issue I hit, and a couple of smaller observations.

### Blocker — `preservedMeta is not defined` on every PG restore

`backend/src/services/restoreService.js:1030` blows up with a `ReferenceError` because the variable is out of scope:

```js
} else {                       // L833 — PostgreSQL branch
  // ...
  const PRESERVED_META_KEYS = [...];
  let preservedMeta = [];      // L850 — declared INSIDE the else
  // ... psql DROP/CREATE/load ...
}                              // L985 — else block closes HERE

// reinitPool + comments about migrate.latest ...

if (preservedMeta.length > 0) { // L1030 — outside the else block scope
```

The actual `psql` restore runs to completion before the JS error fires (which is why the data does come back), but everything after blows up:

```
Database restore failed: preservedMeta is not defined
Install-from-backup: FAILED — Full restore failed: preservedMeta is not defined
```

Knock-on effects:
- Operators see a loud `FAILED` line in `combined.log` even though the data restored — easy to assume the backup is gone
- Trigger file isn't auto-cleaned (the `removed trigger file` success path at L213 of `_installFromBackupBoot.js` never runs)
- The operator-meta replay (`restore_allow_force`, `restore_allow_force_auto_upgraded`) doesn't execute — exactly the chicken-and-egg the snapshot/replay was added to close. `restore_allow_force` silently reverts across every PG restore.

Smallest fix: hoist the `const PRESERVED_META_KEYS` + `let preservedMeta = []` declarations above the `if (db.client.config.client === 'sqlite3')` / `else` split, leaving the SQLite branch a no-op for them. Keeps the snapshot read inside the else (must run before DROP), but the variable lives wide enough to be read at the replay site. Move the early `try { preservedMeta = await db('app_settings')... }` along with the keys constant if you want to centralise it.

### Why CI didn't catch it

The integration tests around `performFullRestore` exercise the SQLite branch (via `db.client.config.client === 'sqlite3'`), so they never traverse the buggy postgres path. Worth adding a unit test that stubs the PG branch (`db.client.config.client = 'pg'`, mock `spawnAsync` for psql, return success) and asserts the function completes without throwing. That would also catch any future drift in the else-block scope.

### Smaller observations (non-blocking)

1. **Install-from-backup boot only writes to `/app/logs/combined.log`, not stdout.** `docker logs picpeak-beta-backend` after the restore boot shows the migration sweep + npm notice and nothing about the restore. Operators tailing container logs see no signal that a restore happened. One-line `console.log` mirror at the trigger-detect / restore-start / restore-end points would make this loud where it matters most.

2. **Fresh-install admin creation fires before the restore boot kicks in.** The boot sequence on an empty volume is:
   - `migrate:safe` runs → 001_init creates the default `admin@localhost / <random>` user + writes `data/ADMIN_CREDENTIALS.txt`
   - All other migrations run → schema is fully fresh
   - `server.js` → `tryInstallFromBackup` → `psql` DROPs the DB and reloads from dump → original admin replaces the fresh one
   
   Net effect: the fresh creds in `ADMIN_CREDENTIALS.txt` are stale junk after restore. Functionally harmless, but it's worth a comment in the file or maybe a guard that skips fresh-admin creation when `RESTORE_ON_INSTALL` is present.

3. **Safety gate works correctly** — restarted the backend without removing the trigger and got `"refusing — install has 1 admin(s) and 1 event(s). [...] Override with INSTALL_FROM_BACKUP_FORCE=true"`. Good.

### Coverage tab + Stage A green

For context on what was validated:
- Backup manifest has `database.backup_file: /backup/database/picpeak-db-postgresql-<ts>.sql.gz`, `size: 93643`, full `row_counts` block (e.g. `admin_users: 1`, `activity_logs: 16`)
- `GET /api/admin/system-health/backup-coverage` returns `database.mode: inline`, `database.ok: true`, `database.lastDumpStale: false`, all 7 paths from migration 109 with correct drift status (caught `previews` as `missing-on-disk` since the dir doesn't exist yet — that's exactly what the diagnostic is for)
- Post-restore, login with the pre-destruction admin password worked first try, all event/setting/activity_log rows present

Fix the scope leak and I'm comfortable merging this. Thanks for the very thorough work on it — the inline DB dump fix is solid and the Stage B/C surfaces are real upgrades.

the-luap:
Re-ran the full DR cycle against the new code. The scope fix is solid — no more ReferenceError, and the `[install-from-backup]` lines show up cleanly in `docker logs` now (nice). ADMIN_CREDENTIALS notice also lands as expected. But one more PG-only landmine surfaced under e2e, and I audited the rest of the backup/restore surface to make sure we're not playing whack-a-mole. Good news: only one more.

### Blocker — `restoreService.js:1243`

```js
if (result.count !== expected.rowCount) {
```

`knex('table').count('* as count').first()` returns `result.count` as a **string** on PG (pg-driver serialises `bigint` to string to avoid precision loss for huge counts). The manifest stores `rowCount` as a number (`parseInt`-ed at `databaseBackup.js:118`). Strict `!==` flags every match as a mismatch on PG.

Confirmed in container:
```
PG count result: {"count":"16"} typeof count: string
```

Symptom is the same shape as `preservedMeta` was — install-from-backup logs FAILED, trigger file isn't cleaned, data is actually intact. Verification message reads like a parody:
```
Table activity_logs row count mismatch: expected 16, got 16
Table admin_users row count mismatch: expected 1, got 1
Table app_settings row count mismatch: expected 165, got 165
... (every table, all matching)
```

**Fix:**
```js
const actual = Number(result.count);
if (actual !== expected.rowCount) {
  verification.errors.push(
    `Table ${table} row count mismatch: expected ${expected.rowCount}, got ${actual}`
  );
}
```
Both the comparison and the interpolated value need the coercion.

### Pre-emptive cleanup — same shape, currently masked by `>`

`restoreService.js:448` and `:458-459` have the same string-vs-number issue but use `>` instead of `!==`:

```js
if (eventCount && eventCount.count > 0) {
  validation.warnings.push(`Database contains ${eventCount.count} existing events that will be overwritten`);
```

The conditional fires correctly on both DBs (JS coerces `>` operands), but the warning text prints `"5"` on PG vs `5` on SQLite — and the moment a future patch changes that to `=== 0` or `!== expectedCount`, it silently breaks on PG. Same coercion + a one-liner comment that strict equality here would regress on PG would close the latent footgun.

### Why this kept surfacing

Both bugs (`preservedMeta`, this one) share the same root cause: **the PG branch has no runtime test traversal** because the integration suite doesn't have a real `psql` binary + cluster. The source-inspection test you added for `preservedMeta` pins THAT specific contract beautifully, but won't catch this class of bug going forward.

Your own "real-PG integration test in CI" follow-up from the original PR body would have caught both. I'd promote it from "follow-up" to either:
- Bundled into this PR (boldest), or
- A hard gate for the next backup/restore-touching PR (still good)

### Confidence boost — I audited the rest

To avoid another round, I grepped every backup/restore file for the same string-vs-number / type-coercion pattern. Audit covered: `restoreService.js`, `_installFromBackupBoot.js`, `_restoreSettingsBoot.js`, `_backupPathsBoot.js`, `backupCoverageService.js`, `backupIntegrityService.js`, `backupService.js`, `databaseBackup.js`.

**Findings: 1 blocker (line 1243), 2 cosmetic (lines 448/458), zero other landmines.**

Safe sites worth calling out (so you know I looked):
- `_installFromBackupBoot.js:124-131` — already pre-coerces with `Number(adminCount?.c || 0)` ✓
- `_restoreSettingsBoot.js:137` — `app_settings.setting_value` is `text` (always JSON-stringified), string-vs-string strict equality is safe ✓
- `backupCoverageService.js` — `parseSettingValue` + `Boolean(v)` + `instanceof Date` handling all correct ✓
- `backupService.js` — `normalizeBoolean(...)` used consistently for boolean comparisons ✓
- `backupService.js:928` — `record.id === currentRunId` is safe because both come from `int4` PK (pg-driver returns int4 as JS number; only `bigint` from `COUNT(*)` returns string)
- `backupIntegrityService.js:192` — hex-string vs hex-string ✓
- `databaseBackup.js:625-654` — version strings ✓

So after the line-1243 fix (+ optional 448/458 pre-empt), the PG-only `bigint`-as-string class is closed in the audited scope. We're not playing whack-a-mole — the rest of the code already handles this case correctly.

Once line 1243 is patched I'll re-run the same DR cycle one more time and confirm. If that comes back green, this is mergeable from my side.

the-luap:
Green from my side. Re-ran the full DR cycle on real Postgres with the v3 fix — clean end-to-end.

### What the round-3 boot log looks like now

```
[install-from-backup] trigger file detected → /backup/manifests/backup-manifest-backup-20260601-202541-1098141d.json
[install-from-backup] starting restore from /backup/manifests/backup-manifest-backup-20260601-202541-1098141d.json
[install-from-backup] restore completed successfully
```

`combined.log` tells the full happy-path story in order:
- `Snapshotted 2 restore-meta setting(s) for post-restore replay` ← snapshot fires
- `Re-initializing knex pool against the restored database…`
- `Replayed 2 restore-meta setting(s) post-restore` ← **operator-meta actually persists now** (the whole reason the snapshot/replay was added — silently dropped on v1/v2)
- `Restore progress: Performing post-restore verification…` ← no row-count mismatch errors
- `Email queued: restore_completed to admin@example.com`
- `Restore completed successfully` `{ databaseRestored: true, filesRestored: 1, errors: [] }`
- `Install-from-backup: removed trigger file /backup/RESTORE_ON_INSTALL` ← **trigger auto-cleaned, no manual rm needed**
- `Install-from-backup: completed from <manifest>. Server will start with restored state.`

### What I verified post-restore

- Login with the pre-destruction password (`TestPass123!`) succeeds first try
- `events` count = 1 (the test event survived)
- `admin_users` email = `admin@example.com` (original, not the fresh-install `admin@localhost`)
- `activity_logs` count = 16 (matches pre-destruction count exactly)
- `/api/admin/system-health/backup-coverage` returns `database.mode: inline`, `database.ok: true`, 7 paths
- `RESTORE_ON_INSTALL` trigger file is gone from `/backup` (no manual cleanup required)
- New `restoreService.pgBranch.test.js` — **3/3 pass** locally including the new `"every .count() result is coerced to Number before comparison"` contract

### The `Number(...)` fix is exactly right

Coerced at both the comparison AND the interpolated value in the verification message, plus the pre-empt at the validation-warning sites (lines 448/458) with the explanatory comment block. The new contract test catches the regression at the source-inspection level the same way the `preservedMeta` test does for the scope leak. Pragmatic and on-brand with the rest of the PR.

### Final note on the "real-PG integration test in CI" follow-up

Two rounds of "looks fine on SQLite tests, breaks on PG" tells the story — that follow-up is the structural answer. Source-inspection tests are great for pinning specific contracts you've already hit (and these two will), but they don't generalise to "the next PG-only landmine nobody's thought of yet." Strongly support promoting it to either bundled-into-this-PR or a hard gate for the next backup/restore-touching PR. Otherwise we're trusting future patches to not re-introduce the same class of bug.

### Verdict

This is the cleanest install-from-backup boot I've seen in three rounds, the post-restore meta replay finally lands, and the audit confirms the PG-only `bigint`-as-string class is closed in scope. Merge from my side.

Thanks for the very fast turnaround on both rounds. The inline-DB-dump fix is the headline, but the discipline around adding regression tests at every fix, the operator-meta snapshot/replay, the trigger auto-cleanup, the `announce()` mirror to docker logs, and the `ADMIN_CREDENTIALS.txt` annotation are all worth more than the headline by themselves.

the-luap:
Re-ran the e2e DR cycle on v4. Reproduced your round-3 trigger condition exactly (deleted \`restore_allow_force_auto_upgraded\` from the live DB → took backup → that backup had 166 \`app_settings\` rows, **no** auto-upgraded flag). On v3 code this is the exact shape that would have failed \`expected 166, got 167\` at verification. On v4: clean success.

### Verified behaviour matches the architectural change

Log signatures from this run (timestamps `2026-06-02 08:56:43`):

```
[install-from-backup] trigger file detected → ...backup-20260602-065405-b0fa6e2e.json
[install-from-backup] starting restore from .../backup-manifest-backup-20260602-065405-b0fa6e2e.json
Snapshotted 2 restore-meta setting(s) for post-restore replay  ← inside performDatabaseRestore (PG branch)
Performing post-restore verification...                         ← runs FIRST now
Replayed 2 restore-meta setting(s) post-verification            ← runs SECOND, your new "post-verification" wording
Applying post-restore migrations to restored database...        ← new migrate:safe step
Post-restore migrations applied
[install-from-backup] restore completed successfully
[install-from-backup] removed trigger file /backup/RESTORE_ON_INSTALL
```

Zero `row count mismatch` lines from this run (`grep "2026-06-02" combined.log | grep "row count mismatch" | wc -l` → 0).

### Verified state

- `app_settings`: 166 in backup, **167** post-restore (= backup + replayed `restore_allow_force_auto_upgraded`). The replay is now layered on top of an already-verified DB instead of inflating it before verification — exactly the ordering you described.
- `restore_allow_force_auto_upgraded`: `true` (correctly inserted by replay)
- All 13 files (events photos, thumbnails, business-docs PDF, uploads logo, download-cache zip) restored byte-for-byte from the same wiped-storage scenario as the round-3 test
- Login with original admin credentials works first try
- Trigger auto-cleaned

### One pre-existing audit-trail quirk worth knowing about (NOT introduced by this PR)

`restore_runs` is empty post-install-from-backup, so the `was_successful=true` patch in `7988c18` doesn't actually persist on the install-from-backup path:

- `restore()` inserts the audit row (`restoreService.js:83`)
- `performDatabaseRestore` does the psql DROP/CREATE and reloads the dump — including the just-inserted `restore_runs` row, which gets replaced by the backup's state of `restore_runs` (typically empty on a freshly-installed system)
- `restore_runs.update(... was_successful: true ...)` at L326 then targets an id that no longer exists → silent no-op

This isn't a regression and the fix is mechanically correct for **admin-UI-triggered** restores where the audit row survives the round-trip differently (or for restore types that don't wipe \`restore_runs\` via the dump). Worth a follow-up either:
- Re-insert the audit row AFTER `performDatabaseRestore` (so the success update has something to land on), or
- Exclude \`restore_runs\` from the dump on backup so the row survives the load.

But that's a separate concern from what this PR set out to close, so 👍 to ship as-is.

### Source-inspection tests

5/5 pass against v4 — both the existing contracts (preservedMeta scope, .count Number coercion, replay-after-verification) and the new ones (was_successful=true on completed, migrate:safe invoked after replay).

### Cumulative scorecard

| Round | Bug | Fixed in commit | Verified |
|---|---|---|---|
| 1 | `preservedMeta` scope leak | `a23fa3b` | ✓ |
| 2 | bigint `result.count` strict-equality | `354fbed` | ✓ |
| 3 | replay runs before verification → inflates count | `20e3092` | ✓ this run |
| 4 | `was_successful` not set on completed update | `7988c18` | ✓ (mechanically; see audit-trail note above) |

### Verdict

End-to-end on real PG: backup captures everything (DB + 7-path file walker), restore reconstructs everything byte-for-byte, replay/verification/migrate sequencing is clean, operator-meta survives, trigger is auto-cleaned, the announce()/docker-logs visibility makes the boot story legible to operators, and the source-inspection tests prevent regression of all four contracts.

From my side: cleared for merge.

PR #603, opened 2026-06-03
CRM improvements: invoicing & payments, hours, email queue/scheduling, branding (dark mode + favicon), country pickers

Luca-Timo:
Bundles the CRM-area work that's been running on the beta channel into one reviewable diff. Scope is **Liechtenstein/Switzerland-first** throughout (ISO country codes incl. LI not "FL", §14/§11 UStG immutability respected — issued invoices are never mutated).

**Scale:** 59 commits · 104 files · +4,676 / −629 · 7 migrations (`110`–`116`).

---

### Invoicing & payments
- **Cash-basis revenue, made trustworthy.** Dashboard "Revenue · last 30/90/365 days" tiles recognise by `paid_at` for every paid invoice, and the mark-paid dialog gains an **optional backdatable Payment date** so the figures reflect when money actually arrived (and are bank-reconciliation-ready — a future matcher just writes the same `paid_at`).
- **Document dates anchored correctly.** Issue date + due date track the actual send date and payment term; the Skonto window anchors to the send date; imported historical invoices keep their original issue/payment dates instead of the import timestamp (migrations `111`, `116` backfill existing rows).
- **Finder-style sortable invoice columns**, default sort by issue date.
- **Date inputs respect `general_date_format`** everywhere, render Postgres full-ISO dates correctly, and commit on change (no lost value when submitting before blur).

### Customers & country handling
- **Country is now an ISO dropdown** (full ISO 3166-1 list, localized labels via `Intl.DisplayNames`) on **every** surface: admin customer create/edit, business profile, quick-add, customer onboarding, and the customer billing-address profile.
- Normalised `FL` → `LI` (migration `110`); dropped the now-redundant free-text "Country (full name)" field (the PDF derives the localized name from the code).
- "Company **or** name required" guard on customer create/edit.

### Quotes & contracts
- **Per-customer Skonto opt-out** (migration `112`).
- **Admin decline-on-behalf** for quotes, with optional reason (migration `115`).
- **Contract PDF preview** before sending.
- Cross-link from the quote/invoice editors to CRM settings (in-app navigation).

### Hours / time tracking
- Install-wide **default hourly rate** with an inline "set a rate" CTA when missing (migration `113`).
- Aggregate **open-hours landing view** at `/admin/clients/hours`.
- "Create invoice from logged hours" opens the editor so the admin can add line items before issuing.

### Events ↔ invoices
- **"Create invoice" action** on the event detail page (pre-fills customer/event).
[...]

Issue #42, opened 2025-10-16
[FEATURE] Passwort protected gallery lockout - suggested improvemnt


mudevio:
When sharing a password-protected gallery link with a community, some people enter the wrong password. After **5 incorrect attempts from any source**, the gallery is locked for **30 minutes**.

I would like to have an option where **login attempts are validated by the combination of gallery slug and IP address**, so that the lockout threshold is **scoped per IP** instead of globally.

In addition, it would be great to make the following values configurable (e.g. in the settings or environment variables):

- Maximum login attempts  
- Lockout duration  
- Attempt window (timeframe)  

---

### Use Case

Anyone who shares a gallery with many people could benefit from this improvement to **avoid frustrating end users**.

Especially when the link is posted in a community, many people try to log in and sometimes make typos. Users who are excited to see the photos get frustrated when they have to wait 30 minutes just because **someone else entered the wrong password multiple times.**

---

### Implementation Ideas

- Add configuration options for the constants `MAX_LOGIN_ATTEMPTS`, `LOCKOUT_DURATION`, and `ATTEMPT_WINDOW`.
- Modify the `checkAccountLockout` function to accept an additional `ipAddress` parameter and apply the lockout per IP.  
  _(both in `backend/src/utils/authSecurity.js`)_

---

If a pull request would be useful, I’d be happy to work on it!

PR #616, opened 2026-06-10
feat(projects): Project Overview cockpit — link (multiple) quotes/contracts/hours into projects

Luca-Timo:
## Summary

Adds a **Project Overview** ("cockpit") area to the CRM: a project groups a
customer's quotes, contracts, hour entries and lifecycle emails into one view,
with a rolled-up project value and an activity feed. The whole feature is gated
behind a new **`projects` feature flag (default OFF)**, so existing installs are
unaffected until an admin opts in under **Settings → Features**.

Also bundles a few CRM email/localization fixes the cockpit's email-preview
relies on (kept together so the diff reviews as one coherent change).

## What's included

**Project Overview feature**
- New **Projects** area: list page + per-project cockpit (`ProjectsListPage`,
  `ProjectCockpitPage`), plus a gated CRM nav entry.
- Link **quotes & contracts** to a project; linking a quote/contract **cascades
  the whole deal** (via `deal_uuid` lineage) into the project for a precise
  rollup. Project pickers added to the quote/contract/hours editors.
- **Hour entries** can be attached to a project; attach-event control in the
  cockpit.
- **Rolled-up project value**: newest stage wins per deal, cumulative across the
  project; PG amount coercion for correct numeric totals.
- **Activity feed**: clickable milestones/feed, email rollup grouped by customer,
  scoped to CRM document types; re-rendered emails flagged in the feed.
- **Email preview** in the cockpit: fully read-only (no clickable links), renders
  with the email's own brand colors (not forced light), long URLs wrapped.
- Permissions: project writes use the real `events.edit` permission; "book to
  project" scoped to the current customer; cockpit rows only link when the target
  feature is enabled.

**Supporting CRM fixes (base of the stack)**
- Outlook-proof email design + PR #603 review follow-ups.
- Recent-activity email placeholder + customer-dashboard locale dates.
- Localized scheduled-send + installment dates + timezone picker.

## Schema / migrations (117–121)

All idempotent (`hasColumn`/`hasTable`-guarded), sequential on top of beta's 116:

| # | Migration | Change |
|---|-----------|--------|
| 117 | `add_projects` | New `projects` table (`name`, nullable `customer_account_id` → SET NULL on customer erase, `status` default `active`, timestamps, indexed) |
| 118 | `add_project_id_to_hour_entries` | `customer_hour_entries.project_id` |
| 119 | `add_rendered_html_to_email_queue` | `email_queue.rendered_html` (persist sent email HTML) |
| 120 | `seed_projects_feature_flag` | Seeds `projects` flag **default OFF** (inserts only if missing) |
| 121 | `add_project_id_to_quotes_contracts` | `project_id` on quotes + contracts; backfills customers that own exactly one project |

## Backward compatibility

- **Default OFF**: no UI/behavior change for existing installs until opted in.
- Migrations are idempotent and additive; `project_id` columns are nullable.
- Backfill in 121 only auto-assigns the unambiguous case (customer with exactly
  one project) — never guesses.

## Testing

- `npm run build` (frontend) + backend `migrate:safe` clean on top of `beta`.
- Verified flag-off path leaves existing CRM surfaces unchanged; flag-on path
  links quotes/contracts/hours and rolls up project value correctly.

## Notes for reviewers

- Merges into `beta` with **no conflicts**; migrations slot in at 117–121 with no
  numbering collisions.
- Translations: only `de.json` / `en.json` updated (both natively reviewed);
  nl/pt/ru untouched.

### DIFF
--- a/backend/migrations/core/117_add_projects.js
+++ b/backend/migrations/core/117_add_projects.js
@@ -47,29 +47,34 @@ exports.up = async function (knex) {
     });
   }
 
-  // 3. Backfill one auto-project per still-unassigned event.
+  // 3. Backfill one auto-project per still-unassigned event. Wrapped in a
+  //    single transaction so a heavy install (10k+ events) can't be left
+  //    half-assigned if the loop dies mid-way — it's all-or-nothing, and a
+  //    re-run still no-ops (gated on whereNull('project_id')).
   if ((await knex.schema.hasTable('events')) && (await knex.schema.hasColumn('events', 'project_id'))) {
-    const events = await knex('events').whereNull('project_id').select('id', 'event_name');
     const hasAssignments = await knex.schema.hasTable('event_customer_assignments');
-    for (const ev of events) {
-      let customerId = null;
-      if (hasAssignments) {
-        const rows = await knex('event_customer_assignments')
-          .where({ event_id: ev.id })
-          .select('customer_account_id');
-        if (rows.length === 1) customerId = rows[0].customer_account_id;
+    await knex.transaction(async (trx) => {
+      const events = await trx('events').whereNull('project_id').select('id', 'event_name');
+      for (const ev of events) {
+        let customerId = null;
+        if (hasAssignments) {
+          const rows = await trx('event_customer_assignments')
+            .where({ event_id: ev.id })
+            .select('customer_account_id');
+          if (rows.length === 1) customerId = rows[0].customer_account_id;
+        }
+        const name = (ev.event_name && String(ev.event_name).trim()) || `Event ${ev.id}`;
+        const inserted = await trx('projects').insert({
+          name,
+          customer_account_id: customerId,
+          status: 'active',
+          created_at: knex.fn.now(),
+          updated_at: knex.fn.now(),
+        }).returning('id');
+        const projectId = (inserted[0] && typeof inserted[0] === 'object') ? inserted[0].id : inserted[0];
+        await trx('events').where({ id: ev.id }).update({ project_id: projectId });
       }
-      const name = (ev.event_name && String(ev.event_name).trim()) || `Event ${ev.id}`;
-      const inserted = await knex('projects').insert({
-        name,
-        customer_account_id: customerId,
-        status: 'active',
-        created_at: knex.fn.now(),
-        updated_at: knex.fn.now(),
-      }).returning('id');
-      const projectId = (inserted[0] && typeof inserted[0] === 'object') ? inserted[0].id : inserted[0];
-      await knex('events').where({ id: ev.id }).update({ project_id: projectId });
-    }
+    });
   }
 };
 
--- a/backend/migrations/core/117_add_projects.js
+++ b/backend/migrations/core/117_add_projects.js
@@ -1,0 +1,85 @@
+/**
+ * Migration: Projects — an admin-only grouping layer ABOVE events
+ * (Project Overview cockpit, Model A).
+ *
+ * A project groups one OR MORE events of (usually) one customer; all the
+ * money documents (quotes/contracts/invoices) stay attached to their EVENT
+ * and the project simply rolls them up. Customers never see projects.
+ *
+ *   projects            id, name, customer_account_id (nullable), status,
+ *                       timestamps.
+ *   events.project_id   FK → projects (nullable, SET NULL on project delete).
+ *
+ * Backfill: every existing event gets its OWN auto-created project (the
+ * 1:1 default) so nothing is unassigned; admins then relink freely (group
+ * several events under one project, move events between projects). The
+ * auto-project's customer = the event's single assigned customer when there
+ * is exactly one, else NULL (admin sets it later). Cardinality is 1:N — the
+ * per-event auto-project is only the starting point, never a hard rule.
+ *
+ * Idempotent: table + column guarded; backfill touches only events whose
+ * project_id is still NULL, so a re-run is a no-op.
+ */
+
+exports.up = async function (knex) {
+  // 1. projects table
+  if (!(await knex.schema.hasTable('projects'))) {
+    await knex.schema.createTable('projects', (table) => {
+      table.increments('id').primary();
+      table.string('name', 255).notNullable();
+      // Nullable: a multi-customer or not-yet-assigned project has no single
+      // customer. SET NULL so erasing a customer doesn't delete the project.
+      table.integer('customer_account_id').unsigned()
+        .references('id').inTable('customer_accounts').onDelete('SET NULL');
+      table.string('status', 24).notNullable().defaultTo('active');
+      table.timestamp('created_at').defaultTo(knex.fn.now());
+      table.timestamp('updated_at').defaultTo(knex.fn.now());
+      table.index(['customer_account_id']);
+    });
+  }
+
+  // 2. events.project_id
+  if ((await knex.schema.hasTable('events')) && !(await knex.schema.hasColumn('events', 'project_id'))) {
+    await knex.schema.alterTable('events', (table) => {
+      table.integer('project_id').unsigned()
+        .references('id').inTable('projects').onDelete('SET NULL');
+      table.index(['project_id']);
+    });
+  }
+
+  // 3. Backfill one auto-project per still-unassigned event.
+  if ((await knex.schema.hasTable('events')) && (await knex.schema.hasColumn('events', 'project_id'))) {
+    const events = await knex('events').whereNull('project_id').select('id', 'event_name');
+    const hasAssignments = await knex.schema.hasTable('event_customer_assignments');
+    for (const ev of events) {
+      let customerId = null;
+      if (hasAssignments) {
+        const rows = await knex('event_customer_assignments')
+          .where({ event_id: ev.id })
+          .select('customer_account_id');
+        if (rows.length === 1) customerId = rows[0].customer_account_id;
+      }
+      const name = (ev.event_name && String(ev.event_name).trim()) || `Event ${ev.id}`;
+      const inserted = await knex('projects').insert({
+        name,
+        customer_account_id: customerId,
+        status: 'active',
+        created_at: knex.fn.now(),
+        updated_at: knex.fn.now(),
+      }).returning('id');
+      const projectId = (inserted[0] && typeof inserted[0] === 'object') ? inserted[0].id : inserted[0];
+      await knex('events').where({ id: ev.id }).update({ project_id: projectId });
+    }
+  }
+};
+
+exports.down = async function (knex) {
+  if ((await knex.schema.hasTable('events')) && (await knex.schema.hasColumn('events', 'project_id'))) {
+    await knex.schema.alterTable('events', (table) => {
+      table.dropColumn('project_id');
+    });
+  }
+  if (await knex.schema.hasTable('projects')) {
+    await knex.schema.dropTable('projects');
+  }
+};
--- a/backend/migrations/core/118_add_project_id_to_hour_entries.js
+++ b/backend/migrations/core/118_add_project_id_to_hour_entries.js
@@ -1,0 +1,30 @@
+/**
+ * Migration: book logged hours to a project.
+ *
+ * Adds customer_hour_entries.project_id (nullable FK → projects, SET NULL).
+ * Hours stay primarily customer-scoped; the optional project link powers the
+ * "book to project" checkbox + the Project Overview hours roll-up. Null =
+ * not booked to a project (existing behaviour preserved).
+ *
+ * Idempotent: column guarded by hasColumn.
+ */
+
+exports.up = async function (knex) {
+  if (!(await knex.schema.hasTable('customer_hour_entries'))) return;
+  if (!(await knex.schema.hasColumn('customer_hour_entries', 'project_id'))) {
+    await knex.schema.alterTable('customer_hour_entries', (table) => {
+      table.integer('project_id').unsigned()
+        .references('id').inTable('projects').onDelete('SET NULL');
+      table.index(['project_id']);
+    });
+  }
+};
+
+exports.down = async function (knex) {
+  if (!(await knex.schema.hasTable('customer_hour_entries'))) return;
+  if (await knex.schema.hasColumn('customer_hour_entries', 'project_id')) {
+    await knex.schema.alterTable('customer_hour_entries', (table) => {
+      table.dropColumn('project_id');
+    });
+  }
+};
--- a/backend/migrations/core/119_add_rendered_html_to_email_queue.js
+++ b/backend/migrations/core/119_add_rendered_html_to_email_queue.js
@@ -1,0 +1,31 @@
+/**
+ * Migration: store the rendered email HTML at send time.
+ *
+ * The Project Overview cockpit previews the ACTUAL email that was sent (not a
+ * re-render from the current template, which may have changed). email_queue
+ * only stored the template variables (email_data), so add a rendered_html
+ * column the sender populates with the final wrapped HTML on dispatch.
+ *
+ * Nullable: rows queued/sent before this column existed have no stored HTML —
+ * the cockpit reconstructs those from email_data with a "reconstructed" note.
+ *
+ * Idempotent: column guarded by hasColumn.
+ */
+
+exports.up = async function (knex) {
+  if (!(await knex.schema.hasTable('email_queue'))) return;
+  if (!(await knex.schema.hasColumn('email_queue', 'rendered_html'))) {
+    await knex.schema.alterTable('email_queue', (table) => {
+      table.text('rendered_html');
+    });
+  }
+};
+
+exports.down = async function (knex) {
+  if (!(await knex.schema.hasTable('email_queue'))) return;
+  if (await knex.schema.hasColumn('email_queue', 'rendered_html')) {
+    await knex.schema.alterTable('email_queue', (table) => {
+      table.dropColumn('rendered_html');
+    });
+  }
+};
--- a/backend/migrations/core/120_seed_projects_feature_flag.js
+++ b/backend/migrations/core/120_seed_projects_feature_flag.js
@@ -1,0 +1,24 @@
+/**
+ * Migration: seed the `projects` feature flag (default OFF).
+ *
+ * Gates the admin-only Project Overview cockpit + the "book to project" hours
+ * control. Off by default so existing installs don't suddenly surface a new
+ * top-level CRM area — the admin opts in under Settings → Features, exactly
+ * like bills/quotes/contracts/hours.
+ *
+ * Idempotent: inserts only when the row is missing (migration 088 already
+ * seeded the original flag set on fresh installs and won't re-run).
+ */
+
+exports.up = async function (knex) {
+  if (!(await knex.schema.hasTable('feature_flags'))) return;
+  const existing = await knex('feature_flags').where({ key: 'projects' }).first();
+  if (!existing) {
+    await knex('feature_flags').insert({ key: 'projects', value: false });
+  }
+};
+
+exports.down = async function (knex) {
+  if (!(await knex.schema.hasTable('feature_flags'))) return;
+  await knex('feature_flags').where({ key: 'projects' }).del();
+};
--- a/backend/migrations/core/121_add_project_id_to_quotes_contracts.js
+++ b/backend/migrations/core/121_add_project_id_to_quotes_contracts.js
@@ -1,0 +1,65 @@
+/**
+ * Migration: link quotes + contracts to a project.
+ *
+ * Quotes and contracts carry no event_id (see migration 107), so the Project
+ * Overview cockpit originally rolled them up by the project's customer — which
+ * is imprecise once a customer has more than one project. This adds an explicit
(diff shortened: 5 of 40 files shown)


the-luap:
Reviewed end-to-end. The feature-flag gating is clean (route layer 403 + frontend `RequireFeature` redirect), migrations 117-121 are textbook-defensive (`hasTable`/`hasColumn`-guarded, additive, idempotent, irreversible `down()` only on the nominal seed), and the PG `Number()` coercion is consistently applied to `total_amount_minor` / `paid_amount_minor` / `duration_minutes` / `event_count` — no #596-style bigint-as-string slip. Two cross-customer leak blockers I'd want closed before this merges, plus a small set of non-blocking concerns.

## 🔴 Blockers

### 1. `assignDocument` lets admins link a quote/contract to a project that belongs to a different customer

`backend/src/services/projectService.js:196-211`:

```js
async function assignDocument(table, projectId, documentId) {
  ...
  if (projectId != null) {
    const project = await db('projects').where({ id: projectId }).first();
    if (!project) throw new AppError('Project not found', 404);
  }
  const doc = await db(table).where({ id: documentId }).first();
  if (!doc) throw new AppError('Document not found', 404);
  await db(table).where({ id: documentId }).update({ project_id: projectId || null });
  if (projectId && doc.deal_uuid) {
    await linkDealToProject(doc.deal_uuid, projectId);
  }
  ...
}
```

No check that `project.customer_account_id === doc.customer_account_id`. The route gate is just `events.edit`, so any admin with that perm can attach Customer A's quote to Customer B's project. Then `linkDealToProject` cascades — re-pointing Customer A's events + invoices into Customer B's cockpit. Customer B's rolled-up project value now reflects Customer A's totals.

The hours endpoint already has this guard (`customerHoursService.js:240-250`, error code `PROJECT_CUSTOMER_MISMATCH`). The same guard belongs here:

```js
if (
  project.customer_account_id != null &&
  doc.customer_account_id != null &&
  project.customer_account_id !== doc.customer_account_id
) {
  throw new AppError('Document belongs to a different customer than the project', 409);
}
```

### 2. `linkDealToProject` re-points events and adopts customers across tenants

`backend/src/services/projectService.js:151-194`. Combined with #1, dropping a quote whose `deal_uuid` lineage spans customers re-points events across customers silently. Even standalone, the "adopt customer onto a still-unassigned project" branch promotes `projects.customer_account_id` from null → arbitrary deal customer without any permission check beyond `events.edit` — customer-assignment is normally a `customer.manage`-class write. Add the same-customer assertion before the bulk `events`/`projects` updates AND before the adoption write.

## 🟡 Concerns (non-blocking, listing so they're tracked)

1. **`projectService.js:466-471` — `sendEmailNow` calls `processEmailQueue({ ignoreSchedule: true, limit: 50 })`** which in `emailProcessor.js` drops both the schedule filter AND the `retry_count < 3` guard. Clicking "Send now" on email #42 force-retries up to 49 OTHER stuck emails that exceeded retry cap. Either narrow the flush to `WHERE id = ?` or surface this in the button copy.

2. **`projectService.js:434-446` — `resendEmail` inserts `row.email_data` verbatim into a `jsonb` (PG) / `text` (SQLite) column.** If the source row's value type and the insert column type ever disagree (jsonb auto-parses to object on PG, stays string on SQLite), the resent row's payload can end up double-encoded. Either explicit `JSON.stringify(...)` when it's an object or `db.raw('?::jsonb', [...])`.

3. **`projectService.js:299-313` — email-rollup scoped by `recipient_email` string rather than `customer_account_id` join.** Two customers sharing one address (family addresses are common) cross-contaminate the cockpit feed. Lower severity since admins see both, but the scoping should mirror the way quotes/invoices are scoped.

4. **Migration 117 backfill is O(N) with per-row inserts and no `knex.transaction` wrapper.** Re-runnable (gates on `whereNull('project_id')`), but on a 10k-event install that's 10k+ round-trips and a half-applied state if it dies mid-loop. Wrap in `knex.transaction`.

5. **`PROJECT_CUSTOMER_MISMATCH` has no client-side translated message.** Raw 422 surfaces as untranslated toast. Worth wiring an `i18n` key + the standard error-toast format.

6. **No `activity_logs` entry for project-side email resend / sendNow / cancel.** Other CRM mutations log uniformly; these admin email actions should too.

7. **`nl/pt/ru/es/sl/fr` translations missing for the cockpit + the new hours `bookToProject` toast.** Admin-only + feature-flagged so acceptable, but worth release-notes mention.

8. **`adminProjects.js:78`** — `body('customerAccountId').optional({ values: 'null' }).isInt(...)` — `{ values: 'null' }` isn't a standard express-validator option name. If you mean "accept JSON null in the body", that's `optional({ nullable: true })`. Verify with a test that sending `{"customerAccountId": null}` actually reaches the handler.

9. **Pre-121-install list page shows valuation = 0 for all projects** because `attachValuations` (`projectService.js:79-83`) only loads quote rows when `hasColumnCached('quotes','project_id')` is true. Detail page falls back to customer-scoping; list does not. Either parallel fallback or surface `—` explicitly.

## 🟢 Nits

- `adminProjects.js:42-44` vs `:130` — list valuation perms gate `bills`+`quotes` but detail also gates `contracts`. Contracts contribute no totals, so inconsistent rather than wrong. Align.
- `projectService.js:367-376` — milestone "first quote" picks `out.quotes[out.quotes.length - 1]`. `Array.at(-1)` reads cleaner.
- `emailProcessor.js:872-874` — `require('../utils/schemaCache')` inside the loop body. Lift to module scope.
- `ProjectsListPage.tsx` shows `0` for `eventCount ?? 0` but `—` for valuation when bucket totals are zero. Pick one convention.

## ✅ Verified

- Feature flag: server-side `requireProjectsFlag` returns 403 when off; frontend `RequireFeature` redirects. API not reachable, URL not reachable. No backdoor.
- Migrations 117-121: idempotent, additive, `hasColumn`/`hasTable`-guarded, sequential on top of beta's 116. No collisions. Down migrations safe.
- PG bigint coercion: `Number(...)` applied at every `*_amount_minor` / `*_count` / `duration_minutes` read. No #596 lurking.
- Hours attachment: `customerHoursService.js:240-250` correctly rejects `PROJECT_CUSTOMER_MISMATCH`. The defense exists for hours — just not for quotes/contracts (which is what blocker #1 is).
- Email preview iframe: `sandbox=""` is the most restrictive value (no scripts, forms, popups, top-nav). Even if a template body is compromised, no XSS via the cockpit preview.
- `htmlToText` (used in the preview text fallback) strips `<style>` + `<script>` before tag-stripping.
- `de.json` + `en.json` cover every cockpit string used in the page; mirror each other.
- No `restore_runs`-style audit-wipe class of bug — N/A here, no restore code.

## Verdict

Close blockers 1 + 2 (the same `customer_account_id` mismatch guard, applied to both `assignDocument` and the adoption branch of `linkDealToProject`) and I'm ready to merge from my side. The 9 concerns are real but each is a follow-up, not individually blocking. Migration discipline is excellent — same pattern as #603's 110-116 set. Feature flag enforcement is the cleanest I've seen in this codebase.

the-luap:
Adding a clarifying question on Blocker #1 and a bit more detail on each concern so you have the failure mode in front of you when you triage.

## On Blocker #1 — design intent or oversight?

Re-reading `linkDealToProject` the "adopt customer onto a still-unassigned project" branch (lines 188-193) clearly shows you thought about *one* edge case: a brand-new project with no customer that picks up its identity from the first dropped deal. That's a real workflow.

What's NOT obvious from the code: did you also *intend* to allow attaching a doc when the project + doc have **different** customers assigned, or did that case just not get guarded? Two reasons it might be intentional:

- B2B parent/subsidiary: one customer holds the master contract, another receives some invoices. Wedding-planner-on-behalf-of-couple is the same shape on the consumer side.
- "Move this doc from Customer A to Customer B" as an admin recovery flow (someone created the doc under the wrong customer).

If either of those is on your roadmap, the right shape is probably not "block at the service layer" but rather "warn at the UI + require an explicit `--cross-customer` confirmation + write a separate `activity_logs` entry" so the cross-customer move is intentional and auditable. The current code allows it silently, which is the part that's worrying — not the existence of the capability.

If neither is on your roadmap, the same-customer guard from `customerHoursService.js:240-250` (`PROJECT_CUSTOMER_MISMATCH`) drops in cleanly.

Same question applies verbatim to Blocker #2 — `linkDealToProject` adopting an arbitrary customer onto an unassigned project: intentional ("first deal wins") or oversight?

## On the concerns — expanded with failure modes

### 1. `sendEmailNow` bypassing the retry cap on OTHER emails

Concrete failure: admin runs out two welcome emails in the morning, both fail to send (SMTP password expired). Each accumulates `retry_count = 3` and is now dead-letter. Admin fixes the password mid-afternoon and clicks "Send now" on the third email in the cockpit. `processEmailQueue({ ignoreSchedule: true, limit: 50 })` picks up *all three* — the targeted one AND the two zombie failures. The two zombies retry and send. Net result: the welcome email those customers thought was lost shows up hours late with the original timestamp.

Fix shape: pass a `whereId` filter through to `processEmailQueue`, or skip the queue helper entirely and run the single send inline.

### 2. `email_data` jsonb double-encoding

Failure window: only on PG (jsonb auto-parses to object), only when the original sender stored the payload as a JS object (not a JSON string). The resent row's `email_data` is now `JSON.stringify(JSON.stringify(payload))` because Knex sees a string going into a jsonb column and quotes it. A subsequent send tries to read `data.recipient_name` and gets undefined because the column value is `"\"{\\\"recipient_name\\\":...\""` instead of `{recipient_name: ...}`. Manifests as a silently broken email.

Fix shape: explicit `db.raw('?::jsonb', [typeof row.email_data === 'string' ? row.email_data : JSON.stringify(row.email_data)])` or normalise on read.

### 3. Email rollup scoped by `recipient_email`

Concrete failure: family wedding photo shoot. Parent A is "customer_a" with email `family@example.com`. Parent B is "customer_b" with same `family@example.com`. Both have CRM quotes and invoices. Open Customer A's project cockpit → the email rollup query at `:299-313` returns EVERY `quote_*` / `contract_*` / `invoice_*` email ever sent to `family@example.com`, including Customer B's. The rest of the cockpit (quotes/contracts/hours) is correctly scoped by `customer_account_id` join, but the email feed leaks.

Fix shape: join the `email_queue` rows back to whatever table the email references (`quote_*` → quotes.customer_account_id, `invoice_*` → invoices.customer_account_id, etc.) and filter on that, not on the recipient string.

### 4. Migration 117 backfill non-transactional

On a clean install this doesn't matter (no existing events). The risk is on heavy installs upgrading from beta. If the loop dies at row 5000 of 10000:
- DB has 5000 events with auto-projects + 5000 without
- The migrations tracking row for 117 is NOT yet inserted (migration didn't finish)
- Next boot's `migrate:safe` re-runs 117 → because the loop is gated on `whereNull('project_id')` it just picks up where it left off ✓

So it's recoverable. But the symptom while it's half-done — half your events show up in projects, half don't — is confusing. Wrapping in `knex.transaction(async trx => { ... })` makes it all-or-nothing.

### 5. Untranslated `PROJECT_CUSTOMER_MISMATCH`

The hours service throws this code, the frontend toast displays it raw because there's no `i18n` key mapping. Admin sees literally `PROJECT_CUSTOMER_MISMATCH` instead of "This hour entry's customer doesn't match the project's customer." Tiny UX hit — but once the Blocker #1 fix lands, the same code will surface on the quote/contract attach path too, so it'll be more visible.

### 6. No `activity_logs` for email-side admin actions

Other CRM mutations (`mark-paid`, `cancel`, `reissue`, `send-reminder`) all `await logActivity(...)`. The new `resendEmail` / `sendEmailNow` / `cancelEmail` in `projectService.js` don't. If a customer asks "who re-sent my welcome email at 2 AM?", the audit trail has no answer. Same shape as the other mutations — `logActivity('project_email_resent', { eventId, queueId }, projectId, adminCtx)`.

### 7. Translations

Just a release-notes mention. The feature is admin-only + flag-gated, so the hit surface is small. Mentioning here mainly so it doesn't surprise anyone with a non-DE/EN-speaking admin team.

### 8. `optional({ values: 'null' })` validator

`express-validator`'s `optional()` options are `{ nullable: true, checkFalsy: true, values: 'falsy' | 'null' | 'undefined' | 'falsy' }`. So `{ values: 'null' }` *is* valid syntax — it means "treat the string `'null'` as if the field were omitted". Probably not what you want if you're posting JSON. If you want to accept JSON `null`, use `{ nullable: true }`. If you want to accept both `null` (JSON) and the string `'null'` (form-encoded), `{ nullable: true, values: 'null' }` is the union. A quick `curl -d '{"customerAccountId": null}'` test reveals which behavior is current.

### 9. Pre-121 list page valuation = 0

Specifically: an admin who's upgraded the codebase but hasn't run migrations yet (or is running them now and is between 120 and 121) opens the Projects list — every row shows `—` for valuation. Detail page works because of the customer-scoping fallback you wrote at `:344-355`. List page doesn't have that fallback so the list shows zero everywhere.

This is a transient state (it goes away the moment 121 lands), but a self-hosted admin running the upgrade interactively will see "all my valuations are zero" for the duration of the migration. Either mirror the customer-fallback into the list query OR explicitly render `Calculating…` until the column exists.

---

Nits 1-4 are pure code-style polish; they don't affect behaviour at all. Happy to leave those off the table if you'd rather ship.

Once you confirm the design intent on Blockers #1/#2 (and ideally fix or scope them out), I'll re-verify on the next push.

Luca-Timo:
## Review follow-up — pushed

@the-luap thanks for the thorough pass. All blockers and actionable concerns are addressed across three commits (`9d13880f`, `4b1e85c8`, `f74d8d4e`). Design-intent answer up front, then the point-by-point.

### Design intent on Blockers #1 & #2

**Projects are single-customer.** A project is pinned to one customer; the cross-customer cascade you flagged was a real oversight, not an intended capability. No B2B/parent-subsidiary flow.

But the *enforcement* is **"addable if one customer matches,"** not strict first-customer equality — because some content legitimately touches more than one account:

- **Events** are many-to-many with customers. An event attaches if **any** of its customers is the project's customer.
- **Quotes/contracts/hours** each carry one customer → must equal the project's customer.
- **Deal cascade** (`linkDealToProject`) collects **all** customers across the quote→contract→invoice lineage and rejects only when **none** matches.

The project itself never holds a second customer: an empty project adopts the first content's customer, and `updateProject` refuses to re-label to a customer that conflicts with content already linked.

While closing this I found the original guard only covered quotes/contracts/hours — **`assignEvent` was unguarded** (you could attach a foreign-customer event) and the project's customer could be **re-labelled** out from under existing content. Both are now closed (`4b1e85c8`). Worth a re-verify since these are beyond the original diff.

### 🔴 Blockers

| # | Status |
|---|---|
| 1 `assignDocument` cross-customer | ✅ Closed — guard at the boundary; editors actually link via `quoteService`/`contractService` → `linkDealToProject`, so the guard lives at that chokepoint too. |
| 2 `linkDealToProject` re-point/adopt | ✅ Closed — membership check before any write; adoption only onto a customer-less project. |

### 🟡 Concerns

| # | Status |
|---|---|
| 1 `sendEmailNow` bypasses retry cap | ✅ `processEmailQueue` gained `onlyId`; the flush targets the single row. |
| 2 `email_data` double-encode | ✅ `resendEmail` re-stringifies the PG-parsed object to match the canonical enqueue. |
| 3 email rollup by `recipient_email` | ✅ Feed now scoped to the project's own document numbers (event_id for gallery mail; `email_data` doc-number match for CRM mail). Shared inbox no longer leaks. |
| 4 migration 117 non-transactional | ✅ Backfill wrapped in `knex.transaction` (adds atomicity on SQLite; PG already wraps the whole migration). |
| 5 untranslated `PROJECT_CUSTOMER_MISMATCH` | ✅ `projects.error.customerMismatch` (de+en), wired into hours/quote/contract/event handlers. |
| 6 no `activity_logs` for email actions | ✅ resend/cancel/retry/sendNow log `project_email_*` with adminId. |
| 7 nl/pt/ru/es/sl/fr translations | ⚠️ **Open by design** — only de/en touched (native-reviewed). Admin-only + flag-gated; flagging for release notes. |
| 8 `optional({ values: 'null' })` validator | ✅ → `optional({ nullable: true })`. |
| 9 pre-121 list valuation = 0 | ✅ List falls back to customer-scoped quotes during the upgrade window. |

### 🟢 Nits

- #2 `Array.at(-1)` ✅ · #3 in-loop `require` removed ✅
- #1 perms split & #4 `0` vs `—` — kept as-is with clarifying comments: contracts carry no totals (a contract just renders the quote's line items), and a zero *count* shows `0` while a zero *value* shows `—` (since `CHF 0.00` would imply a real zero-value deal).

Build + `node --check` green; migrations clean on top of beta. Ready for your re-verify.

the-luap:
Re-verified the three commits. Spot-checks confirm all blockers + concerns landed, and the two bonus catches you found while fixing (`assignEvent` + `updateProject` re-label) are exactly the kind of follow-on holes that would have surfaced on the next post-merge bug report.

## Design intent — crisp

The "single-customer project + 'one customer matches' for many-to-many content" rule is the right shape and it reads cleanly:

- **`projectLinkedCustomerIds`** (`projectService.js:147-165`) — a single helper that collects distinct customer ids from `event_customer_assignments` ∪ quotes ∪ contracts already linked to the project. Used by `updateProject` to refuse a re-label that would conflict, and as the conceptual source of truth for "what does this project hold."
- **Strict equality** on documents (quotes / contracts / hours) — they carry one customer, equality is the right check (`assignDocument:311-320`).
- **Membership** on events + deal lineage — `assignEvent` uses `eventCustomerIds` (`:204-211`) to handle many-to-many; `linkDealToProject` collects `dealCustomerIds` across the whole quote→contract→invoice lineage and rejects only when *none* matches (`:262-276`).
- **Empty project adopts** the first content's customer in both `assignEvent` (when the event has exactly one customer) and `linkDealToProject` (taking the first lineage customer). Adoption is the only path that mutates `projects.customer_account_id` and it's gated on `customer_account_id == null`. Clean.

## 🔴 Blockers — closed

| # | Verified |
|---|---|
| 1 `assignDocument` cross-customer guard | ✅ `projectService.js:311-320` — strict customer equality, throws `PROJECT_CUSTOMER_MISMATCH` (422) before the document write. |
| 2 `linkDealToProject` membership + adoption-only-onto-null | ✅ `:262-276` — collects every deal customer BEFORE any write, rejects when none matches the project's customer. Adoption gate (`project.customer_account_id == null`) at `:285-289`. |

## Bonus finds you closed while fixing

- **`assignEvent` was unguarded** — verified the new code at `:213-231`: project with a customer rejects a non-matching event; empty project adopts the event's single customer; event with multiple customers + an empty project is intentionally NOT adopted (would otherwise pick arbitrarily). Right call.
- **`updateProject` re-label** — verified at `:166-191`: any re-label is checked against `projectLinkedCustomerIds(id)`, rejected with the same `PROJECT_CUSTOMER_MISMATCH` code if there's a conflict. Clearing to null is allowed (you can "park" a project), which is the right escape hatch.

Both of these were exactly the kind of "fix one, find two" follow-on that would have shown up on the next incident report. Good catch.

## 🟡 Concerns — closed

Spot-checked the load-bearing ones:

- **#1 `sendEmailNow`** — `processEmailQueue({ ignoreSchedule: true, onlyId: emailId })` at `:474-475`. Comment block at the call site explicitly references the "don't force-retry OTHER dead-lettered emails" motivation. ✅
- **#2 `email_data` re-stringify** — `if (emailData != null && typeof emailData !== 'string') emailData = JSON.stringify(emailData);` at `:442-444`. Handles both PG-object and SQLite-string cases. ✅
- **#4 migration 117 transactional** — `await knex.transaction(async (trx) => { ... })` wrapping the per-event backfill. Half-applied state on heavy installs now impossible. ✅
- **#5 `PROJECT_CUSTOMER_MISMATCH` i18n** — `projects.error.customerMismatch` in en.json + de.json with the right message shape, wired into the hours/quote/contract/event handlers. ✅
- **#6 `activity_logs`** — `logEmailAction` helper called from `resend` / `cancel` / `retry` / `sendNow` with `project_email_*` activity types. Best-effort wrap so a logging failure doesn't block the action itself. ✅
- **#8 validator** — UPDATE route uses `optional({ nullable: true })` with the explanatory comment; CREATE route still uses `{ values: 'falsy' }` (treats empty string / 0 / false as omitted). The split is deliberate and reads correctly — accepting JSON `null` only makes sense on update (clear the existing customer), not on create.

I trust #3 (email rollup scoped to project's own doc numbers), #7 (open by design), and #9 (pre-121 list fallback) at the description level — they're well-explained and consistent with the rest of the architecture.

## 🟢 Nits

- `Array.at(-1)` ✅
- Module-scope `require` ✅
- Perms split + `0` vs `—` empty-state kept with explanatory comments — both defensible (contracts don't carry totals; count-zero vs value-zero are semantically distinct). Fine either way.

## Verdict

Two real blockers closed, two bonus single-customer holes closed alongside them, all actionable concerns addressed. Architecture documentation in the comments is some of the cleanest I've seen in this codebase — `projectLinkedCustomerIds` as a single source of truth for "what does this project hold" makes the invariant easy to reason about and easy to extend.

From my side: merge cleared.

Issue #607, opened 2026-06-07
[BUG] Umlauts dropped on download (filenames and gallery names)


patchingfailed:
**Version:** 3.44 (stable)
**Browser/OS:** Chrome on Windows

### Description
Umlauts are handled inconsistently between display/URL and file download.
On upload, umlauts are transliterated to ASCII for the URL (e.g. `Ä` → `A`),
but on download the umlaut character is silently dropped entirely instead of
being transliterated or properly encoded. This affects both the image
filename and the gallery name.

### Steps to Reproduce
1. Create a gallery and upload an image, both containing an umlaut, e.g.
   gallery `Ägypten` / file `Ägypten_individual_0050.jpg`
2. Open the gallery — the URL slug renders the name as `Agypten_...` (Ä → A)
3. Download the image

### Expected
The downloaded name preserves the character consistently — either as a proper
UTF-8 name (`Ägypten_individual_0050.jpg`) or the same ASCII transliteration
used in the URL (`Agypten_...` / `Aegypten_...`).

### Actual
The leading umlaut is dropped entirely; the downloaded file is named
`gypten_individual_0050.jpg` (the `Ä` produces no character at all). The same
dropping occurs for the gallery name.

### Notes
Likely a `Content-Disposition` encoding issue: filenames set via
`filename="..."` with raw non-ASCII bytes get stripped, rather than using the
RFC 6266 `filename*=UTF-8''...` form (with an ASCII `filename="..."` fallback).
The URL slug path transliterates, so display and download handle non-ASCII
differently — but the exact cause should be confirmed.

the-luap:
Thanks for the careful report — your bisect on the symptom ("leading umlaut dropped entirely, not transliterated, while the URL transliterates to `Agypten`") was the key clue.

The root cause sits one layer earlier than the Content-Disposition encoding you hypothesised. The download header is actually fine — `buildContentDisposition` in `backend/src/utils/filenameSanitizer.js` emits both an ASCII fallback `filename="…"` AND an RFC 5987 `filename*=UTF-8''…` unicode form, which Chrome correctly honours. The bug is that by the time we hit the download path, the string we're encoding has *already* lost the umlaut at the DB layer.

[...]

PR #622, opened 2026-06-16
Accounting suite + CRM hardening

### LINKED ISSUE #620: [BUG] some of CRM tab is following Dark mode if you on light mode
go to CRM tab in light mode :

Image

Some options are following Dark mode like this : 

Image

Luca-Timo:
Adds a dedicated **Accounting** area to picpeak and hardens the surrounding CRM. Everything new sits behind the `accounting` master feature flag (and sub-toggles), **off by default** — existing installs are unchanged until an admin opts in. Tax/VAT scope is **Swiss/LI rails** (QR-bill, MWST, Banana/bexio export — not German DATEV/ELSTER). All legal/financial defaults carry "verify with your Treuhänder / lawyer / bank" disclaimers.

### 1. New Accounting section
A top-level **Accounting** nav (separate from CRM), gated by a master flag with sub-toggles `incomingInvoices`, `expenses`, `taxReport`. The Tax report moved permanently out of CRM into here.

- **Inbound supplier invoices** — capture by upload or camera; **PDFs are rasterised server-side in an isolated step and never served raw to the browser** (CSP-locked images only); triage view opens at the QR-bill; QR decoded server-side and treated as untrusted; disposition actions (re-bill to client / company expense / etc.) + supplier-payment toggle.
- **Internal expenses** — mileage / per-diem / cash booked to an event or the company, optional proof file; rates configured in Settings → Accounting; rate snapshotted at entry time.
- **Tax report** — one unified, signed, typed, **sortable** ledger combining outgoing invoices + incoming invoices + expenses; period/currency filters; PDF + CSV export. Cancelled invoices shown but excluded from totals. **VAT-payable honours VAT registration + reclaim-country settings** (output VAT − reclaimable input VAT; foreign non-reclaimable excluded).
- **Treuhänder / collective-journal export** — lives on the Tax page; **Banana-importable** (tab-separated `.txt`, UTF-8 BOM, ISO dates; double-entry **and** Income&Expense/cash-book formats) plus generic + bexio.
- **Chart of accounts + VAT codes (Layer A)** — Swiss/LI KMU-Kontenrahmen CRUD + the category/account/VAT mappings the export relies on. All of this now lives in **Settings → Accounting** (chart of accounts + VAT codes + rate maps + registration), so accounting config is in one place; the Accounting section is purely operational.

### 2. VAT consolidation
Single source of truth = the `vat_codes` registry (Swiss/LI seeds). Invoices/quotes **snapshot the chosen code** for export immutability (export prefers the snapshot, falls back to the rate→code map). Editor VAT field is a dropdown of configured output codes + "custom". The "VAT code by revenue rate" map is **data-driven** (rows derived from your output codes' rates) so it works for **any country's percentages** — retype codes to local rates and the rows follow. New **VAT registration + reclaim-countries** settings drive the report math.

### 3. Incoming mail (IMAP) — standalone `incomingMail` flag
Poll a dedicated mailbox; drop invoice attachments into Accounting → Incoming invoices. Config block mirrors SMTP styling, "Test connection" + send-to-self round-trip test, all received mail logged (90-day lookback + dedup).

### 4. CRM feature-flag audit fixes
Closed gating gaps where a feature was actable with its flag off: hours-logging routes (read/edit/delete/**bill**), `deals/installment-plan` (now `bills`-gated), customer invoice/contract PDF endpoints. Added a shared `requireFeatureFlag` middleware. New **per-customer `feature_contracts` override** (migration 131, defaults on → preserve-visuals) — the missing counterpart so admins can hide Contracts per customer like quotes/bills.

### 5. Dark/light mode + force color mode
- **Fix #620** — admin CRM/accounting surfaces were reading the gallery theme's CSS vars (`--color-surface`, `text-theme` utilities), so modals rendered dark in light mode. Repointed ~23 files to Tailwind `dark:` classes so they follow the admin toggle.
- **Force color mode** — force light/dark now only locks the light/dark palette. The **Branding page stays the full global preset** (colors/fonts/style); it just hides the redundant per-theme light/dark picker (light/dark = Force). **Per-event gallery editors** hide their colour + light/dark pickers when a site-wide lock is active (a gallery can't override it).

### 6. Settings bug fix
`PUT /admin/settings/accounting` (and customer-surface settings) inserted a non-existent `created_at` into `app_settings`, breaking first-time saves of new keys (e.g. the VAT-registration toggle). Fixed.

### Database — migrations 122–131
`122` accounting flag · `123` accounting permissions · `124` inbound_documents + expenses · `125` contract expense surcharge · `126` split incoming-vs-expenses · `127` expenses flag + accounting settings · `128` incoming mail · `129` ledger_accounts + vat_codes · `130` vat_code snapshot on quotes/invoices · `131` per-customer feature_contracts. All idempotent / hasColumn-guarded.

### Notes for review
- **Opt-in & non-destructive:** all new surfaces gated behind `accounting` (+ sub-flags), off by default.
- **i18n:** EN + DE authored natively; **nl/pt/ru/es/sl/fr fall back to English** and need native review.
- **Tests:** backend accounting/tax/feature-flag + frontend theme tests pass; `npm run build` green.
- **Scope reminder:** Swiss/LI-first; tax/VAT treatment is guidance only, surfaced with Treuhänder/lawyer/bank disclaimers.


the-luap:
Reviewed end-to-end. Strong overall — the multer `diskStorage` filename whitelist defeats the #607-class path-traversal at intake, every rasterised-PDF download route carries `default-src 'none'; img-src 'self' data:` + `nosniff` (so even a compromised PDF can't get its rendered PNG to execute), inbound raw PDFs are served `attachment; nosniff` (never as inline), IMAP host is blocked from RFC1918 on save+test+folder-list, `imap_pass` is correctly masked on GET + preserved on PATCH, the cancelled-invoice ledger treatment is correct (subtraction via storno negatives, kept-in-ledger with replacedBy link), and all the paginated counts are wrapped in `parseInt(... || 0, 10)` — no #596-class bigint trap lurking in the new code. Migrations 122-131 follow the same defensive pattern as 110-121 (hasColumn/hasTable guards, idempotent, sane `down`s).

Two blockers worth closing before merge, plus the usual non-blocking concerns. Both blockers verified in source.

## 🔴 Blockers

### 1. CSV / Banana export formula injection

`backend/src/services/ledgerService.js:393-396` (`csvEscape`) and `backend/src/services/taxReportService.js:999-1004` (`escape`) only RFC-4180 quote — neither prefixes risky leading characters:

```js
function csvEscape(cell) {
  const s = cell === null || cell === undefined ? '' : String(cell);
  return `"${s.replace(/"/g, '""')}"`;
}
```

A cell starting with `=`, `+`, `-`, `@`, `\t` or `\r` executes as a formula in Excel/Numbers when the Treuhänder opens the export. The taxReport comment even claims *"We always wrap, simpler + bulletproof for Excel"* — wrapping in quotes doesn't stop formula evaluation; only a leading-char prefix does. Vectors are real: `supplier_name`, `invoice_number`, `payment_reference`, `description` are all admin-editable (and once incoming-mail ingestion is live, sender-controlled too). The Banana export amplifies the risk because it's **tab-separated unquoted** per the PR body, so even the surrounding quotes don't apply.

```js
function csvEscape(cell) {
  let s = cell === null || cell === undefined ? '' : String(cell);
  // Excel/Numbers/Banana formula-injection defence — prepend a single quote
  // when the cell starts with a character that would otherwise be evaluated
  // as a formula. Stays compatible with the RFC 4180 quote wrap.
  if (/^[=+\-@\t\r]/.test(s)) s = `'${s}`;
  return `"${s.replace(/"/g, '""')}"`;
}
```

Apply identically in `taxReportService.escape`. Add a unit test pinning one of each leading char (worth a `.test.js` next to the service — same source-inspection style as the #596 `pgBranch` test was the right shape for that class).

### 2. `received_emails.message_id` is `INDEX` not `UNIQUE`

`backend/migrations/core/128_incoming_mail.js:42`:

```js
table.string('message_id', 512);
table.index(['message_id']);
```

The IMAP poller serialises within a single process (`let polling = false`), but that lock evaporates on a second backend replica or a k8s rolling deploy with overlap. Two workers racing the same message will both pass the check-then-insert in `expenseService.recordInboundDocument` (no row lock between SELECT and INSERT) and double-ingest the attachment.

Fix: `table.unique(['message_id'])` on the column, and switch the ingest to `INSERT ... ON CONFLICT DO NOTHING` (PG) / `INSERT OR IGNORE` (SQLite) so the race loses cleanly instead of throwing. While there: `inbound_documents.file_sha256` is the second dedup key — a partial unique index there closes the same race on the SHA path.

## 🟡 Concerns

3. **`requireFeatureFlag` does a DB read per request** — `backend/src/middleware/requireFeatureFlag.js:14-30`. Accounting routes are 10+ endpoints; the dashboard polls a few of them. 5–10s in-memory TTL cache or startup-warmed map + invalidation on `PUT /admin/feature-flags`.

4. **Customer routes only check the per-customer flag, not the master** — `backend/src/routes/customer.js:594/639/700` for `/invoices/:id/pdf`, `/contracts`, `/contracts/:id/pdf`. Admin disables `bills` globally → customer with `feature_bills=true` still downloads PDFs. Admin paths are double-gated; customers should be too. Shared helper that checks both.

5. **Tax-report "foreign VAT non-reclaimable" reads `tax_treatment` column only**, not derived from supplier-country vs. company-country (`taxReportService.js:214`). PR body claims it honours reclaim-country settings; the code honours per-row classification only. Either wire `reclaim_countries` into the row classifier or correct the docs.

6. **`accounting_vat_registered` fallback flips on `grandTotalVat > 0`** — `taxReportService.js:507`. Quarter with zero output VAT (all exempt cross-border) silently flips to "not registered" and hides the reclaim. Treat null as "not configured, refuse to compute payable" + surface a UI warning.

7. **`created_at` on `app_settings` fixed at the route layer but not at the schema layer** — `adminSettings.js:225/294` (good), migration 127 (good). Future route authors can re-introduce the bug. Either add the column to the schema or factor a shared `upsertAppSetting()` helper. The fix as-shipped is locally correct but the class isn't closed.

8. **PDF rasterisation: per-file resource cap missing** — `rasterizeService.js:51-55` has a 25s timeout + 4MB stderr cap, but `pdftoppm` is invoked per-page on demand. A 1000-page malicious PDF still walks pages one request at a time. Either cap `row.page_count` at ingest (e.g. 200) or `prlimit`-wrap the subprocess.

9. **`emailIntake.saveAttachment` stores `original_filename` to the DB unsanitised** — render-side React `{}` interpolation escapes by default, but any later `dangerouslySetInnerHTML` (e.g. an email-preview surface) would surface stored XSS from arbitrary senders. Worth one paranoia pass on the frontend.

10. **`imap_pass` stored plaintext in `email_configs`** — same pre-existing class as `smtp_pass`. This widens the inbound-mailbox-read blast radius if `app_settings`/`email_configs` ever leak (backup, DB dump). Not a regression — but worth flagging in the PR body so this is a known operator concern, not an oversight.

## 🟢 Nits

11. **`backend/storage/business-docs/quote/2026/Q-2026-0001.pdf` committed** — stray test artifact. Delete + `.gitignore`.
12. **Two inline `requireFlag` copies still in `adminExpenses.js:51-60` + `adminLedger.js`** — audit didn't migrate them to the shared middleware. Polish.
13. **Dashboard + events pages left out of the dark-mode swap** — known per commit `d3266a0` message, document as follow-up.
14. **`roundTripTest` polls every 3s for 30s** = up to 10 IMAP folder locks per test. Some servers throttle. IDLE or backoff is cleaner.

## ✅ Verified

- Multer `diskStorage` whitelisted filenames (`${subdir}-${Date.now()}.ext`) defeat #607-class traversal at intake (`adminExpenses.js:42`, `emailIntake.js:75`).
- `assertPathInside([storagePath/business-docs])` guards every download/render route (`adminExpenses.js:118/136/138/201`).
- Rasterized PNG served with `Content-Type: image/png`, `nosniff`, `Content-Disposition: inline`, strict `default-src 'none'; img-src 'self' data:` CSP (`adminExpenses.js:139-142`). No SVG path.
- Raw inbound PDF served `attachment; nosniff` — forces download, never inline render.
- IMAP host blocked from RFC1918 on save + test + folder-list (`adminEmail.js:153/186/207`); `imap_pass` masked on GET, preserved on PATCH (`adminEmail.js:130/165`).
- Tax-report cancelled invoices excluded from totals, kept in ledger with `replacedByInvoiceNumber` link (`taxReportService.js:434-450`).
- Late-fee gross-up + VAT-payable computed in minor units, integer math throughout (`computeReportedAmounts:104-114`, line 522).
- Paginated counts wrapped in `parseInt(... || 0, 10)` — no #596-class bigint trap in new code.
- Migration 131 `feature_contracts` defaults to TRUE — existing customer visibility preserved.
- Customer quotes-PDF silent-no-op (`req.customer.feature_quotes` → DB lookup) fixed at `customer.js:559`.
- IMAP poller fail-fast timeouts (10s connect/greeting, 30s socket) prevent the previous opaque-502 mode.
- Migrations 122-131 all `hasColumn`/`hasTable`-guarded, idempotent, with sane `down`s.

## Verdict

Close blockers 1 + 2 (CSV-injection prefix + UNIQUE on `received_emails.message_id`, both small) and I'm comfortable merging. Concerns 3-10 are real but each is a follow-up. CI green doesn't cover the formula-injection class (no fuzz on export inputs) nor the multi-replica race (single-instance test env) — both reasons the blockers slipped past green CI.

Luca-Timo:
@the-luap thanks for the quick and thorough review. All three review groups addressed across cd6d5783 (blockers), a93b6dc2 (concerns), d6da89f4 (nits), plus a couple of additions at the bottom.

## 🔴 Blockers
1. **Formula injection** — fixed. New `neutralizeSpreadsheetFormula()` prefixes `'` on a leading `= + - @ TAB CR`, wired into `csvEscape`, the tax-report `escape`, **and** the unquoted Banana tab cell formatter (the most exposed one). Unit test pins one of each char.
2. **IMAP double-ingest** — `received_emails.message_id` is now `UNIQUE`, and the poller **claims** the row (`status='processing'`) *before* ingesting, so a second replica loses the unique-constraint race and skips cleanly (stale claims reclaimed after 10 min).
   - On `file_sha256`: I **didn't** add the unique index — it's a *soft* dedup key by design. `recordInboundDocument` intentionally keeps duplicate-SHA rows (`status='duplicate'` + `duplicate_of_id`) for the manual **Duplikat** disposition; a unique index would silently drop the second copy and break that feature. The SHA race only yields an extra `'unsorted'` row (caught by the existing manual Duplikat backstop), not a double-count — the real email double-ingest is closed by the `message_id` claim above. Happy to dedup hard there if you'd prefer.

## 🟡 Concerns
- **1** 10s flag cache + invalidate-on-`PUT /feature-flags`.
- **2** Customer routes now gate on master **and** per-customer via `getEffectiveFeaturesForCustomer` (shared helper).
- **4** Unset registration → `vatPayableMinor: null` + UI warning, no more `grandTotalVat>0` guess.
- **5** Shared `upsertAppSetting()` closes the `created_at` class.
- **6** `getRenderedPagePath` refuses pages > 200 + `page_count` capped at ingest.
- **7** Paranoia pass clean — `original_filename` only renders via auto-escaped JSX; the two `dangerouslySetInnerHTML` sites are admin-authored content.
- **3** Corrected the claim: the report honours per-row `tax_treatment`, not an auto-derivation from supplier-vs-company country. Full country-derived classification needs a `supplier_country` column — tracked as a follow-up, not in this PR.
- **8** Flagging `imap_pass` plaintext as a known operator concern (same class as `smtp_pass`): widens the inbound-mailbox blast radius if `email_configs` leaks. Not a regression; encryption-at-rest for both is a separate hardening item.

## 🟢 Nits
1 (stray PDF removed + `backend/storage/business-docs/` gitignored), 2 (`adminLedger`/`adminExpenses` use the shared cached middleware), 4 (round-trip poll uses ×1.5 backoff). 3 (dashboard/events dark-mode) tracked as a follow-up.

## ➕ Added since the review
- **Test coverage** for the classes you flagged CI didn't reach: `spreadsheetSafe.test.js` (formula injection, one of each leading char), `dbErrors.test.js` (`isUniqueViolation` — the multi-replica race detector), `rasterizePageCap.test.js` (out-of-range page rejected before `pdftoppm`), and the export-scope filter.
- **Tax-report export scope** — a `Complete / Income only / Cost only` selector on the PDF + CSV export (filters the ledger + the summary; income-only drops the incoming-invoice + expense rows, cost-only drops the income-by-rate breakdown). Handy in Liechtenstein where, under the income threshold, a flat 20 % Gewinnungskosten deduction is sometimes better than actual costs — you can hand the Treuhänder just the income (or just the cost) basis. The 20 % calc itself stays out of the app, applied by the Treuhänder.
- **README** — added a "CRM & Accounting (Beta)" subsection to Key Features, expanded the Roadmap row to cover the accounting module, broadened the disclaimers with a tax/VAT "verify your own jurisdiction" paragraph, and refreshed the contributor entry.

## ❓ One thing for you
The docs link in the Roadmap row / README — https://docs.picpeak.app/features/crm — currently 404s. Could you point me at the right URL (or stub the page)? I left the link as-is for now since it's on your docs domain.

PR #636, opened 2026-06-18
feat(accounting): incoming-invoice workflow v2 + VAT/financial settings consolidation

Luca-Timo:
Bundled accounting work on top of the incoming-invoices foundation already on `beta`. Migrations **132–134** (sequential after `131`). All VAT/tax surfaces are guidance-only and carry "verify with your Treuhänder" disclaimers.

### Incoming invoices (the inbox)
- **Re-categorise after the fact** — a categorised invoice can be changed again (e.g. pass-through → company expense), even after the supplier is paid (payment and classification are independent axes). `categorizeInbound` is re-runnable: it unwinds any prior re-bill line + recomputes the invoice totals before re-applying, and refuses (`INVOICE_LOCKED`) only when the re-bill sits on an already-issued invoice (Storno required).
- **Note field** on categorisation (migration 132).
- **Re-bill like hours** — rebill/pass-through now attach to a customer and consolidate onto their bill: monthly/manual customers append to the running draft immediately; per-event customers accumulate in a **"Pending re-bills"** pool that bundles into one invoice via "Bill these" (mirrors `billUnbilledEntries`).
- **Pass-through (durchlaufender Posten)** can attach to a customer but **never carries a markup** (billed at cost / VAT-neutral) — enforced in UI + service.
- **Per-disposition inline explainer** under the dropdown (re-bill vs pass-through vs company expense weren't obvious from labels).
- **Supplier-country tax default** (migration 134) — `categorizeInbound` auto-derives `tax_treatment` from the supplier country against the install's VAT-reclaim list (in-list → domestic; outside → foreign non-reclaimable). Finally consumes the previously-stored-but-unused `accounting_vat_reclaim_countries`.

### VAT / financial settings consolidation
- All VAT/financial config now lives under **Settings → Accounting** (two cards, one Save):
  - The orphaned **"Default VAT rate %"** is removed from Business profile (the real rates are the Accounting VAT codes).
  - Invoice/quote **VAT picker is code-only** — options are exactly the configured output VAT codes (no free-text rate); legacy/off-list values on existing documents are preserved as a read-only option so issued invoices aren't silently changed.
  - **VAT label** moved to Accounting **and wired onto the invoice/quote PDF** VAT-line label (it was previously ignored), falling back to the locale default when blank.
  - **Default hourly rate** moved to Accounting (storage stays on `business_profile`).
  - **Default currency** is now a normalising dropdown (an old free-text `chf` auto-selects `CHF`; unknown values preserved).
  - New **default output VAT code** setting — new invoices/quotes seed their VAT from it instead of silently starting at 0%.

### Feature flags
- **Invoices force-enable the Accounting master** (invoice VAT settings live there). Applied in `applyDependencyRules` (backend + frontend); migration 133 corrects existing installs so the `requireFeatureFlag` middleware (which reads the stored row) agrees. The Features tab shows the Accounting card locked-on while Invoices is enabled.

### Bug fixes
- **`vat_code` lost on invoice edit** — the GET-route serializer (`transformInvoice`) dropped `vat_code`, so the editor fell back to rate-matching and silently lost a custom-rate code on edit. Now returned.
- **SQLite write-in-transaction deadlock** — `logActivity` writes via the global db; inside a `db.transaction` it deadlocks against the held write lock on a SQLite-backed install. Moved out of the transactions in `expenseService` and `customerHoursService` (logged after commit). *Known limitation:* the monthly/billing paths additionally route through `createInvoice`, whose own internal `logActivity` is the shared root case — documented in code; full fix is a separate, broader change.

### Tests & docs
- New integration test (`incomingInvoiceRebill.test.js`, 8 cases) covering the disposition state machine, pending pool, pass-through-no-markup, unwind, and `INVOICE_LOCKED`. Unit tests for `buildInboundLineItem`, `isInvoiceMutable`, `resolveTaxTreatment`. Existing CRM/hours/ledger suites green.
- `docs/accounting-inbound-invoices.md` rewritten to the current implementation.

### Migrations
- **132** — `inbound_documents.note` + `customer_account_id`
- **133** — set stored `accounting`=true where `bills` is on
- **134** — `inbound_documents.supplier_country`

### i18n
New strings are **EN + DE** (DE authored natively). Other locales fall back to the English inline defaults.

### Testing notes
- Capture/triage an incoming invoice → categorise → re-categorise (incl. after marking the supplier paid); set a supplier country and confirm the tax treatment flips for a foreign supplier.
- Per-event customer: categorise two rebills → "Bill these" → one bundled invoice.
- Toggle **Invoices** in Settings → Features → Accounting flips on + locks.
- Set a default output VAT code → open a new invoice/quote → VAT picker is pre-selected.
- Set a VAT label → the invoice PDF VAT line uses it.


the-luap:
Reviewed end-to-end. Strong overall — migrations 132/133/134 are all guarded + idempotent (133 deliberately no-down with documented rationale), the `bills → accounting` rule is server-enforced at every flag PUT (not just UI), `categorizeInbound` re-runs unwind the prior line cleanly and recompute totals from line-items (no additive drift across many re-categorisations), `INVOICE_LOCKED` correctly refuses unwinding `sent` invoices, the pending-rebill bundling scopes strictly to `customer_account_id = customer.id` so no cross-customer leak, `transformInvoice` returns `vatCode` on every admin-side serializer (GET/POST/PUT/storno), and the PDF VAT label is read fresh per render — no caching surprise.

Six concerns worth closing or acking before merge, plus a few nits. None of them is a hard blocker.

## 🟡 Concerns

### 1. `resolveTaxTreatment` mis-classifies on fresh installs

`backend/src/services/expenseService.js:75-80` — when `accounting_vat_reclaim_countries` is empty (admin hasn't configured the setting yet) AND `supplier_country` is set on the doc, every supplier — including the admin's own domestic one — defaults to `foreign_vat_non_reclaimable`. The "unknown country → domestic" fallback at line 78 only fires when `cc` is empty, not when the *list* is empty. So a brand-new install with migration 134 deployed but the reclaim-countries setting still unset auto-classifies CH suppliers as foreign.

```js
function resolveTaxTreatment(payloadTreatment, supplierCountry, reclaimCountries) {
  if (TAX_TREATMENTS.includes(payloadTreatment)) return payloadTreatment;
  const cc = String(supplierCountry || '').toUpperCase();
  if (!cc) return 'domestic';
  if (!reclaimCountries || reclaimCountries.length === 0) return 'domestic'; // <- add this
  return reclaimCountries.includes(cc) ? 'domestic' : 'foreign_vat_non_reclaimable';
}
```

Defer the auto-classification until the setting is configured. Admins who haven't told us their reclaim countries shouldn't have us guessing on their behalf.

### 2. `listPendingRebillSummary` orphans pending re-bills when the customer is deleted

`backend/src/services/expenseService.js:491` — INNER JOIN to `customer_accounts`. `inbound_documents.customer_account_id` is a loose-FK (you explicitly chose that for parity with `event_id` / expenses), so deleting the customer makes the doc vanish from the pending pool and become unreachable via `billPendingRebills` (the customer-lookup 404s). Doc silently orphaned.

Suggest LEFT JOIN + `whereNotNull('c.id')` for the visible pool, AND a periodic "orphaned pending re-bills" admin surface, OR null the column on customer delete via the existing customer-delete service.

### 3. `logActivity` is still inside the trx on the `createInvoice` path

You document this as a known limitation in the PR body — moving `logActivity` out of the `expenseService` + `customerHoursService` transactions, but `createInvoice` still calls its own `logActivity` inside the inherited trx, which deadlocks on SQLite-backed installs. The wrapping try/catch swallows it producing a silent audit gap on every `categorizeInbound → billInboundNow → createInvoice` and `billPendingRebills` chain.

Production is exclusively Postgres per `docker-compose`, so this is dev/test only — explicit ack that's the intended scope (and the broader fix is a separate change) is fine. Just want to make sure the audit-gap-on-SQLite caveat is in our heads for when someone reports it.

### 4. `VatRateSelect` rate-fallback can collide

`frontend/src/components/admin/VatRateSelect.tsx:36` — when stored `vat_code` is null, the picker falls back to `codes.find(c => Number(c.rate) === Number(rate))`. If two configured codes share the same rate (e.g. "Standard" and "B2B Standard" both at 8.1%), the first wins and gets silently saved on next edit, quietly replacing one with the other.

Low practical risk but not guaranteed safe. Suggest: when `codes.filter(... rate).length > 1`, fall through to the legacy-option path so the form shows "8.1% (not configured)" instead of silently picking one.

### 5. `unwindBilledLine` leaves an empty-invoice survivor

`backend/src/services/expenseService.js:313-328` — when the unwound rebill line was the only line item on its invoice, the invoice survives at `net_amount_minor=0` in its prior status. Net-zero invoices accumulate over many re-categorisations.

```js
const allItems = await trx('invoice_line_items').where({ invoice_id: invoice.id });
if (allItems.length === 0) {
  // Last line removed — invoice has no reason to exist anymore. For drafts
  // delete outright; for scheduled drop the schedule and cancel.
  await trx('invoices').where({ id: invoice.id }).del();
  return;
}
```

### 6. `isInvoiceMutable` locks plain `'draft'` invoices

`backend/src/services/expenseService.js:287-293` — only `is_monthly_draft` and not-yet-armed `'scheduled'` invoices are mutable. A standalone manual draft with one rebill on it refuses re-categorisation with `INVOICE_LOCKED`, even though the draft hasn't been sent.

Either intentional — in which case the function comment should add "Plain `'draft'` invoices also lock; admins must delete the rebill line manually before re-categorising" so the next reader knows — or a missed case where `'draft'` should slot in next to monthly drafts. Your call.

## 🟢 Nits

- `applyDependencyRules` duplicated between backend (`adminFeatureFlags.js:123-167`) and frontend (`FeatureFlagsContext.tsx:81-122`). Worth a shared spec doc the next time someone touches dependency rules.
- `frontend/src/constants/currencies.ts:21` — `normalizeCurrency` has a tautological ternary, both branches return `cleaned`. Collapse to `return cleaned;`.
- `BillEditorPage.tsx` VAT-seed effect guards with `if (vatCode || vatRate) return;` — `vatRate=0` is falsy, so a quote conversion that legitimately set 0% would still seed if render order changes. Render order makes it safe today but it's a fragile invariant; an `everTouched` ref would harden it.

## ✅ Verified

- Migrations 132 / 133 / 134 directly inspected — additive, `hasTable`/`hasColumn`-guarded, sane downs (133 deliberately no-down with documented rationale).
- `applyDependencyRules` consistency between backend `adminFeatureFlags.js:123-167` and frontend `FeatureFlagsContext.tsx:81-122` — server enforces on every PUT; direct API call with `bills=true, accounting=false` gets force-corrected.
- `invalidateFeatureFlagCache()` called from the flag PUT handler at `adminFeatureFlags.js:237`; cache empty at boot so post-migration-133 reads see the corrected value on first hit.
- `categorizeInbound` re-runnability + unwind atomicity in `expenseService.js:282-328` — in-trx delete + recompute of totals from line-items, no additive drift; confirmed by `incomingInvoiceRebill.test.js:175-181`.
- `INVOICE_LOCKED` refusal for `status='sent'`; confirmed by `incomingInvoiceRebill.test.js:185-194`.
- `logActivity` moves in `customerHoursService.js` (`createEntry/updateEntry/deleteEntry/billUnbilledEntries`): staged inside the trx callback, fired AFTER `await db.transaction(...)` returns, wrapped in try/catch so logging failure doesn't bubble. Order is commit-then-log.
- `transformInvoice` (`adminInvoices.js`) now returns `vatCode`; piped through GET `/:id`, POST `/`, PUT `/:id`, and storno endpoints (all four use `transformInvoice`). Customer-facing `customer.js` does NOT return vat_code, but customers can't edit so the snapshot rate is sufficient — not a regression.
- PDF VAT label (`pdfService.js` + `_renderContext.js`) reads `profile.vat_label` fresh per render via `buildIssuerBlock`; DB column capped at 64 chars; PDFKit `text()` — not a markup vector.
- Currency dropdown (`SettingsBusinessProfilePage.tsx`) preserves unknown stored codes via `currencyOptions(profile.defaultCurrency)`.
- `billPendingRebills` (`expenseService.js:553`) scopes the trx query strictly to `customer_account_id = customer.id` — no cross-customer bundling possible.

**Trusted at PR-body level** (not independently re-derived): the markup-math unit tests in `expenseService.markup.test.js` (test file presence + name; didn't re-derive each case), and your claim that the VAT-codes lookup endpoint is intentionally un-gated.

## Verdict

Close concerns 1, 2 and 5 (each is a small, mechanical fix), get an explicit ack on 3 (SQLite audit-gap = documented limitation) and 6 (plain-draft locking = intentional or oversight). Then I'm comfortable merging. Concern 4 (rate-fallback collision) can be a separate small follow-up.

Luca-Timo:
I am still testing and will ping you when I am ready

Thanks for the thorough review — really appreciate the depth. Pushed two commits: `707c5d02` (review fixes) and `db9e41d1` (two pre-existing accounting bugs my own pass caught). Per-concern:

### ✅ Closed

**#1 — `resolveTaxTreatment` on fresh installs.** Added the empty-list guard exactly as suggested: an unset `accounting_vat_reclaim_countries` now returns `domestic` instead of classifying every supplier (including the domestic one) as foreign. Auto-classification only kicks in once the admin has configured the list. Unit test added.

**#2 — pending re-bills on customer delete.** One clarification first: picpeak has **no hard customer delete** — `customerAccountsService.eraseCustomer` anonymizes in place (sentinel email, keeps the row) precisely to avoid the FK-RESTRICT problem you'd expect. So the `customer_account_id` loose-FK never dangles, the INNER JOIN never drops the doc, and `billPendingRebills` never 404s — the orphan/404 path can't actually occur on this codebase. That said, your underlying point stands: a pending re-bill left pointing at an *anonymized* account shouldn't quietly remain billable. So `eraseCustomer` now returns the customer's not-yet-billed inbound docs to the inbox (`customer_account_id = null`, `status = 'unsorted'`) for re-triage. Guarded with `hasColumn` for pre-132 installs.

**#5 — `unwindBilledLine` empty-invoice survivor.** Implemented your snippet: when the unwound re-bill was the only line item, the (mutable, never-issued — already checked) invoice is deleted outright instead of left at `net=0`. For a monthly draft this just means the next append re-creates one.

**#6 — `isInvoiceMutable` and plain `'draft'`.** It's the first branch of your either/or: invoices have **no `'draft'` status** in this codebase (only quotes do) — the editable, not-yet-sent invoice state *is* `status='scheduled'` with no `scheduled_send_at`, which the function already treats as mutable. So there's no `'draft'` case to slot in; I added a comment to that effect so the next reader doesn't go looking for it.

**#4 — `VatRateSelect` rate-fallback collision.** Decided to do this now rather than defer. When the stored `vat_code` is null and we fall back to rate-matching, the picker now only auto-selects a code if **exactly one** configured code has that rate. If two or more share it (your "Standard" + "B2B Standard" at 8.1%), it falls through to the legacy `"8.1% (not configured)"` option so the admin picks explicitly — no silent swap on the next save.

**Nits:** collapsed `normalizeCurrency`'s tautological ternary; fixed a VAT-picker i18n key (`t('vat.legacyRate')` had no matching namespace — the key lives at `ledger.vat.legacyRate`, so the legacy label was always falling back to English); removed dead i18n keys left by the settings refactor. The `applyDependencyRules` backend/frontend duplication I've left as-is for now — agree a shared spec is the right call next time someone touches dependency rules, but it's out of scope here.

### 🔖 Acked (documented limitation, not fixed in this PR)

**#3 — `logActivity` inside the inherited `createInvoice` trx (SQLite only).** Confirming scope explicitly: the deadlock is `createInvoice` / `appendToMonthlyDraft` calling their **own** `logActivity` via the global `db` handle. When my accounting paths pass an open transaction (`categorizeInbound → billInboundNow → createInvoice(…, trx)`, and `billPendingRebills`), that internal log issues a *second* write connection which blocks on the held SQLite write lock → deadlock; the wrapping `try/catch` swallows it, so on a SQLite-backed install those chains silently drop their `incoming_invoice_rebilled` / `monthly_billing_items_queued` audit-log row.

- **Functional/financial impact: none.** The invoice creation + document stamping commit fine; only the best-effort audit-trail row is lost.
- **Scope: dev/test only.** Production is Postgres exclusively per `docker-compose.production.yml` (MVCC + pooled connections, no lock contention), so it doesn't bite there. The integration test bypasses it by stubbing `logActivity`.
- **Why not in this PR:** the proper fix is to thread an optional `trx` through `logActivity` (or have `createInvoice`/`appendToMonthlyDraft`/`spawnInstallmentInvoices`/`createStorno` return what to log and let the outermost caller log after commit). That's ~18 call sites across the invoice/quote/contract/hours/expense services — a codebase-wide change well beyond this feature. I moved the `expenseService` + `customerHoursService` *own* `logActivity` calls out of their transactions here; the nested-`createInvoice` case is the remaining root cause and I'll open it as a separate change. Caveat noted so it's in our heads if a SQLite user ever reports a missing audit row.

### ➕ Also folded in (pre-existing, accounting surface — `db9e41d1`)

While reviewing I caught two existing beta bugs worth fixing here since they're on the same surface:
- **Tax-report storno double-subtraction.** The income totals excluded only `status='cancelled'`, never `kind='storno'`. A Storno (`status='sent'`, amounts stored negative) netted into the totals on top of the already-excluded cancelled original → the revenue got deducted twice, so a cancel-and-reissue read as **0 income** instead of the reissued amount. Now excludes storno rows from `grandTotal*` / `byRate` (still visible in the row list for the gap-free audit trail). Added a regression test that reproduces the real cancel→storno→reissue three-row flow — the old test only modelled cancelled+reissue without the storno row, which is why it passed.
- **Hours-line date corrupt on Postgres.** `buildLineItemFromEntry` did `String(entry.entry_date).slice(0,10)` on a `date` column → PG hands it back as a JS `Date`, baking `"Wed Apr 06"` into the invoice line + PDF (SQLite returns the bare string, so SQLite-only tests passed). Normalised via the `Date` branch like every other date read.

Happy to split `db9e41d1` into its own PR if you'd rather keep this one feature-scoped — just say the word.

Luca-Timo:
@the-luap — pushed three follow-up commits addressing the review; ready for another pass.

**[`707c5d0`](https://github.com/the-luap/picpeak/pull/636/commits/707c5d027798bdafb9fe09d7efcbd9ea65330076) — your concerns + nits**
- **Concern 1** `resolveTaxTreatment`: empty reclaim-countries list now returns `domestic` instead of auto-classifying every supplier (incl. domestic) as foreign — defer until configured. +test.
- **Concern 2** `eraseCustomer` returns the customer's not-yet-billed inbound docs to the inbox. (Note: there's no hard customer delete — erase anonymizes in place — so the orphan/404 can't actually occur; this is hardening.)
- **Concern 4** `VatRateSelect`: when >1 code shares a rate, fall through to the "(not configured)" option instead of silently picking the first. (Did it now rather than deferring.)
- **Concern 5** `unwindBilledLine` deletes the (mutable, never-issued) invoice when the unwound re-bill was its only line, instead of leaving a net-zero survivor.
- **Concern 6** clarified in a comment that invoices have no `'draft'` status (editable state is `scheduled` w/o send-at) — no behaviour change.
- nits: collapsed `normalizeCurrency`'s tautological ternary; fixed a VAT-picker i18n key (`vat.legacyRate` → `ledger.vat.legacyRate`, it was always falling back to English); removed dead i18n keys from the settings refactor.
- **Concern 3** acked as documented — `createInvoice`'s own `logActivity` still runs in the inherited trx, deadlocks only on SQLite (prod is Postgres), worst case a silently-dropped audit-log row, no functional/financial impact. Proper fix (`trx` through `logActivity`, ~18 sites) is a separate change.

**[`db9e41d`](https://github.com/the-luap/picpeak/pull/636/commits/db9e41d19846b31b29c5c1be2ee06a7958bb43b0) — two pre-existing accounting bugs on the same surface**
- Tax-report **storno double-subtraction**: totals excluded only `status='cancelled'`, never `kind='storno'`, so a cancel-and-reissue read as **0 income** instead of the reissued amount. Now excludes storno from totals (kept visible in the row list). Regression test reproduces the real cancel→storno→reissue 3-row flow.
- Hours-line **date corrupt on Postgres**: `String(entry.entry_date).slice(0,10)` baked "Wed Apr 06" into the invoice line on PG. Normalised via the Date branch.

**[`e9b297c`](https://github.com/the-luap/picpeak/pull/636/commits/e9b297c162a19da31d53de377b90bfd5cda1b0a7) — editor VAT display**
- The create/edit-invoice **totals box** showed VAT 100× too small (CHF 0.63 vs 63.18) — `Math.round(subtotal * vatRate) / 100` was missing the `*100` for cent-rounding. **Display-only**: backend `computeTotals`, the PDF, and the tax report were always correct; it only surfaced now that new invoices seed a non-zero default VAT code instead of 0%.

All green — backend suites (83 incl. the new storno + tax-treatment cases) + `npm run build`.

PR #667, opened 2026-06-23
feat: admin-configurable workflow engine + dunning/Mahngebühr rework (RFC — feedback welcome)

Luca-Timo:
## What this is

An admin-configurable **visual workflow engine** for picpeak (React Flow canvas:
triggers → conditions → branches → loops → waits → approval gates → actions),
plus the **dunning / Mahngebühr** rework it grew out of. It converts the
previously hardcoded automations (reminder ladder, gallery-expiry emails,
pre-event reminders) into **editable built-in flows**, so photographers can see
and tweak their automation as blocks instead of digging through settings.

This is a large, cohesive feature (31 commits, ~5k LOC) and I'd genuinely value
your review on the design calls below before we consider it for beta. 

## Why

Today's reminder ladder + expiry mailers are hardcoded; every new automation
(deposits, booking pipeline, pre-event nudges) means more bespoke cron/email
code. The engine makes those data-driven so future work is "drop in a block,"
not a refactor.

## What's included

- **Engine** — graph executor (branch/loop/wait/gate/action), idempotent runs
  (dedup key), persisted suspension (waits resume on the existing scheduler
  tick; gates resume on admin action), crash recovery (heartbeat + attempts cap).
- **Editor** — React Flow canvas, structured per-node config panel, dagre
  auto-layout, JSON import/export, and **test-fire** (safe dry-run of a whole
  flow with side effects mocked).
- **Approval gates** — admin gets a confirm/deny email **and** a webview
  approvals inbox; a pending-approvals card on the dashboard.
- **Built-in flows** — invoice dunning, gallery-expiring, gallery-expired,
  pre-event reminder (shipped enabled, see cutover), plus booking
  quote→contract→event→invoice and quote→event→invoice (shipped disabled —
  the prepare_*/send_document actions are still stubs).
- **Dunning / Mahngebühr** — 3-reminder ladder, flat/percent fee accumulating on
  every reminder after the first, **Mahnung as a separate reminder letter** (the
  issued invoice stays immutable — §14 UStG), VAT-on-fee jurisdiction toggle,
  collections handoff after the ladder exhausts.
- **Lifecycle/time triggers** wired (additive, fail-closed): invoice.sent/paid/
  overdue, quote.sent/accepted/declined, contract.sent/signed, gallery.published/
  expiring/expired, event.date_approaching, customer.created.
- **Quote event type** — quotes now carry an event type (dropdown from the
  event-types catalog) so quote→event no longer hardcodes 'wedding'.

Migrations 142–146; gated behind a new `workflows` feature flag + a
`workflows.manage` / `workflows.view` permission split.

## Design decisions I'd like your eyes on

1. **`workflows` flag is default-OFF.** The whole cutover is gated on it; the UI
   that moves settings into flows is conditional on the flag, so flag-off
   installs keep the legacy controls and lose nothing. Do you want it default-on
   for beta, or opt-in?
2. **Hard cutover via existence-based mutual exclusion.** Once a built-in is
   seeded (flag on) the engine OWNS that automation — the legacy path stands
   down whether the flow is enabled (it sends) or disabled (off). So the flow
   toggle is the single switch and there's no double-send. OK with this model?
3. **Cutover built-ins ship ENABLED**, delegating to the existing send functions
   so behaviour is byte-identical. Booking flows ship disabled (stubs).
4. **Mahngebühr = reminder letter, not a new invoice or a mutation** of the
   issued invoice. Fee tracked on the row + rendered on a separate Mahnung PDF.
5. **VAT-on-fee is a toggle** (CH yes / DE-AT no), no-ops when the org has no VAT
   rate. The tax-report/Banana booking of that VAT is deliberately deferred
   pending Treuhänder input — flagged in-code.
6. **Installments stay scheduler-driven** (no flow) — the scheduler has no clean
   workflow insertion point and it already works; didn't want to risk it.

## Legal/tax — needs a second opinion

The Mahngebühr / AGB / VAT bits are modelled for CH/LI (OR-based) and every
surface that touches them carries a "verify with your Treuhänder" disclaimer.
I'm not qualified to sign off the legal treatment — would appreciate your /
your accountant's sanity check, especially the VAT-on-fee handling.

## Translations

New UI strings are authored natively in **EN + DE only** (the two languages I'm
fluent in). NL/PT/RU fall back to English and need a native pass before they're
considered localized — flagging rather than shipping machine translations.

## Safety & tests

- Flag gating fails CLOSED everywhere; emits never throw into callers; engine
  writes happen after commit (no SQLite global-write-in-trx deadlock).
- Backend suite green (~925 tests, incl. new engine/dunning/cutover coverage);
  frontend tsc + build clean.

## How to try it

Enable the `workflows` flag (Settings → Features), open Admin → Workflows, and
use **Test run** on the Invoice dunning built-in to watch a dry-run walk the
flow with no real mail sent.

## Open questions for you

- Default-on vs opt-in for the flag in beta?
- Comfortable with the enabled-by-default cutover + existence-based stand-down,
  or would you prefer the built-ins ship disabled and you flip them?
- Any concerns on the Mahngebühr letter model or the VAT toggle?


the-luap (changes requested):
RFC ack — and substantive feedback below. I direct-inspected all 5 migrations, the engine state machine, `publicWorkflowApprovals.js`, `adminWorkflows.js`, `approvals.js`, and the head of `actions.js`. A subagent covered the cutover / mutual-exclusion in all 4 legacy services, the dunning math + Mahnung PDF + VAT toggle, every lifecycle emit site, the frontend XSS/i18n surface, and the test suite. I re-verified both blockers at file:line before posting.

## 🔴 Blockers

### 1. Approval-link GET silently consumes the gate decision on email-client prefetch

`backend/src/routes/publicWorkflowApprovals.js:24-49` — `GET /:token/:action` mutates state on first hit. Outlook Safe Links, Gmail/Workspace, Proofpoint, corporate antivirus link-checkers all GET email URLs *before the human clicks*. For payment-confirm gates (the seeded dunning flow), a prefetcher hitting `/confirm` silently advances the "payment received" path → stops dunning on an unpaid invoice. The "already recorded" fallback only shows to the human — the prefetcher's first GET already executed `finalizeApproval` → `resumeRun`.

```js
router.get('/:token/:action', async (req, res) => {
  // …
  const result = await actByToken(token, action);  // ← already mutates here
```

**Fix**: confirmation interstitial. `GET /:token/:action` renders an HTML page with "Confirm payment received" / "No payment received" buttons that POST the actual decision. Token is still single-use, single-shot — POST defeats the prefetcher class.

### 2. Gate `deny` with a missing deny edge silently marks the run `done`

`backend/src/services/workflows/engine.js:227-237` — `resumeRun` calls `outEdge(edges, current_node, decisionHandle)`. When the flow author imports/edits a graph that lost the `deny` edge, `outEdge` returns null → `nextKey = null` → `finishRun(runId)` flags the run as `status: 'done'`, indistinguishable from a clean completion. Run history shows green; the deny was lost.

```js
const e = outEdge(edges, run.current_node, decisionHandle);
const nextKey = e ? e.to_node : null;
// …
if (!nextKey) { await finishRun(runId); return; }  // ← should be failRun for missing handle
```

**Fix**: `if (decisionHandle != null && !e) { await failRun(runId, `gate decision ${decisionHandle} has no matching edge`); return; }`. Fail loudly. Same pattern as the existing `failRun` for unknown nodes.

## 🟡 Concerns

1. **`validateGraph` accepts any string for `n.type`** (`adminWorkflows.js:36-49`). A typo'd `actoin` ships an enabled flow that no-ops every cycle (`engine.js:193` records a `skipped` step). Whitelist against `{trigger, action, condition, branch, loop, wait, gate, webhook}`.
2. **`matchFilter` loose equality** (`engine.js:61-72`, eslint-disabled `==`). Filter `{value: 0}` matches `false` and `''`. Switch to `===`; drop the eslint-disable.
3. **No size cap on graph payload** (`adminWorkflows.js`). `workflows.manage` user can DoS the DB. Cap nodes + edges count + `JSON.stringify(config).length` per node.
4. **Engine-owns-disabled = silent send-nothing foot-gun.** When admin clicks "disable" on a seeded built-in, `isBuiltinFlowPresent` still returns true → legacy stands down, engine doesn't fire (`enabled:false`) → automation goes dark. An admin disabling the dunning flow expecting to roll back to legacy is silently turning off dunning entirely. Fix: list-page warning copy when toggling a built-in off, or wire a "Restore legacy" action that flips `is_builtin=false`.
5. **Booking flows ship `enabled:false` but reference stub actions** (`actions.js` registers `prepare_*` + `send_document` as `{skipped:true}`). If an admin enables `booking_full` thinking it works, every quote-acceptance silently records 7 skipped steps and drops the contract. Fix: render an "Unimplemented" panel error in the UI when a built-in's actions list contains any DOCUMENT_ACTIONS stub, and refuse the enable toggle on that flow until it's wired.

## 🟢 Nits

- `_workflowSeedBoot.js` re-seed branch re-enables a built-in on the next `SEED_VERSION` bump if the admin had disabled it. Document, or honour an `admin_disabled_at` sentinel.
- `applyReminder` doesn't roll back the Mahnung PDF write if `queueEmail` throws — orphan file on disk. Admin-visible, low-risk.
- DE i18n authored natively (PR body says so). NL/PT/RU fall back to EN via `fallbackLng: 'en'` — no build break; flagged for future native pass.

## ✅ Verified

- **All 5 migrations (142–146)** — `hasTable`/`hasColumn`-guarded, idempotent, conservative defaults preserve current behaviour, sensible indexes. Permission seed mirrors migration 123.
- **Existence-based mutex** wired correctly in `eventReminderService.js:126-141`, `expirationChecker.js:30-46`, `invoiceService.js:3447-3501`. Flag-gated, fail-closed. Tested.
- **Dunning math** — `applyReminder`: level 1 = 0 fee, level 2 = 1× fee, level 3 = 2× fee. Re-applying same level idempotent. Level-3 cap enforced. Tested at `invoiceDunning.test.js:81-119`.
- **Invoice immutability** — `total_amount_minor` never mutated; fee lives in `late_fee_amount_minor` + `late_fee_vat_minor`. Original invoice PDF untouched. Mahnung renders to a **separate file** under `storage/business-docs/mahnung/<year>/<invoice>_mahnung_L<level>.pdf`. QR suppressed on Mahnung (`pdfService.js:1710`). §14 UStG posture is correct.
- **VAT toggle** — `resolveLateFeeVatRate` returns 0 when toggle off OR when `business_profile.vat_rate_default` is 0/null. DE/AT and small-business orgs correctly get no fee VAT; CH-style orgs with VAT get it.
- **Lifecycle emits** — all after DB commit, all wrapped in try/catch with `logger.warn`, none throw into caller. `invoice.paid` only on the `sent→paid` transition. No race where event emits before row persists.
- **Version pinning** — `emitWorkflowEvent` snapshots `wf.version` into the run; `runDueWaits`, `loadGraph`, `resumeRun` all filter by `(workflow_id, version)`. Mid-flight edits don't disturb running flows.
- **XSS** — no `dangerouslySetInnerHTML` or `innerHTML` in `WorkflowEditorPage.tsx`, `NodeConfigPanel.tsx`, `WorkflowApprovalsPage.tsx`. JSON import goes through `setNodes`/`setEdges` and renders as React text — escaped.
- **Installments stay scheduler-driven** — no `emitWorkflowEvent` on installments paths. PR body's claim holds.
- **Approval token** — 256-bit `crypto.randomBytes(32)`, SHA-256 hashed at rest, UNIQUE constraint on `token_hash`, idempotent `status !== 'pending'` check. Strong.

## My take on the three RFC questions

- **Flag default**: keep `workflows: false` for beta. Blockers #1 + #2 are non-trivial; an opt-in posture lets contributors hit them cleanly. Default-on would expose unaware operators on upgrade.
- **Enabled-by-default cutover** for the four built-ins: ship them seeded with **`enabled: false`** for the first beta. The existence-based mutex is correct and tested, but blocker #1 (prefetched payment confirmation) makes "live on upgrade" too risky for the dunning gate specifically. Once #1 and #2 are closed, flip to enabled-by-default in a follow-up beta. Booking flows already ship disabled — keep that.
- **Mahngebühr / VAT model**: comfortable. Invoice immutability is enforced + tested, separate-Mahnung-PDF is the right §14 UStG posture, VAT toggle correctly no-ops when the org has no VAT rate (DE Kleinunternehmer won't accidentally tack on VAT). AGB-disclaimer warnings are in the right spots. The "verify with your Treuhänder" copy is the right CYA — happy with the legal shape.

## Verdict

Close blockers #1 (POST confirmation interstitial) + #2 (`failRun` on missing handle in `resumeRun`), flip the four cutover built-ins to `enabled: false` for the first beta, and the rest is acknowledge-and-ship. Engine architecture (graph + dedup + version-pin + crash-recovery + dry-run + the test suite) is genuinely impressive work — the blockers are narrow execution gaps in two specific surfaces, not architectural issues. **Not safe to merge today; merge cleared once #1 + #2 land and the four cutover built-ins ship `enabled: false`.**

the-luap (approved):
Re-verified the 8 follow-up commits. Every blocker + concern landed with the exact fix shape I suggested, plus a bonus latent-bug catch and a small new built-in. Tests pin the contract changes.

## Blockers — closed

| # | Fix | Verified |
| - | --- | -------- |
| 1 | `publicWorkflowApprovals.js` split into a **read-only `GET`** (renders interstitial via new `peekApproval`, HTML-escapes `prompt` text) and **`POST`** (the only mutator → `actByToken`). Prefetchers issue GET, not POST → can't consume the gate decision. | ✅ commit `98ab717` |
| 2 | `engine.js` `resumeRun` now matches `decisionHandle` EXACTLY — no fall-back to `outEdge`'s sole-edge heuristic. A `deny` decision on a graph with only a `confirm` edge calls `failRun('gate decision deny has no matching edge…')` instead of `finishRun`. | ✅ commit `98ab717` |

## Concerns — closed

| # | Fix | Verified |
| - | --- | -------- |
| 1 | `validateGraph` whitelists `{trigger, action, condition, branch, loop, wait, gate, webhook}`; a typo'd `actoin` → 400 "Unknown node type". | ✅ commit `d927464` |
| 2 | `matchFilter` switched to `===` (and `!==` for `neq`); eslint-disable removed. | ✅ commit `98ab717` |
| 3 | `validateGraph` caps `MAX_NODES=200`, `MAX_EDGES=500`, per-node `MAX_NODE_CONFIG_BYTES=16KB`. | ✅ commit `d927464` |
| 4 | **Cutover posture flipped to enabled-based mutex**: `eventReminderService.js:139`, `expirationChecker.js:35-36`, `invoiceService.js:3465` all switched from `isBuiltinFlowPresent` → `isBuiltinFlowActive`. Disabling a built-in now re-engages legacy, not silence. Frontend list-page has a `window.confirm` dialog with i18n'd "reverts to the previous built-in behaviour — it does not turn the automation off" copy. | ✅ commits `5893ecb` + `c5f131c` |
| 5 | `adminWorkflows.js` create/update/PATCH refuse `enabled:true` when the graph references any action in `DOCUMENT_ACTIONS` — returns 409 with explicit "uses actions that aren't implemented: prepare_quote, …" message. | ✅ commit `d927464` |

## RFC postures applied

- ✅ **Built-ins ship `enabled:false` for first beta** (`5893ecb`). The four cutover flows (`invoice_dunning`, `gallery_expiring`, `gallery_expired`, `pre_event_email`) all seeded with default-OFF; admin opt-in.
- ✅ **`admin_toggled_at` sentinel** (migration 148, nullable, idempotent) stamps on enable/disable/edit. The re-seeder honours it (`_workflowSeedBoot.js` `if (adminOwned || storedVersion >= def.version) return`), so a shipped default-flip propagates to never-touched installs without clobbering an admin choice. Closes my earlier nit on the re-seed branch.
- ✅ Flag stays `workflows: false` for beta.

## Bonus catches

- **`5fbe514`** — pre-event reminder recipient resolved from the event row (not a non-existent column). Latent bug Luca surfaced while doing the review fixes; worth a separate test if it isn't already covered.
- **`d14f1d8`** — feature add: per-quote booking-workflow picker + `quote→invoice (no gallery)` built-in. New scope, not part of my original review; ships disabled per the cutover posture so the existing-stub-actions guard catches any misuse. ✅

## Tests

8 new cases in `440b2b3` covering EVERY contract change:
1. Built-ins seeded disabled (v6/v2/v3) for first beta.
2. Re-seed flips never-touched defaults but preserves an `admin_toggled_at`-owned flow.
3. Gallery, pre-event + booking built-ins all disabled.
4. `isBuiltinFlowActive` reflects enabled state (enabled-based mutex).
5. Legacy event-reminder pass stands down ONLY when `pre_event_email` is enabled.
6. Gate decision with no matching edge fails the run.
7. Rejects an unknown node type.
8. Refuses enabling a flow with unimplemented actions.

## Verdict

From my side, **merge cleared.**

the-luap:
One loose end I missed on the first review pass — surfacing it explicitly here so it doesn't get lost. **Not a merge-blocker for this beta** (my approval still holds), but should land before `workflows` flips default-on.

## `webhook` node type passes validation but has no handler → silent skip

`adminWorkflows.js`:
```js
const VALID_NODE_TYPES = new Set(['trigger', 'action', 'condition', 'branch', 'loop', 'wait', 'gate', 'webhook']);
```

`actions.js` registered actions: `send_email`, `queue_payment_check`, `escalate_to_collections`, `notify_gallery_expiring`, `notify_gallery_expired`, `notify_pre_event`, plus the 7 `DOCUMENT_ACTIONS` as `{skipped: true}` stubs. **No `webhook` action is registered.**

`engine.js:184-186` dispatches the `webhook` node type to `registry.getAction('webhook')` → undefined → records `{skipped: true, reason: 'unknown action webhook'}`:

```js
case 'action':
case 'webhook': {
  const actionKey = node.config?.action || (node.type === 'webhook' ? 'webhook' : 'noop');
  const action = registry.getAction(actionKey);  // ← undefined for 'webhook'
  const result = action ? (await action(ctx)) || {} : { skipped: true, reason: `unknown action ${actionKey}` };
```

So an admin can drag a webhook node onto the canvas, save the flow, enable it, and every run silently no-ops with a `skipped` step. Same shape as concern #5 (booking stub actions) but the `UNIMPLEMENTED_ACTIONS` guard in `adminWorkflows.js` only covers `DOCUMENT_ACTIONS` — a webhook walks past it.

## Three fix options

| Option | Effort | When |
| --- | --- | --- |
| **1. Wire a real webhook handler** that goes through `services/webhookService.js` (gets the NAT64-SSRF protection from GHSA-wmjx-pc37-272r / PR #643 for free). | ~½ day | The right end state. Before flipping `workflows` flag default-on. |
| **2. Drop `'webhook'` from `VALID_NODE_TYPES`** until (1) ships. | 1 line | Safest interim. |
| **3. Extend `UNIMPLEMENTED_ACTIONS` to also catch `node.type === 'webhook'`** — refuses enable until handler exists. | ~5 lines | Matches the booking-stub pattern. |

(1) is the right long-term path; (2) or (3) is the safe interim if (1) doesn't ship before this beta.

## Why this isn't blocking the merge

- `workflows` flag stays default-off this beta, so no install hits this without an admin opt-in.
- Symptom is "skipped step in run history" — visible, not silent data corruption.
- Same trust boundary as concern #5: admin's `workflows.manage` is the only one who can drop the node.

Just calling it out so the gap doesn't slip through when the flag default flips. Otherwise the PR's still merge-cleared from my side.

Luca-Timo:
@the-luap All review items are addressed — folding the second-pass webhook fix into this same update so it's one place.

## 🔴 Blockers (first review) — closed

1. **Prefetch-safe approvals** (`98ab7170`) — `GET /workflow-approvals/:token/:action` no longer mutates. A new read-only `peekApproval` renders a confirm/deny interstitial; only **POST** calls `actByToken`. Email-client/scanner prefetches can't advance a payment-confirm gate anymore.
2. **Loud gate-edge failure** (`98ab7170`) — `resumeRun` matches the decision handle **exactly** (no fall-back to `outEdge`'s sole-edge heuristic) and `failRun`s when a gate decision has no matching edge, so a lost `deny` shows as a failed run instead of a green `done`.

**Merge prerequisite:** the four cutover built-ins now ship `enabled: false` (`5893ecb2`).

## 🟡 Concerns — all done

1. **Node-type whitelist** in `validateGraph` (`d9274647`) — a typo'd `actoin` is rejected.
2. **`matchFilter` strict equality** (`dee8d40b`) — `===`/`!==`, eslint-disable dropped; `{value:0}` no longer matches `false`/`''`.
3. **Graph payload caps** (`d9274647`) — max 200 nodes / 500 edges / 16 KB per-node config.
4. **Engine-owns-disabled foot-gun** (`5893ecb2`) — reverted the mutex to **enabled-based**, so legacy paths run until a built-in is enabled and **disabling reverts to legacy** (no silent dark). Plus a list-page confirm when disabling a built-in (`c5f131ce`).
5. **Refuse enabling stub flows** (`d9274647`) — create/update/PATCH return `409` if a graph uses unimplemented actions; surfaced via the toggle's error toast.

## 🟢 Nits

- **`admin_toggled_at` sentinel** (`5893ecb2`, migration 148) — the boot re-seeder applies a default-flip only to never-touched built-ins, never overriding an admin's enable/disable/edit.
- **Orphan Mahnung PDF** — `applyReminder` unlinks the rendered file if `queueEmail` throws.
- **NL/PT/RU** fall back to EN via `fallbackLng` — flagged for a native pass.

## Second review — webhook node, now full Option 1

The `webhook` silent-skip is closed **through the delivery pipeline**:

- `af7eea8b` — interim real, SSRF-guarded one-shot handler.
- `675e41a2` — **supersedes it** with the pipeline version you asked for.

The node now **references a configured webhook subscription** (Settings → Webhooks) and enqueues a real `webhook_deliveries` row via `webhookService.enqueueForWebhook(webhookId, eventType, data)`. Delivery rides the existing worker, so it **inherits, not reimplements**: per-delivery SSRF re-validation (`validateExternalUrl` / GHSA-wmjx-pc37-272r), HMAC signing with the subscription's secret, retry/backoff, and the audit log. No schema change; `event_type` = `workflow.<trigger>`; unset/missing/inactive → observable `skipped`; dry-run doesn't enqueue. The editor's webhook field is now a subscription dropdown.

**Trade-off:** the node can no longer hit an ad-hoc URL — it must point at a managed subscription. That's the cost of signing/retries/audit, and it's what makes it ride your pipeline.

---

Backend suite **933 / 82 suites**, frontend `tsc` + build clean. Two pre-existing bugs this branch also fixes, independent of the engine — happy to peel them into their own PR against `beta` if you'd prefer: `5fbe514d` (pre-event reminder queried the non-existent `events.customer_account_id`) and `415c93a5` (event-type slug rename orphaned events/quotes + the per-type template; now cascades).

the-luap:
Re-verified `675e41a` at file:line. Everything in your comment is in the diff — `af7eea8` is correctly superseded, and the webhook node is now end-state by my measure.

## Webhook concern — closed

| | |
|---|---|
| Action no longer silent no-op | `backend/src/services/workflows/actions.js` registers `webhook`; test asserts `typeof webhook === 'function'` ✅ |
| Subscription-based (was raw URL) | reads `config.webhookId`; missing/unset → observable skip ✅ |
| Dry-run does not enqueue | returns `{ dryRun, would: 'webhook' }` before the enqueue call; test asserts `webhook_deliveries.count = 0` ✅ |
| Real run enqueues a delivery row | `webhookService.enqueueForWebhook` inserts into `webhook_deliveries` with `status='pending'`, `event_type='workflow.<trigger>'`; test asserts both ✅ |
| Inactive subscription skipped | `enqueueForWebhook` lookup is `{ id, active: true }`; test toggles `active=false` and asserts skip ✅ |
| **SSRF re-validation inherited** | `webhookDeliveryWorker.js:97` runs `validateExternalUrl(webhook.url)` per delivery row — not reimplemented in the workflow path ✅ |
| **HMAC signing inherited** | `webhookDeliveryWorker.js:123` runs `signPayload(webhook.secret, rawBody)` per delivery ✅ |
| Editor: subscription dropdown + hint | `WorkflowEditorPage.tsx` fetches `/admin/webhooks`; `NodeConfigPanel.tsx` renders `<select>` with `(inactive)` suffix and a hint pointing at Settings → Webhooks ✅ |
| Never throws | `enqueueForWebhook` try/catch returns `{ enqueued: false, reason }` ✅ |

The architecture is exactly what was asked: the flow author picks a configured webhook subscription, and the delivery rides the *same* worker that every other webhook uses. SSRF, HMAC, retries/backoff, and the deliveries audit log are inherited, not reimplemented — which means future hardening to the webhook pipeline automatically covers workflow-emitted deliveries too. Nice.

CI is green across all 12 checks (backend 933/82, frontend, both build matrices, fresh + upgrade installs, GitGuardian) and `mergeStateStatus: CLEAN` on `675e41a`.

## Verdict

Merge cleared from my side.

PR #674, opened 2026-06-26
fix: event-reminder, email-language & gallery-publish bugs surfaced during workflow testing

Luca-Timo:
Small, self-contained bug-fix PR. While testing the workflow reminder/booking pieces (already upstreamed), these **pre-existing** bugs turned up in the event-reminder, email and gallery-publish paths — independent of the engine.

## Pre-event reminder
- **`250b240`** — "Invalid Date" in the reminder email. `composePayload` pre-formatted `event_date` to `DD.MM.YYYY`, but the email processor runs date variables through `formatDate(value, language)` → `new Date("25.06.2026")` can't parse → "Invalid Date". Now passes the **raw** date and lets the processor localise it (same contract as the expiry mailer).
- **`3ccaed0`** — reminder skipped events that had **no inline `customer_email`/`host_email` but an assigned customer account**. It now falls back to the active assigned account(s) via `event_customer_assignments`, sent in the customer's preferred language. (The gallery-ready mail deliberately does *not* fall back — only the reminder.)

## Email language
- **`10559fd`** — recipient language is resolved from `email_data.eventId`, which only `queueEmail` injects. Direct `email_queue` inserts (e.g. the gallery-publish notification) set the `event_id` **column** but not the payload field, so those mails skipped `event.language` and fell through to the default — a gallery-ready mail in EN while the same event's reminder was DE. The processor now backfills `eventId` from the authoritative `event_id` column before rendering, so every send path resolves language consistently.

## Gallery publish
- **`c657892`** — publishing a gallery with **no inline email but assigned customer account(s)** sent nothing. It now reuses the existing `customer_gallery_assigned` "your galleries" email (per active account, in their language). The publish dialog also reflects who'll be notified (and shows a plain "Publish" when no one will).
- **`aa3471e`** — the publish dialog **hung** for a password-protected gallery with no inline email: the password field only renders on the inline-email path, but submit required a password whenever the gallery was protected, so it silently bailed. Password collection + validation are now gated on the same `needsPassword = requirePassword && customerEmail`; the no-inline-email path publishes without re-entering the password (existing hash kept; customer reaches it via the portal).

## Notes
- Branch was synced with the latest `beta` (`3356b22`).
- Tests: new coverage for the reminder recipient/date logic (raw `event_date`, assigned-account fallback); full backend suite **977 / 87 suites** green; frontend `tsc` + build clean. The email-language backfill, publish-route notification and dialog fix were verified manually — there's no existing harness for those send/route/UI paths.
- New UI strings are EN + DE; NL/PT/RU fall back to EN (flagged for a native pass).


the-luap (commented):
Reviewed all 9 files directly. Three real, well-contained bug fixes + the dialog UX polish, with a test for each backend behaviour change. No migrations, no middleware, no server.js / Dockerfile touch — surfaces I'd normally inspect first are clean. Skipped the deep-audit subagent because the PR is moderate-size and I'd read every changed line plus the unchanged helpers the new code depends on.

## 🔴 Blockers

None.

## 🟡 Concerns

None.

## 🟢 Nits

### `eventReminderService.js:326` — bare `true` vs codebase `formatBoolean(true)` convention

`where('c.is_active', true)` works (CI's fresh-install on SQLite passes), but the rest of `customerAccountsService.js` (e.g. line 1023, 1388) and the gallery-publish hook in this same PR (`adminEvents.js:1199` filtering on `a.is_active !== false && a.is_active !== 0`) use the explicit `formatBoolean(true)` / dual-shape pattern for SQLite/PG portability. The bare `true` here will silently cargo-cult into future code. Not worth a re-spin on its own, but worth a one-line tweak if you push another commit.

## ✅ Verified

- **`emailProcessor.js:868-871`** — eventId backfill from the queue row's `event_id` column is guarded on `emailData.eventId == null && email.event_id != null` (both-sides null check), so it can't clobber a deliberately set value. Fixes the publish-mail-shipped-EN-while-reminder-shipped-DE shape Luca described.

- **`eventReminderService.js:108-114`** — `composePayload` now passes raw `event.event_date`. Test asserts `'2026-06-25'` survives, not the previous DD.MM.YYYY pre-format that broke `formatDate`'s `new Date(...)` parse.

- **`eventReminderService.js:318-340`** — `resolveReminderRecipients` joins `event_customer_assignments` ⨝ `customer_accounts`, filters active + non-null email, dedupes case-insensitively. Both `runEventReminderPass` and `sendReminderForEvent` route through it. Each recipient gets a queue row tagged `fromEvent: true|false`, and the queueEmail call passes `row.id` for inline (so event language wins) or `null` for assigned-account (so the language resolver's step-2 lookup picks the customer's `preferred_language` by email match). Spot-checked `getRecipientLanguage` at `emailProcessor.js:112-144` — that step-2 lookup is the right one to land on for the assigned-account case.

- **`eventReminderService.js:306`** — `event_reminder_sent_at` stamp stays outside the recipients loop, so the per-event idempotency sentinel still works (the field marks the event, not the recipient). Good.

- **`adminEvents.js:1192-1207`** — publish hook only fires the assigned-account notification on the no-inline-email branch, fanning out to active assignments. Reuses `notifyCustomerOfNewAssignments(c.id, [eventId])` — that function itself re-checks `is_active` + email presence and filters archived/expired events (`customerAccountsService.js:1022-1052`), so the outer iteration is doubly safe. Wrapped in try/catch; warnings logged with `customerId` + `eventId` for debuggability.

- **`PublishGalleryDialog.tsx`** — `needsPassword = requirePassword && !!customerEmail` is now the single source of truth for both the password-field render and the submit gate, which closes the silent-bail dead state for password-protected galleries with no inline email. `willNotify` flips the CTA between "Publish & Notify" + Send icon and "Publish" + no icon. Send icon vanishing also makes the no-recipient case visually distinct.

- **`EventDetailsPage.tsx:2664`** — the `assignedCustomerCount` prop is derived from the event payload's `customer_accounts` array via an inline type assertion. Grungy but minimal — the proper fix would be threading the type through `events.service.ts`, out of scope here.

- **`customerAccountsService.js`** — just exposes the existing `notifyCustomerOfNewAssignments` via `module.exports`. No behavioural change to that function.

- **EN/DE locale** — both `descriptionAssignedAccount` and `justPublish` strings added in parallel, with German matching the established formal-Sie tone of the rest of the dialog. NL/PT/RU fall back to EN per the PR notes (acknowledged for a native pass).

- **Tests** — `pre-event payload passes the RAW event_date` and `pre-event falls back to the assigned customer account when the event has no inline email` are both at the right altitude — they pin the contract that was actually broken, not just shape-test the helpers. Second test additionally asserts `mail.event_id == null` in `email_queue` so future drift won't silently break the language-resolution path.

## Verdict

Merge cleared from my side. The bool-consistency nit is take-it-or-leave-it.

PR #679, opened 2026-06-28
feat(workflows): booking cutover — wire booking actions + hold documents behind approval gates

Luca-Timo:
## Summary

Makes the workflow engine actually drive a booking end-to-end. The built-in
`booking_full` / `booking_simple` flows previously had stubbed actions, so they
couldn't be enabled. This implements those actions and — crucially — has every
document side-effect created in a **held/draft** state behind the workflow's
review gate instead of firing immediately.

Builds on the merged `feat/workflow-engine` (#337).

## What changed

**Implemented the previously-stubbed workflow actions** (`workflows/actions.js`)
so `booking_full` / `booking_simple` are enableable:
- `prepare_event` — create the event from the booking
- `prepare_quote`
- `prepare_invoice` / `prepare_contract`
- `prepare_gallery`
- `reserve_date`
- `send_document`

**Documents are prepared, not auto-sent (the "cutover"):**
- `prepare_invoice` creates invoices in **draft/held** mode — `scheduled_send_at`
  stays `NULL` and status is `'scheduled'` (not `'pending_delivery'`) — so the
  invoice scheduler never auto-issues them before the review gate + an explicit
  `send_document` step. Pinned down by the new `bookingCutover.test.js`.
- `quote.accepted` / `quote.declined` emit is **deferred past the 15-minute
  response window** so the workflow doesn't convert a quote on the customer's
  first click and defeat the grace period the public quote page promises.
  Migration `149_add_quote_workflow_emitted_at` adds
  `quotes.workflow_response_emitted_at` as a once-only idempotency marker; the
  scheduler sweeps locked-but-not-yet-emitted responses.

**Approval UX:**
- Pending-approval rows (`WorkflowApprovalsPage`) and the dashboard
  pending-approvals card items (`AdminDashboard`) are now clickable, opening the
  underlying document.

## Why

The engine could define a booking flow but couldn't run one — the actions were
no-ops and documents would have auto-sent the moment they were created. This
closes both gaps: the flow produces real documents, and an admin reviews them
before anything leaves the building.

## Testing

- New `backend/__tests__/integration/bookingCutover.test.js` — asserts
  `prepare_invoice` creates the invoice with `scheduled_send_at = NULL` (held).
- Updated `workflowRoutes.test.js`.

## Notes

- Migration **149** (`quotes.workflow_response_emitted_at`), additive + idempotent.
- ⚠️ This branch was cut from an earlier `beta`, so the raw diff also shows
  version/CHANGELOG/route reversions. **Rebase onto current `beta` before merge**
  to drop that noise — the substantive change is the workflow actions + held
  documents + deferred emit above.

### DIFF
--- a/backend/__tests__/integration/bookingCutover.test.js
+++ b/backend/__tests__/integration/bookingCutover.test.js
@@ -1,0 +1,68 @@
+/**
+ * Booking cutover — prepare_invoice's draft seam. convertToInvoiceOnly({draft})
+ * must create the invoice(s) but leave scheduled_send_at NULL so the scheduler
+ * never auto-sends them before the workflow's review gate + explicit
+ * send_document.
+ */
+const crypto = require('crypto');
+const { bootCrmDb, seedMinimal } = require('./helpers/crmDb');
+
+jest.setTimeout(30000);
+
+describe('booking cutover — draft invoices on hold', () => {
+  let db; let cleanup; let adminId; let customerId; let quoteService;
+
+  beforeAll(async () => {
+    ({ db, cleanup } = await bootCrmDb());
+    ({ adminId, customerId } = await seedMinimal(db));
+    quoteService = require('../../src/services/quoteService');
+  }, 120000);
+
+  afterAll(async () => { if (cleanup) await cleanup(); });
+
+  async function acceptedQuote() {
+    const dealUuid = crypto.randomUUID();
+    const [id] = await db('quotes').insert({
+      quote_number: `Q-${dealUuid.slice(0, 8)}`,
+      customer_account_id: customerId,
+      status: 'accepted',
+      currency: 'CHF',
+      issue_date: '2026-01-01',
+      net_amount_minor: 100000, vat_amount_minor: 0, shipping_amount_minor: 0, total_amount_minor: 100000,
+      // A non-delivery installment so the contrast (scheduled date vs null) is meaningful.
+      payment_term_snapshot: JSON.stringify({ installments: [{ percent: 100, trigger: 'quote_accepted', offset_days: 0, label: 'Total' }], net_days: 30 }),
+      deal_uuid: dealUuid,
+      created_by_admin_id: adminId,
+    });
+    return id;
+  }
+
+  it('draft mode creates the invoice with scheduled_send_at = NULL (held), and returns its id', async () => {
+    const quoteId = await acceptedQuote();
+    const res = await quoteService.convertToInvoiceOnly(quoteId, adminId, { draft: true });
+    expect(Array.isArray(res.invoiceIds)).toBe(true);
+    expect(res.invoiceIds.length).toBeGreaterThanOrEqual(1);
+
+    const inv = await db('invoices').where({ id: res.invoiceIds[0] }).first();
+    expect(inv.status).toBe('scheduled');     // editable + sendInvoice can issue it
+    expect(inv.scheduled_send_at == null).toBe(true); // held — scheduler won't auto-send
+  });
+
+  it('without draft, the same installment IS scheduled (scheduled_send_at set)', async () => {
+    const quoteId = await acceptedQuote();
+    const res = await quoteService.convertToInvoiceOnly(quoteId, adminId);
+    const inv = await db('invoices').where({ id: res.invoiceIds[0] }).first();
+    expect(inv.status).toBe('scheduled');
+    expect(inv.scheduled_send_at == null).toBe(false); // normal convert → auto-send date set
+  });
+
+  it('prepare_contract path (createFromQuote) completes under SQLite — no in-trx deadlock', async () => {
+    const contractService = require('../../src/services/contractService');
+    const quoteId = await acceptedQuote();
+    const res = await contractService.createFromQuote(quoteId, adminId);
+    expect(res.contractId).toBeGreaterThanOrEqual(1);
+    expect(res.alreadyConverted).toBe(false);
+    const c = await db('contracts').where({ id: res.contractId }).first();
+    expect(c).toBeTruthy();
+  });
+});
--- a/backend/__tests__/integration/bookingCutover.test.js
+++ b/backend/__tests__/integration/bookingCutover.test.js
@@ -99,6 +99,40 @@ describe('booking cutover — draft invoices on hold', () => {
     expect(inv.scheduled_send_at == null).toBe(true); // still held — no auto-send
   });
 
+  it('finalizeQuoteResponses only fires once the 15-min response window has locked', async () => {
+    const mk = async (lockOffsetMs) => {
+      const dealUuid = crypto.randomUUID();
+      const [id] = await db('quotes').insert({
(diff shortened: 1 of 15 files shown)


the-luap (commented):
Reviewed all 15 substantive files directly. This PR overlaps with #674 (those 5 commits will fall out of the diff naturally once #674 merges), so my review focuses on the booking-cutover work: migration 149, workflow actions, held documents, deferred quote emit, the approval UX, and the deadlock-prevention pattern.

The design is sound and the testing is at the right altitude — `bookingCutover.test.js` pins the actual contracts (status, `scheduled_send_at`, `event_id` presence, idempotent claim) rather than just shape-testing the helpers. Skipped the deep-audit subagent because the surfaces are interconnected enough that a fresh read would essentially re-do mine.

## 🔴 Blockers

None.

## 🟡 Concerns

None blocking. Two minor things worth flagging but not worth a re-spin:

### 1. `prepare_invoice` crash-recovery via `deal_uuid`

`workflows/actions.js` crash-path: when `convertToInvoiceOnly` throws (e.g. quote already in `'converted'` state from a prior workflow tick), it adopts every invoice with the quote's `deal_uuid`. Today that's safe because `duplicateQuote → createQuote` mints a fresh `deal_uuid` (verified at `quoteService.js:2008+`), so a deal_uuid uniquely identifies one quote's invoice set. If any future flow ever shared `deal_uuid` across quote rounds (re-quote-with-same-deal pattern), this lookup would over-adopt. Not actionable today — flagging in case the deal_uuid scope ever broadens.

### 2. `finalizeQuoteResponses` throughput cap

`quoteService.js:1167-1175` fetches up to 200 deferred-but-now-locked responses per scheduler tick and filters in JS (justified by the comment — avoids SQLite/PG date-string comparison pitfalls). For normal volume that's plenty. If the scheduler is down for an extended window and a backlog accumulates, throughput is ~200/min. Probably not a real-world concern at PicPeak's scale.

## 🟢 Nits

### Same `formatBoolean(true)` consistency nit from #674

`workflowEngine.test.js:464` inserts `is_active: true` and `eventReminderService.js:326` continues `where('c.is_active', true)`. Both work (CI green confirms SQLite + PG paths), but the codebase convention elsewhere is `formatBoolean(true)`. Pure consistency — not blocking.

## ✅ Verified

**Held-state guarantee (the core promise of the cutover):**
- `invoiceService.js:3358-3362` — scheduler's flush query is `.where({ status: 'scheduled' }).andWhere(function() { this.whereNotNull('scheduled_send_at').andWhere('scheduled_send_at', '<=', now); })`. A held invoice has `scheduled_send_at = NULL`, so `whereNotNull` filters it out — auto-send is unreachable.
- `invoiceService.js:1935` — `sendInvoice` accepts `['scheduled', 'sent', 'overdue']` (rejects `pending_delivery`). The PR's `hold` flag at `invoiceService.js:1080-1089` switches `after_delivery` rows from `pending_delivery` to `scheduled` so `send_document` can issue them.
- Test `draft mode with the DEFAULT (after_delivery) payment term yields a SENDABLE scheduled invoice, not pending_delivery` pins exactly this — the contract that would otherwise be silently broken if someone tweaks the `pending_delivery` branch later.

**Deferred quote emit + atomic claim (no double-fire):**
- `quoteService.js:1117-1127` (`maybeEmitQuoteResponse`) — single-row atomic claim: `UPDATE quotes SET workflow_response_emitted_at = now WHERE id = ? AND workflow_response_emitted_at IS NULL`. The `claimed` rowcount gates the actual emit, so a concurrent scheduler sweep returns 0 and skips.
- `quoteService.js:1150-1180` (`finalizeQuoteResponses`) — same atomic claim per row. Locked-time comparison done in JS after candidates are fetched.
- Migration 149 — additive `workflow_response_emitted_at` timestamp, both `hasTable`/`hasColumn`-guarded, sane down.
- Window-edge behaviour: 0-minute window or `null` lockedAt → `locked = true` → emits immediately. Verified at `maybeEmitQuoteResponse` line 1119.

**Deadlock-prevention pattern (single-connection SQLite pool):**
- `db.js:648-654` + `appSettings.js:28-31` — both grew optional `executor` / `conn` params, defaulting to the global `db` so the rest of the codebase is unaffected.
- `quoteService`, `invoiceService`, `contractService` — every `getAppSetting` and `logActivity` call inside a transaction now passes `trx`. `createFromQuote` resolves the actor BEFORE opening the trx (line 1667). `convertToEvent` / `convertToInvoiceOnly` move `logActivity` AFTER commit. `createQuote` hoists `hasColumnCached` calls above the trx.
- The pattern is consistent across the changed services — no half-applied gaps spotted.

**Registry-driven `unimplementedActionsIn` (replacing the hardcoded `DOCUMENT_ACTIONS`):**
- `adminWorkflows.js:65-72` — now `workflows.registry.getAction(action)`. `getAction` exported from `registry.js:19` via `services/workflows/index.js:21`. The `n.type === 'action'` filter avoids false positives on trigger/condition/webhook nodes (webhook nodes carry `config.webhookId`, not `config.action`).
- `workflowRoutes.test.js` adds both halves of the new contract: an unregistered action gets 409, a real `prepare_invoice → send_document` flow now enables successfully.

**Idempotent re-entry:**
- `ctx.vars.prepared*` (`Event`/`Invoice`/`Contract`/`Quote`) short-circuit re-runs in `actions.js`.
- `convertToEvent` idempotent re-entry: when `converted_event_id` is already set, returns the existing `{ eventId, alreadyConverted: true, invoiceIds: [...] }` so `prepare_event` adopts (not re-creates).
- `prepare_invoice` crash-recovery lookup by `deal_uuid` (safe given `duplicateQuote → createQuote` mints a fresh `deal_uuid` — see Concern 1 above for the only edge case).

**Approval click navigation:**
- `WorkflowApprovalsPage.tsx:39-47` + `AdminDashboard.tsx:101-109` — `entityHref` maps `quote/invoice/event/contract/customer → /admin/{quotes|bills|events|contracts|customers}/:id`. Routes verified in `App.tsx:222-278`. No new information leak — the row already displays `entity_type #entity_id` before the click.

**Test depth:**
- `bookingCutover.test.js` (7 tests): held vs scheduled contrast, after_delivery → sendable in hold mode, `finalizeQuoteResponses` window logic + idempotency (the rerun-update returning 0), reserve_date zero-invoice, all 6 actions registered, no-deadlock on the duplicateQuote + createFromQuote paths.
- All checks green: backend (982 / 88 suites likely, based on the additions), build matrices, fresh-install, upgrade-from-bootstrap, GitGuardian. `mergeStateStatus: CLEAN` on `6e20d58`.

## Note on the branch

Author flagged the branch was cut from an earlier `beta`, so the raw diff currently shows the 5 commits from #674 (`3356b22..250b240`). Once #674 merges, those drop out and the diff cleans to just the booking-cutover work. Diff hygiene, not correctness.

## Verdict

Merge cleared from my side. Both yellow items are real-edge-case observations, not asks. Recommend landing #674 first so the diff narrows to just this PR's actual surface.

PR #680, opened 2026-06-28
Fix/invoice pdf multipage

Luca-Timo:
## Summary

Two related invoice/quote PDF fixes, split into two commits.

1. **Multi-page layout** — long invoices/quotes broke badly: page 1 only held
   ~6 line items with a large blank gap, a stray blank trailing page appeared,
   and the "Seite X von Y" footer was misnumbered (page 1 unnumbered, the blank
   page labelled "Seite 1 von 3"). Also dropped a duplicate IBAN block under the
   totals when a Swiss QR-bill is attached.
2. **Sub-cent rounding** — optional reconciliation so a total matches qty × rate
   arithmetic, surfaced as a "Rundung" row.

## 1. Multi-page layout (`707b676a`)

**Root cause:** before drawing the line-items table the renderer inflated page 1's
bottom margin to reserve space for the bottom-pinned totals block, but the
`finally` restored the margin on whichever page the *table ended on* — leaving
page 1 permanently short (maxY ≈ 570 instead of 802). Effects on any multi-page
document:
- the table broke far too early (huge blank gap on page 1);
- the page-number stamp landed below page 1's phantom bottom margin, which
  auto-paginated a stray blank trailing page and desynced the page numbers.

**Fix:** let the table paginate with the document's normal margins; the existing
`desiredTotalsY` check already advances to a fresh page when items would collide
with the pinned totals. Affects both invoices and quotes (shared `renderDocument`).

Also: suppress the IBAN block under the totals when `qrFormat === 'swiss'` — the
QR slip already prints the account/IBAN, so it was duplicated. EPC QR keeps the
block (its QR is on a trailing page).

| | Before | After |
|---|---|---|
| Pages (21-item invoice) | 4 (one blank) | 3 |
| Page 1 items | ~6 | 14, fills page |
| Page numbers | misnumbered | 1/3, 2/3, 3/3 |
| IBAN under total (Swiss QR) | duplicated | removed |

## 2. Sub-cent rounding reconciliation (`3fb3f0b8`)

Per-line totals are each rounded to the cent before the net is summed, so a long
time-based invoice can drift a few Rappen from qty × rate — e.g. 68 h × 32.25 =
2193.00 but the 21 rounded lines sum to 2193.02. This is the standard "sum of
rounded lines" convention (Stripe/QuickBooks/Xero do the same) and it foots, but
some issuers want the clean total.

New per-issuer setting **`crm_invoice_round_total`** (Settings → CRM → Invoices,
**default OFF**, no migration — read via `getAppSetting(..., false)`). When on,
the create paths store the full-precision net rounded **once**, and the drift is
shown as an explicit row:
Betrag Netto 2'193.02 ← Σ visible line totals (still foots)
Rundung -0.02
Gesamtbetrag 2'193.00 ← QR encodes this too

- New `src/utils/invoiceRounding.js` (`cleanNetMinor`) mirrors the migration-119
  hierarchy (priced sub-items override their parent) but sums at full precision;
  rate-agnostic, so mixed hourly rates reconcile to one clean net.
- `computeTotals` (quotes) + `createInvoice` + payload-preview gain the toggle.
- Render contexts derive the row as `storedNet − Σ(line totals)`; legacy/off
  documents have equal values ⇒ adjustment 0 ⇒ **byte-identical output**.
  Suppressed on Storno/Mahnung.
- Storno & tax report stay correct: both use the stored net scalar (createStorno
  negates `net_amount_minor`; neither re-sums lines).
## Testing
- New `backend/__tests__/utils/invoiceRounding.test.js` — real 68 h invoice,
  mixed rates, discounts, sub-item hierarchy, no-op case.
- Full backend sweep (pdfService / invoiceService / quoteService / rounding):
  **113 tests pass**. Frontend + backend build clean. OFF path verified
  byte-identical.
## Notes
- `nl` / `pt` / `ru` translations of "Rundung" are machine-generated — **flag for
  native review**; de/en/fr are solid.
- Not covered (deliberately — edge flows): the monthly-billing running-draft
  append and installment-split paths.


Luca-Timo:
@the-luap 

**Update — editor totals preview now reconciles too**

Closed the last surface that wasn't rounding-aware. The quote/invoice editor's live totals panel was still summing the per-line rounded totals, so with `crm_invoice_round_total` on it previewed e.g. **CHF 322.52** while the saved invoice + PDF correctly showed **322.50** with a Rundung row.

The editor now mirrors the backend (`cleanNetMinor`): clean net + a "Rundung" row folded into the VAT base and total, gated on the same setting and hidden when there's no drift. Both the bill and quote editors are wired.

So all three surfaces are now consistent:

- **Calc** (`computeTotals` / `createInvoice`) — stores the clean net
- **PDF** (`drawTotals`) — Rundung row
- **Editor preview** (`LineItemsTable`) — Rundung row ✅ (this update)

Default still **off**; zero-drift invoices render exactly as before (no row).

the-luap (commented):
Reviewed all 12 files directly. **Important diff-reading note up front**: this branch was cut from beta before #679 merged, so `git diff origin/beta..pr-680` makes it look like the deadlock-prevention pattern + booking `hold` flag + clickable-approval i18n strings are being reverted. They're not — a non-destructive three-way merge preview (`git merge-tree`) shows GitHub's `mergeStateStatus: CLEAN` is accurate and #679's `maybeEmitQuoteResponse`, `finalizeQuoteResponses`, `convertToEvent({hold, skipInvoices})`, the `trx || db` threading, `hold = false` on `spawnInstallmentInvoices`, and the `openEntity` translation key all survive intact. No regression risk.

That said, posting the diff against a current-beta base before merge would make this review cleaner for anyone else looking — purely diff hygiene, not correctness.

## 🔴 Blockers

None.

## 🟡 Concerns

None blocking. One note:

### `nl` / `pt` / `ru` machine-generated translations

Author flagged this in the PR body — "Afronding" / "Arredondamento" / "Округление" are technically correct words but may not match native legal/accounting terminology for the rounding line on a tax document. Worth a native pass post-merge. Not a blocker for the de/en/fr install base.

## 🟢 Nits

None this round.

## ✅ Verified

**Multi-page layout fix (the smoking gun):**
- `pdfService.js:1610-1648` before: `_origBottomMargin = doc.page.margins.bottom; doc.page.margins.bottom = _itemsBottomReserve; try { drawLineItems } finally { doc.page.margins.bottom = _origBottomMargin }`. The `finally` ran on whichever page the table ENDED on, so page 1's bottom margin was inflated and never restored — exactly the root cause the PR body describes. Page-number stamps landed below the phantom margin → auto-pagination spawned the blank trailing page → "Seite X von Y" desynced.
- New code at the same site removes the margin hack entirely and trusts the existing `desiredTotalsY` check (further down `renderDocument`) to advance to a fresh page when the last item row would collide with the pinned totals. The table's `header: true` flag handles continuation-page header repetition. Clean fix.

**IBAN dedup for Swiss QR:**
- `pdfService.js:931` — `showIbanHere = !isQuote && ctx.qrFormat !== 'swiss'`. EPC QR keeps the block (its QR is on a trailing page so the bank details on the invoice page still help). Swiss QR drops it because the slip itself prints "Konto / Zahlbar an" in human-readable form. Correct asymmetry.

**Byte-identical output when the toggle is OFF (the crucial safety property):**
- `invoiceService.js:790-794` + `quoteService.js:139-150` — `roundTotal` defaults to `false` from `getAppSetting('crm_invoice_round_total', false)`. When false, `cleanNetMinor` is never called; `netMinor` stays the sum of rounded line totals. Stored value unchanged.
- `invoiceService.js:1764-1777` + `quoteService.js:842-855` — render-time `displayedNetMinor` is derived as `Σ over (parent_line_item_id == null && parent_position == null/'') of line_total_minor`. For legacy persisted invoices both columns are null together (invariant maintained by `insertLineItemsHierarchical`), so `displayedNetMinor === stored.net_amount_minor` → `roundingAdjustmentMinor === 0`.
- `pdfService.js:870` — `if (roundingMinor !== 0) { ... }`. Zero → row omitted → PDF byte-identical. Verified.

**Storno + Mahnung suppression:**
- `invoiceService.js:1766` — `isReversalDoc = invoice.kind === 'storno' || invoice.kind === 'mahnung'`. Both negate the stored net and flip line-total signs at render, so the forward `storedNet − Σ lines` derivation would generate a fake reconciliation row. Suppressed: `roundingAdjustmentMinor = 0` for these, and `displayedNetMinor` falls back to the stored net. Storno + tax report use the stored net scalar (verified at the existing `createStorno` path that negates `net_amount_minor` directly), so neither downstream consumer is affected.

**Migration-119 hierarchy preserved in `cleanNetMinor`:**
- `invoiceRounding.js:67-81` — same parent-with-priced-children override rule as `computeTotals` phase 2. A top-level row with at least one priced sub-item contributes its children's exact contributions, not its own. Test "parent with priced sub-items derives from the children" pins this exact contract: `parent.own_price = 9999` is correctly ignored.

**Frontend `LineItemsTable` editor preview:**
- `LineItemsTable.tsx:228-247` — extends the `childPricingByParent` memo with `pricedSumExact` (un-rounded). The render-time `cleanExact` mirrors backend math: hierarchy override per top-level row, parent uses `pricedSumExact` when priced children exist, else its own un-rounded product. `roundingAdjustment = cleanSubtotal - subtotal` when `roundTotal` is on, else 0. Folded into `netForTotals` before VAT. The Rundung row at line 505-507 only renders when nonzero — same byte-identical-when-off shape as the PDF.
- Quote + Bill editor pages thread `roundTotal={appSettings?.crm_invoice_round_total === true}` into the table. `CrmSettingsPage.tsx:31` adds the key to `SETTING_KEYS` so the value reaches the editors.

**Test depth:**
- `invoiceRounding.test.js` (6 tests) pins:
  - The real 68 h × 32.25 case: sum-of-rounded = 2193.02, cleanNetMinor = 2193.00, drift -0.02 ✅
  - No-op when every line is cent-exact (adjustment 0)
  - Rate-agnostic mixed hourly rates
  - Per-line discounts at full precision
  - Migration-119 hierarchy override
  - `exactLineMinor` shape
- All at the right altitude — pins what the production paths must produce, not just helper shape.

**i18n:**
- `pdf-i18n.js` — `totals_rounding` added across all 6 locales (en/de/fr/nl/pt/ru). Used by `drawTotals`.
- Frontend en/de get `crm.lineItems.rounding` and the settings label. The fr/nl/pt/ru frontend strings flagged for native pass per PR body.

**Three-way merge verification (in case anyone else gets confused by the diff):**
- `git merge-tree --write-tree origin/beta pr-680` confirms the merged tree preserves:
  - `nextInvoiceNumber` with `getAppSetting('...', null, trx || db)`
  - `spawnInstallmentInvoices(..., hold = false)` parameter
  - `maybeEmitQuoteResponse` + `finalizeQuoteResponses` + `workflow_response_emitted_at` claim
  - `convertToEvent({hold, skipInvoices})` options
  - `openEntity` translation key
- All while integrating pr-680's `cleanNetMinor`, `roundingAdjustmentMinor`, `crm_invoice_round_total` toggle. Both PRs land cleanly together.

All 11 CI checks green. `mergeStateStatus: CLEAN` on `c2bc2b0`.

## Verdict

Merge cleared from my side. The diff display is confusing but the actual merge result is correct — verified with merge-tree before signing off.

PR #682, opened 2026-06-29
CRM: held invoices read as "Draft", + mark-paid / dashboard / reminder fixes

Luca-Timo:
A small bundle of invoicing + dashboard fixes that came out of real usage on a
manual-cadence, hours-billed setup. One coherent diff (11 files) rather than
five PRs.

### 1. Held / accumulator invoices now read "Draft", not "Scheduled"
`createInvoice` mints held invoices as `status='scheduled'` with
`scheduled_send_at = NULL` (hours billing, monthly/manual accumulators, the
booking-cutover holds). The scheduler never ships those — but they showed up as
"Scheduled" everywhere, which reads as "about to go out" when it isn't.

- New shared `isDraftInvoice(inv)` helper (`bills.service`): `scheduled` + no
  send date, **or** `is_monthly_draft`. A scheduled invoice *with* a future send
  date keeps "Scheduled".
- Applied to the Bills list, the invoice detail header, and the customer
  profile's invoice panel.
- The monthly/manual accumulator draft (`is_monthly_draft`, migration 128) was
  also **hidden from the Bills list** — it carries a real number and stamps the
  hours ("Billed: R-…"), so it looked lost. The list now opts in via a new
  `includeDrafts` query param (pickers/sub-lists stay clean); the hours "Billed"
  chip links straight to its invoice.

*Design note for you:* this labels **every** held invoice (no send date)
"Draft" app-wide — I think that's the honest reading, but if you'd rather scope
it tighter (e.g. accumulators only) say so and I'll narrow the helper. The
customer-facing portal is untouched (customers never see held invoices).

### 2. Mark-paid: add "bank transfer"
The mark-paid method list had Cash / Card / PayPal / TWINT but not bank
transfer — the default for the QR-bill / IBAN invoices picpeak issues
(`createInvoice` even falls back to `'bank_transfer'`). Added as the first
option; backend already takes `paymentMethod` as free text, so no API change.

### 3. Dashboard: revenue "year" tile toggles 365d ↔ calendar YTD
Instead of a fifth tile, the "Revenue · last 365 days" card is now clickable and
flips in place to calendar year-to-date (since Jan 1). Backend adds
`revenue.calendarYearMinor` using the same cash-basis `paid_at` window logic as
the existing trio; `StatCard` gains an optional `onClick`.

### 4. Reminder nit (your note on #674 / #679)
`eventReminderService` used bare boolean literals in its knex `.where()` calls
instead of `formatBoolean()`. Wrapped the four (`events.is_active` /
`is_archived` / `event_reminder_disabled`, `c.is_active`).

### Testing
- Backend `invoiceService` + `eventReminderService` + rounding suites green.
- Frontend + backend build clean.
- Held/off cases verified: a `scheduled` invoice with a send date still reads
  "Scheduled"; sent/paid/overdue/cancelled unchanged.
- i18n: `bills.payment.methods.bankTransfer` (de "Überweisung"),
  `bills.status.draft` (de "Entwurf"). Dashboard labels follow the existing
  `crmOverview.*` fallback-only pattern (no German keys there yet).

### Note
Branch was cut from my fork's `beta`, so the commit list shows the already-merged
booking/PDF merges — they contribute nothing to the diff against `beta`. Happy to
rebase to a clean linear history if you'd prefer.

### DIFF
--- a/backend/src/routes/adminDashboard.js
+++ b/backend/src/routes/adminDashboard.js
@@ -439,6 +439,10 @@ router.get('/crm-stats', adminAuth, async (req, res) => {
     const monthCutoff   = new Date(now -  30 * DAY);
     const quarterCutoff = new Date(now -  90 * DAY);
     const yearCutoff    = new Date(now - 365 * DAY);
+    // Calendar year-to-date (Jan 1 of the current year, local time) —
+    // the dashboard's revenue "year" tile can toggle between this and
+    // the trailing-365-day window.
+    const calendarYearCutoff = new Date(new Date(now).getFullYear(), 0, 1);
 
     // ---- quotes: counts by status ---------------------------------
     let quoteCounts = { draft: 0, sent: 0, accepted: 0, declined: 0, expired: 0, converted: 0 };
@@ -459,6 +463,7 @@ router.get('/crm-stats', adminAuth, async (req, res) => {
     let revenueMonthMinor = 0;
     let revenueQuarterMinor = 0;
     let revenueYearMinor = 0;
+    let revenueCalendarYearMinor = 0;
     let outstandingTotalMinor = 0;
     let outstandingCount = 0;
 
@@ -504,9 +509,10 @@ router.get('/crm-stats', adminAuth, async (req, res) => {
             .first();
           return Number(row?.total || 0);
         };
-        revenueMonthMinor   = await winSum(monthCutoff);
-        revenueQuarterMinor = await winSum(quarterCutoff);
-        revenueYearMinor    = await winSum(yearCutoff);
+        revenueMonthMinor        = await winSum(monthCutoff);
+        revenueQuarterMinor      = await winSum(quarterCutoff);
+        revenueYearMinor         = await winSum(yearCutoff);
+        revenueCalendarYearMinor = await winSum(calendarYearCutoff);
 
         // Outstanding: every invoice that's been sent but not fully
         // paid (sent + overdue). Outstanding = total - paid. We sum
@@ -568,9 +574,10 @@ router.get('/crm-stats', adminAuth, async (req, res) => {
       quotes: quoteCounts,
       invoices: invoiceCounts,
       revenue: {
-        monthMinor:   revenueMonthMinor,
-        quarterMinor: revenueQuarterMinor,
-        yearMinor:    revenueYearMinor,
+        monthMinor:        revenueMonthMinor,
+        quarterMinor:      revenueQuarterMinor,
+        yearMinor:         revenueYearMinor,
+        calendarYearMinor: revenueCalendarYearMinor,
       },
       outstanding: {
         totalMinor: outstandingTotalMinor,
--- a/backend/src/routes/adminInvoices.js
+++ b/backend/src/routes/adminInvoices.js
@@ -342,6 +342,7 @@ router.get(
     query('customerAccountId').optional({ values: 'falsy' }).isInt({ min: 1 }),
     query('sourceQuoteId').optional({ values: 'falsy' }).isInt({ min: 1 }),
     query('unpaidOnly').optional({ values: 'falsy' }).isBoolean(),
+    query('includeDrafts').optional({ values: 'falsy' }).isBoolean(),
     query('q').optional({ values: 'falsy' }).isString().isLength({ max: 255 }),
     query('sort').optional({ values: 'falsy' }).isIn(['newest', 'oldest', 'issue_asc', 'issue_desc', 'due_asc', 'due_desc', 'value_asc', 'value_desc', 'customer_asc', 'customer_desc']),
     query('page').optional({ values: 'falsy' }).isInt({ min: 1 }),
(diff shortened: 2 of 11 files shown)


the-luap (commented):
Reviewed all 11 substantive files directly. Same diff-staleness pattern as #680 — branch was cut before #680 merged, so the raw `git diff origin/beta..pr-682` view shows #680's PDF + rounding work duplicated. The actual pr-682-only surface is small: 11 files / +135 / -39, and `mergeStateStatus: CLEAN` means the three-way merge resolves cleanly now that #680 is on beta.

## 🔴 Blockers

None.

## 🟡 Concerns

None.

## 🟢 Nits

### `StatCard onClick` button a11y

`CrmOverviewSection.tsx:283` — the new `onClick` wrapper uses `<button type="button" className="block w-full text-left hover:opacity-90 transition-opacity">`. No `focus:outline-*` ring, so keyboard users tabbing to the tile get no visible focus indicator. The `to` variant uses `<Link>` which inherits the global link focus styles, so it's only the new button branch that lacks them. One-line tailwind tweak if you want.

## ✅ Verified

**`isDraftInvoice` predicate matches the held-state semantics from #679:**
- `bills.service.ts:107-110` — `inv.isMonthlyDraft || (inv.status === 'scheduled' && !inv.scheduledSendAt)`.
- The booking-cutover #679 creates held invoices with `status='scheduled'` + `scheduled_send_at = NULL`. Those are picked up. ✅
- A scheduled invoice with a future `scheduled_send_at` keeps the "Scheduled" badge (second branch false). ✅
- `pending_delivery`, `sent`, `paid`, `overdue`, `cancelled`, `skipped` all unaffected. ✅
- Backend mapping confirmed at `adminInvoices.js:165` — `i.is_monthly_draft === true || i.is_monthly_draft === 1` handles both PG bool and SQLite int, so the frontend's `isMonthlyDraft` check works on both DB paths.

**`includeDrafts` query param wiring:**
- `adminInvoices.js:362` adds the query validator; passes `includeMonthlyDrafts: true|false` to `listInvoices`.
- Service already supported the flag at `invoiceService.js:448-452` (existing code, no change needed). The default-hide invariant from migration 128 is preserved — only the Bills list opts in (`BillsListPage.tsx:50`), pickers + `CustomerCrmPanels` leave it off.

**Dashboard YTD toggle:**
- Backend: `adminDashboard.js:442-444` adds `calendarYearCutoff = new Date(year, 0, 1)`. Reuses the existing `winSum(cutoff)` aggregator with `paid_at >= cutoff`, so it inherits the same cash-basis logic as the existing trio. Same time-zone behaviour as the rolling windows (server local), so consistent.
- Frontend: tile flips between `'rolling'` and `'calendar'` via `useState`. Label + value both swap. Hint sub-text added to telegraph the click.

**Bank transfer in mark-paid:**
- `BillDetailPage.tsx:528` — added as the first option, value `'bank_transfer'`. Backend `paymentMethod` is free text (verified: `createInvoice` already falls back to `'bank_transfer'`), so no API surface change.
- EN/DE strings added under `bills.payment.methods.bankTransfer`.

**HoursSection "Billed" chip link:**
- `HoursSection.tsx:451-468` — wraps the chip in `<Link to="/admin/clients/bills/${e.invoiceId}">` when `invoiceId` is present. Falls back to `<span>` for legacy entries without invoice linkage.
- Route verified: `App.tsx:266` mounts `bills/:id` under the `clients` parent layout, so `/admin/clients/bills/:id` is the correct fully-qualified path.

**`eventReminderService` formatBoolean wrap (the nit I raised on #674 + #679):**
- All four bare-true wheres at `eventReminderService.js:180-182` + `:332` now use `formatBoolean()`. Import added at line 58. Closes the consistency thread cleanly — thanks for picking that up.

**Held invoice on the customer profile + detail header:**
- `CustomerCrmPanels.tsx:207-220` and `BillDetailPage.tsx:267-270` both use `isDraftInvoice(inv)` as the gate — same predicate as the list, so the three surfaces stay in lockstep.

**Customer-facing portal untouched:**
- Grepped — no changes under public routes. The PR body's claim that "customers never see held invoices" holds: there's no held-invoice surface on the public side, so this is admin-only UX.

**i18n:**
- `bills.status.draft` (EN "Draft" / DE "Entwurf"), `bills.payment.methods.bankTransfer` (EN "Bank transfer" / DE "Überweisung"). Dashboard labels (`crmOverview.revenue.yearCalendar`, `crmOverview.revenue.toggleHint`) use fallback-only strings, consistent with the existing `crmOverview.*` keys.

**Test depth:**
- No new tests in pr-682. The `isDraftInvoice` helper is a 2-condition predicate — testing it would be tautological, and the PR body documents manual verification of the held/scheduled/sent/paid/overdue/cancelled cases. The behavioural surface this PR touches (badge labels, dashboard toggle, mark-paid option) doesn't have meaningful unit-testable shape beyond what the rendered components do, and there's no existing UI test harness to extend. Acceptable.

All 11 CI checks green. `mergeStateStatus: CLEAN` on `e4367e0`.

## Verdict

Merge cleared from my side. The a11y nit is take-it-or-leave-it.

PR #694, opened 2026-06-30
feat(updates): "What's New" highlights after update + pre-update teaser

Luca-Timo:
## feat(updates): "What's New" highlights after update + pre-update teaser

Shows admins what changed, sourced from the GitHub release notes — **no AI at runtime.** Short bullets are written once per release in CI via GitHub Models into a `<!-- whatsnew -->` block; the app reads that block and **falls back to the changelog `### Features`** for releases without it, so it works against today's releases immediately.

- `utils/whatsNew.parseWhatsNew(body)` — curated block else Features section, strips scope/PR-links, de-dups, caps at 8. Unit-tested.
- `GET /admin/system/updates/whatsnew` — highlights for every version moved through since a **per-instance** marker (`whatsnew_last_seen_version` in `app_settings`); fresh installs self-anchor silently. Best-effort, never throws.
- `POST /admin/system/updates/whatsnew/seen` — advance the marker.
- `GET /admin/system/updates` also returns `latestHighlights` (pre-update teaser).
- Frontend: `WhatsNewBanner` (green bar → modal w/ "Full changelog" link) on the dashboard via `adminService`; `UpdateNotification` gets a "New features include:" teaser. i18n de/en.

### Review surface
- **No migration** — single `app_settings` row.
- **No `server.js` change** — extends the already-mounted `adminSystem` router.
- Reuses `adminAuth` + `requirePermission('settings.view')`. Admin-only.

### Testing
- `parseWhatsNew` unit test; end-to-end against live `PicPeak/picpeak` releases.
- The CI generation step (`docs/ci/whatsnew-highlights.yml`) was **validated end-to-end on a fork**: extract Features → `openai/gpt-4o-mini` → inject the `<!-- whatsnew -->` block into real release notes → `parseWhatsNew()` reads it back. Build + suites green.

### Follow-up (not in this PR)
Move `docs/ci/whatsnew-highlights.yml` into `.github/workflows/`. It needs **GitHub Models enabled for the org** (owner setting) — the one thing a fork can't verify. Failure is isolated by `continue-on-error` + a deterministic fallback, so it can never break a release; until enabled, the app uses the Features-fallback bullets.

### DIFF
--- a/backend/__tests__/utils/whatsNew.test.js
+++ b/backend/__tests__/utils/whatsNew.test.js
@@ -1,0 +1,54 @@
+const { parseWhatsNew } = require('../../src/utils/whatsNew');
+
+describe('parseWhatsNew', () => {
+  it('prefers the curated <!-- whatsnew --> block', () => {
+    const body = [
+      '<!-- whatsnew -->',
+      '- Invoice drafts in list',
+      '- Bank transfer payments',
+      '<!-- /whatsnew -->',
+      '',
+      '### Features',
+      '* **invoices:** something long that should be ignored ([#1](http://x))',
+    ].join('\n');
+    expect(parseWhatsNew(body)).toEqual(['Invoice drafts in list', 'Bank transfer payments']);
+  });
+
+  it('falls back to the Features section, stripping scope + commit links', () => {
+    const body = [
+      '## [3.73.0-beta.0](http://x) (2026-06-29)',
+      '',
+      '### Features',
+      '',
+      '* **dashboard:** revenue tile toggles 365 days ([d1c9e02](http://c))',
+      '* **invoices:** surface monthly drafts in the Bills list ([e457656](http://c))',
+      '',
+      '### Bug Fixes',
+      '',
+      '* **invoices:** add bank transfer ([e96ef4c](http://c))',
+    ].join('\n');
+    expect(parseWhatsNew(body)).toEqual([
+      'revenue tile toggles 365 days',
+      'surface monthly drafts in the Bills list',
+    ]);
+  });
+
+  it('excludes Bug Fixes from the fallback', () => {
+    const body = '### Features\n* **a:** feature one\n### Bug Fixes\n* **b:** fix one';
+    expect(parseWhatsNew(body)).toEqual(['feature one']);
+  });
+
+  it('caps at 8 bullets and de-dups', () => {
+    const lines = Array.from({ length: 12 }, (_, i) => `- bullet ${i % 9}`);
+    const body = `<!-- whatsnew -->\n${lines.join('\n')}\n<!-- /whatsnew -->`;
+    const out = parseWhatsNew(body);
+    expect(out.length).toBe(8);
+    expect(new Set(out).size).toBe(8);
+  });
+
+  it('returns [] for empty / non-string input', () => {
+    expect(parseWhatsNew('')).toEqual([]);
+    expect(parseWhatsNew(null)).toEqual([]);
+    expect(parseWhatsNew(undefined)).toEqual([]);
+  });
+});
--- a/backend/src/routes/adminSystem.js
+++ b/backend/src/routes/adminSystem.js
@@ -7,7 +7,9 @@ const path = require('path');
 const os = require('os');
 const { formatBoolean } = require('../utils/dbCompat');
 const logger = require('../utils/logger');
-const { checkForUpdates, getCurrentChannel, getReleasesSince } = require('../services/updateCheckService');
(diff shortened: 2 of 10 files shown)


the-luap (approved):
Tested locally — `git checkout pr-694` + spun up the backend + Vite dev pointing at it, seeded the `whatsnew_last_seen_version` setting to an older version to trigger the banner, exercised the full lifecycle. The feature does exactly what the PR description claims.

## What I verified end-to-end

| | Verified |
|---|---|
| Banner appears on the admin dashboard when `current > last_seen_version` | ✅ — green bar with sparkles + 3-bullet teaser + "What's new" CTA |
| Modal expands per-version with "Full changelog" external link | ✅ — `v3.74.0-beta.0` (1 bullet) and `v3.73.0-beta.0` (2 bullets) listed separately |
| "Got it" advances the server marker | ✅ — verified in SQLite: `whatsnew_last_seen_version` went from `"3.72.0-beta.0"` → `"3.74.0-beta.0"` |
| Banner disappears after the marker advances + stays gone on reload | ✅ |
| Bullets sourced from the real GitHub release notes via the `### Features` fallback path | ✅ — the `<!-- whatsnew -->` curated block isn't there yet (that's your follow-up CI workflow), and the fallback correctly extracted bullets from `v3.74.0-beta.0` and `v3.73.0-beta.0`'s actual release bodies |
| Fresh install self-anchors silently (`whatsnew_last_seen_version` missing → upsert + `hasNews: false`) | ✅ — `adminSystem.js:`L_running set on first call without a marker |
| Per-instance dismissal (not per-admin) | ✅ — by design, called out in PR body |
| EN + DE i18n | ✅ — both render cleanly |
| Layout: sits below the existing migration banner (#687) | ✅ — clean stacking, no overlap |

## Backend review

- `backend/src/utils/whatsNew.js` — `parseWhatsNew(body)` is well-bounded: curated-block-first, falls back to the `### Features` section, strips scope prefixes + PR-links, de-dups, caps at 8. Unit tests at `__tests__/utils/whatsNew.test.js` cover the obvious shapes. ✅
- `backend/src/routes/adminSystem.js` — `/updates/whatsnew` is wrapped in try/catch (`best-effort, never throws` — verified), checks `UPDATE_CHECK_ENABLED`, scopes by `adminAuth + requirePermission('settings.view')`. No migration needed — single `app_settings` row. ✅
- `getReleasesSince(lastSeen, channel)` is already in `updateCheckService` (pre-existing) — reused, not reimplemented. ✅

## Frontend review

- `WhatsNewBanner.tsx` — local `hidden` state for instant UX feedback, React Query `staleTime: 5min`, mutation invalidates the query on success. Mount point chosen correctly (above `UpdateNotification`). ✅
- `UpdateNotification.tsx` teaser — `slice(0, 4)` cap on `latestHighlights` is reasonable. Only renders when the array is non-empty + present. ✅
- `admin.service.ts` — `getWhatsNew()` + `markWhatsNewSeen()` shape matches the endpoint. ✅

## 🟢 Nits (non-blocking)

**Teaser bullets aggregate across versions without grouping.** The banner shows the first 3 bullets via `data.versions.flatMap((v) => v.bullets).slice(0, 3)` — so if v3.74 has 1 bullet and v3.73 has 2, the user sees "3 highlights" without realizing they span two versions. Mostly harmless because the modal makes the per-version split obvious. Could add a small `Highlights from v3.74.0-beta.0 and v3.73.0-beta.0` subhead under the title if you want, but I'd say ship it as-is.

## Bystander issue (NOT in this PR's scope)

While running `npm run dev` for the local test, I hit a pre-existing Babel error: **`SettingsPage.tsx:58:41 — Identifier 'Mail' has already been declared.`** `Mail` is imported twice in that file (line 20 + line 58). Not introduced here, not something this PR touches. esbuild (`npm run build` + Docker production) silently deduplicates same-source same-name imports, which is why CI is green; Babel (Vite dev's fast-refresh) is stricter and rejects. Filing as a separate 1-line fix-PR — flagging here only so it's not blamed on this PR if anyone else hits it while reviewing.

## Notes on the follow-up

You called this out in the PR body but worth restating: `docs/ci/whatsnew-highlights.yml` needs to move to `.github/workflows/` AND **GitHub Models enabled at the org level**. Until then, the Features-fallback path carries the feature — which is fine, the bullets I saw locally were extracted cleanly. Worth doing the Models enable + workflow-move as a small follow-up after this merges.

## Verdict

Merge cleared from my side. The size-gate (`bypass-size-gate` failing at 463 lines vs 300 cap) is working exactly as designed — needs a maintainer review for features this large, can't self-merge. Approving via review unblocks the merge button.

Issue #26, opened 2025-09-19
[BUG] view gallery URL generation does not work properly


fiskare:
**Describe the bug**
The generated link under "View Gallery" is wrong an it is not possible to view the gallery without changing the URL. This is the URL which is generated:
http://mypicpeakinstance/gallery//gallery/wedding-test-2025-09-18/eb7faf2aab385f1b3f65aaeb20afd782

**To Reproduce**
1. Go to 'Admin Panel'
2. Click on 'Events'
3. Click on 'View Gallery' (under actions)

**Expected behavior**
URL should be like this:
http://mypicpeakinstance/gallery/wedding-test-2025-09-18/eb7faf2aab385f1b3f65aaeb20afd782

**Environment (please complete the following information):**
 - OS: [Debian 12]
 - Browser: [Firefox 142]
 - PicPeak Version: [e.g. 1.0.128]
 - Deployment Method: [native via steup.sh]
 - Database: [standard]

the-luap:
should be fixed in latest commit

Issue #25, opened 2025-09-19
[BUG] company logo upload doesn't work properly


davidcodes26:
**Describe the bug**
After uploading the custom logo the field where it should be displayed is only showing the standard image missing icon and when opening the image in a new tab it says: Cannot GET /uploads/logos/logo-1758292181023.png
The "broken" image also cannot be removed via the red circle, it appears again after saving.

**To Reproduce**
1. Upload logo in the tab branding
2. save the settings

**Expected behavior**
The logo is displayed correctly in the gallery and the settings.

**Environment (please complete the following information):**
 - OS: Debian 13.1
 - Browser: Arc/Chromium
 - PicPeak Version:  v1.0.128
 - Deployment Method: Native via Script
 - Database: [e.g. PostgreSQL 15, SQLite]

davidcodes26:
Still the same error in v1.1.4

Cannot GET /uploads/logos/logo-1758881298014.png

ls /opt/picpeak/app/backend/storage/uploads/logos/
shows the correct file
logo-1758881298014.png

Issue #24, opened 2025-09-19
[BUG] not all values in settings and branding are saved


davidcodes26:
**Describe the bug**
When I try to disable the "Require password for all galleries" it resets itself to active after I save the settings and when I set a company name for example, it won't be displayed in the preview after saving. 

**Expected behavior**
The value should be set correctly

**Environment (please complete the following information):**
 - OS: Debian 13.1
 - Browser: Arc/Chromium
 - PicPeak Version:  v1.0.128
 - Deployment Method: Native via Script
 - Database: [e.g. PostgreSQL 15, SQLite]

the-luap:
should be fixed in latest commit

the-luap:
I never tested this feature but I have redesigned it to be able to unset password requirements per gallery. It is included in latest changes.

Issue #23, opened 2025-09-19
[BUG] no admin login possible after gallery opened


davidcodes26:
**Describe the bug**
When I open a gallery with individual password in a new tab I will be logged out from the admin panel. After login I'm immediately redirected to the login page again. Only after I logout from the gallery view I am able to use the admin panel again.

**To Reproduce**
1. Open a gallery via admin panel that is secured by a password
2. Use the password to enter the gallery
3. Try to login to the admin panel again

**Expected behavior**

**Environment (please complete the following information):**
 - OS: Debian 13.1
 - Browser: Arc/Chromium
 - PicPeak Version:  v1.0.128
 - Deployment Method: Native via Script

maximilianbiebl:
For me it’s quite similar: after installing the latest PicPeak version, I can log in and briefly see the dashboard, but I’m immediately redirected back to the login page. No visible error message is shown.

the-luap:
should be fixed in latest commit

PR #703, opened 2026-06-30
ci(whatsnew): generate release highlights via GitHub Models

Luca-Timo:
Activates the **What's New highlights** step so each release's bullets are generated automatically, completing the What's New feature shipped in #694.

### What it does
When release-please publishes a release, this condenses that release's `### Features` into ≤8 short, user-facing bullets via **GitHub Models** (free tier, `models: read`) and injects a `<!-- whatsnew -->` block at the top of the release notes. The app reads that block (`backend/src/utils/whatsNew.parseWhatsNew`) and shows it in the admin What's New banner/modal.

### Why a reusable `workflow_call`, not a `release: published` trigger
release-please creates releases with the default `GITHUB_TOKEN`, and **GitHub never starts new workflow runs from token-generated events** — a standalone `on: release` workflow would simply never fire. So the step runs as a `needs:` job *inside* the release-please run, and is shared as a reusable `workflow_call` so the `stable` and `main`/beta channels stay in sync.

### Safety
Best-effort only: `continue-on-error` + a deterministic fallback (first 8 Features lines, de-scoped) mean it **can never break a release**. If Models is unavailable the AI step returns nothing and the fallback is used. Idempotent — re-runs strip any prior block before re-injecting.

### 🙏 One thing only you can do, @the-luap
**Please enable GitHub Models for the PicPeak org** (Org → Settings → Models, or the owner-level "Models" toggle). It's free and read-only here. Until it's on, releases still get the deterministic fallback bullets — flipping it on upgrades them to the AI-condensed version. That's the single setting I can't reach from a fork.

Validated end-to-end on a fork (extract → `openai/gpt-4o-mini` → inject into real release notes → `parseWhatsNew()` reads it back). `actionlint` clean. 3 files, +20/−76.

### DIFF
--- a/.github/workflows/release-please-beta.yml
+++ b/.github/workflows/release-please-beta.yml
@@ -35,3 +35,16 @@ jobs:
           echo "" >> $GITHUB_STEP_SUMMARY
           echo "Docker images will be built and tagged with this beta version." >> $GITHUB_STEP_SUMMARY
 
+  # Best-effort "What's New" highlights on the freshly-created release. Runs in
+  # this same workflow run (not a `release:` trigger) because release-please
+  # creates the release with GITHUB_TOKEN, which never starts new workflow runs.
+  whatsnew:
+    needs: release-please
+    if: ${{ needs.release-please.outputs.release_created }}
+    permissions:
+      contents: write   # edit the release body
+      models: read       # GitHub Models (free tier)
+    uses: ./.github/workflows/whatsnew-highlights.yml
+    with:
+      tag: ${{ needs.release-please.outputs.tag_name }}
+
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -34,3 +34,16 @@ jobs:
           echo "" >> $GITHUB_STEP_SUMMARY
           echo "Docker images will be built and tagged with this version." >> $GITHUB_STEP_SUMMARY
 
+  # Best-effort "What's New" highlights on the freshly-created release. Runs in
+  # this same workflow run (not a `release:` trigger) because release-please
+  # creates the release with GITHUB_TOKEN, which never starts new workflow runs.
+  whatsnew:
+    needs: release-please
+    if: ${{ needs.release-please.outputs.release_created }}
+    permissions:
+      contents: write   # edit the release body
+      models: read       # GitHub Models (free tier)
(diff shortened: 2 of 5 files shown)


the-luap:
@Luca-Timo tbh I am not able to find that settings for a model. neither within Org or repository or personal level. On the Org level - if it is related to copilot - there is just a message that copilot is deactivated for free tier at the moment.

Luca-Timo:
No problem — you don't need to find the setting. Turns out org-level GitHub Models is a GitHub Enterprise Cloud feature (an enterprise owner has to enable it for the enterprise first), so it isn't exposed for a standalone org at all — which matches the "Copilot deactivated for free tier" message you saw. It's not a wrong-menu thing; the toggle genuinely isn't there for us.

Good news: it doesn't matter. I pushed a fix so the AI step fails soft — with Models unavailable it degrades to a deterministic fallback that builds the bullets straight from the release's ### Features. So the feature is complete and safe to merge as-is.

What admins actually see, with vs without Models — given these release Features:

* **invoices:** badge held (unsent, no send date) invoices as "Draft" (#26)
* **dashboard:** revenue "year" tile toggles 365 days ↔ calendar YTD (#25)
* **invoices:** add bank transfer to the mark-paid method list (#24)
* **invoices:** surface monthly/manual accumulator drafts in the Bills list (#23)

Without Models (what ships today — deterministic fallback):
- badge held (unsent, no send date) invoices as "Draft"
- revenue "year" tile toggles 365 days ↔ calendar YTD
- add bank transfer to the mark-paid method list
- surface monthly/manual accumulator drafts in the Bills list

With Models (the AI polish — ≤8 bullets, 3–4 words each):
- Draft invoice badges
- Revenue YTD toggle
- Bank-transfer payments
- Drafts in Bills list
- 
So the fallback is fully readable — the raw commit subjects — just longer and a bit more technical; Models only tightens them into punchy highlights.

I don't have an AI API key I can hand over. But if you've got one for any provider we can use, I'm happy to adapt the workflow to call it instead of GitHub Models — it'd just be a repo/org Actions secret (e.g. AI_API_KEY), no code from you. Otherwise the deterministic version is honestly fine to ship.

Luca-Timo:
@the-luap just saw this, still a bit messy, but a great start.

image
image

PR #708, opened 2026-06-30
Surface which files failed during photo upload

andredlng:
## Description

A partial photo upload only told the admin **"some files failed"** with no way to find out *which* ones — even though the information already existed on both sides and was being discarded:

- the upload response already returns per-file rejections as `errors: [{ filename, error }]`, but the frontend only read `successCount` and dropped `errors` entirely;
- `useUploadProgress` already exposes `failedPhotos` (`{ id, filename, error }`), but the component only used the failed **count**.

This adds a **dismissible failure report** to the upload modal that names every file that didn't make it into the gallery, grouped by the stage where it failed, each with its reason:

- **Rejected** — per-file validation rejections from the upload response (previously discarded).
- **Transfer** — whole-chunk request failures (now captured *with* the error, not just the filename).
- **Processing** — background-worker failures from `useUploadProgress.failedPhotos`.

The count-only "some files failed" toast now points at the list. New `upload.failures.*` keys added for English and German.

The upload modal previously auto-closed the moment the transfer finished, which would have unmounted the report before it could be read. Closing is now split out (`onUploadSettled`): a clean upload still auto-closes, but a partial failure keeps the modal open so the report stays visible. As a side effect this also fixes a pre-existing case where an upload in which *every* file failed left the modal spinning forever — the transfer UI now resets when nothing was queued.

Fixes # (no tracking issue — UX improvement)

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] Bug fix (per-file `errors` returned by the backend were silently dropped)

## How Has This Been Tested?

- [x] Unit tests pass (`npm test`) — `PhotoUpload.failureReport` (drives a real upload, asserts rejected + processing rows render with filename + reason, and dismissal) and `PhotoUploadModal` (auto-close on clean upload vs. stay-open on partial failure). `tsc -b` adds no new type errors over the baseline; `npm run lint` reports 0 errors on changed files.
- [x] Manual testing completed — Playwright pass on a live Docker deployment: uploaded a valid image alongside two corrupt PNGs; the valid one queued, the modal stayed open, and the report listed both failures by name with reasons ("Invalid image file: … pngload: invalid IHDR chunk size"), tagged `Rejected`. Dismiss verified.
- [x] Tested on Docker deployment — full stack via `docker compose`, frontend image built from this branch.

> Verified end-to-end on a local Docker deployment built from this branch — the failure report renders with each failed file + reason, the modal stays open on partial failure, and a clean upload still auto-closes. (Screenshots available on request.)

**Test Configuration**:
* PicPeak Version: `main` (development)
* Node.js Version: 22.x
* Database: n/a (frontend-only change)
* Browser: jsdom (unit/component)

## Checklist:

- [ ] I have made corresponding changes to the documentation — n/a (no docs reference the upload modal internals)
- [x] I have added tests that prove my feature works
- [ ] Any dependent changes have been merged and published — n/a
- [ ] I have updated the CHANGELOG.md file — intentionally skipped per CONTRIBUTING.md (`release-please` owns it)

_Can add a before/after on request — see note under testing._

## Additional Notes:

- Frontend-only; branched off and targets `main` per the branch model in CONTRIBUTING.md.
- All new `t()` calls pass inline English fallbacks, so untranslated locales degrade gracefully.
- While here, fixed a pre-existing case where an upload in which *every* file failed (no `upload_id` returned) left `isUploading` stuck and the modal spinning forever — the transfer UI now resets in that path.
- The host's existing post-upload `toast.success` still fires on partial failures (it predates this PR); the new warning toast + report make the failure clear, but tightening that success toast could be a small follow-up.

### DIFF
--- a/frontend/src/components/admin/PhotoUpload.tsx
+++ b/frontend/src/components/admin/PhotoUpload.tsx
@@ -13,7 +13,14 @@ import { useUploadProgress } from '../../hooks/useUploadProgress';
 
 interface PhotoUploadProps {
   eventId: number;
+  /** Refresh the photo grid. Called early (as bytes land) and again when
+   *  processing finishes. Never closes the modal. */
   onUploadComplete?: () => void;
+  /** Fired once the transfer stage is done, reporting whether any file
+   *  failed. The host (modal) uses this to decide whether to auto-close:
+   *  a clean upload closes as before; a partial failure keeps the modal
+   *  open so the failure report stays visible. */
+  onUploadSettled?: (result: { hasFailures: boolean }) => void;
 }
 
 const DEFAULT_MAX_FILES_PER_UPLOAD = 500;
@@ -43,7 +50,7 @@ interface UploadFailure {
   kind: UploadFailureKind;
 }
 
-export const PhotoUpload: React.FC<PhotoUploadProps> = ({ eventId, onUploadComplete }) => {
+export const PhotoUpload: React.FC<PhotoUploadProps> = ({ eventId, onUploadComplete, onUploadSettled }) => {
   const { t } = useTranslation();
   const [isUploading, setIsUploading] = useState(false);
   const [selectedFiles, setSelectedFiles] = useState<File[]>([]);
@@ -234,6 +241,8 @@ export const PhotoUpload: React.FC<PhotoUploadProps> = ({ eventId, onUploadCompl
     // Accumulates transfer-stage failures (per-file rejections + whole-chunk
     // failures) with their reasons, so the report can name each one.
     const collected: UploadFailure[] = [];
+    // Whether at least one chunk was accepted for background processing.
+    let anyQueued = false;
 
     try {
       for (let chunkIndex = 0; chunkIndex < chunks.length; chunkIndex++) {
@@ -311,6 +320,7 @@ export const PhotoUpload: React.FC<PhotoUploadProps> = ({ eventId, onUploadCompl
           // Backend returns a per-request upload_id. Track it so the
           // processing-status hook can poll/stream live progress.
           if (response.data?.upload_id) {
+            anyQueued = true;
             const newId = response.data.upload_id as string;
             setUploadIds((prev) => (prev.includes(newId) ? prev : [...prev, newId]));
           }
@@ -359,6 +369,22 @@ export const PhotoUpload: React.FC<PhotoUploadProps> = ({ eventId, onUploadCompl
       // below will refresh again on completion.
       if (onUploadComplete) {
         onUploadComplete();
+      }
+
+      // Tell the host the transfer stage settled. A clean upload lets the
+      // modal auto-close (unchanged behaviour); a partial failure keeps it
+      // open so the failure report below stays visible.
+      onUploadSettled?.({ hasFailures: collected.length > 0 });
+
+      // Nothing was queued for background processing (every chunk failed,
+      // or a pre-async backend) — there's no processing phase to wait on,
+      // so reset the transfer UI here. The failure report persists.
+      if (!anyQueued) {
+        setIsUploading(false);
+        setUploadProgress(0);
+        setCurrentChunk(0);
+        setTotalChunks(0);
+        setPhase({ kind: 'idle' });
       }
 
       // If the backend never returned an upload_id (e.g. only failures
--- a/frontend/src/components/admin/PhotoUpload.tsx
+++ b/frontend/src/components/admin/PhotoUpload.tsx
@@ -68,6 +68,11 @@ export const PhotoUpload: React.FC<PhotoUploadProps> = ({ eventId, onUploadCompl
   // validation rejections (response.errors) and whole-chunk failures.
   // Processing failures are merged in from the progress hook below.
   const [transferFailures, setTransferFailures] = useState<UploadFailure[]>([]);
+  // Processing failures are captured into state (not read live) because the
+  // completion effect clears uploadIds, which empties the progress hook's
+  // failedPhotos — reading live would make the rows vanish the instant they
+  // appear.
+  const [processingFailures, setProcessingFailures] = useState<UploadFailure[]>([]);
   const [failuresDismissed, setFailuresDismissed] = useState(false);
   const fileInputRef = useRef<HTMLInputElement>(null);
 
@@ -76,16 +81,12 @@ export const PhotoUpload: React.FC<PhotoUploadProps> = ({ eventId, onUploadCompl
   });
 
   // Single source of truth for the "which files failed" report: transfer
-  // stage failures (collected during handleUpload) plus processing
-  // failures (live, from the progress hook). Both carry filename + reason.
-  const failures = useMemo<UploadFailure[]>(() => {
-    const processing: UploadFailure[] = processingAggregate.failedPhotos.map((p) => ({
-      filename: p.filename,
-      reason: p.error || t('upload.failures.unknownReason', 'Unknown error'),
-      kind: 'processing',
-    }));
-    return [...transferFailures, ...processing];
-  }, [transferFailures, processingAggregate.failedPhotos, t]);
+  // stage failures (collected during handleUpload) plus processing failures
+  // (captured on completion). Both carry filename + reason.
+  const failures = useMemo<UploadFailure[]>(
+    () => [...transferFailures, ...processingFailures],
+    [transferFailures, processingFailures]
+  );
   
   // Fetch categories for this event
   const { data: categories = [] } = useQuery({
@@ -203,6 +204,7 @@ export const PhotoUpload: React.FC<PhotoUploadProps> = ({ eventId, onUploadCompl
     setUploadIds([]);
     // Clear any prior failure report before this run.
     setTransferFailures([]);
+    setProcessingFailures([]);
     setFailuresDismissed(false);
 
     // For large uploads, chunk the files by both count AND size to prevent memory/network issues.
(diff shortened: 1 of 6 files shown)


the-luap:
Inspected `PhotoUploadModal.tsx` + i18n changes myself; spawned a deep-audit subagent for `PhotoUpload.tsx` (+156) and both new test files; re-verified the live blocker at file:line.

Solid scope discipline overall — recovers per-file error data that was already being thrown away, splits transfer-finished from modal-close cleanly, and the de-dup analysis is correct (rejected files don't get a photos row so they can't double-list via `useUploadProgress.failedPhotos`). One blocker around the "every file rejected" path that the PR explicitly tried to fix but can't reach.

## 🔴 Blockers

### 1. Spinner still hangs forever when every file is validation-rejected
The PR description claims this case is fixed via the new `if (!anyQueued)` reset at `PhotoUpload.tsx:382`. The fix doesn't reach the actual path because the backend emits `upload_id` unconditionally:

- `backend/src/routes/adminPhotos.js:412` — `response = { upload_id: uploadId, count: 0, errors: [...], ... }` is sent regardless of `uploadedPhotos.length`
- `PhotoUpload.tsx:322-326` — `if (response.data?.upload_id)` is truthy → `anyQueued = true`
- `PhotoUpload.tsx:382` — `if (!anyQueued)` reset path SKIPPED
- `useUploadProgress.ts:150` — `isComplete = ... && totals.total > 0`, never true with zero photos queued
- `PhotoUpload.tsx:407-409` — completion effect gated on `processingAggregate.isComplete` → never fires → `setIsUploading(false)` never called

Modal hangs at "uploading…" forever; the user sees the failure report populate (good) but can't get the modal back to a clean state.

**Fix shape** (one-line, frontend-only, backwards-compatible with older API versions):
```diff
- if (response.data?.upload_id) {
+ if (response.data?.upload_id && (response.data?.count ?? 0) > 0) {
    anyQueued = true;
```
Belt-and-braces alternative: also early-reset when `processingAggregate.isReady && processingAggregate.total === 0`.

## 🟡 Concerns

### 1. Processing-only failures arrive too late to be seen
`PhotoUpload.tsx:374` — `onUploadSettled({ hasFailures: collected.length > 0 })` reports only **transfer-stage** failures. When every chunk transfers cleanly but the background worker later fails N photos, the modal already auto-closed (`PhotoUploadModal.tsx:33-37`), and the report — even though it would render correctly — is never seen. The post-processing `useEffect` (`PhotoUpload.tsx:404-430`) only fires a `toast.warning`.

Either delay `onUploadSettled` until processing settles, or fire a second `onUploadSettled` from the post-processing effect with the live `processingAggregate.failed > 0` flag.

### 2. No test coverage for the `kindTransfer` failure class
`PhotoUpload.failureReport.test.tsx` mocks the chunk POST as always-resolving. The whole-chunk-failure path (network drop, 5xx → `catch` block at `PhotoUpload.tsx:327-336` → `kind: 'transfer'`) is the most user-impacting failure in practice and has no test. One `postMock.mockRejectedValueOnce(...)` test asserting the orange "Transfer failed" badge + filename closes the gap.

### 3. `PhotoUploadModal.test.tsx` stubs `PhotoUpload` entirely
So the test only validates the modal's local `if (!hasFailures) onClose()` wiring — it can't catch contract drift between the two (e.g. renaming `onUploadSettled` to `onSettled` in one place but not the other would still pass both tests). Not a blocker, but it does less than its name suggests.

## 🟢 Nits

- `PhotoUpload.tsx:577` — failure card lacks `role="status"` / `aria-live="polite"`. Screen readers won't announce its arrival.
- Card markup is `<div>`; an `<aside>` or `<section aria-labelledby>` with the title as heading would be more semantic.
- `kindRejected: "Rejected"` in EN reads slightly terse for non-technical staff. "Validation rejected" or "Refused" reads clearer. (DE "Abgelehnt" is fine.)

## ✅ Verified

- **Backend response shape**: `errors: [{filename, error}]` confirmed at `adminPhotos.js:265, 288, 393, 414, 442`. Frontend parses defensively (`Array.isArray` check, optional chaining, fallbacks for both fields at `PhotoUpload.tsx:312-318`)
- **State reset on fresh upload**: `setTransferFailures([])` + `setFailuresDismissed(false)` fire at `handleUpload` start (`PhotoUpload.tsx:204-206`), not at component mount — correct
- **`onUploadSettled` fires exactly once per session**: only after the chunk loop ends (`PhotoUpload.tsx:374`), gated by `disabled={selectedFiles.length === 0 || isUploading}` on the submit button — no flicker risk
- **De-dup across rejected + processing**: cannot collide — rejected files never reach the photos table, so `useUploadProgress.failedPhotos` can't list them. No double-render
- **Error message rendering safety**: plain JSX text nodes, no `dangerouslySetInnerHTML`. Backend error strings render auto-escaped
- **i18n**: every visible string passes `t(key, inlineFallback, params)`; EN+DE both contain all 8 keys symmetrically
- **Title count**: `count: failures.length` (`PhotoUpload.tsx:585`) is the merged array, not a sum of buckets — correct
- **Dismiss button** has an `aria-label` via `t('common.dismiss', 'Dismiss')`
- **Modal change in isolation**: clean. Only caller is `EventDetailsPage.tsx:2445`, whose `onUploadComplete` only refreshes data (doesn't close), so the modal owns its own close — no regression for the host

## Verdict

Close blocker #1 (one-line gate change) and I'm comfortable merging. Concern #1 (processing-only failures unseen) is worth a clear answer — whether it's in scope for this PR or a follow-up issue — since it leaves the feature's promise half-fulfilled in the most common partial-failure shape on healthy networks. Concerns #2-#3 are test-coverage gaps worth filling but not blockers.

andredlng:
Thanks — excellent review, both the blocker and concern #1 were spot on. Addressed in 1be871c.

## 🔴 Blocker #1 — spinner hang
Fixed exactly as you suggested: gate `anyQueued` on `count > 0`, since the backend emits `upload_id` unconditionally.
```ts
if (response.data?.upload_id && (response.data?.count ?? 0) > 0) {
  anyQueued = true;
  ...
```
A zero-photo response now takes the `!anyQueued` terminal path → `setIsUploading(false)` + report stays. Backwards-compatible: `count` is one of the same async fields as `upload_id`, so any backend sending one sends the other.

## 🟡 Concern #1 — processing-only failures unseen
Fixed by **deferring the settle/close decision to the completion effect** rather than firing it at transfer-end. Now:
- transfer failures → modal stays open immediately (report shows them);
- clean transfer + queued work → modal stays open through processing and auto-closes **only if** `transferFailures === 0 && processingAggregate.failed === 0`;
- any processing failure → stays open with the rows.

While doing this I hit a second bug: the completion effect calls `setUploadIds([])`, which empties the progress hook's `failedPhotos` — so reading them live made the rows vanish the instant they appeared. I now **persist** them into a `processingFailures` state before clearing `uploadIds`.

**One UX tradeoff to flag for you:** the modal now stays open during background processing for *clean* uploads too (auto-closing when it finishes), instead of closing right after transfer. The 'Processing photos… you can leave this page' message is shown and the user can close manually. I think that's acceptable/clearer, but happy to revert to close-on-clean-transfer if you'd rather keep the fast close and accept processing failures only surfacing as grid tiles.

## 🟡 Concern #2 — transfer-failure test
Added: `postMock.mockRejectedValue(...)` → asserts the orange **Transfer failed** badge + filename + reason.

## 🟡 Concern #3 — modal test stubs PhotoUpload
The real-component test now passes an `onUploadSettled` spy and asserts it's called with `{ hasFailures: true }` (failure path) and `{ hasFailures: false }` (clean path). So renaming the prop on either side breaks a test now — the modal test covers its wiring, the PhotoUpload test covers the contract.

## 🟢 Nits
- Report card now has `role="status"` + `aria-live="polite"`.
- Left `<div>` (vs `<aside>`) and the "Rejected" wording as-is for now — easy to change if you feel strongly.

Test count on the feature is now 6 (rejected+processing, transfer, clean-settle, dismiss, + 2 modal). `tsc`/`lint` clean. Re-verified logic under unit tests; happy to re-run the live Docker pass on the blocker path if useful.

the-luap (approved):
Re-verified the fix commit `1be871c` at file:line. Blocker + all three concerns closed; ran both suites locally (6/6 pass).

## Blocker — closed
`PhotoUpload.tsx:327` — `anyQueued` now gated on `(response.data?.count ?? 0) > 0`. A zero-photo response (every file rejected) leaves `anyQueued = false` → hits the `!anyQueued` terminal reset at `:386` → `setIsUploading(false)`, report persists. No more hang. Backwards-compatible.

## Concern #1 — closed (+ a real bonus catch)
Settle/close decision moved into the completion effect (`:436`), so `hasFailures` now folds in `processingAggregate.failed`, not just transfer failures. Good catch on the secondary bug: the effect's `setUploadIds([])` empties the hook's live `failedPhotos`, so you snapshot them into `processingFailures` state (`:71`) before clearing — rows no longer vanish the instant they appear. Verified the two settle paths (`!anyQueued` vs. the effect) are mutually exclusive via the `uploadIds.length === 0` guard — no double-settle.

## Concerns #2 / #3 — closed
Transfer-failure test asserts the "Transfer failed" badge + filename + reason; the real-component modal test now spies `onUploadSettled` with `{hasFailures:true/false}`, so prop-rename drift breaks a test.

## Nit — done
Report card has `role="status"` + `aria-live="polite"`.

## UX decision — accepted
The modal now stays open through background processing for clean uploads too, auto-closing when the worker finishes. That's the clearer behavior (processing failures land in the report, not just as grid tiles) — keeping it.

Merge cleared.

Issue #705, opened 2026-06-30
[FEATURE] Simplify deployment and add first-class NAS installs for Synology, UGREEN, QNAP


the-luap:
## Summary

PicPeak is already strong as a self-hosted client-gallery product, but the current deployment story still feels too much like a developer-oriented Docker app. For adoption by photographers and small studios, especially those with Synology/UGREEN/QNAP NAS devices, we should make installation feel closer to a NAS app / managed self-hosted product.

Goal:

> Make PicPeak feel like a 10–15 minute install on a VPS or NAS, with generated secrets, sane defaults, HTTPS options, first-run guidance, and platform-specific NAS paths/guides.

This issue collects findings from reviewing the current repository deployment files and comparing PicPeak against SaaS gallery products and self-hosted photo apps.

---

## Why this matters

Photographers compare PicPeak against SaaS tools like PicDrop, Pixieset, Pic-Time, ShootProof, and SmugMug. Those products win on onboarding:

1. Create account
2. Create gallery
3. Upload images
4. Choose design/access settings
5. Send client link

No Docker, no env vars, no reverse proxy, no SMTP, no database, no ports.

PicPeak's differentiator is different and valuable:

- own server / own NAS
- own domain
- no platform lock-in
- client galleries, proofing, expiry, downloads, branding, webhooks
- potential direct access to existing NAS photo folders

But for this to convert, deployment needs to feel much less technical.

---

### 1. Quickstart uses source-build path

`README.md` currently says:

```bash
git clone https://github.com/PicPeak/picpeak.git
cd picpeak
cp .env.example .env
nano .env
docker compose up -d
```

But the default `docker-compose.yml` builds backend/frontend locally from source. For self-hosters this should be a developer setup, not the main user quickstart.

Recommendation:

- Make the user quickstart use prebuilt GHCR images.
- Rename/split:
  - `compose.yml` or `docker-compose.yml`: production, prebuilt images
  - `compose.dev.yml`: source build/dev
  - `compose.caddy.yml`: optional HTTPS

---

### 2. `.env.example` is too large for first install

The current env example asks users to think about many things immediately:

- `JWT_SECRET`
- `DB_PASSWORD`
- `REDIS_PASSWORD`
- `ADMIN_PASSWORD`
- `FRONTEND_URL`
- `ADMIN_URL`
- `API_URL`
- SMTP
- storage paths
- branding
- analytics
- S3 options
- ports

For a first install, ideally this should be closer to:

```env
[...]

PR #714, opened 2026-07-01
feat: zero-config first run — in-browser admin bootstrap + auto-generated secrets

Luca-Timo:
## Summary

Makes a fresh PicPeak install need **nothing in `.env`**. Today a new user has to hand-edit `.env` (JWT/DB/Redis secrets + admin password) before the app will even start — a big adoption barrier for non-developers (see #705). This PR removes that entirely:

```
docker compose up -d
  → open http://<host>/admin
  → paste the one-time setup token (from the logs / data/SETUP_TOKEN)
  → set your admin email + password
  → done
```

Everything is **backward-compatible**: any secret you *do* set in `.env` is still used exactly as before.

### 1. In-browser first-run admin (no `ADMIN_PASSWORD`)
- The app boots with **no admin account**; the first visit to `/admin` redirects to a new **`/setup`** screen that creates the initial `super_admin` in the browser.
- Guarded by a **one-time setup token**, generated at first boot while no admin exists, printed to the logs **and** written to `data/SETUP_TOKEN`, and **burned on first use** (endpoint returns `409` afterwards). The token is **always required** — no "trusted LAN" bypass, since remote/VPS installs are public-IP anyway and `X-Forwarded-For` is spoofable behind a proxy. Constant-time comparison, rate-limited.
- Migration `001_init`'s legacy default-admin seed is now **gated behind `ADMIN_PASSWORD`** — set it and you keep the old auto-created admin (+ `ADMIN_CREDENTIALS.txt`); leave it unset and you get the wizard.

### 2. Auto-generated machine secrets (JWT/DB/Redis)
- A tiny **`secrets-init`** service generates `jwt_secret`/`db_password`/`redis_password` onto a shared `picpeak-secrets` volume before Postgres/Redis start. Each is **seeded from the matching env var when set** (backward-compatible), otherwise a strong random value.
- Postgres reads `POSTGRES_PASSWORD_FILE`, Redis reads `--requirepass` from the file, and the backend entrypoint (`wait-for-db.sh`) sources the files into env **only when unset** — explicit `.env` values always win.
- **Idempotent, never-regenerate**: the init never overwrites an existing secret file, so the DB password can't drift out from under an already-initialised Postgres volume.

## Backward compatibility
- Set `ADMIN_PASSWORD` → legacy auto-admin, unchanged.
- Set `JWT_SECRET` / `DB_PASSWORD` / `REDIS_PASSWORD` → used as-is (seeded into the secret files).
- Existing installs already ran migration `001` and already have their secrets in `.env`, so nothing changes for them.

## Design decisions worth a look
- **Editing `001_init` in place** (gating the seed) rather than a compensation migration — only affects *fresh* installs, and only those without `ADMIN_PASSWORD`.
- **`$$`-escaping in the `secrets-init` entrypoint** — `$VAR` in a compose command string is interpolated by Compose itself; escaping restores the shell reading the container env (and silences the "variable not set" warnings).
- **`picpeak-secrets` volume** must be backed up alongside `postgres-data` — deleting it orphans the DB password. Documented in the compose files.

## Testing
- **13 backend tests** (`setupService.test.js`): status, one-time token lifecycle, gates (wrong token / weak password / already-set → 400/409), `super_admin` role, token burn, file lifecycle, **a Postgres-safe `jsonb` guard** (the token is stored JSON-encoded — a SQLite-only test missed this originally), and **route-level supertest** coverage (`GET /status`, `POST /admin` 400/201+cookie/409).
- **Validated end-to-end on Docker + Postgres with an empty `.env`**: secrets auto-generated, DB/Redis connect, wizard + admin creation (JWT signed with the generated secret) work, no compose warnings.

## Docs
- README Quick Start: "First run — create your admin account".
- `.env.example`: `ADMIN_PASSWORD` / `JWT_SECRET` / `DB_PASSWORD` / `REDIS_PASSWORD` now **OPTIONAL** (commented out) with notes.
- `SIMPLE_SETUP.md`: "First Login" section.

## Translations
i18n added in **en** and **de** (both native-reviewed).


the-luap:
Deep-reviewed `415bffa`. Inspected the security core myself — `setupService.js`, `routes/setup.js`, the `001_init` gate, `server.js` boot ordering, and the `secrets-init` / `wait-for-db.sh` chain in both compose files. Spawned a subagent for the frontend/tests/i18n/docs and spot-verified at file:line. Strong PR — the token model, boot ordering, and never-regenerate idempotency are all correct.

## 🔴 Blockers

### 1. Client password rule is weaker than the server → confusing bounce
`frontend/src/pages/SetupPage.tsx:53` gates on `password.length < 8` only, but the server runs `validatePassword()` (`setupService.js:98`) which requires upper + lower + digit (+ zxcvbn floor). A password like `password` passes the browser, the server then rejects it, and the raw English `{ error }` (`routes/setup.js:48`) is rendered verbatim (`SetupPage.tsx:92`) with no i18n. The security boundary is fine — no weak password can actually be set — but the UX is broken. Mirror the real rule client-side and map the server failure to a translated key.

### 2. `picpeak-setup.sh` isn't updated — the scripted install dead-ends at the token prompt
The migration seed is now gated on `ADMIN_PASSWORD`, which the setup script never writes (it only writes `ADMIN_EMAIL`, script:460). So script installs get no seeded admin — correct, they should hit the wizard — but the script's post-install block still copies and prints `data/ADMIN_CREDENTIALS.txt` (script:570-583), which no longer exists. Result: the operator sees "Login URL: …/admin" with no password and no mention of the setup token, lands on `/setup`, and is asked for a token they were never given. Fix: replace that block with the setup token (from `data/SETUP_TOKEN` / `docker compose logs backend`) and point at `/setup`. This should land with this PR or an immediately-following companion PR so the scripted simple-setup path doesn't ship broken.

## 🟡 Concerns

### 1. Double-submit race on the one-time gate
`createInitialAdmin` does `noAdminExists()` (count) → insert, non-transactionally, burning the token after insert (`setupService.js:86-122`). Two concurrent valid-token submits with different emails could both create `super_admin` rows. The `admin_users.username`/`email` unique constraints (`db.js:360-361`) catch the realistic same-email double-click, and the true race needs the operator's token — so low severity, but there's no concurrency test. Either add a DB guard (`INSERT … WHERE NOT EXISTS` / advisory lock) or note the race is accepted.

## 🟢 Nits
- Transient `/setup/status` 500 briefly renders the create-admin form on a configured instance (`SetupPage.tsx:37-40` only redirects on `status && !status.needsAdmin`); self-corrects via the 409 on submit.
- `secrets-init` writes the secret files `chmod 644` — world-readable inside the containers sharing the volume; `640` + a shared gid is tighter (volume isn't host-exposed, so minor).
- `docker-compose.yml` still bakes `DATABASE_URL=…:${DB_PASSWORD:-}@…` with an empty password on the zero-config path. Harmless — nothing reads `DATABASE_URL` (knex uses `process.env.DB_PASSWORD`, `knexfile.js:103`, sourced by `wait-for-db.sh`) — but drop it to avoid confusing a debugger.

## ✅ Verified
- Setup token: generated at boot only while no admin exists, jsonb-encoded, logged + `data/SETUP_TOKEN` (0600), burned on first use, constant-time compare, 409 after completion.
- Migration gate: legacy seed only when `ADMIN_PASSWORD` set; fresh installs → empty `admin_users` → wizard. Existing installs already migrated → unaffected.
- Boot ordering: `ensureSetupToken()` runs after `tryInstallFromBackup` (`server.js:1010` vs `997`) — restored instances never print a stale token.
- Rate limiting on `/api/setup/admin` (`server.js:401`).
- `secrets-init`: `$$` escaping correct, idempotent never-overwrite, `service_completed_successfully` gating, `POSTGRES_PASSWORD_FILE`, redis reads from file. `wait-for-db.sh`: env-wins / file-fallback before the su-exec drop. Backward-compat for explicit `.env` secrets + `ADMIN_PASSWORD`.
- Frontend auth: `{ user }` only, HttpOnly cookie, no token in body/storage; routing guards have no dead-end; i18n 22 `setup.*` keys symmetric en/de; docs accurate, stale `nano .env` removed.

## Note — UI PRs need a screenshot
Going forward I'm updating CONTRIBUTING so that **any PR touching the UI must attach a screenshot of the change** (before/after where it helps). This PR adds the `/setup` wizard + the `/admin`→`/setup` redirect — please add a screenshot of the setup screen so the visual side is reviewable. This applies to every UI PR from here on, ours included.

## Verdict
Close blockers 1 and 2 (client password rule + i18n; script token messaging), give me a call on the race, and I'm comfortable merging. Architecture is right — the wizard becoming the single admin-creation path across both compose and scripted installs is exactly the shape we want.

Luca-Timo:
@the-luap Thanks for the deep review — all points addressed in `286975dc` (on top of the reviewed `415bffa0`).

## 🔴 Blockers

**1. Client password rule + i18n** — `SetupPage` now mirrors the server rule (≥8 with upper/lower/digit) before submit, so a valid client never gets bounced. The server error carries a `field` (`routes/setup.js`) that the client maps to a translated key instead of rendering the raw English — new `setup.invalidToken` / `setup.passwordRequirements` (en + de). Live client-side message in the screenshots below.

**2. `picpeak-setup.sh`** — the post-install block no longer dead-ends. On the wizard path (no `ADMIN_PASSWORD`) it fetches the one-time token from `data/SETUP_TOKEN` (falling back to `docker compose logs backend`) and prints a step-by-step "open `/admin` → paste token → set email + password". The legacy credentials print is kept for when `ADMIN_PASSWORD` is set.

## 🟡 Concern — double-submit race

`createInitialAdmin` now creates the admin **and burns the token in a single transaction**, atomically claiming the token (`UPDATE … WHERE setting_value IS NOT NULL`, expecting exactly one row) so two concurrent valid-token submits can't both create a `super_admin`. Cross-DB (no jsonb comparison; `trx`-only writes to avoid the SQLite in-transaction deadlock). Added a concurrency test (`Promise.allSettled` of two submits → exactly one admin).

## 🟢 Nits
- **Status-500 form flash** — `SetupPage` redirects to `/login` when `/setup/status` errors instead of rendering the create-admin form.
- **`chmod 644`** — kept, with a comment on why: the readers run as three different users (postgres, redis, nodejs), so `640` would break Redis's non-root `cat`; the volume is private to these containers and never host-exposed.
- **`DATABASE_URL`** — dropped from `docker-compose.yml`.

## Verified
14/14 backend tests (incl. the new concurrency guard), clean frontend build, `bash -n` on the script, both compose files valid with **0 "variable is not set" warnings**, i18n symmetric (24 `setup.*` keys en/de). Re-ran the full zero-config flow end-to-end on Docker + Postgres with an empty `.env` — secrets generated, wizard + admin creation work.

## Screenshots
Screenshot 2026-07-02 at 13 19 33

Screenshot 2026-07-02 at 13 20 01

the-luap:
Re-verified `286975dc` at file:line and re-ran the suite — **14/14 backend tests pass**, including the new double-submit guard. All blockers + the race closed, nits addressed, screenshots attached. From my side: **merge cleared.**

## Blockers — closed
1. **Password UX + i18n** — client mirrors the server rule (`SetupPage.tsx:56`); the server returns a `field` (`routes/setup.js:48`) mapped to `setup.invalidToken` / `setup.passwordRequirements` (both present in en **and** de). Note: the server also enforces a zxcvbn floor the client regex can't fully mirror, so a complex-but-guessable password can still be bounced — but now with the translated message, not raw English. Fine as-is.
2. **Script token** — `picpeak-setup.sh` now surfaces `data/SETUP_TOKEN` (falling back to the logs) with browser steps.

## Race — closed
Atomic token claim inside a transaction (`setupService.js` — `whereNotNull('setting_value').update({…null})` expecting exactly one row, then a re-check, all on `trx`, no jsonb comparison → cross-DB safe). New concurrency test verified passing.

## Nits — addressed
Status-500 → login redirect; `chmod 644` kept with a valid justification (redis's non-root `cat`); `DATABASE_URL` dropped.

## Two UI items from the screenshots (not blocking, but this is the first screen a new admin ever sees)

**1. Use the PicPeak logo, not the star icon.** `SetupPage.tsx:120` renders a lucide `<Sparkles>` as the header mark. Swap it for the PicPeak brand logo (the same asset the login page uses) — the first-run screen should be on-brand, since branding is exactly what the user sees first.

**2. Give the setup-token step room to breathe.** The "Can't find it? Run: `docker compose logs backend | grep -i "setup token"`" hint (`setup.tokenLocationHint`) is easy to miss under a screen that also holds email + password. Two options, in order of preference:
- **Preferred — a two-step wizard:** step 1 asks for **only the setup token** (full width, with the `docker compose logs …` command shown prominently, plus a link to the docs for when the logs have already rotated away), then step 2 collects email + password. Cleaner, and gives the token-recovery guidance the space it needs.
- **Minimum:** move the "Can't find it?" line to sit **directly below the token input box**, adjacent to the field it explains.

Please attach a screenshot of the reworked screen (per the new CONTRIBUTING rule).

## Verdict
Merge cleared. Pick up the two UI items in this PR or an immediate follow-up.

Luca-Timo:
I am working on the UI changes now

Luca-Timo:
@the-luap Picked up both UI items from the review, plus tightened step-1 validation.

1. On-brand logo. Swapped the lucide Sparkles mark for the bundled PicPeak logo (the same /picpeak-logo-transparent.png the login page falls back to), on the cream brand plate. It renders through the shared resolveLoginLogoClasses helper at the login default (medium, 200×150), so /setup and /admin/login show the logo at identical size.

2. Room for the setup token — two-step wizard.

Step 1 — token only. Full-width token field with the docker compose logs backend | grep -i "setup token" recovery command shown prominently right under it (with a copy button), plus a "Logs already rotated away? Read the setup guide" link to the docs for when the logs are gone.
Step 2 — email + password. A rejected token bounces the user back to step 1 so the error lands on the field that owns it.
3. Step-1 token is now validated server-side (new). Previously "Continue" only checked the field was non-empty, so a wrong token slid through and only failed after the user had filled in email + password. Added a non-burning check:

Backend: POST /setup/verify-token constant-time compares the token without consuming it (createInitialAdmin still claims it atomically on final submit). Gated on "no admin exists" and rate-limited exactly like /setup/admin, so it can't be used to brute-force — and the token is 24 random bytes regardless.
Frontend: "Continue" verifies before advancing; a wrong token shows the invalidToken error on the field (429 → too-many-attempts, 409 → redirect to login).
3 new integration tests (accept-without-burn / reject-wrong / closed-once-admin-exists); full suite (17) passes.
en + de strings added; other locales fall back to en.

Screenshots of the reworked screen (per the CONTRIBUTING screenshot rule):

Step 1 — setup token (with server-side validation)
Screenshot 2026-07-02 at 15 04 59 Screenshot 2026-07-02 at 15 22 46

Step 2 — create admin account
Screenshot 2026-07-02 at 15 05 07

Verified end-to-end on a locally-built Docker image: wrong token (last char dropped) → rejected at Continue; correct token → advances and is not consumed by the check; vite build + backend Jest suite both green.

the-luap:
that looks better now and many thanks for implementation and testing! I will approve it now

the-luap (approved):
all approved

PR #718, opened 2026-07-02
feat: setup wizard + argument-driven unattended install

the-luap:
> **Draft — stacked on #714.** The browser-first admin + zero-secret behavior needs #714's setup-token backend and `secrets-init`. This PR only touches `scripts/picpeak-setup.sh` (a clean diff on `main`), but should merge **after** #714. Needs a real-host smoke test before un-drafting (can't run a full Docker install in CI).

## What

Restructures `picpeak-setup.sh` around **two clear modes**, addressing #705's "simple setup" goals and the script gap I flagged on #714.

### 1. Interactive wizard (default)
`run_wizard()` walks step-by-step, then shows a **review + confirm** before touching anything:

```
method (docker/native) → install dir → release channel → domain →
HTTPS handling → admin email → SMTP (optional) → review → install
```

Each value already passed as a flag is respected and its question skipped.

### 2. Unattended (`--unattended` + flags)
`validate_unattended()` fills defaults and **fails fast** on impossible combos (e.g. `--enable-ssl` without `--domain`). New flags: `--admin-password`, `--install-dir`, `--channel`.

```bash
# browser-first admin, headless stack
sudo ./picpeak-setup.sh --docker --unattended --email admin@example.com
# fully headless incl. seeded admin
sudo ./picpeak-setup.sh --docker --unattended --admin-password 'S0me-Str0ng-Pass'
```

## Alignment fixes (came out of the #714 review)

- **Uses the committed `docker-compose.production.yml`** (prebuilt GHCR images) via `COMPOSE_FILE` in `.env`, instead of hand-generating a now-divergent compose file (removed ~100 lines). Fixes the #705 §4 divergence and gives script users #714's zero-config secrets.
- **Browser-first admin**: no password written by default; the one-time `/setup` token is surfaced (from `data/SETUP_TOKEN` or the logs) with clear browser instructions. `--admin-password` keeps the legacy seeded-admin flow for headless runs. This closes the "scripted install dead-ends at the token prompt" blocker from #714.
- **Removed the broken `setup_ssl_docker` call** (referenced but never defined → silent no-op on `--enable-ssl --docker`). Docker HTTPS is now honest: assumes a fronting reverse proxy; native still gets automatic Caddy HTTPS.
- Update path `pull`s images instead of building.

## HTTPS handling
- **Native + domain**: automatic HTTPS via Caddy (unchanged, works).
- **Docker + domain**: assumes your own reverse proxy (Traefik/nginx/Cloudflare); a Caddy-sidecar profile for built-in Docker TLS is a planned follow-up.

## Testing
- `bash -n` clean; heredoc rendering spot-checked for both admin paths.
- ⚠️ **Not yet run end-to-end on a real host** — needs a Docker + a native smoke test (and #714 merged) before this leaves draft.

### DIFF
--- a/scripts/picpeak-setup.sh
+++ b/scripts/picpeak-setup.sh
@@ -44,8 +44,13 @@ INSTALL_METHOD=""  # docker or native
 OS_TYPE=""
 OS_VERSION=""
 PACKAGE_MANAGER=""
-ADMIN_EMAIL="admin@example.com"
+ADMIN_EMAIL=""              # empty until set by the wizard, --email, or the unattended default
+ADMIN_PASSWORD=""          # optional. When set, the admin is seeded directly (headless).
+                           # Empty = browser-first: the app prints a one-time /setup token.
 DOMAIN_NAME=""
+INSTALL_DIR=""             # override for the install directory (else per-method default)
+PICPEAK_CHANNEL="stable"   # GHCR image tag for the production compose (stable|beta)
+HTTPS_MODE=""              # caddy | proxy | none — how TLS is terminated
 SMTP_HOST=""
 SMTP_PORT=""
 SMTP_USER=""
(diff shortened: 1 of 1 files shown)


the-luap:
## Rebased onto merged #714 — dependency resolved

#714 is now on `main`, so this PR's runtime dependency (setup-token backend + `secrets-init` in the production compose) is satisfied. Rebased; the only file touched is `scripts/picpeak-setup.sh`.

### How this reconciles with what #714 already shipped
- #714 added setup-token surfacing to the script **for the Docker path only**, and kept the hand-generated compose. This PR's restructure **supersedes** that block with a unified version that:
  - surfaces the **one-time `/setup` token at the end of install for BOTH Docker and native** (reads `data/SETUP_TOKEN`, waits for the backend to write it, falls back to `docker compose logs backend | grep -i 'setup token'` / `journalctl`), matching the backend's `…One-time setup token: <token>` output — so the operator pastes it straight into `/admin` → `/setup`;
  - switches Docker installs to the committed **`docker-compose.production.yml`** (now backed by #714's `secrets-init` → zero-config secrets), dropping the divergent generated compose;
  - adds the full **step-by-step wizard** and **argument-driven unattended** mode (`--admin-password`, `--install-dir`, `--channel` + validation);
  - removes the broken `setup_ssl_docker` call.

### Still needed before un-drafting
- **Real-host smoke test** (Docker + native), which I can't run in CI. `bash -n` is clean; the token/`.env` heredocs were dry-rendered. shellcheck would be worth a pass on a host.

PR #731, opened 2026-07-02
feat: first-run setup wizard (feature selection + config) and portable .picpeak backup roundtrip

Luca-Timo:
## Summary

Two interlocking additions that share the first-run / reinstall flow:

1. **First-run setup wizard** — extends the in-browser bootstrap so a new admin can, right after creating their account: pick which features they'll use, get prompted for just the config those features need, **or** restore a full backup instead of setting up fresh.
2. **Portable `.picpeak` backup** — a single self-contained file you download from one instance and upload on another, entirely through the web UI, to clone/migrate an instance.

Everything here is **additive** — new services, endpoints, a component, and wizard steps. No existing backup/restore path or setup behaviour is changed.

## 1 · Setup wizard

After the token → account steps:

- **"How will you use PicPeak?"** — grouped feature ticks (Client management / Accounting / Automation). Saved via the existing `PUT /admin/feature-flags`, whose server-side `applyDependencyRules` resolves dependencies (e.g. Invoices → Accounting), so the wizard only sends raw ticks. `galleries`/`analytics`/`userManagement` stay always-on. Labels reuse `settings.features.<key>.title/description`.
- **Per-feature config step** — shown only when the picked features need it, collecting the hard-required bits and leaving everything else on seeded defaults:
  - *Invoices* → company/legal name, address, **VAT-ID or tax number**, IBAN, currency (→ `business_profile` + a default bank account), with a bank/VAT legal disclaimer.
  - *Email features* → SMTP host/port/user/pass/from (→ `email_configs`).
  - Each section saves only if started; **"Skip for now"** is always available.
- **Restore branch** — "Migrating from another PicPeak?" uploads a `.picpeak` to clone another instance onto this fresh one, preserving the account just created.
- Pinned the first-run screen to its intended **light** look (it previously inherited `var(--color-background)` and went dark under `.dark`).

### Screenshots

**"How will you use PicPeak?" (feature selection + restore entry)**

Screenshot 2026-07-02 at 22 22 21

**Per-feature config step (Invoices + Email)**

Screenshot 2026-07-02 at 22 22 57

**Restore-from-backup branch**

Screenshot 2026-07-02 at 22 22 38

## 2 · Portable `.picpeak` backup

- **Format** — a zip (`.picpeak`) with `manifest.json` + engine-neutral per-table NDJSON + `files/` (invoice/contract PDFs & business-docs; original photos optional via a toggle). Tables are introspected at runtime, so it never rots as tables are added. Engine-neutral because the target rebuilds its own schema via migrations, then loads rows — enabling **pg↔pg / sqlite↔sqlite** and **forward-only** auto-migrate (a newer-schema backup is refused).
- **Restore = full override, except the current account** — all data is replaced by the backup, but the logged-in admin is preserved (an email-colliding backup admin is overwritten with the current creds), so the operator is never locked out. Behind an explicit destructive-action confirmation.
- **Endpoints** — `GET /admin/backup/picpeak/export` (download, `backup.create`, sets `X-Picpeak-Contains-Secrets`) and `POST /admin/backup/picpeak/import` (multipart upload, `backup.restore`).
- **UI** — download lives on **Backup → Dashboard** (a "Portable backup" card with a plaintext-secrets warning + include-photos toggle); restore lives under **Backup → Restore → Upload Backup**, split into **`.picpeak backup`** (working) and **Manifest + files** (legacy, still "coming soon"). Kept separate from the multi-step RestoreWizard's scheduled-backup flow.

### Screenshots

**Restore → Upload Backup, split into `.picpeak backup` vs `Manifest + files`**

Screenshot 2026-07-02 at 21 52 52

**Destructive-action confirmation**

Screenshot 2026-07-02 at 21 52 37

**Successful restore**

Screenshot 2026-07-02 at 21 52 52

## Safety / compatibility

- Restore is **same-engine only** (surfaced in the UI) and **forward-only** (refuses a newer-schema backup).
- The `.picpeak` contains **plaintext secrets** (SMTP password, admin hashes, API keys) — the download UI warns, and the manifest carries a `contains_secrets` flag.
- FK handling on the destructive import: single transaction with FK enforcement suspended (pg `session_replication_role=replica`, reset before commit; sqlite `defer_foreign_keys`). `knex_migrations` is never touched, so the target's schema/migration state is preserved.

## Testing

- 22 backend integration tests: export archive shape + row-count consistency; roundtrip (data restored, **current account survives a full override**, email-collision keeps operator creds).
- **Verified end-to-end on live Docker Postgres**, which caught two pg-only bugs SQLite couldn't and are fixed here: knex `.stream()` needing the unbundled `pg-query-stream` (export), and json/jsonb columns needing re-serialization on insert (import).

## i18n

`en` + `de` added throughout. Other locales fall back to `en` and would need native translation.

## Notes for reviewers

- No new dependencies.
- Two follow-ups intentionally left out of scope: scheduled/History backups optionally emitting a `.picpeak` per run, and a couple of soft config sections (branding/site basics) that stay in Settings for now.


the-luap:
Reviewed the `.picpeak` import service, routes, and node-stream-zip version directly; spawned a subagent for the setup-step gating, export cleanup, frontend, and tests, and spot-checked the security-critical claims. Solid design overall — transactional full-replace, operator-preserve, forward-only, engine-match, secret warnings.

## 🔴 Blockers

None. The main worry — that the setup config/restore step runs pre-auth — does not hold (see Verified).

## 🟡 Concerns

### 1. Import trusts `manifest.tables` — no `knex_migrations` guard
`picpeakImportService.js:226` — the restore derives its table list straight from the uploaded manifest (`Object.keys(manifest.tables)`) and runs `trx(table).del()` on each. The **export** excludes `knex_migrations` / `knex_migrations_lock` (`picpeakExportService.js:32` `EXCLUDED_TABLES`), but the **import** never re-applies that filter. A crafted or corrupted `.picpeak` that lists `knex_migrations` (or any existing table) would wipe it inside the transaction → schema-version state corrupted on next boot. Admin-gated so not remotely exploitable, but the fix is cheap and also protects against bad backups: apply the same `EXCLUDED_TABLES` filter on import, and ideally validate `tables ⊆` the actual DB tables before touching anything.

### 2. `SET session_replication_role = 'replica'` requires a Postgres superuser
`picpeakImportService.js:132` — fine on the bundled postgres (its `picpeak` role is a superuser), but on **managed/external Postgres** (RDS, Cloud SQL, etc.) the app user usually isn't a superuser, so this `SET` fails and the whole restore transaction aborts with a cryptic permission error. Please either catch it and surface a clear message, document the limitation, or use a superuser-free FK-deferral. Worth an explicit answer since it silently breaks restore for a chunk of deployments.

### 3. Test gaps on the core invariants
`picpeakRoundtrip.test.js` covers roundtrip data + current-admin preservation + email-collision overwrite, but the validation contracts that are actually implemented — **engine-mismatch rejection** and **forward-only newer-refused** (`:52/:56/:68`) — plus **files-restored** (`filesRestored` returned but never asserted) are untested. Add three cases: a mismatched `manifest.database.engine`, a `latest_migration` newer than target, and a file in `files/` asserted on disk after import.

## 🟢 Nits
- `picpeakExportService.js:171-190` — on an `archiver` error the out temp dir (a partial plaintext-secret archive) is orphaned; only the NDJSON scratch is cleaned. Low severity (`os.tmpdir()`, not web-served) — track and `rm` the out dir on reject too.
- Download temp cleanup is best-effort (`adminBackup.js:151`): a crash mid-download leaves a secret-bearing `.picpeak` in `os.tmpdir()`. A startup sweep of `picpeak-*` would close it.

## ✅ Verified
- **No pre-auth privileged action.** Restore/config steps are only reachable from the `usage` step, set exclusively *after* `createInitialAdmin` + `login()` (step machine `token → account → usage → {restore|config}`, `SetupPage.tsx:168`). And the API boundary holds regardless: `POST /picpeak/import` is `adminAuth` + `requirePermission('backup.restore')` (`adminBackup.js:177`), export `backup.create` — an unauthenticated `/setup` visitor can't trigger a restore.
- **Zip-slip guarded** by `node-stream-zip@1.15.0` (`throw 'Malicious entry'`), so `extract()` rejects `../` entries.
- **Temp cleanup**: import upload always unlinked (`finally`, `:194`); export scratch always removed (`:198`); both under `os.tmpdir()`, never web-served.
- **Secrets flagged**: `contains_secrets:true` in the manifest + `X-Picpeak-Contains-Secrets` header + amber UI warning; destructive-restore **confirmation modal** present before the wipe.
- **Operator-preserve** prevents lockout (parameterized `whereRaw`), knex identifier quoting blocks table-name SQLi, no XSS from manifest fields (only numbers/bools + React-escaped filename rendered), i18n symmetric (69 keys, en == de).

## Verdict
No blockers. Two I'd like closed before merge: the **`knex_migrations` import filter** (small, real safety win) and a **decision on the superuser / managed-PG limitation** (fix or document). Test gaps + temp-cleanup are follow-up polish.

Luca-Timo:
@the-luap Thanks for the thorough review — all addressed in fa7665c5.

🟡 1 · manifest.tables trust / knex_migrations guard — fixed. Import no longer derives its table list straight from the manifest. It now intersects the manifest's tables with the real data tables of this DB via listDataTables() (which already excludes knex_migrations/_lock), and drops anything else; skipped tables are logged. So a crafted or corrupted .picpeak listing knex_migrations — or any table not present here — can't be touched.

🟡 2 · session_replication_role superuser — fixed + surfaced. The SET is now wrapped in a try/catch. On a non-superuser (managed PG), it fails before any del() — the transaction rolls back, nothing is wiped — and the operator gets a clear 400: "Restore needs a PostgreSQL superuser to suspend foreign-key checks during the full replace… restore onto the bundled Postgres, or grant the role superuser." Decision: document the limitation + fail clearly rather than a superuser-free deferral (deferring non-deferrable FKs isn't possible without SET CONSTRAINTS-compatible constraints, and a topological reload is a much bigger change). Happy to revisit if you'd prefer the ordered-reload approach.

🟡 3 · Test gaps — added (+4, suite now 26): engine-mismatch rejection, forward-only newer-refused (via a seeded knex_migrations), a files/ entry asserted on disk after import with filesRestored, plus a not-a-picpeak rejection.

🟢 Nit (export orphan) — fixed: on an archiver error the temp out-dir (partial plaintext-secret archive) is now removed instead of orphaned.

🟢 Nit (download temp startup sweep) — acknowledged, leaving as the noted follow-up; the download temp is unlinked in the normal path, so this only bites on a hard crash mid-download.

PR #752, opened 2026-07-04
fix(invoices): correct payment-check email template key so dunning email sends

### LINKED ISSUE #751: [BUG] Dunning payment-check admin email never sends — wrong template key (invoice_payment_check_admin)
## Summary

The built-in **Invoice dunning** workflow's admin "payment check" email is **never sent** — it fails with `Email template 'invoice_payment_check_admin' not found`, retries to the cap, and sits `pending` in the queue.

## Root cause

`invoiceService.queuePaymentCheckEmail()` queues the email with template key **`invoice_payment_check_admin`** (`invoiceService.js:3059`), but no template with that key exists. The actual template — the admin "Paid in full / Partially paid / Not paid yet" email with the `paid_url`/`partial_url`/`unpaid_url` action buttons — is keyed **`invoice_payment_check`** (`crmEmailTemplates.js:217`, seeded by migration 116). The email processor resolves templates by an exact `template_key` lookup (`emailProcessor.js:715`) and throws "not found".

## Impact

Every payment-check admin email in the dunning flow fails, on any install that enables the invoice-dunning built-in. Because the dunning built-in is new and ships disabled, it's gone unnoticed. Net effect: enabling dunning produces **no** reminders — the admin never gets the "did they pay?" nudge, so the customer reminder (which is only queued after the admin clicks "Not paid") is never triggered either.

## Steps to reproduce

1. Enable the built-in **Invoice dunning** workflow.
2. Let an invoice reach the payment-check step (or call `invoiceService.queuePaymentCheckEmail(id)`).
3. Inspect `email_queue`: the `invoice_payment_check_admin` row is `status=pending`, `retry_count=3`, `error_message = "Email template 'invoice_payment_check_admin' not found"`.

## Fix

Queue the correct key: `invoice_payment_check_admin` → `invoice_payment_check` (`invoiceService.js:3059`). One-word change — PR attached.

## Note (separate, cosmetic — not fixed here)

The dev-tools test-send list still references `invoice_payment_check_admin` (`adminDev.js` `TEMPLATES_KEYS` + the CRM Development page). That surface already greys out absent templates (`/email-templates` returns a `present` flag), so it's a naming inconsistency, not a functional break — left as a follow-up.

## Environment

- Version: 3.79.1-beta.0
- DB: PostgreSQL (bundled)

Luca-Timo:
Fixes #751.

`queuePaymentCheckEmail` queued the admin payment-check email with template key **`invoice_payment_check_admin`** (`invoiceService.js:3059`), but no such template exists — the only one is **`invoice_payment_check`** (`crmEmailTemplates.js:217`, seeded by migration 116), which IS the admin "Paid / Partial / Not paid" email. The processor does an exact `template_key` lookup (`emailProcessor.js:715`) and throws `Email template '…' not found`, so **every** dunning admin payment-check email fails, retries to the cap, and gets stuck `pending`.

**One-word fix:** queue `invoice_payment_check`. Unbreaks the built-in invoice-dunning flow's email step. Verified in the field: the queued row's `error_message` was exactly `Email template 'invoice_payment_check_admin' not found`.

Left as a follow-up (cosmetic, not a break): the dev-tools test-send list still references `invoice_payment_check_admin` — that surface already greys out absent templates.

### DIFF
--- a/backend/src/services/invoice/payments.js
+++ b/backend/src/services/invoice/payments.js
@@ -278,7 +278,7 @@ async function queuePaymentCheckEmail(invoiceId, { skipThrottle = false } = {}) 
     : null;
 
   await emailProcessor.queueEmail(invoice.event_id || null, adminContact.email,
-    'invoice_payment_check_admin', {
+    'invoice_payment_check', {
       invoice_number: invoice.invoice_number,
       customer_name: customer?.company_name
         || customer?.display_name


the-luap:
Verified the diagnosis end to end — correct fix.

`invoice_payment_check_admin` isn't a real template: it only exists in the dev test-send list (`adminDev.js:88`), never seeded. `emailProcessor` does an exact `template_key` lookup (`emailProcessor.js:716`) and throws at `:720`, so every dunning admin payment-check email failed, retried to the cap, and stuck `pending`.

The correct key `invoice_payment_check` (`crmEmailTemplates.js:217`, seeded by migration 116) is the admin "Paid / Partial / Not paid" email — and the payload here already passes exactly its variables (`paid_url` / `partial_url` / `unpaid_url`, skonto, late-fee). So this renders the email correctly, not just stops the throw; the payload was always meant for this template, only the key string was wrong.

Confirmed pre-existing, not a decomposition artifact: `git log -S` traces the wrong key to the original CRM commit (d543949); the refactor moved it verbatim.

Follow-up note is accurate — `adminDev.js:88` still lists the phantom key, cosmetic only (that surface greys out absent templates).

One non-blocking nit: worth a one-line test asserting `queuePaymentCheckEmail` queues `invoice_payment_check`, since there's no coverage on the key today and this is exactly the kind of typo that silently regresses.

Merge cleared.

the-luap (approved):
approved

Issue #761, opened 2026-07-06
[FEATURE] Per-customer language override in the customer form


Luca-Timo:
The customer form has no field to set a customer's **preferred language**, even though the backend already supports it (`adminCustomers.js` validates `preferred_language` on create/update; `customerAccountsService` defaults it from the business profile). A global default exists (Settings → General → Language, `general_default_language`), but no per-customer override in the UI. Add a language select to the customer create/edit form writing `preferred_language`. Pairs with the billing-email-language bug so customer-facing emails actually honor it.

the-luap:
Fixed in #763 and shipped in **v3.45.0**, now promoted to stable. Thanks for the report — closing.

Issue #760, opened 2026-07-06
[BUG] Billing/dunning emails use the event gallery language instead of the customer/invoice language


Luca-Timo:
Customer-facing billing/dunning emails render in the **event's gallery language**, not the customer's/invoice's language. `emailProcessor.getRecipientLanguage()` resolves event-first: (1) `events.language`, (2) `customer_accounts.preferred_language`, (3) `general_default_language`, (4) email-config, else `en`. Invoice reminders (`invoice/reminders.js`) pass the invoice's `event_id`, so an event whose gallery is English forces the dunning email to English even for a German customer/invoice (`invoices.language` defaults `de`). For CRM/billing emails the customer/invoice language should win. Fix: let senders pass an explicit language override that skips the event-first resolution; have the invoice reminder pass the invoice/customer language.

the-luap:
Fixed in #763 and shipped in **v3.45.0**, now promoted to stable. Thanks for the report — closing.

PR #763, opened 2026-07-06
fix(email,ui): billing emails follow customer language + readable payment-check confirmation

### LINKED ISSUE #759: [BUG] Payment-check confirmation view ("Action recorded") unreadable in dark mode
The public payment-check **confirmation** card ("Action recorded") in `frontend/src/pages/public/PaymentCheckPage.tsx` is hard to read in dark mode. It hardcodes a dark-green title (`#166534`) over a `color-mix(in srgb, #dcfce7 50%, var(--color-surface))` background — in a dark theme `--color-surface` is near-black, so the mix is a muddy mid-tone and the dark-green title has almost no contrast. The payment-check page itself is fine; only the result card. Fix: make the confirmation card theme-adaptive (readable title + legible tint on both light and dark surfaces).

Luca-Timo:
Fixes three issues found while testing the invoice-dunning flow. Tested on the dev instance.

## #760 — billing/dunning emails used the event's gallery language
`emailProcessor.getRecipientLanguage()` resolves language **event-first**, so an invoice tied to an English-gallery event sent the customer an **English** dunning email even for a German customer/invoice.
- `emailProcessor` now honors an explicit **`__language`** in the email data (both the queue-processor and preview paths); otherwise it falls back to the existing event-first resolution — no change for gallery emails.
- The invoice reminder (`invoice/reminders.js`) resolves the locale as `customer.preferred_language || invoice.language || 'de'` and passes it as `__language`.

## #759 — payment-check confirmation card unreadable in dark mode
The "Action recorded" card hardcoded a dark-green title over a light-green `color-mix` background — muddy + illegible on dark surfaces. Now theme-adaptive: a green tint over `--color-surface` with `--color-text` title/body, readable on both light and dark.
Screenshot 2026-07-06 at 19 59 27

## #761 — per-customer language
The per-customer **"Preferred language"** dropdown already exists (`CustomerDetailPage`), as does the global business-profile default. The gap was that emails ignored it (#760). Updated the field's helper text (en + de) to note billing emails now honor it.

Fixes #759, #760, #761.

### DIFF
--- a/backend/src/services/emailProcessor.js
+++ b/backend/src/services/emailProcessor.js
@@ -726,9 +726,12 @@ async function sendTemplateEmail(to, templateKey, variables) {
       throw new Error('Email configuration not found');
     }
 
-    // Determine recipient language (pass eventId if available in variables)
-    const language = await getRecipientLanguage(to, variables.eventId || null);
-    
+    // Determine recipient language. An explicit `__language` in the email data
+    // wins (CRM/billing emails set it to the customer/invoice language so a
+    // gallery event's language can't override a dunning notice — see #760);
+    // otherwise fall back to the event-first recipient resolution.
+    const language = variables.__language || await getRecipientLanguage(to, variables.eventId || null);
+
     // Process template with variables
     const { subject, htmlBody, textBody } = await processTemplate(template, variables, language);
 
@@ -783,7 +786,7 @@ async function sendTemplateEmail(to, templateKey, variables) {
 async function renderQueuedEmail(templateKey, variables = {}, to = '') {
   const template = await db('email_templates').where('template_key', templateKey).first();
   if (!template) return null;
-  const language = await getRecipientLanguage(to, variables.eventId || null);
+  const language = variables.__language || await getRecipientLanguage(to, variables.eventId || null);
   const { subject, htmlBody } = await processTemplate(template, variables, language);
   return { subject, html: htmlBody };
 }
--- a/backend/src/services/invoice/payments.js
+++ b/backend/src/services/invoice/payments.js
@@ -185,6 +185,8 @@ async function queueInvoicePaidAdminNotification({
         || [customer?.first_name, customer?.last_name].filter(Boolean).join(' ')
         || customer?.email || '',
       event_name: invoice.event_name || '',
+      // Keep the body language consistent with the locale-formatted amounts.
+      __language: locale,
       total_amount: formatMajor(invoice.total_amount_minor, invoice.currency, locale),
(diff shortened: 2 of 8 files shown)


the-luap:
Verified the whole chain — the fix is correct.

`__language` is honored on both the queue (`sendTemplateEmail`) and preview (`renderQueuedEmail`) paths and falls back to the existing event-first resolution, so gallery emails are untouched. Leak-safe too: `safeTemplateReplace` only substitutes `{{tokens}}` present in the template, and nothing references `{{__language}}`, so the control key can't surface in the body. `customer.preferred_language` is a real column (loaded via the `select *` in `applyReminder`), and the precedence reads right. #759 card + #761 helper text (en/de/inline) all good.

One gap worth closing in the same PR — the sibling billing emails have the identical event-first bug and don't set `__language`:

- `invoice_sent` (`invoice/sending.js:121`) — it even re-syncs `invoice.language` from the customer and formats amounts in `ctx.locale`, but the email body still resolves event-first, so a German customer on an English-gallery event gets an English invoice email with German-formatted amounts.
- `storno_issued` (`invoice/sending.js:365`) — same shape.
- payment-check (`invoice/payments.js:180/280`) — lower priority (admin-facing), but same pattern.

Each already computes the locale, so it's a one-liner per call: `__language: ctx.locale` (invoice_sent/storno) / `__language: locale` (payment-check). Worth folding in so billing-email language is consistent, not just dunning.

Otherwise merge cleared.

Luca-Timo:
@the-luap Folded in — c0008be3. All the sibling billing emails now set `__language` and match the locale their amounts are formatted in, instead of resolving event-first:

- `invoice_sent` (`sending.js:132`) → `__language: ctx.locale`
- `storno_issued` (`sending.js:375`) → `__language: ctx.locale`
- payment-check (`payments.js:292`) → `__language: locale`
- `invoice_paid_admin_notification` (`payments.js:189`) — the other `:180` site → `__language: locale`

Leak-safe (nothing references `{{__language}}`) and each falls back to the existing event-first resolution when unset. Ready for another look.

the-luap:
Re-verified c0008be3. All four sibling emails now carry `__language`, each matching the locale their amounts are formatted in:

- `invoice_sent` + `storno_issued` → `ctx.locale` (`sending.js:132` / `:371`)
- payment-check + `invoice_paid_admin_notification` → `locale` (`payments.js:189` / `:292`)

Swept the invoice `queueEmail` sites — every event-scoped one (reminder, invoice_sent, storno) now sets it, so there's no event-first path left for billing. Nice catch folding in `invoice_paid_admin_notification` — I hadn't listed that one. CI green.

Merge cleared.

PR #764, opened 2026-07-06
fix(workflows): backfill existing invoices + anchor dunning grace to due date when enabled (#750)

### LINKED ISSUE #750: [BUG] Enabling the invoice_dunning workflow silently stops dunning for already-sent unpaid invoices
## Summary

Enabling the built-in **Invoice dunning** workflow makes the engine take over dunning and **suppresses the legacy time-based reminder ladder** — but the workflow only creates runs for invoices sent *after* it was enabled. As a result, **every invoice that was already sent and still unpaid at the moment the workflow is enabled gets dunned by neither mechanism.** Dunning goes dark for those invoices, silently, with no error.

## Steps to reproduce

1. Send an invoice with a due date (status `sent`), leave it unpaid.
2. *Afterwards*, enable the built-in **Invoice dunning** workflow (`trigger_type: invoice.sent`).
3. Wait past `due_date + reminder_first_days`.

## Expected

The already-open invoice keeps getting dunned — either the legacy ladder still covers it, or enabling the workflow backfills a run for it.

## Actual

No payment-check / reminder is ever sent for that invoice. It is enrolled in **no** dunning path.

## Root cause

- `invoiceService.runScheduledTasks()` gates the legacy ladder off when the built-in is active:
  ```js
  engineDrivesDunning = await workflows.isBuiltinFlowActive('invoice_dunning'); // ~invoiceService.js:3505–3514
  ```
  (intentional mutual-exclusion, to avoid double-dunning — see the comment on the `queue_payment_check` action in `workflows/actions.js`).
- The workflow only ever gets a run from the `invoice.sent` emit, which fires **only at send time** (`invoiceService.js` ~L2095). There is **no backfill** for invoices that were already sent when the workflow was enabled.
- Net: legacy ladder = skipped; workflow run = never created → the invoice is orphaned.

## Impact

Real-world: an invoice sent before the workflow was enabled (due date passed) received **zero** reminders. Affects **all** pre-existing unpaid `sent`/`overdue` invoices at enablement time — a photographer who turns on dunning to chase old unpaid invoices gets the opposite of what they expect.

## Suggested fix (either)

1. **Backfill on enable** — when the `invoice_dunning` built-in is activated, create an `invoice.sent` run for every open `sent`/`overdue` unpaid invoice (idempotent via the existing per-(flow,entity) dedup). *Preferred.*
2. **Scope the ladder skip** — in `runScheduledTasks`, only skip an invoice when it actually has an active dunning run; invoices without one still fall through to the legacy ladder.

## Workarounds (current)

- Backfill one invoice: `emitWorkflowEvent('invoice.sent', { entityType:'invoice', entityId })`.
- Or one-shot: `invoiceService.queuePaymentCheckEmail(id, { skipThrottle:true })`.

## Environment

- Version: 3.79.1-beta.0
- DB: PostgreSQL (bundled)

Luca-Timo:
## Root cause

Enabling the `invoice_dunning` built-in stands the legacy time-based reminder ladder down (the `isBuiltinFlowActive` mutual-exclusion guard in `runScheduledTasks`), but the flow is only triggered by the `invoice.sent` event. Invoices that were **already sent** before dunning was enabled never emitted that event under the new flow, so they get dunned by *neither* path — the exact silent gap #750 reports.

## Fix

**1. Backfill on enable.** Turning the dunning built-in ON now enrolls every open, unpaid invoice by emitting `invoice.sent` for it (`engine.backfillDunningRuns()`, wired into `PATCH /:id/enabled` in `adminWorkflows.js`). Idempotent via the existing per-`(workflow, entity)` `dedup_key`, so re-enabling never double-enrolls. Best-effort — a backfill error is logged and never fails the toggle.

**2. Grace anchored to the due date.** `computeWakeAt` now treats `{ untilVar, delayDays }` as *var + offset* (was var-only OR now+offset), and the built-in's grace wait becomes `{ untilVar: 'dueDate', delayDays: firstDays }` (seed v6 → v7). An already-overdue invoice therefore duns on its **real** timeline (due + N days, already elapsed → fires on the next scheduler tick) instead of restarting a fresh grace clock from the enable moment.

Existing throttle still applies: `queuePaymentCheckEmail` 24h-throttles on `last_payment_check_at`, so a backfilled invoice whose admin payment-check was sent manually within the last day is not double-sent.

## Files

- `backend/src/services/workflows/engine.js` — `computeWakeAt` untilVar+offset; new `backfillDunningRuns()` + export.
- `backend/src/services/_workflowSeedBoot.js` — `waitGrace` anchored to `dueDate`; dunning seed `version` 6 → 7.
- `backend/src/routes/adminWorkflows.js` — enable toggle calls `backfillDunningRuns()` for the dunning built-in.
- `backend/__tests__/integration/workflowEngine.test.js` — seed-version assertions updated to v7.

## Testing

`invoiceDunning` + `workflowEngine` suites green. `computeWakeAt` change is backward compatible (untilVar-only → the var; delay-only → now+delay).

## Caveat (reviewer note)

The due-date-anchored graph applies to **freshly seeded** built-ins. An **already-admin-enabled** dunning workflow is admin-owned, so the boot seeder won't re-seed its graph on the version bump — it still enrolls existing invoices via the backfill, but keeps its current (relative) grace timing until re-seeded. A "reset built-in to default" affordance would let admins pull the v7 graph on demand; flagged as a follow-up rather than bundled here.

Closes #750

### DIFF
--- a/backend/__tests__/integration/workflowEngine.test.js
+++ b/backend/__tests__/integration/workflowEngine.test.js
@@ -239,7 +239,7 @@ describe('workflow engine', () => {
     expect(again.already).toBe(true);
   });
 
-  test('seeds the invoice-dunning built-in as the delegation graph (v6, disabled for first beta)', async () => {
+  test('seeds the invoice-dunning built-in as the delegation graph (disabled for first beta)', async () => {
     const { seedBuiltinWorkflowsAtBoot, DUNNING_KEY } = require('../../src/services/_workflowSeedBoot');
     const noopLogger = { info() {}, warn() {} };
     await seedBuiltinWorkflowsAtBoot(db, noopLogger);
@@ -248,7 +248,7 @@ describe('workflow engine', () => {
     expect(wf).toBeTruthy();
     expect(!!wf.is_builtin).toBe(true);
     expect(!!wf.enabled).toBe(false); // first beta: ships disabled; legacy ladder runs until enabled
-    expect(JSON.parse(wf.trigger_config).seedVersion).toBe(6);
+    expect(JSON.parse(wf.trigger_config).seedVersion).toBe(7);
 
     const nodes = await db('workflow_nodes').where({ workflow_id: wf.id, version: wf.version });
     expect(nodes.filter((n) => n.type === 'trigger')).toHaveLength(1);
@@ -273,7 +273,7 @@ describe('workflow engine', () => {
     await seedBuiltinWorkflowsAtBoot(db, noopLogger);
     const reseeded = await db('workflows').where({ id: wf.id }).first();
     expect(reseeded.version).toBe(wf.version + 1); // bumped
-    expect(JSON.parse(reseeded.trigger_config).seedVersion).toBe(6);
+    expect(JSON.parse(reseeded.trigger_config).seedVersion).toBe(7);
     expect(!!reseeded.enabled).toBe(false); // seed default re-applied (not admin-owned → flips enabled→disabled)
     const newNodes = await db('workflow_nodes').where({ workflow_id: wf.id, version: reseeded.version });
     expect(newNodes.some((n) => n.type === 'gate')).toBe(false); // legacy graph replaced
--- a/backend/src/routes/adminWorkflows.js
+++ b/backend/src/routes/adminWorkflows.js
@@ -254,6 +254,17 @@ router.patch('/:id/enabled', requirePermission('workflows.manage'), async (req, 
     // enabled state on the next SEED_VERSION bump (review nit #1).
     if (await hasColumnCached('workflows', 'admin_toggled_at')) patch.admin_toggled_at = db.fn.now();
     await db('workflows').where({ id }).update(patch);
+    // Turning dunning ON enrolls existing open/unpaid invoices (anchored to
+    // their due date) so it starts chasing current debtors, not only invoices
+    // sent after enabling (#750). Best-effort — never fail the toggle over it.
+    if (enabled && wf.builtin_key === 'invoice_dunning') {
+      try {
+        const n = await require('../services/workflows').backfillDunningRuns();
+        require('../utils/logger').info('[workflow] dunning enabled — enrolled existing invoices', { enrolled: n });
+      } catch (e) {
+        require('../utils/logger').warn('[workflow] dunning backfill failed', { error: e.message });
+      }
+    }
     res.json({ id, enabled });
   } catch (e) { next(e); }
 });
--- a/backend/src/routes/adminWorkflows.js
+++ b/backend/src/routes/adminWorkflows.js
@@ -256,10 +256,12 @@ router.patch('/:id/enabled', requirePermission('workflows.manage'), async (req, 
     await db('workflows').where({ id }).update(patch);
     // Turning dunning ON enrolls existing open/unpaid invoices (anchored to
     // their due date) so it starts chasing current debtors, not only invoices
-    // sent after enabling (#750). Best-effort — never fail the toggle over it.
+    // sent after enabling (#750). Scoped to this flow's id so the backfill only
+    // enrolls dunning, not any custom invoice.sent flow. Best-effort — never
+    // fail the toggle over it.
     if (enabled && wf.builtin_key === 'invoice_dunning') {
       try {
-        const n = await require('../services/workflows').backfillDunningRuns();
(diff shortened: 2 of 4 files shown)


the-luap:
Verified the whole chain — solid fix, tests green. One thing to tighten before merge.

## 🟡 The backfill over-enrolls beyond dunning

`backfillDunningRuns()` emits `invoice.sent` without `targetWorkflowId`, and `emitWorkflowEvent` fans out to *every* enabled workflow whose `trigger_type` is `invoice.sent` (`engine.js:277`). `invoice.sent` is an exposed trigger for admin-created custom workflows (`WorkflowEditorPage.tsx:31`) — so if a customer has any custom `invoice.sent` flow enabled and then turns dunning on, the backfill enrolls *all historical open invoices* into that flow too, firing its actions for every past invoice.

Dunning is the only built-in on `invoice.sent` today, so it may not bite yet, but it's a footgun the moment someone builds a custom `invoice.sent` flow. The plumbing already exists — scope it to the flow being enabled:

```js
// caller already has `id`:
await require('../services/workflows').backfillDunningRuns(id);
// backfillDunningRuns(targetWorkflowId) → emitWorkflowEvent('invoice.sent', { …, targetWorkflowId });
```

`emitWorkflowEvent` already restricts to a single flow when `targetWorkflowId` is set (`engine.js:278`) — exactly right for a function named `backfillDunningRuns`.

## 🟢 Nit — `computeWakeAt` both-fields case

The rewrite changes one case the "backward compatible" note skips: a node with **both** `untilVar` and `delayDays`. Old code returned the var and silently ignored `delayDays`; new code returns `var + delayDays`. No seeded node hits this (only the new `waitGrace`, which is the intended behaviour), but an admin custom wait node with both would shift timing. Worth a one-line mention in the note.

## ✅ Verified

- Export resolves — the `workflows` barrel spreads `...engine`, so `.backfillDunningRuns` is live.
- Ordering — `enabled=true` is written before the backfill, so dunning itself gets enrolled.
- Idempotency — `dedup_key` per `(workflow, version, entity)`; re-enabling skips the existing run.
- Backfill query — `['sent','overdue']` + `paid_amount_minor < total_amount_minor` catches open and partially-paid invoices (`partial` is a payment action, not a status), excludes `scheduled`/`paid`/`cancelled`.
- No email storm — the run parks at `waitDue`; the dunning email only fires after grace + not-paid, and `queuePaymentCheckEmail`'s 24h throttle guards double-sends.
- The already-enabled-keeps-relative-timing caveat is accurate.

## Verdict

Scope the backfill to the dunning workflow (`targetWorkflowId`) and I'm comfortable merging; the nit is optional.

Luca-Timo:
@the-luap Thanks for the thorough review! Both points addressed in da3a77da.

🟡 Backfill over-enrollment — backfillDunningRuns now takes a targetWorkflowId and threads it into emitWorkflowEvent, and the caller in adminWorkflows.js passes the id of the flow being enabled:

const n = await require('../services/workflows').backfillDunningRuns(id);
// → emitWorkflowEvent('invoice.sent', { …, targetWorkflowId });
So enabling dunning enrolls historical open invoices only into the dunning flow — a custom invoice.sent flow an admin builds is never touched. The param is optional (targetWorkflowId != null gate), so calling it without an id keeps the old fan-out behavior and nothing else changes.

🟢 computeWakeAt both-fields note — expanded the comment to call out the behavior change: a node with both untilVar and delayDays used to return just the var and now adds them (intended waitGrace semantics; no seeded node relied on the old path).

Ready for another look. 🙏

the-luap (approved):
Re-verified da3a77da — both points closed.

- **Scoped backfill:** `backfillDunningRuns(id)` now forwards `targetWorkflowId` to `emitWorkflowEvent`, which restricts the fan-out to that one flow (`engine.js:281`). `id` is the dunning workflow being enabled (`adminWorkflows.js:238`, and the call is guarded by `wf.builtin_key === 'invoice_dunning'`), so the backfill enrolls dunning only — a custom `invoice.sent` flow no longer gets swept in.
- **computeWakeAt nit:** the both-fields behaviour change is now documented in the comment; no seeded node relied on the old path.

Idempotency (`dedup_key`), the enable-before-backfill ordering, and the `['sent','overdue']` + `paid < total` query all still check out. Tests 30/30, CI green.

Merge cleared.

PR #769, opened 2026-07-07
feat(messages): unified Messages email client (flag-gated, default off)

Luca-Timo:
A new admin **Messages** section — a mail-client view over the mail PicPeak already sends and receives, plus a customer mailbox, an editable-template composer, one-click "create/reference a CRM document from a message", search, and archive/trash. Entirely behind the existing **`messaging`** feature flag (default **off**); an install sees no change until it's enabled. No behavior change to any existing send/receive path.

## Screenshots

**Messages — the mail client**
Screenshot 2026-07-07 at 15 34 29

**Customer mailbox settings (incoming IMAP + outgoing SMTP)**
Screenshot 2026-07-07 at 15 35 38

**Create / reference a Quote from a message**
Screenshot 2026-07-07 at 15 34 41

**Contract**
Screenshot 2026-07-07 at 15 34 49

**Gallery (event editor)**
Screenshot 2026-07-07 at 15 34 52

**Invoice**
Screenshot 2026-07-07 at 15 35 02

## What it is

Sidebar groups map to real mailbox identities (addresses read from the mail config, not hardcoded):
All mail Inbox · Sent
Customers (hello@) Inbox · Sent
Accounting (rechnungen@) Inbox
Automated (no-reply@) Sent
─────────────────
Archived · Deleted (cross-account system folders)

Inbound is filed by receiving address; the Sent stream splits by origin (`system` = automated/template mail, `manual` = human-composed → Customers ▸ Sent).
## Feature breakdown
- **Viewer** over `email_queue` (Sent/Automated, bodies from `rendered_html`, migration 119) and `received_emails` (Accounting inbox). Reuses the accounting rasterized-PDF viewer for attachments.
- **Customer `hello@` mailbox** — a second inbound IMAP mailbox with its own outgoing SMTP identity. Its mail appears under Customers with the body captured; attachments are **not** routed to Accounting.
- **Composer** — Reply / New message open an editable composer pre-loaded with the rendered template (or a reply stub); the body is fully editable (add a note anywhere) and sent as-is, recorded as a `manual` send. Manual mail sends from the customer mailbox's SMTP identity (falls back to the global from-address).
- **Document actions** — Quote / Contract / Invoice / Gallery buttons resolve the customer from the sender (or let you pick/create one), then **create a new** document in the real CRM editor (prefilled via `?customerAccountId=`) or **reference an existing** one in a reply. Gated by the `quotes` / `contracts` / `bills` flags.
- **Search** — filters the current folder (sender/subject), incl. the merged Archived/Deleted views.
- **Archive / Delete** — soft moves to Archived / Deleted folders; Restore moves back; deleting **from** the Deleted folder is the only permanent removal.
## Backend
- **Migrations** (all additive, `hasColumn`/`hasTable`-guarded, sane `down()`):
  - `154` — `mail_accounts` table (extra inbound mailboxes) + `received_emails.{account_key, to_address, body_html, body_text}`.
  - `155` — `email_queue.origin` (`system` | `manual`).
  - `156` — `mail_accounts.{smtp_*, from_*}` (per-mailbox outgoing identity).
  - `157` — `{email_queue, received_emails}.mailbox_state` (`active` | `archived` | `deleted`).
- **`emailIntakeService`** refactored to `pollAccountOnce(cfg, {accountKey, routeToExpenses})`; polls the accounting mailbox **and** every enabled `mail_accounts` row. The accounting attachment→expenses flow is preserved byte-for-byte behind `routeToExpenses`; customer mail is logged with its (sanitized) body and not routed to accounting.
- **`emailProcessor.sendRawEmail({..., accountKey})`** sends admin-edited HTML via the per-account SMTP identity (falls back to global).
- **Routes** (`email.view`/`edit`/`send`): `GET /queue/:id`, `GET /received/:id`, `GET /identities`, `GET/POST /accounts` (+ `/test`), `POST /send`, `POST /item/:kind/:id/state`, `DELETE /item/:kind/:id`, plus `account`/`origin`/`state` filters on the list endpoints.
- Adds the missing `?customerAccountId=` prefill to `ContractEditorPage` (quotes + bills already had it).
## Security
- Inbound HTML is **sanitized server-side** (`sanitize-html`) on ingest **and** rendered in a strict, script-less, no-same-origin sandboxed iframe (`sandbox=""`) — both the inbound and the sent-body panes. The `/send` sanitizer uses the same strict allowlist (no `<style>`, no `data:`).
- New mailbox `/accounts` + `/accounts/test` reject private/internal IMAP/SMTP hosts via `isPrivateIP()`, matching `/config` + `/incoming-config` (SSRF).
- Attachment **disk paths are never exposed** — only filenames. IMAP/SMTP passwords are masked on read; per-account SMTP transport verifies TLS. Credentials are stored plaintext (consistent with the existing `email_configs`).
## Reuse / conventions
`RequireFeature` + `PermissionGate` (`email.view` / `email.send`), Tailwind `dark:` theming, the branding accent (`bg-accent-soft` / `accent-dark`), TanStack Query, the existing CRM editors + `CustomerPicker` + list/PDF endpoints, and the accounting inbound blob viewer. No duplicated document or email logic.
## i18n
Only `navigation.messages` is added to `en.json` + `de.json`. The page's own strings use inline `t(key, 'English fallback')`, so the **`messages.*` namespace has no German translations yet** — they render in English until added. Flagged for native review.
## Testing
`npm run build` (frontend) and migration boot (154–157) pass. **No automated test** was added for the `pollAccountOnce` intake refactor (no existing intake test harness) — verified by preserving the accounting path behind the flag + build/boot. A `pollAccountOnce` regression test (accounting still routes to expenses; a customer mailbox captures the body and does not) is the recommended pre-merge add.
## Deferred (roadmap, intentionally out of scope)
- **Two-way IMAP write-back** (mark-seen / move / delete on the server) — behind its own flag, sequenced last.
- Customer-mail **attachment storage/viewing** (counted, not yet stored).
- "Reference existing document" inserts the **number**; attaching the PDF / public quote/contract link is a follow-up (endpoints exist).
- Gallery is create-only (events are multi-customer); no "select existing gallery".
- Server-side search (current search is client-side over the loaded page); per-customer feature-flag gating of the doc buttons; full thread-linking of documents sent via the CRM editor.


the-luap:
Reviewed the whole thing — migrations + backend routes + intake refactor directly, frontend deep-audited and cross-checked at source. The architecture is solid and the sandboxed body rendering is done right. One critical blocker slips through the single path that bypasses the iframe.

## 🔴 Blocker — stored XSS via inbound sender name

An attacker who emails the `hello@` customer mailbox can take over an admin session. Verified end-to-end:

1. `emailIntakeService.js:374` stores `from_address: ((parsed.from && parsed.from.text) || '').slice(0, 512)` — the full `From` header incl. the **attacker-controlled display name**, unsanitized.
2. `MessagesPage.tsx:530` interpolates it **raw** into the reply stub:
   ```js
   const quoted = `<p><br></p><p style="…">${t('messages.onWrote','On')} ${fmt(it.received_at)}, ${it.from_address}:</p>`;
   ```
   → passed as `html` (`:531`).
3. `MessageComposer.tsx:38` — `bodyRef.current.innerHTML = init.html || ''` on an **unsandboxed** `contentEditable` div in the admin origin.

A display name of `` executes the moment the admin clicks **Reply** — session/JWT theft. The body *reading* panes are correctly sandboxed; this is the one path that bypasses them.

**Fix:** HTML-escape `it.from_address` before interpolation. The reading pane already renders it safely via JSX (`:631`) — only the composer stub is raw.

## 🟡 Concerns

1. **Backend routes aren't gated by the `messaging` flag.** The PR says "entirely behind the messaging flag," but the only gate is the frontend route (`App.tsx:245`). The backend routes have `adminAuth` + `requirePermission('email.*')` and no `requireFeatureFlag('messaging')` — so `POST /admin/email/accounts` and `/send` are live for any admin with email perms even when the feature is off. Every other flagged feature gates the backend (`adminWorkflows.js:29`, `adminWhatsapp.js:34`, bills/accounting/expenses). Add `requireFeatureFlag('messaging')` to the **new** routes only — `/email` is a shared mount, so gating the whole mount would break pre-existing email-config admin.
2. **`/messages` route has no permission gate** — `RequireFeature` only; a user without `email.view` can deep-link it and fire the list/detail fetches. Server-side `email.view` *is* enforced, so it's defense-in-depth/consistency, not a leak — but the sidebar (hidden on `email.view`) and the route disagree on the permission dimension.
3. **`DocumentActionModal.tsx:52`** auto-picks `rows[0]` from `customerAdminService.search(senderEmail)`. Confirm `searchCustomers` matches email **exactly** — if it's substring/fuzzy, a spoofed sender could prefill a CRM doc against the wrong customer (the admin sees the picker, so not a leak).

## 🟢 Nits
- Search (`MessagesPage.tsx:443`) filters only the loaded page client-side; larger mailboxes are silently truncated before search.
- `to: it.from_address` may carry `"Name <email>"` as the recipient — cosmetic.
- `messages.*` has English inline fallbacks on every `t()`, so DE-missing renders English (not blank) — fine to follow up.

- **Migrations 154–157:** additive, `hasTable`/`hasColumn`-guarded, idempotent, sane `down()`.
- **Both body iframes genuinely `sandbox=""`** (`MessagesPage.tsx:583` + `:643`); no `dangerouslySetInnerHTML` in the messages components.
- **Sanitizers sound** (`/send` + inbound `sanitizeBody`): schemes limited to http/https/mailto/cid (no `javascript:`/`data:`), `on*`/`<script>`/`<style>` stripped.
- **`/item/:kind`** allowlisted to `received`/`queue` (no table injection); **`/accounts`** SSRF-guards both imap+smtp hosts; **`/accounts/test`** only connects IMAP; passwords masked on read, preserved on masked write.
- **Intake refactor preserves accounting byte-for-byte** behind `routeToExpenses: true`; customer mail is `false` (body captured, not routed to expenses).
- **Contract `?customerAccountId=` prefill validated** (`parseInt` → `Number.isFinite && > 0` → server-authorized get); **permanent delete only from the Deleted folder**; **doc buttons flag-gated** (quotes/contracts/bills).
- `email.view/edit/send` are existing permissions (no provisioning gap). Minor: `isPrivateIP` doesn't do async DNS resolution, but that's the pre-existing shared guard, not new here.

## Verdict
Close the XSS blocker (escape `from_address`) and add the backend `messaging` flag gate, and I'm comfortable merging; a one-line answer on the search-match semantics clears concern #3. Nice work on the sandboxed rendering and the intake preservation.

Luca-Timo:
@the-luap Fixed in bb235e72 + 1e08a4fb.

**🔴 XSS — closed.** `from_address` is HTML-escaped before the reply-stub HTML, and the composer `DOMPurify.sanitize()`s before `innerHTML`. Reading panes were already `sandbox=""`.

**🟡 Backend flag gate — added.** `requireFeatureFlag('messaging')` per-route on all 8 new endpoints (queue/:id, received/:id, identities, accounts, accounts/test, send, item/:kind/:id/state, DELETE item/:kind) — not the shared `/email` mount, so the pre-existing config/queue/received endpoints stay reachable with the feature off.

**🟡 #3 — fixed.** `DocumentActionModal` auto-picks only on an exact email match now (search is prefix-based); otherwise it leaves the picker to the admin.

**🟢 Nits — fixed.**
- Search now hits the backend (`/received` gained a `q` filter; frontend passes the debounced term to every list query), so it's no longer truncated to the loaded page.
- Reply recipient extracts the bare address from `Name <addr>` (`extractEmail`).
- Added full `messages.*` + `email.customerMailbox.*` de/en namespaces (were English inline-fallbacks).

**🟡 #2 route permission gate.** Kept as-is: no route in `App.tsx` is permission-gated — routes are feature-gated and permission is enforced server-side + via the sidebar (which hides on `email.view`). Adding a one-off wrapper on `/messages` only would break that convention. The server 403s every messages route without `email.view`, so there's no data exposure — happy to add a route wrapper if you'd prefer it there.

Remaining: a `pollAccountOnce` regression test — I'll add it as a follow-up. Thanks for the review.

the-luap:
Re-verified bb235e72 + 1e08a4fb — blocker and all concerns closed, spot-checked at source.

## 🔴 Blocker — closed (defence in depth)
The reply-sender XSS is fixed at both ends:
- `MessagesPage.tsx` — new `escapeHtml()` applied at the interpolation site: `${escapeHtml(it.from_address || '')}` (`:533`), so the attacker-controlled display name is neutralised where the raw HTML stub is built.
- `MessageComposer.tsx:38` — `DOMPurify.sanitize(init.html || '')` before the `contentEditable` `innerHTML`, so the sink is safe even if some other path feeds it untrusted HTML. `dompurify ^3.2.6` is a declared dep, and the sanitizer keeps the `<p style>` stub so the reply still renders.

## 🟡 Concerns — closed
1. **Backend flag gate** — `messagingGate = requireFeatureFlag('messaging')` now sits on every NEW route (`/received/:id`, `/item/*`, `/accounts` GET+POST, `/accounts/test`, `/identities`, `/queue/:id`, `/send`), placed `adminAuth → messagingGate → requirePermission`. The pre-existing list/config endpoints (`/received`, `/queue`, `/config`) stay ungated — exactly right for the shared mount.
2. **Search** — now server-side: `/received` takes `?q=` and filters `from_address`/`subject` across the full table (`like %q%`, knex-parameterised, capped 255) instead of the loaded page.
3. **Customer auto-pick** — `DocumentActionModal` only auto-selects on an EXACT lowercased email match (`rows.find(r => r.email === target)`), otherwise leaves the picker. A spoofed/partial sender can't prefill the wrong customer.

## 🟢 Nits — closed
Bare-email recipient handled, and a full German `messages.*` + `customerMailbox` block landed (no longer English-only).

## Verdict
Merge cleared from my side. Nice — the double-layer XSS fix and the per-route gating are the right shape. (Heads up: `bypass-size-gate` shows red, but that's the size-label gate, not a code failure — `size-gate` itself passed.)

the-luap (approved):
Merge cleared. Re-verified the two fix commits (bb235e72 + 1e08a4fb) at source: the reply-sender XSS is closed at both ends (escapeHtml at the interpolation site + DOMPurify before the composer innerHTML), the messaging flag now gates every new backend route (pre-existing endpoints left ungated), search is server-side, and the doc-action customer auto-pick requires an exact email match. Migrations, sandboxed body rendering, sanitizers, SSRF guards and the intake preservation all checked out in the earlier pass. Nice work.

Issue #749, opened 2026-07-04
[FEATURE] Epic: gallery delivery & multi-photographer improvements


Luca-Timo:
Umbrella for a batch of gallery/delivery feature ideas — parked, **no timeline**. Motivated by real photographer workflows (multi-shooter events, richer client proofing, editing-tool integration). Each item ships independently as its own PR/release line.

Not in scope: card payments / print store (tracked separately). CRM/accounting suite is unaffected.

### Sub-features
- [ ] Multi-photographer "festival" upload — #743
- [ ] Frame-level proofing (per-image comments + approve/reject) — #744
- [ ] Lightroom round-trip (re-import approved selections) — #745
- [ ] Client-activity notifications — #746
- [ ] Team workspaces / creative-team roles — #747
- [ ] Cloud-storage source ingestion (nice-to-have) — #748
- [ ] Category reordering within events — #782

### Rough priority (effort-adjusted)
notifications (S) → frame-level proofing (L, high value) → festival + team roles (L) → Lightroom round-trip (L) → cloud ingestion (nice-to-have)

PR #788, opened 2026-07-10
feat(slideshow): per-event play order + category filter (#202)

the-luap:
Closes #202.

The **Live Slideshow** already ships the core of #202 — a token-only fullscreen kiosk that live-appends new uploads during an event, with timing, transitions, colour filters and a logo watermark, opt-in per event via the share link. This PR adds the two customization dimensions the reporter also asked for: **play order** and a **content filter**.

## What's new
- **Play order** (`show_order`): `chronological` (upload order, default) or `random` — the kiosk shuffles the initial set (Fisher–Yates) so live-appended uploads keep landing correctly.
- **Category filter** (`show_category_id`): restrict the slideshow to a single photo category (NULL = all photos, default). **Enforced server-side** on the slideshow's `/photos` access *and* mirrored in the `/session` + `/state` `photo_count`, so the kiosk viewer can't widen the set and the new-upload poll stays consistent.

Per-event **enable/disable (default off)** is the existing *Generate / Disable slideshow link* flow — no token = no slideshow. Unchanged.

admin controls

## Implementation
- **Migration 158** — `show_order` (default `chronological`) + `show_category_id` (nullable). Additive/guarded, PG + SQLite.
- **Backend** — `slideshowSettings` returns `order`; `/photos` + `slideshowPhotosQuery` (count) apply the category filter for `accessLevel:slideshow`; the PATCH `/:id/slideshow` accepts + validates both, with a category-ownership check (must belong to the event or be global).
- **Admin** — Play-order dropdown + "Show only category" picker in the Live Slideshow card (picker auto-hides for events with no categories); EN + DE i18n.
- **Kiosk** — `SlideshowPage` shuffles when `order === 'random'`.

## Verified
- Migration on SQLite + PG.
- Live API: category filter → 3 / 2 / 5 photos across two categories + `photo_count` matching; `order` propagates to the session.
- Running kiosk requests exactly the filtered set (photo ids 4/5/6 for the 3-photo category, not all 5).
- `tsc` clean, 136 backend tests pass, i18n en/de parity.

PR #790, opened 2026-07-10
feat(categories): per-event category ordering — global default + override (#782)

### LINKED ISSUE #782: [FEATURE] Category reordering within events
Manual **reordering of categories / pages within an event**, so a gallery follows the flow of the day instead of alphabetical order.

### Motivation
From #780 (@liamrotheram): categories currently display **A–Z**, so a wedding reads *Ceremony, Couple's Shoot, Evening, Group Shots, Post-Ceremony, Pre-Ceremony, Reception* instead of the actual order of the day:

> Pre-Ceremony → Ceremony → Post-Ceremony → Group Shots → Couple's Shoot → Reception → Evening

Parity with Pixieset's per-event category ordering.

### Proposal
- Add an explicit per-event **`position`** (sort order) to categories.
- **Drag-to-reorder** in the admin category / gallery editor.
- The gallery viewer renders categories by `position`, falling back to name when unset.

### Notes
- **Preserve existing galleries:** on migration, seed `position` from the current (alphabetical) order so nothing reshuffles for existing events — custom order is opt-in by dragging.
- Per-event ordering (not a global default).

Part of #749 · originally raised in #780.

Luca-Timo:
## Summary

Photographers can order a gallery's categories in the **flow of the day** (Pre-Ceremony → Ceremony → Reception …) instead of A–Z. Closes #782, raised in #780, part of #749.

Galleries are usually built from the **shared global** categories (Ceremony, Reception, Portraits …), so ordering is layered — **resolved per event: per-event override → global default → name**:

- **Global default** — reorder the shared categories once (Settings → Photo Categories); every gallery uses this order.
- **Per-event override** — an event's Categories tab shows one **combined** list (globals + event-specific, interleaved) so a single gallery can be arranged independently, with a **Reset to default**.

## Screenshots

**Global default order** — Settings → Photo Categories. Up/down arrows set the order every gallery uses by default.
Screenshot 2026-07-10 at 11 55 37

**Per-event override** — an event's Categories tab. Shared globals (SHARED badge) + any event-specific categories form one reorderable list; reordering overrides the global default for this gallery only.
Screenshot 2026-07-10 at 11 55 20

## Changes
- **migration 158** — `photo_categories.display_order` (global default), backfilled from the current alphabetical order so existing galleries don't reshuffle.
- **migration 159** — `event_category_order (event_id, category_id, position)` — the per-event override; no backfill, every event starts on the default.
- **utils/categoryOrder** — shared resolution used by the admin event view *and* the public gallery; fails safe to the global default if the table is absent.
- **adminCategories** — `POST /reorder` sets a per-event override (accepts globals, rejects out-of-scope); `DELETE /reorder/:eventId` resets; `POST /reorder-global` sets the global default; ordered list endpoints + create-append.
- **gallery** renders the resolved order.
- **EventCategoryManager** combined list + reset; **CategoryManager** global reorder. Up/down buttons — no drag-and-drop dependency.

## Tests
Real-DB integration suite: migration 158 backfill; global reorder + a non-customised event following it; per-event override + isolation from other events; override accepts globals / rejects a foreign event's category; reset clears the override; create appends. 6/6 green.

## Verified
- ✅ frontend `npm run build` + ESLint clean; backend ESLint clean
- ✅ full core-migration chain (through 159) boots on fresh SQLite
- ✅ `jest categoryReorder` 6/6

_DE strings authored; other locales untouched._


the-luap:
Solid feature — cross-event isolation, validation, atomicity and i18n all check out. Two blockers before merge, plus a read-scope concern worth folding in.

## 🔴 Blockers

### 1. Per-event reorder endpoints have no event-ownership check
`adminCategories.js:271` (`POST /reorder`) and `:320` (`DELETE /reorder/:eventId`) are gated only by `adminAuth, requirePermission('settings.edit')`. The scope check validates that the *categories* belong to the event (global or event-specific) — but **not that the admin owns the event**. Since `settings.edit` is a DB-seeded permission (grantable to non-super_admin roles), a lower-privilege admin can pass an arbitrary body `event_id` and rewrite — or `DELETE` — another admin's gallery category order.

The direct analog in this repo gates every per-event mutation on ownership:
```js
// adminEvents/slideshow.js:46
router.post('/:id/slideshow/generate', adminAuth, requirePermission('events.edit'), …, requireEventOwnership, …)
```
Fix: enforce ownership for non-super_admins. `requireEventOwnership` reads `req.params`, so it drops onto `DELETE /reorder/:eventId` directly — but `POST /reorder` takes `event_id` in the **body**, so it needs an inline `created_by` check (mirror `slideshow.js:35`: `q.where('created_by', req.admin.id)` for non-super_admins).

### 2. Migration number collision
`158_add_category_display_order.js` collides with `158_slideshow_order_category.js`, which merged to `main` a bit ago (#788). Two migration-158s. Renumber this PR's migrations to **159 + 160**, and update the `159` migration's header comment (it says "builds on migration 158" — that becomes 159).

## 🟡 Concern
`GET /event/:eventId` (`adminCategories.js:29`, `settings.view`) also lacks ownership scoping — a lower-privilege admin can read another event's category set. Read-only + lower severity (and pre-existing hero/download routes on this file share the gap), but this PR makes `/event/:eventId` a first-class surface, so worth closing in the same pass.

## 🟢 Nits
- `EventCategoryManager.tsx:456` — the inline EN default reads "Drag the arrows" but the control is click-only up/down buttons. The `en.json` value already says "Use the arrows", so the key resolves fine; just fix the stale inline fallback.
- `bg-accent-dark/150` — `/150` isn't a valid Tailwind opacity step (max 100). Pre-existing style, visible in this file.

## ✅ Verified
- **Cross-event isolation is correct** — `event_category_order` rows are keyed by `event_id`; `categoryOrder.js` joins with `andOnVal('o.event_id','=',eid)`; per-event reorder never touches the global `display_order`. Reordering event A can't affect B or the global default (the `override-a/override-b` test confirms B stays `null`).
- **Scope validation** on `/reorder` (`:288`) + `/reorder-global` (`:349`) rejects foreign/non-global ids (400).
- **Resolution order** (`categoryOrder.js:56-63`): override-position → `is_global` → `display_order` → `name` — full tiebreaker, single query, no N+1, with a fail-safe fallback if `event_category_order` is absent (half-applied migration) so the gallery can't 500.
- **Reset** deletes override rows, idempotent; both reorders wrapped in `db.transaction` (atomic).
- **Gallery** (`gallery.js:577`) uses the verified `req.event.id`, orders only categories-with-photos; no new auth surface, and no conflict with the slideshow category filter from #788.
- **i18n** en+de parity across all 9 new keys.

Close the two blockers (ownership + renumber) and I'm comfortable merging.

Luca-Timo:
Thanks for the thorough review — all addressed in `a4b4485d`.

**🔴 1. Event ownership** — `GET /event/:eventId` and `DELETE /reorder/:eventId` now run `requireEventOwnership` after the permission gate. `POST /reorder` takes `event_id` in the body (so the middleware's `req.params` read doesn't apply) → equivalent inline check: super_admin bypasses, other admins limited to owned/ownerless events, mirroring `slideshow.js`. Added a test: a non-super_admin role *holding* `settings.edit` is blocked (403) on all three per-event routes.

**🔴 2. Migration renumber** — bumped to **159 + 160** (upstream #788's `158_slideshow_order_category.js` is now on main); header comments ("builds on migration 159") and the test's `require` path updated.

**🟡 Read scope** — folded in: `GET /event/:eventId` is ownership-gated too (same middleware).

**🟢 Nits** — stale inline `"Drag the arrows"` → `"Use the arrows"` (matches en.json; control is click-only up/down); `bg-accent-dark/150` → `bg-accent-dark`.

Verified: 159+160 boot clean alongside #788's 158; **7/7** integration tests (incl. the new ownership case); frontend build + backend/frontend lint clean.

the-luap (approved):
Re-verified `a4b4485d` at source — both blockers + the read-scope concern closed.

- **Event ownership** — `requireEventOwnership` on `GET /event/:eventId` (`:30`) and `DELETE /reorder/:eventId` (`:335`); `POST /reorder` has the equivalent inline check (`:290-296`: super_admin bypass, else `event.created_by !== req.admin.id → 403`), exactly right for the body-`event_id` route. The regression test hits the precise case — a non-super_admin *holding* `settings.edit` gets 403 on all three per-event routes.
- **Migrations** — renumbered to `159_add_category_display_order` + `160_add_event_category_order`; no more `158` collision with the merged slideshow migration, and the header comment + test require path track.
- **Read scope** — `GET /event/:eventId` is now owner-gated.

Category tests: 7/7 green (incl. the new ownership test). The isolation / resolution-order / atomicity / i18n from the first pass still hold.

Merge cleared.

Issue #613, opened 2026-06-10
[BUG] File limit for uploads is not applied to guest uploads


Zszywany:
**Describe the bug**
"Max Files per Upload" setting does not affect guest uploads.
When accessing the event gallery as a guest, the "Upload Photos" modal does not limit the number of files and shows {{limit}} parameter instead of number set under Settings -> General -> Max Files per Upload.

Limit is enforced when uploading files from admin panel.

**To Reproduce**
1. Set "Max files per Upload" to 10
2. Create new Event
3. Check "Allow guests to upload photos"
4. Select "Group Photos" in Upload Category
5. Publish the gallery
6. Access the gallery as guest
7. Click "Upload Photos", select 16 files and click "Upload"
8. Result: The upload goes through. 

Expected behavior: Error is shown "Upload limit reached (10 files per batch)"

**Screenshots**
Guest upload:
<img width="684" height="426" alt="Image" src="https://github.com/user-attachments/assets/6feda935-76f2-4356-b2f7-bd6321ab6584" />

Admin upload:
<img width="684" height="748" alt="Image" src="https://github.com/user-attachments/assets/dbb38d63-f645-46d5-b575-268fd0e35aa0" />

**Environment (please complete the following information):**
 - OS: Ubuntu noble 24.04 x86_64
 - Browser: Firefox 151.0.2, Chromium 149.0.7827.53, Brave 1.91.168
 - PicPeak Version: v3.44.0
 - Deployment Method: Docker Compose
 - Database: PostgreSQL 15

the-luap:
Confirmed and tracked down. Two separate misses on the guest path:

The backend route at `gallery.js:1641` hardcoded `limits: { files: 10 }` AND `.array('photos', 10)` for the multer config — admin path at `adminPhotos.js:131` has been resolving the value dynamically from `general_max_files_per_upload` via `getMaxFilesPerUpload()` since #214, but the guest route just never used it. Setting your limit to 10 had no effect on the guest endpoint because the multer caps were already 10 from the literal — uploading 16 worked because the literal was rewritten to the actual setting nowhere along the path.

[...]

PR #823, opened 2026-07-17
fix(uploads): apply configured max file size to guest uploads (#613 follow-up)

the-luap:
Follow-up to #613/#614, from the new comment by `mat1990dj`: the admin's **Settings → General → "Max File Size (MB)"** value (`general_max_file_size_mb`) never applied to **guest** gallery uploads, so a guest couldn't upload a large video even when the admin raised the limit.

## Root cause
The guest upload path hardcoded multer's per-file cap at **50MB** (`gallery.js:1865`, literally commented *"separate concern from #613"*), and the guest UI hardcoded the same 50MB client-side guard **and** the "max 50MB" hint text. It's the same class of miss as the file-*count* bug fixed in #614 — a settings value never wired into the guest multer config — but for *size*.

(For the record: the reporter also mentioned formats — those already work; the guest path resolves `getAllowedMimeTypes()` from settings. And `general_max_upload_batch_size_mb` is a working client-side chunk control for the admin uploader, not a gap. So this PR is scoped to the one genuine bug: guest per-file size.)

## Changes
- **`uploadSettings.js`** — new `getMaxFileSizeMb()` / `getMaxFileSizeBytes()` reading `general_max_file_size_mb` (default 50MB, cached 60s, clamped to a 10GB ceiling), mirroring `getMaxFilesPerUpload`.
- **`gallery.js`** (guest upload) — multer `limits.fileSize` now resolves from the setting; a `LIMIT_FILE_SIZE` error returns an actionable *"max N MB"* message instead of multer's generic "File too large".
- **`publicSettings.js`** — exposes `general_max_file_size_mb` (default 50) so the gallery UI renders the real limit and can guard client-side before an oversized POST.
- **`UserPhotoUpload.tsx`** — reads the limit, uses it for the client-side size guard, and passes it to the requirements hint. The hardcoded "max 50MB" in `upload.fileRequirements` is now interpolated (`{{sizeLimit}}`) across **all 8 locales**; adds `upload.fileTooLarge` (en/de; others fall back to en).

## Scope decision
**Guest path only.** The admin path keeps its generous 10GB cap — admins are trusted, and enforcing the 50MB default there would regress large admin video uploads. This directly fixes the reported scenario: admin raises the setting → guests can upload the larger file (500MB is well under nginx's 1G `client_max_body_size`).

## Verification
- **Live**, on a booted instance: admin `PUT /api/admin/settings/general {general_max_file_size_mb:500}` → persisted → `GET /api/public/settings` returns `general_max_file_size_mb: 500` → the guest multer sources its cap from the getter.
- **Unit tests** (SQLite-backed): default 50MB, honours a configured 500MB, clamps nonsense/absurd values, respects the 60s cache. 4/4 pass.
- Backend loads + lints clean; frontend typechecks clean; all 8 locale JSON valid.

Bug fix → **main** (per request). No stable backport unless you want one.

### DIFF
--- a/backend/__tests__/services/uploadSettingsMaxFileSize.test.js
+++ b/backend/__tests__/services/uploadSettingsMaxFileSize.test.js
@@ -1,0 +1,65 @@
+/**
+ * Unit tests for the per-file upload size limit getter (general_max_file_size_mb),
+ * added so the admin's "Max File Size (MB)" setting applies to guest uploads
+ * (#613 follow-up — mat1990dj). Real in-memory SQLite app_settings so the
+ * read/parse/cache path runs exactly as in production.
+ */
+const knex = require('knex');
+
+let db;
+let svc;
+
+beforeEach(async () => {
+  db = knex({ client: 'sqlite3', connection: { filename: ':memory:' }, useNullAsDefault: true });
+  await db.schema.createTable('app_settings', (t) => {
+    t.increments('id');
+    t.string('setting_key').notNullable().unique();
+    t.text('setting_value');
+    t.string('setting_type');
+    t.timestamp('updated_at');
+  });
+  jest.resetModules();
+  jest.doMock('../../src/database/db', () => ({ db }));
+  svc = require('../../src/services/uploadSettings');
+  svc.clearMaxFileSizeCache();
+});
+
+afterEach(async () => {
+  jest.dontMock('../../src/database/db');
+  await db.destroy();
+});
+
+async function setLimit(mb) {
+  await db('app_settings')
+    .insert({ setting_key: 'general_max_file_size_mb', setting_value: JSON.stringify(mb), setting_type: 'general', updated_at: new Date() })
+    .onConflict('setting_key').merge({ setting_value: JSON.stringify(mb) });
+  svc.clearMaxFileSizeCache();
+}
+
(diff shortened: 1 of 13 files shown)


the-luap:
Verified end-to-end before requesting review:

- **Live**: on a booted instance, `PUT /api/admin/settings/general {"general_max_file_size_mb":500}` → persisted → `GET /api/public/settings` returns `general_max_file_size_mb: 500` → the guest multer sources its `limits.fileSize` from `getMaxFileSizeBytes()`. Default (unset) stays 50MB, so **no behaviour change for existing installs** until an admin raises it.
- **Unit tests** (SQLite-backed): default 50MB, honours a configured 500MB, clamps 0/absurd values to default/ceiling, respects the 60s cache. 4/4 pass.
- Backend loads + lints clean; frontend typechecks; all 8 locale JSONs valid.

Scope note for reviewers: **guest path only.** The admin path keeps its 10GB cap on purpose — enforcing the 50MB default there would regress admins who upload large videos. Two things from the original report are intentionally untouched because they already work: **allowed formats** (the guest route already resolves `getAllowedMimeTypes()` from settings) and **`general_max_upload_batch_size_mb`** (a working client-side chunk-size control for the admin uploader, not a backend gap).

Issue #821, opened 2026-07-17
[BUG] Guest upload ignores backend config


mat1990dj:
**Describe the bug**
The guest upload page ignores the global admin backend configuration settings for allowed file types (MAX_FILE_SIZE, MAX_TOTAL_UPLOAD_SIZE, and allowed extensions). Instead, the frontend client-side validation appears to hardcode allowed extensions and enforces a strict file size limit (failing at files over 50MB).

This prevents guests from uploading high-quality formats like Apple ProRAW (.dng) or larger video files (.mov / .mp4), even when the backend configurations are explicitly set to allow them. Paradoxically, .mov isn't listed in the upload popup's supported formats, but small .mov files under 50MB successfully upload anyway, confirming the backend supports it while the frontend limits it.

**To Reproduce**
Log in as an admin and configure the backend settings to allow large files (e.g., max file size 200MB) and add dng and mov to the allowed extensions.
Open the guest upload portal.
Attempt to upload a 51MB .mov or .mp4 video file, or any .dng raw image.
See client-side validation error blocking the upload or rejecting the file type before it hits the backend.

**Expected behavior**
The guest upload interface should dynamically respect the environment variables or backend configuration settings for file size and extension restrictions. If the admin configures a 100MB limit and enables DNG/MOV, guests should be able to upload those files seamlessly.

**Screenshots**
<img width="1290" height="999" alt="Image" src="https://github.com/user-attachments/assets/fa9c621f-ef3f-4465-8b11-ec5d71bc6ef6" />
<img width="1290" height="1262" alt="Image" src="https://github.com/user-attachments/assets/69d65a47-c512-4fed-8286-09412bd11a5a" />
<img width="1290" height="934" alt="Image" src="https://github.com/user-attachments/assets/ed2113fd-83c6-4661-9cf6-7a4b8e730fc0" />

**Environment**
OS: Windows Server
Browser: Chrome
PicPeak Version: 3.45.1
Deployment Method: Docker Compose

**Additional context**
This issue heavily impacts live event setups (like weddings) where guests frequently take videos or use iPhone ProRAW settings (.dng), which quickly exceed the hardcoded 50MB limit. Smaller .mov files (~20MB) successfully upload despite .mov not being explicitly listed in the UI text, indicating a discrepancy between the UI text, client validation rules, and the actual backend capability.

**Possible Solution**
[...]

PR #832, opened 2026-07-17
feat(uploads): HEIC/HEIF support + dynamic format hint on guest upload (#821)

the-luap:
Addresses two of the three items from #821 (the third — RAW/DNG — is a separate PR, see below).

## 1. HEIC/HEIF support
Sharp's bundled libvips decodes `heif` input — verified on this build (`sharp.format.heif.input.file === true`, Sharp 0.34.3 / libvips 8.17.1) — so thumbnails generate fine. Added `heic`/`heif` to `EXTENSION_TO_MIME` in **both** the backend (`uploadSettings.js`) and frontend (`fileTypes.ts`) maps (kept in sync). An admin can now add `heic,heif` to allowed file types and guests can upload them. (As the reporter noted, iOS Safari usually transcodes HEIC→JPEG at file selection, so it often never arrives as HEIC — but a genuine `.heic` is now handled when it does.)

## 2. Dynamic requirements hint
The `upload.fileRequirements` hint hardcoded "JPEG, PNG or WebP" regardless of the configured types (that's why the reporter saw `.mov` upload despite the text not listing it). New `extensionsToLabel()` renders the actually-configured, supported formats (e.g. `JPG, PNG, WEBP, MOV`); `fileRequirements` now interpolates `{{formats}}` across all 8 locales. Unsupported extensions are dropped from the label so it never advertises something the backend rejects.

## Deliberately out of scope: DNG / camera RAW
Sharp's libvips has **no raw loader** (input formats: jpeg, png, webp, tiff, gif, svg, heif, vips — nothing for raw/dng). So adding `dng` here would let it upload and then **fail thumbnailing** (`processing_status: 'failed'`, no preview, broken photo in the gallery). Proper RAW support needs embedded-preview extraction (exiftool/libraw) + a Docker dependency — that's its own PR.

## Verification
- `sharp.format` confirms `heif` decodable; vitest covers `extensionsToLabel` + the HEIC mapping (7/7).
- All 8 locale JSONs valid; backend lint clean; no new TS errors introduced (the two pre-existing `general_max_file_size_mb` TS2339s are from #823 and fixed separately in #830).

Follow-up on #823 (main-only) → targets **main**.

### DIFF
--- a/backend/src/routes/gallery.js
+++ b/backend/src/routes/gallery.js
@@ -38,6 +38,24 @@ const {
 } = require('../services/downloadFilenameService');
 const { buildContentDisposition } = require('../utils/filenameSanitizer');
 const { getStorage } = require('../services/storage');
+
+// Formats whose ORIGINAL bytes a browser can't render in an <img> (HEIC/HEIF,
+// camera RAW/DNG). For these the lightbox must be served the generated JPEG
+// preview instead of `url` (the original) — otherwise it shows a broken image.
+// So we force `preview_url` for them regardless of the lightbox_preview_enabled
+// toggle. Detection is by MIME first, extension as a fallback (browsers report
+// these MIMEs inconsistently). EXPERIMENTAL: whether a preview actually renders
+// still depends on the backend being able to decode the source (HEVC-in-HEIC on
+// the prod image; exiftool for DNG) — see #821.
+const NON_DISPLAYABLE_ORIGINAL_EXT = new Set(['heic', 'heif', 'dng']);
+const NON_DISPLAYABLE_ORIGINAL_MIME = new Set(['image/heic', 'image/heif', 'image/x-adobe-dng']);
+function originalNeedsPreview(photo) {
+  const mime = (photo.mime_type || '').toLowerCase();
+  if (NON_DISPLAYABLE_ORIGINAL_MIME.has(mime)) return true;
+  const name = photo.original_filename || photo.filename || '';
+  const ext = name.includes('.') ? name.split('.').pop().toLowerCase() : '';
+  return NON_DISPLAYABLE_ORIGINAL_EXT.has(ext);
+}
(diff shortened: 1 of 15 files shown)


the-luap:
**Codex review applied — 3 rounds.** Updated per the review + a design decision:

- **Round 1 (fixed):** registered `image/heic`/`image/heif` in `ALLOWED_MEDIA_TYPES` (they were rejected by the file validator before Sharp); fixed the admin `PhotoUpload` caller that rendered `{{formats}}`/`{{sizeLimit}}` literally.
- **Round 2 (fixed):** the lightbox served the **original** for HEIC/HEIF/DNG when `lightbox_preview_enabled` is off (the default) → broken image. Now forces `preview_url` for non-browser-displayable originals so the lightbox always gets the generated JPEG. (Covers DNG too — forward-compatible with #833.)

**⚠️ Experimental — HEIC support is not fully verified:**
- **HEVC decode:** `sharp.format.heif.input.file=true` proves the HEIF *container* loader exists, but the registered `fileSuffix` is `.avif`-only and I couldn't confirm the **prod Alpine** prebuilt libvips decodes **HEVC** (what iPhone HEIC uses). Needs a real-iPhone-HEIC test on a beta build. If it can't decode, a HEIC upload fails processing (photo → `failed`) — nothing else regresses.
- **Empty `File.type`:** browsers that don't register HEIC send an empty type → the client guard rejects the file. Extension-based acceptance is a follow-up.

**Known minor nit (not fixed):** the admin upload hint now shows `general_max_file_size_mb`, but the admin path still enforces a hardcoded 10 GB — so for admins the number is cosmetic. Either make the admin path honor the setting or drop the size from the admin hint; deferred as it's a separate design call (admins are intentionally unthrottled today).

Nothing here regresses existing uploads; the HEIC path is additive and gated behind an admin opt-in of the format.

PR #843, opened 2026-07-19
fix(security): read the password-complexity key the settings UI writes

the-luap:
## Summary

The admin's password-complexity setting has never had any effect: the settings UI saves it as `security_password_complexity` (`useSettingsState.ts:340` prefixes `security_` onto the `password_complexity` field), but `getPasswordComplexitySettings()` in `backend/src/utils/passwordValidation.js:127` queried `security_password_complexity_level` — a key nothing ever writes. The read always came back empty, so validation silently ran on the `'moderate'` default regardless of what the admin configured.

One-word fix: read the key the UI writes. No migration needed — nothing ever wrote the `_level` key, so there is no data to migrate.

Credit: spotted in the [filpgame fork](https://github.com/filpgame/picpeak) (their main, 2026-07-14) during a fork sweep.

## Changes

- `backend/src/utils/passwordValidation.js` — query `security_password_complexity` instead of `security_password_complexity_level`.
- `backend/__tests__/utils/passwordValidation.complexityKey.test.js` — regression test: mock DB serves the row exactly as the settings UI writes it (JSON-stringified value under the `security_password_complexity` key) and asserts the configured level (`very_strong`) is returned instead of the default.

## Testing

- `npx jest __tests__/utils/passwordValidation.complexityKey.test.js` — 1 passed.

No UI change — no screenshot.

Note: a twin PR targets `stable` — this bug affects both lines.

### DIFF
--- a/backend/__tests__/utils/passwordValidation.complexityKey.test.js
+++ b/backend/__tests__/utils/passwordValidation.complexityKey.test.js
@@ -1,26 +1,33 @@
 /**
- * Regression test for the password-complexity setting key mismatch.
+ * Regression tests for the password-complexity setting read path.
  *
- * The settings UI saves the admin's choice as `security_password_complexity`
- * (useSettingsState.ts prefixes every security field with `security_`), but
- * getPasswordComplexitySettings() queried `security_password_complexity_level`
- * — a key nothing writes — so the configured level was silently ignored and
- * validation always ran on the 'moderate' default.
+ * Bug 1 (key mismatch): the settings UI saves the admin's choice as
+ * `security_password_complexity` (useSettingsState.ts prefixes every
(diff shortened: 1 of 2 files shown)


the-luap:
Codex review round on this branch surfaced a real gap in the original one-word fix: on **Postgres** (production default) `app_settings.setting_value` is a `json` column, so the driver returns the value already decoded (`very_strong`, no quotes) — the bare `JSON.parse` threw and the outer catch silently fell back to `moderate` again. The fix only worked on SQLite's TEXT shape.

`3828ed7` parses with fallback (mirroring `getAppSetting`'s documented pattern) and the regression test now covers both driver shapes plus the empty-value default. Same commit cherry-picked to the stable twin #844.

PR #844, opened 2026-07-19
fix(security): read the password-complexity key the settings UI writes (stable)

the-luap:
Cherry-pick of #843 for the stable line — same one-word key fix (`security_password_complexity_level` → `security_password_complexity`) plus the regression test. The settings UI writes `security_password_complexity` on stable too (`useSettingsState.ts:231` verified on `origin/stable`), so the admin's complexity choice was equally ignored there. No migration needed — the old key was never written.


the-luap:
Same Postgres json-column gap as found on #843 — fixed here via cherry-pick `973ad71` (parse with fallback + both-driver-shape tests). See the #843 comment for details.

PR #845, opened 2026-07-19
fix(uploads): keep videos when thumbnail generation fails

the-luap:
## Summary

A failing `processUploadedVideo()` (ffmpeg probe + thumbnail extraction) currently loses the video — in both pipeline paths — while the image branch sitting right next to each call already survives its thumbnail failures:

- **Sync path** (`processUploadedPhotos`, `photoProcessor.js:144`): the throw fails the whole upload; the video is gone.
- **Async worker path** (`processPhoto`, `photoProcessor.js:474` — the path real uploads take via `backgroundProcessor`): the throw marks the row `failed`, and the guest gallery only lists `complete` rows — so a fully uploaded video becomes **permanently invisible** whenever ffmpeg chokes on it (exotic pixel formats, HEVC variants, truncated moov atoms…).

## Fix

Mirror the image branch's resilience at both call sites: catch the failure, fall back to `extractVideoMetadata()` alone, and let the video complete without a preview. If even the plain probe fails, keep the video with no metadata — downloadable and streamable either way, just no thumbnail in the grid.

Idea spotted in the [munin92 fork](https://github.com/munin92/picpeak) (2026-07-02, never submitted upstream); reimplemented here to cover both paths.

## Testing

- New regression test: `processUploadedVideo` rejects → row still reaches `processing_status: 'complete'` with metadata from the probe fallback and no `thumbnail_path` (a `failed` row would be invisible to guests).
- `npx jest __tests__/services/photoProcessor.processPhoto.test.js __tests__/services/imageProcessorRaw.test.js` — 10 passed.

No UI change in this PR — no screenshot.

**Known cosmetic caveat:** the grid tile renders `thumbnail_url || url` (`PhotoGrid.tsx:266`), so a rescued video without thumbnail shows a broken-image tile rather than a designed placeholder (the lightbox is fine — it branches on `media_type === 'video'`). Deliberately out of scope here: this backend fix is about not losing the video; a video-placeholder tile for the grid would be a small frontend follow-up.

### DIFF
--- a/backend/__tests__/services/photoProcessor.processPhoto.test.js
+++ b/backend/__tests__/services/photoProcessor.processPhoto.test.js
@@ -71,6 +71,7 @@ jest.mock('../../src/services/imageProcessor', () => {
   const mockExtractCaptureDate = jest.fn();
   return {
     generateThumbnail: mockGenerateThumbnail,
+    generateVideoPlaceholder: jest.fn(async (filename) => `thumbnails/thumb_${filename.replace(/\.[^.]+$/, '')}.jpg`),
     extractCaptureDate: mockExtractCaptureDate,
     withLocalCopy: jest.fn(async (key, fn) =>
       fn(`/tmp/local-copy-${require('path').basename(key)}`)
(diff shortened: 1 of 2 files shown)


the-luap:
Codex review (2 rounds): round 1 sharpened the PR-body caveat into a real finding — a rescued video with NULL thumbnail made the grid fetch the **original video file** as an `` blob (`thumbnail_url || url`), a potentially multi-GB download for a broken tile. `42214af` fixes this: both fallback paths now generate the existing sharp-rendered play-button placeholder (`generateVideoPlaceholder`, ffmpeg-free), so rescued videos get a real tile. Round 2's residual concern (placeholder generation itself failing → pre-PR baseline behavior) was withdrawn after discussion — failing the row instead would reverse the rescue semantics. The frontend caveat paragraph in the PR body is thereby resolved.

PR #846, opened 2026-07-19
fix(file-watcher): bound concurrent photo processing

the-luap:
## Summary

The file watcher processes photos with unbounded concurrency. chokidar fires `add` once per file — and since our watcher sets no `ignoreInitial` option, the **boot scan fires it for every existing file** under `events/active`; a bulk drop into the watch folder fires it for every new file at once. Each handler runs DB lookups plus (for new files) a full sharp pipeline. `sharp.concurrency(2)` only caps libvips threads *within* one operation, not the number of parallel pipelines — the OOM vector on small hosts is many concurrent decode buffers.

## Fix

Gate the watcher through a shared `p-limit` (default 2, `FILE_WATCHER_CONCURRENCY` env, floored at 1):

- `add` → `processNewPhoto` — the sharp/DB-heavy path.
- `unlink` → `removePhoto` — included too (deviation from the original): mass deletes otherwise burst DB work and ZIP-cache invalidation the same way.

`p-limit` is pinned to `^3.1.0` — deliberately the last CommonJS release (4+ is ESM-only).

Adapted from the [filpgame fork](https://github.com/filpgame/picpeak) (`426ca491`, 2026-07-14) — thanks @filpgame. Changes vs. the original: `unlink` shares the limiter, the env var is documented in `backend/.env.example`, and no `processNewPhoto` export was added (our tests don't need it).

## Trade-off

The boot-time re-scan of a large existing library now runs its per-file dedupe checks at 2 in parallel instead of all at once — slower to complete, deliberately so. Raise `FILE_WATCHER_CONCURRENCY` on hosts with headroom.

## Testing

- `fileWatcher.concurrency.test.js` — limiter config matrix (default / explicit / floor / invalid), `add` and `unlink` both routed through the shared limiter, handler failures logged instead of unhandled.
- `imageProcessor.sharpConfig.test.js` — locks the existing `sharp.cache(false)` + `sharp.concurrency(2)` caps this bound relies on (also from the fork commit).
- `npx jest` on both suites — 9 passed. Pre-push local E2E smoke — 13 passed.

No UI change — no screenshot.

Note (pre-existing, untouched): the chokidar `ignored` regex carries a `no-useless-escape` ESLint error that exists on `main` already.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -106,6 +106,11 @@ VITE_API_URL=/api
 # DB_PORT=5432
 # REDIS_PORT=6379
 
+# File watcher (watch-folder auto-import, local storage only)
+# Max photos processed in parallel — raise on hosts with memory headroom,
+# lower to 1 on very small hosts. Default: 2
+# FILE_WATCHER_CONCURRENCY=2
+
 # Release Channel
 # Options: 'stable' (default), 'beta', or specific version like 'v2.3.0'
 # 'stable' uses the :stable tag (same as :latest on main)
--- a/backend/.env.example
+++ b/backend/.env.example
@@ -106,6 +106,12 @@ ARCHIVE_PATH=/app/storage/events/archived
 # EVENTS_PATH=./storage/events
 # ARCHIVE_PATH=./storage/events/archived
 
+# File watcher (auto-import from the events/active folder, local storage only)
+# Max photos processed in parallel by the watcher. The boot scan and bulk
+# folder drops fire one handler per file — this bound keeps thumbnail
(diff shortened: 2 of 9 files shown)


the-luap:
Codex review (2 rounds): round 1 found that the documented `FILE_WATCHER_CONCURRENCY` override never reached the container — the backend service in both compose files uses an explicit `environment` list (no `env_file`), so a `.env` value was silently ignored. `b6f9a25` passes it through in `docker-compose.yml` + `docker-compose.production.yml` and documents it in the root `.env.example`. Round 2: converged, no findings.

PR #847, opened 2026-07-19
feat(events): gallery QR code + printable table-card/poster PDFs

the-luap:
## Summary

Closes #836 — every guest-photo competitor offers QR access as the primary way guests reach a gallery; PicPeak had no gallery QR at all (`qrcode` was only used for MFA enrollment and Swiss QR-bills).

## What's in

**Backend** — new `adminEvents/qr.js` sub-module (registered after `logo` in `adminEvents/index.js`; `/:id/...` literals, so registration order is not sensitive):

- `GET /api/admin/events/:id/qr` — QR for the event's share URL (via `getEventShareToken` + `buildShareLinkVariants`, so short-URL mode is respected). `?format=png|svg`, `?size=128..2048` (PNG), `?download=1` toggles attachment disposition. Guarded by `adminAuth` + `events.view` + `requireEventOwnership`.
- `GET /api/admin/events/:id/qr-print?template=table-card|poster&lang=xx` — print-ready PDF built with pdfkit (already a dependency): A6 table card / A4 poster with event name, centered QR, localized caption (all 8 gallery locales) and the share URL as a footer. PDFs render with bundled complete IBM Plex Sans TTFs (Latin + Cyrillic + Greek, OFL license alongside) — Cyrillic event names and captions render properly; the QR target is the stored `share_link` (single source of truth with the card) with an `?origin=` fallback for unconfigured installs.
- `409` when the event has no share link, `404` for unknown events.

**Frontend** — QR section in the `ShareLinkCard` on the event detail page: live QR preview (fetched as a blob — the admin API uses Bearer auth, so a plain `` would 401) plus four downloads: PNG (1024px), SVG, table card PDF, poster PDF. The print language follows the admin's UI language. New i18n keys in all 8 locales.

## Screenshots

Event detail — share-link card with the new QR section (DE):

QR section

Table card (A6) and poster (A4) PDFs render with title, QR, caption, URL footer — verified by downloading both from the dev stack and inspecting the output.

## Testing

- `adminEventQr.test.js` (real-SQLite harness, same as `adminEvents.smoke.test.js`): 401 without token, PNG magic bytes + content type, SVG output, attachment disposition, both PDF templates render `%PDF`, 409 without share link, 404 unknown event.
- `tsc --noEmit` clean; ESLint clean on all touched files.
- Pre-push local E2E smoke — 13 passed.

PR #848, opened 2026-07-19
feat(slideshow): guest-scannable share-link QR overlay

### LINKED ISSUE #837: [FEATURE] Live slideshow: guest-upload QR code overlay
## Motivation

shared.click's live wall permanently shows an upload QR code as an overlay on the projected slideshow, so guests can scan straight off the screen and their photos appear moments later. That feedback loop is the strongest driver for guest uploads during an event.

PicPeak's live slideshow already supports transitions, color filters, image-fit modes and a logo watermark overlay — but no QR overlay. Since watermark overlay positioning already exists, this is a natural extension.

## Scope

- New slideshow setting: **QR overlay** (off by default).
  - Encodes the gallery/guest-upload URL (share-link token aware).
  - Position (corners), size, opacity — same option shape as the existing slideshow watermark settings.
  - Optional short caption line under the QR (e.g. "Scan to add your photos"), localized.
- Available both as global slideshow setting and per-event override, matching the existing global + per-event override pattern.
- Only offered/rendered when the event has guest uploads enabled (or gallery viewing makes sense for guests).

## Non-goals

- Animated/rotating call-to-action banners.
- QR for anything other than the gallery/upload URL.

## Implementation notes

- Backend: extend slideshow settings shape (`getSlideshowGlobals`, per-event slideshow overrides in `adminFeatureFlags.js` / settings).
- Frontend: render QR client-side in the slideshow view (corner overlay, above the photo layer, below nothing else); reuse positioning logic from the slideshow watermark.

Depends on / related to: gallery QR code generation (#836) — the QR payload logic should be shared.

the-luap:
## Summary

Closes #837 — the live slideshow can now show the gallery share link as a QR code, so guests scan it straight off the beamer and their photos land in the gallery moments later (the shared.click "live wall" pattern). Independent of #847 — no shared code; the QR payload here is generated server-side in the state endpoint.

## Design

Mirrors the watermark cascade exactly:

- **Global look** (Settings → Slideshow): `slideshow_qr_enabled` / `position` / `opacity` / `size`, validated in the existing `PUT /admin/settings/slideshow` whitelist, cached via `getSlideshowGlobals()` (same 5s TTL + invalidate-on-write).
- **Per-event tri-state** `show_qr` (migration 163, NULL = inherit) — editable in the per-event slideshow card next to the watermark mode, `PATCH /:id/slideshow`.
- **Delivery:** the `/show/:token/state` payload carries the QR as a **PNG data URI** (`qr: { data_url, position, opacity, size }`), generated with the existing `qrcode` dependency and cached per share URL — the 3s projector poll never re-encodes, and the public kiosk needs no QR library and no extra authenticated request.
- **Rendering:** corner-positioned like the watermark, in a white padded rounded box so the code stays scannable on any photo.

## Screenshot

Running slideshow with the QR overlay (bottom-left, e2e test image as slide content):

Slideshow QR overlay

## Deviation from the issue

The issue suggested offering the overlay only when guest uploads are enabled. I didn't hard-gate: a QR to *view* the gallery is just as useful for view-only events, and the admin toggle already carries the decision. Say the word if you want the coupling.

## Testing

- `slideshowPublic.test.js` +4: inherits global when `show_qr` NULL (position/opacity/size + `data:image/png` prefix), null by default, per-event OFF beats global ON, per-event ON beats global OFF — 23/23 on the branch.
- Verified live on the dev stack: enabled the global setting, generated a slideshow link, confirmed the overlay renders and the QR encodes the share URL.
- `tsc --noEmit` clean; i18n en + de (the slideshow namespace has no other locales yet, matching the watermark keys).

### DIFF
--- a/backend/__tests__/routes/slideshowPublic.test.js
+++ b/backend/__tests__/routes/slideshowPublic.test.js
@@ -99,7 +99,12 @@ describe('public Live Slideshow routes', () => {
     await setFlag(db, 'slideshow', true);
   });
 
-  const stateUrl = (token = TOKEN) => `/api/gallery/${SLUG}/show/${token}/state`;
+  // QR overlay: supertest's Host is loopback, and a loopback base is now
+  // suppressed rather than encoded — the kiosk passes its reachable
+  // window.location.origin, so the QR tests do the same.
+  const KIOSK_ORIGIN = 'https://gallery.example.com';
+  const stateUrl = (token = TOKEN) => `/api/gallery/${SLUG}/show/${token}/state?origin=${encodeURIComponent(KIOSK_ORIGIN)}`;
+  const stateUrlNoOrigin = (token = TOKEN) => `/api/gallery/${SLUG}/show/${token}/state`;
 
   describe('resolveSlideshow guards', () => {
     it('200 + per-event display settings on a live link', async () => {
@@ -269,6 +274,14 @@ describe('public Live Slideshow routes', () => {
       expect(res.body.qr.data_url).toMatch(/^data:image\/png;base64,/);
       // Look falls back to the global defaults.
       expect(res.body.qr.position).toBe('bottom-left');
+    });
+
+    it('suppresses the QR when no guest-reachable origin exists (loopback base, no kiosk origin)', async () => {
+      await insertEvent(db, { show_qr: 1 });
+      const res = await request(app).get(stateUrlNoOrigin());
+      // Encoding localhost would send scanning phones to THEIR localhost —
+      // no QR beats a broken QR (codex review of #848, confirmation round).
+      expect(res.body.qr).toBeNull();
     });
   });
 
--- a/backend/__tests__/routes/slideshowPublic.test.js
+++ b/backend/__tests__/routes/slideshowPublic.test.js
@@ -228,6 +228,50 @@ describe('public Live Slideshow routes', () => {
     });
   });
 
+  describe('slideshowSettings — QR overlay cascade (#837)', () => {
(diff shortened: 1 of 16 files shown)


the-luap:
Codex review (3 rounds, stopped at cap with all findings fixed):
- **R1:** stored per-event `show_qr` override was silently reset on save (OverviewTab/Event type never carried it); QR hidden exactly when an empty gallery needs it most (now renders for any running show); unbounded QR cache (insertion-order eviction at 50).
- **R2:** compose-default `FRONTEND_URL=localhost` sent scanning phones to *their* localhost — server now falls back to a reachable origin.
- **R3:** sharpened R2 — `req.get('host')` loses the port behind the nginx proxy (`$host`), so the kiosk now passes `window.location.origin` with its session/state calls (validated, same pattern as #847), Host-derived origin as second fallback.
23/23 slideshow tests green throughout.

the-luap:
Confirmation rounds (2 more after the cap): **round 4** — loopback kiosk origins are rejected and the overlay is suppressed when no guest-reachable URL exists (new test); QR cache re-keyed per event with a 60s regeneration throttle against origin-driven CPU exhaustion. **Round 5** — the throttle itself enabled cache poisoning (a token holder's attacker-origin QR was served to the legitimate projector for the rest of the window): mismatched cached artifacts are never served now, and concurrent cold-cache polls share a single in-flight encode. One finding **rejected with evidence**: the claimed loopback-regex port bug — the regex matches `localhost:3000`/`127.0.0.1:port` (verified empirically; the suppression test runs green). 24/24 tests.

Issue #799, opened 2026-07-14
[FEATURE] Microservice deployments


eopo:
**Is your feature request related to a problem? Please describe.**
Currently, PicPeak Backend runs as monolith container, although it contains various workers with very different workload profiles. Whilst for small to medium sized instances API worker would most probably not hold a lot of load, a thumbnail generation workload would be ideal for Kubernetes Autoscaling

**Describe the solution you'd like**
The image could stay as a monolith with capabilities for all workers, however, using env var only certain workers could be activated or vice versa, an ENV could deactivate certain workers, so containers only do a chosen subset of tasks.

**Describe alternatives you've considered**
Custom Image builds

**Use Case**
DevOps with smart Cluster deployments for efficient resource usage
Advanced security (only allow SMTP enabled pod to communicate on those ports, lock down other pods)
Run resource-heavy tasks like thumbnail generation or video rendering on other nodes than easy tasks like mail or API.

**Similar Features**
- OCIS / Opencloud

**Implementation Ideas**
An exemplary syntax could be `ENABLED_WORKERS=smtp,api` or `DISABLED_WORKERS=thumbnails`

the-luap:
interesting point. In the beginning I got even the vice versa comments that it should be a big monolit container without splitting into front and backend. I will think about the idea to put workers.

eopo:
I think the versatility of a monolith image, which can be deployed either as monolith or as microservices is the best of two worlds. Someone using compose to run it on a pi or whatever can have the simplicity of just one container, someone having a refined cluster setup can tweak it the way they need.

the-luap:
I went through the codebase to scope what your `ENABLED_WORKERS` proposal would actually take. Sharing the full breakdown for transparency, plus my decision at the end.

## Current state

The architecture is closer to your proposal than it looks — everything is already DB-coordinated:

- All "queues" are DB-polling loops, no in-memory queue infra: `email_queue` (60s poll), `whatsapp_queue` (30s), `webhook_deliveries` (5s), and photo processing via `photos.processing_status='pending'` with an atomic claim + stuck-job janitor (`backgroundProcessor.js`). No BullMQ/Redis queues anywhere.
- All workers start unconditionally in `startServer()` (`server.js:878–1025`): file watcher, expiration checker, invoice/workflow scheduler, email/WhatsApp/webhook/IMAP pollers, S3 auto-importer, two backup crons, photo processor pool.
[...]

PR #849, opened 2026-07-19
feat(notifications): surface guest activity in the admin bell

### LINKED ISSUE #746: [FEATURE] Client-activity notifications (gallery opened / favorited / downloaded)
Notify the photographer on "gallery opened / client favorited / downloaded." PicPeak has email + webhooks but no unified activity-notification surface.

- [ ] Notify on gallery open / favorite / download

Effort: S — mostly wiring existing events into a notification surface. Good first candidate.

_Parked idea — no timeline._

the-luap:
## Summary

Closes #746 — "notify the photographer on gallery opened / client favorited / downloaded". The unified notification surface already exists: the admin bell reads `activity_logs` (`adminNotifications.js`), and favorites already flow into it via `feedbackService`. What was missing: gallery opens and downloads only landed in `access_logs` — analytics-visible, bell-invisible. This PR wires them in; no new tables, no new settings, no email (a digest can be a follow-up if wanted).

## What's in

**Backend (`gallery.js`):**
- `gallery_opened` on the guest photo-list route — **debounced in-memory to one notification per event per 6h**: the endpoint fires on every gallery page load, so per-hit notifications would spam the bell. Losing the map on restart merely allows one extra notification; the hot view path pays zero DB reads. Slideshow traffic stays excluded, matching the existing analytics exclusion (migration 138).
- `gallery_downloaded` on all four download paths (streamed / pre-zipped / presigned `download-all`, `download-selected`) with `{ scope, photo_count }` metadata. No debounce — downloads are rare, high-signal events.

**Frontend:**
- Locale entries `galleryOpened` / `galleryDownloaded` in all 8 languages, resolved through the existing smart camelCase fallback in `formatNotificationMessage` — no new switch cases. Also added **`photoFavorite`**: `photo_favorite` has been logged since the feedback system landed but fell through to the generic "system activity" line; now it renders properly ("A guest favorited a photo in …").
- Distinct bell styles per type (Eye / Download / Heart).

## Screenshot

Admin bell after a guest opened the gallery and downloaded photos (DE):

Notification bell

## Testing

- `notifications.activity.test.ts` — the three types resolve via real locale keys (no generic fallback) with the event name interpolated, plus the style mapping. 4/4.
- End-to-end on the dev stack: fresh guest share-login → photo list → `gallery_opened` row; `download-all` → `gallery_downloaded` row; both render localized in the bell (see screenshot).
- The backend hooks are two `logActivity` calls beside existing `access_logs` inserts plus a debounce helper — honest note: no backend unit test for the debounce (the helper is module-internal); verified live instead.
- `tsc --noEmit` + ESLint clean. Pre-push local E2E smoke — 13 passed.

Webhook event types (`gallery.opened` etc.) were deliberately left out to keep this small — say the word and I'll add them to the webhook registry as a follow-up.


the-luap:
Codex review (3 rounds, stopped at cap with fixes applied):
- **R1:** the most common download path (single-photo Save) never reached the bell — now emits debounced (1h/event) `gallery_downloaded`; the per-type icons existed but AdminHeader hard-coded Bell — icon map wired.
- **R2:** notify only after the response finishes (<400) so failed downloads neither false-report nor burn the debounce window; icon map completed. **Deliberately rejected:** DB-backed debounce for multi-worker deployments — the backend's deployment contract is single-process (see #799 analysis); worst case under future replicas is N notifications per window, which degrades rather than breaks.
- **R3:** client sessions are attributed as `customer` instead of mislabeled `guest` (#746 explicitly covers client activity); cached-ZIP path logs on `finish` like the single-photo path; Trash2 added to the icon map.
4/4 notification tests + tsc green.

the-luap:
Confirmation rounds (2 more after the cap): **round 4** — dashboard Recent Activity rendered raw type identifiers (`admin.activities.*` keys added ×8 locales); customer-portal opens no longer double-notify; `gallery_downloaded` wording is actor-aware; both archiver ZIP paths notify on response `finish`. **Round 5** caught that my portal dedup was **inverted** — portal tokens run as `guest` (`via:'customer'`, no accessLevel) while `accessLevel:'client'` is the PIN flow with no other signal: `verifyGalleryAccess` now surfaces `req.viaCustomer`, dedup keys on that, and attribution treats both customer shapes as `customer`. Favorite wording is actor-neutral (feedbackService logs favorites without an actor). tsc + 4/4 tests green.

PR #850, opened 2026-07-19
test(crm): mint-path integration tests — quote send, invoice storno, contract countersign

### LINKED ISSUE #587: [FEATURE] CRM mint-path integration tests (#570 follow-up)
## Context

PR #586 closed the original #570 by adding **HTTP-layer route tests** (auth, permissions, validation, token-guard outcomes) for every CRM admin + public route. Combined with the pre-existing **service-layer unit tests** under \`backend/__tests__/services/\`, the CRM module now has solid coverage at both endpoints of the call stack.

The remaining gap: **integration tests that exercise the full chain** — HTTP request → route handler → service → DB writes → token mints → email queue → file-on-disk. A regression in *how the two halves wire together* (e.g., route handler passing the wrong \`adminId\`, feature-flag check ordered before vs. after permission check, scheduled\\_send\\_at not flowing into the email queue) would slip past both existing test layers.

## Scope

Three highest-value mint-and-send paths, ~6-9 tests total:

### 1. \`POST /api/admin/quotes/:id/send\`
- 200 with valid super-admin + quote in \`draft\` status
- Assert: \`quotes.status\` → \`sent\`, \`quotes.sent_at\` populated, PDF file exists on storage path, \`quote_action_tokens\` row inserted with the right \`quote_id\` + future \`expires_at\`, \`email_queue\` row enqueued with \`email_type='quote_sent'\` and the customer's primary email as \`to\`.
- Negative: 400 when quote is already \`sent\` (re-send must go through a separate path).

### 2. \`POST /api/admin/invoices/:id/storno\`
- 201 with valid super-admin + invoice in \`sent\` / \`paid\` status
- Assert: original \`invoices.status\` → \`cancelled\`, a NEW invoice row exists with \`kind='storno'\`, \`cancels_invoice_id\` pointing at the original, same \`deal_uuid\` carried forward, negated totals, fresh sequence-claimed \`invoice_number\`, original row's \`cancellation_storno_id\` populated.
- Negative: 409 when the invoice is already \`cancelled\` (no double-Storno).

### 3. \`POST /api/admin/contracts/:id/countersign\`
- 200 with valid super-admin + contract in \`signed\` status
- Assert: \`contracts.status\` → \`fully_signed\`, \`contracts.integrity_hash\` populated with a 64-hex SHA-256 of the rebuilt PDF, new admin signature persisted (path on storage), \`email_queue\` row enqueued with \`email_type='contract_fully_signed'\` + customer primary email as \`to\` + signed PDF attached.

## Test harness

Build on the existing \`__tests__/integration/helpers/crmDb.js\` extensions from PR #586 (mintAdminToken, buildRouteApp, assignAdminRole, etc.). New helper needed: \`stubEmailProcessor()\` that no-ops \`emailProcessor.initializeTransporter\` and stubs \`sendQueuedEmail\` so the email-queue row write happens but no SMTP attempt is made. Same for \`stubPdfGeneration()\` if PDF rendering is too slow under the test timeout — though ideally the test asserts the real PDF lands on disk, since "PDF wasn't actually written" is exactly the integration regression worth catching.

## Acceptance

- Each mint path covered by at least one happy-path test asserting **every** observable outcome (DB state + file system + email queue), not just the HTTP status code.
- Each mint path covered by at least one negative test asserting the right idempotency / status-guard behaviour.
- Tests run under the existing Jest config without new infrastructure (no test SMTP server, no real PDF rendering against external fonts that might not be installed).

## Out of scope

- End-to-end lineage walk (quote → contract → invoice → Storno across all three services in one test). Belongs as a separate \`crmLineage.integration.test.js\` if desired — that's a different test shape (long-running, asserts cross-service state).
- Email rendering correctness (template content, i18n fallback) — that's a template-rendering test, separate concern.
- Frontend Vitest coverage for the CRM pages — separate follow-up if anyone wants to file it.

## Priority

Low-medium. The marginal value over the existing two-layer coverage is real but narrow — most mint-path bugs would surface in either the service or the HTTP layer tests we now have. File primarily so it's tracked; don't block other work for it.

the-luap:
## Summary

Closes #587 — integration tests for the three highest-value CRM mint-and-send paths, exercising the full chain the existing route- and service-layer tests can't see: HTTP request → route handler → service → DB writes → token mints → email queue → file on disk. 7 tests, real full-migration SQLite (`helpers/crmDb`), real pdfkit/pdf-lib rendering, no mock-fs, no network — and **not** in the CI exclude list (verified: runs green under the plain `npx jest` CI invocation, ~15s).

## Covered

1. **`POST /api/admin/quotes/:id/send`** — draft → `sent` + `sent_at`, PDF exists on disk (size > 0), `quote_action_tokens` row with future expiry, `quote_sent` email to the customer's primary address with the right `quote_number`. Negative: re-send → 409.
2. **`POST /api/admin/invoices/:id/cancel`** (Storno mint) — original → `cancelled` + back-linked `cancellation_storno_id`; new row with `kind='storno'`, `cancels_invoice_id`, same `deal_uuid`, negated net/VAT/total, freshly sequenced number, line items snapshotted. Both `sent` and `paid` originals. Negative: double-cancel → 409 `ALREADY_CANCELLED`.
3. **`POST /api/admin/contracts/:id/countersign`** — `signed_by_customer` → `fully_signed`, admin signature asset persisted, stamped PDF written and its stored `signed_pdf_sha256` **re-hashed against the actual file**, `contract_fully_signed` email to the customer with the `-signed.pdf` attachment. Negative: draft → 409.

## Deviations from the issue spec (tests pin the code)

| Spec said | Code does |
|---|---|
| `POST /:id/storno`, 201 | `POST /:id/cancel`, 200 `{ cancelled, stornoId }` |
| re-send → 400 | 409 |
| contract status `signed` | `signed_by_customer` (countersign also accepts `sent` → `signed_by_admin`) |
| `integrity_hash` | `pdf_sha256` / `signed_pdf_sha256` |
| PDFs under `STORAGE_PATH` | `process.cwd()/storage/business-docs/…` (test isolates via chdir) |

## Bug found along the way (pre-existing, not fixed here)

`createContract` and `createStorno` call `logActivity()` **inside** a knex transaction without passing the trx as executor — exactly the pattern `db.js:648`'s comment warns about. On single-connection SQLite the audit insert deadlocks against the open trx, stalling each call for the full 60s acquire timeout before `logActivity`'s catch swallows the failure (audit row silently lost). Postgres deployments are unaffected. The test shrinks the pool acquire timeout to keep the suite fast; the fix (pass the trx) deserves its own small PR — say the word.

## Testing

- `npx jest __tests__/integration/crmMintPaths.test.js` — 7/7, verified in two independent environments (agent worktree + main checkout). ESLint clean. Pre-push local E2E smoke — 13 passed.

### DIFF
--- a/backend/__tests__/integration/crmMintPaths.test.js
+++ b/backend/__tests__/integration/crmMintPaths.test.js
@@ -43,6 +43,26 @@ const toMillis = (v) => {
 
 const sha256 = (buf) => crypto.createHash('sha256').update(buf).digest('hex');
 
+// Count embedded image XObjects per page via pdf-lib — used to prove BOTH
+// signature stamps (customer + admin) made it into the final PDF instead of
+// only asserting file existence/hash (codex review of #850 round 2).
+async function countImagesPerPage(pdfPath) {
+  const { PDFDocument, PDFName, PDFDict } = require('pdf-lib');
+  const doc = await PDFDocument.load(fs.readFileSync(pdfPath));
+  return doc.getPages().map((page) => {
+    const resources = page.node.Resources();
+    const xobjects = resources && resources.lookupMaybe(PDFName.of('XObject'), PDFDict);
+    if (!xobjects) return 0;
+    let images = 0;
+    for (const [, ref] of xobjects.entries()) {
+      const stream = page.doc.context.lookup(ref);
+      const subtype = stream && stream.dict && stream.dict.get(PDFName.of('Subtype'));
+      if (subtype && subtype.toString() === '/Image') images += 1;
+    }
+    return images;
+  });
+}
+
 let db;
 let cleanup;
 let tmpDir;
@@ -353,6 +373,13 @@ describe('POST /api/admin/contracts/:id/countersign', () => {
     expect(fs.existsSync(row.signed_pdf_path)).toBe(true);
     expect(row.signed_pdf_sha256).toMatch(/^[0-9a-f]{64}$/);
     expect(sha256(fs.readFileSync(row.signed_pdf_path))).toBe(row.signed_pdf_sha256);
+
+    // BOTH stamps must be embedded in the final document — a regression
+    // stamping the admin onto the unsigned base PDF would keep every
+    // path/hash assertion above green (codex review of #850 round 2).
+    const imagesPerPage = await countImagesPerPage(row.signed_pdf_path);
+    const maxImagesOnAPage = Math.max(...imagesPerPage);
+    expect(maxImagesOnAPage).toBeGreaterThanOrEqual(2);
 
     // contract_fully_signed email to the customer's primary address,
     // carrying the signed PDF as attachment (plus the audit cert).
--- a/backend/__tests__/integration/crmMintPaths.test.js
+++ b/backend/__tests__/integration/crmMintPaths.test.js
@@ -1,0 +1,366 @@
+/**
+ * CRM mint-and-send paths — integration tests (#587).
+ *
+ * Pins the three document "mint" flows end-to-end through the real
+ * HTTP → route → service → DB → email-queue → file pipeline:
+ *
+ *   1. POST /api/admin/quotes/:id/send        (draft → sent + PDF + token + email)
+ *   2. POST /api/admin/invoices/:id/cancel    (issued → cancelled + Storno row)
+ *      — the issue spec named this /:id/storno; the real route is
+ *        /:id/cancel (invoiceService.cancelInvoice → createStorno).
+ *   3. POST /api/admin/contracts/:id/countersign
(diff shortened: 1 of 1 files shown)


the-luap:
Codex review (3 rounds, **converged**): round 1 asked to pin the two legs the suite's own report had flagged as uncovered — sendStorno side effects (status/PDF/queued email, since cancelInvoice swallows delivery failures by design) and the real customer-sign flow instead of a status UPDATE. Round 2 pushed further: existence/hash checks can't catch admin-stamps-onto-wrong-base regressions, so the suite now counts embedded image XObjects via pdf-lib and asserts the signature page carries **both** stamps. Round 3: no findings. 7/7 green.

PR #851, opened 2026-07-19
fix(crm): pass trx to logActivity inside transactions — audit rows silently lost on SQLite

the-luap:
## Summary

Fixes the pre-existing bug surfaced by the #587 integration-test work (#850): four CRM write paths call `logActivity()` from **inside** a knex transaction without passing the trx as executor — the exact pattern the comment at `db.js:648` warns about. On single-connection SQLite the audit insert waits on a second pool connection while the transaction holds the only one: **a 60s acquire-timeout stall per call**, after which `logActivity`'s catch swallows the failure and **the audit row is silently lost**. Postgres deployments are unaffected (pool > 1), which is why this never showed up in production reports — every SQLite install has been paying the stall + losing these audit entries.

## Affected call sites (all four fixed)

| Site | Activity type |
|---|---|
| `contract/crud.js` `createContract` | `contract_created` — also resolved `adminActor()` (a global-db read) inside the trx: **two** deadlocking awaits |
| `contract/crud.js` `updateContract` | `contract_updated` — same double pattern |
| `invoice/sending.js` `createStorno` | `invoice_cancelled_via_storno` |
| `invoice/sending.js` `reissueInvoice` | `invoice_reissued` |

The fix mirrors the one call site that already did it right (`contract_created_from_quote` in `conversions.js`, which carries the explanatory comment): resolve the audit actor **before** the transaction opens, and pass `trx` as `logActivity`'s executor so the insert rides the transaction's connection. Side effect worth naming: the audit row now commits/rolls back atomically with the business row — no more audit entries for rolled-back documents.

**Verified not affected** (checked each site's context — `logActivity` runs outside any transaction, unchanged): `cancelContract`, `contract_converted_to_event`, `contract_signed_by_customer`, `contract_sent`, `invoice_sent`, draft-path `invoice_cancelled`, `invoice_released_for_delivery`, `monthly_bill_triggered_manually`.

## Testing

- New `logActivityTrx.test.js` (full-migration SQLite via `helpers/crmDb`): `createContract`, `updateContract` and `cancelInvoice`-Storno each assert the **audit row now exists** with correct metadata/actor — previously silently lost. The suite shrinks the pool acquire timeout to 2s so a reintroduced deadlock fails fast instead of appearing to pass after a stall. 3/3.
- `reissueInvoice` shares the identical one-line pattern; not separately harnessed (needs a full cancel-then-reissue setup for one more identical assertion).
- Pre-existing `no-empty` ESLint errors on the file's `catch (_) {}` blocks are untouched (8 on `main`, 8 here).

Once this and #850 both land, the `acquireTimeoutMillis` workaround comment in `crmMintPaths.test.js` can be simplified — happy to do that as a follow-up nit in whichever lands second.

### DIFF
--- a/backend/__tests__/services/logActivityTrx.test.js
+++ b/backend/__tests__/services/logActivityTrx.test.js
@@ -1,0 +1,125 @@
+/**
+ * Regression tests for logActivity calls inside transactions (#850 review
+ * find). createContract / updateContract / createStorno / reissueInvoice
+ * called logActivity() (and contract paths also adminActor()) from inside
+ * a knex transaction WITHOUT the trx executor. On single-connection SQLite
+ * the audit insert then waits on a second pool connection while the trx
+ * holds the only one — a 60s acquire-timeout stall per call, after which
+ * logActivity's catch swallows the failure and the audit row is silently
+ * lost. Postgres was unaffected.
+ *
+ * The observable fix: the activity_logs rows now exist, and the calls
+ * complete without waiting on the pool. The shrunken acquire timeout
+ * below makes any reintroduced deadlock fail the test quickly instead
+ * of appearing to pass after a long stall.
+ */
+const path = require('path');
+const {
+  bootCrmDb, seedMinimal, assignAdminRole,
+} = require('../integration/helpers/crmDb');
+
+jest.setTimeout(120000);
+
+let db;
+let cleanup;
+let tmpDir;
+let adminId;
+let customerId;
+let contractService;
+let invoiceService;
+
+const prevCwd = process.cwd();
+
+beforeAll(async () => {
+  ({ db, cleanup, tmpDir } = await bootCrmDb());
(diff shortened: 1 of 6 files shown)


the-luap:
Codex review (3 rounds, **converged**): round 1 found the fix was moot for `reissueInvoice` — `createInvoice`'s internal global-connection reads (getProfile, getAppSetting, bank resolution) deadlocked the outer trx before the audit line was ever reached, aborting the replacement after the Storno had committed. Round 2 rightly rejected my first answer (dropping the transaction traded the deadlock for orphan drafts + burnt sequence numbers on validation failures). Final shape: the transaction is back and every read inside createInvoice rides it — getProfile/resolveBankAccountForCurrency gained optional conn params, getAppSetting calls pass trx (incl. the resolveNetDays default the new regression test flushed out), audit uses the trx executor. Round 3: no findings. The reissue regression test now proves a full cancel+reissue commits atomically on single-connection SQLite (4/4).

PR #863, opened 2026-07-23
fix(gallery): block password form in Instagram in-app browser and unmask login errors

### LINKED ISSUE #654: [BUG] The instagram webview says "Incorrect Password" when trying to access an password protected Album
**Describe the bug**
The instagram webview (android/ios) says "Incorrect Password" when trying to access an password protected Album. 

**To Reproduce**
1. Go to Instagram
2. Click on the Album Link
3. Enter correct password
4. See error "Incorrect Password"

**Expected behavior**
The correct password should not be incorrect.

- Instagram App (Android/IOS)

**Additional context**
I just wanted to post this here, if it can be fixed it would be awesome :D. If not, that's no problem either, since Instagram is weird.

the-luap:
Follow-up to #656 for #654 — the reporter confirms password login **still** fails inside Instagram's in-app browser with the input-attribute/trim defenses shipped in v3.45.0, and asked for the warning to become blocking. This PR does that, and fixes two error-reporting paths that were masking the real failure as "incorrect password".

## 1. Instagram IAB: blocking state instead of advisory banner

`GalleryPage.tsx` — when `detectInAppBrowser()` matches the Instagram UA, the password form is now **hidden** and replaced with a red alert: platform-specific "open in external browser" instructions (⋯ on iOS / ⋮ on Android) plus a **copy-link button** (async clipboard API, `execCommand` textarea fallback for webviews that block it). A small "try entering the password here anyway" link restores the form — escape hatch for a UA false positive, or for the day Instagram fixes their webview. The old amber banner and its i18n keys are removed (superseded; they only existed in EN/DE).

## 2. Unmask non-password login failures

Two paths in the login error handler rendered as "Incorrect password" while having nothing to do with the password — exactly the silent-failure shape reported on #654:

- **No response at all** (offline, proxy/webview killed the request): `error.response?.data?.error || 'Invalid password'` defaulted the message to the wrong-password branch. Now reports a connection error (`auth.networkError`).
- **reCAPTCHA 400** (`auth.js:383`): the widget regularly fails to load inside in-app webviews; the rejection fell into the generic "Invalid password" branch. Now reports a verification failure (`auth.recaptchaFailed`).

This matters beyond Instagram: any instance with reCAPTCHA enabled was telling IAB guests their correct password was wrong.

## 3. Invisible-Unicode compare fallback (server-side)

Passwords relayed through chat apps (Instagram DMs/bios especially) pick up invisible characters on copy-paste: zero-width space/joiners (U+200B–U+200D), word joiner (U+2060), BOM (U+FEFF), soft hyphen (U+00AD). `.trim()` only handles the string edges, so mid-string invisibles survived and failed the byte-exact bcrypt compare.

`gallery/verify` now does the fallback **inside the same request** (`backend/src/utils/passwordInput.js`): exact submitted bytes are compared first — so stored passwords that legitimately contain these characters (e.g. a ZWJ emoji sequence) keep working — and only on mismatch is the sanitized form compared. Single request means the fallback costs no reCAPTCHA token (they're single-use) and no failed-attempt lockout quota; `trackFailedAttempt` fires only when both compares fail. (Earlier revisions did this client-side / as a client retry — reworked after external review flagged the reCAPTCHA-token and lockout-quota interactions.)

## Tests

- 7 backend integration tests (`galleryPasswordSanitize.test.js`, real-SQLite migrations) pin the contract: exact-bytes-first, ZWSP/BOM/trailing-space rescue, ZWJ-containing stored password works, no invented matches, no `login_fail` recorded for a rescued login.
- Existing 8 `detectInAppBrowser` tests unchanged and passing; full frontend suite 115/115; `tsc --noEmit` and ESLint clean; pre-push E2E smoke suite 13/13.

## Verified in the running app

Vite dev build against the local backend, Playwright with a real Instagram-iOS UA string: blocked state renders with copy-link + try-anyway; clicking "try anyway" restores the form; a plain mobile-Safari UA gets the normal form with no banner. Screenshots in the comment below.

Closes #654.


the-luap:
Screenshots — Vite dev build against the local backend, Playwright with a real Instagram-iOS UA (`… Mobile/15E148 Instagram 327.0.0.42.122`), iPhone viewport. Instance language is DE.

**Instagram IAB → blocking state** (form hidden, copy-link button, "try anyway" escape hatch):

Instagram IAB blocked state

**After "try anyway"** — the form is restored:

Try-anyway restores the form

**Control: plain mobile-Safari UA** — normal form, no banner:

Normal browser unaffected

(The washed-out card heading is a pre-existing theme quirk of this repro event — visible in the control shot too, unrelated to this PR.)

Issue #798, opened 2026-07-14
[FEATURE] SSO (OIDC, LDAP, ...)


eopo:
**Is your feature request related to a problem? Please describe.**
I don't want to manually create users in PicPeak, as I already have an IdP running.

**Describe the solution you'd like**
SSO with OIDC, Autoprovisioning or LDAP Backchannel, Assign roles via OIDC roles or LDAP groups

**Describe alternatives you've considered**
Having some kind of cronjob sync users, however that wouldn't allow passwords to be synced.

**Use Case**
- Users already having some kind of IdP would benefit from a tighter integration, reducing manual admin work

**Similar Features**
- I don't know

**Mockups or Examples**
Usual SSO Implementation as found in a lot of Self-Hostable apps

the-luap:
which IdPs would you like to have in the system to be able to use them?
Regarding LDAP I am not sure if that is little too much for the moment as I am not sure if users really have ldap servers on their environments. I see the app targeting single photographers or small agenturs

eopo:
I for myself use Authentik for all my services, I guess most homelab users have one of Keycloak, Authentik or Authelia, all supporting LDAP and OAuth. 
LDAP would have the benefit, that galleries could be shared to users, which have not yet signed in using OAuth - Which would be the case most of the time. Some regular clients have their own account(s) in my IdP, so they can access my file sharing cloud and similar services, so I'd like to give them access to picpeak without creating another account. "Only" Supporting OAuth/OIDC would require them to login to PicPeak using SSO first, so their accounts are provisioned, before I can assign galleries to them. LDAP would make all existing users searchable, regardless wether they already signed in or not.

elimae17:
I'd love to see this implemented so PicPeak can be used with authentik!

the-luap:
Admin SSO via OIDC is implemented and on the beta line — three phases, all merged:

- **Phase 1** (#806) — SSO login, JIT auto-provisioning, identity bound to the IdP's stable `(issuer, sub)` pair (never email alone), client secret AES-256-GCM encrypted at rest.
- **Phase 2** (#854) — role/group mapping from an ID-token claim (IdP as source of truth on every login) plus login policy: optionally disable local password login entirely, with an `OIDC_BREAK_GLASS` escape hatch and a "last active super_admin is never demoted" invariant.
- **Phase 3** (#865) — RP-initiated logout, so logging out of PicPeak also ends the IdP session.

[...]

PR #865, opened 2026-07-23
feat(auth): OIDC logout-to-IdP — phase 3 (#798)

the-luap:
Phase 3 of #798 — RP-initiated logout, behind a new `oidc_logout_from_idp` setting (default off). Closes out the implementation phases of #798; phases 1 (#806) and 2 (#854) are already merged.

## Why

Without it, logging out of PicPeak leaves the IdP session alive — the next click on the SSO button signs straight back in, which is surprising on shared machines and effectively makes "logout" a no-op for SSO-only instances.

## How it works

- **Callback** (`auth.js`): `handleCallback` now also returns the raw ID token; the callback stores it in an HttpOnly cookie (`oidc_id_token`, path `/api/auth`, 24h = admin session lifetime). The cookie is both the `id_token_hint` for later and the marker that **this browser session came in via SSO** — local-password sessions never get bounced to the IdP. Tokens over ~3.9KB (huge group lists) skip the cookie to stay under the 4KB limit; logout then falls back to a client_id-only end-session request (one extra IdP confirmation click).
- **Logout** (`auth.js` `/logout`): when the cookie is present, it is cleared and `oidcService.buildEndSessionUrl(idToken)` is consulted — end-session endpoint from discovery metadata, `id_token_hint` + `post_logout_redirect_uri` (`<frontend-base>/admin/login`) + `client_id`. Non-null → returned as `ssoLogoutUrl`. **Every failure path returns null** (feature off, SSO unconfigured, discovery down, no `end_session_endpoint` advertised) — a logout must never fail because the IdP is unreachable.
- **Frontend**: `authService.adminLogout()` navigates to `ssoLogoutUrl` when present, `/admin/login` otherwise. Session **timeouts** deliberately stay local-only — an idle timeout is not an explicit logout intent.
- **Settings**: toggle in the Login policy card + the computed **post-logout redirect URI** displayed for copy-paste (must be registered at the IdP, e.g. Keycloak "Valid post logout redirect URIs"). Exposed via GET/PUT `/admin/settings/sso` like the other phase-2 policy flags; EN/DE strings.

## Tests

6 new integration tests (`oidcLogout.test.js`) over the mock in-process IdP (which now advertises a configurable `end_session_endpoint`):

- callback stores the raw ID token in the HttpOnly cookie (scoped `/api/auth`)
- logout returns the correct end-session URL (hint/redirect/client_id pinned) and clears the cookie
- feature off → no `ssoLogoutUrl` even for an SSO session
- no cookie (local-password session) → no `ssoLogoutUrl` even with the feature on
- IdP without `end_session_endpoint` → no `ssoLogoutUrl`, logout still 200
- config surface (`logoutFromIdp`, `getPostLogoutRedirectUri`)

Existing OIDC suites (32 tests, phases 1+2) pass unchanged. Frontend 115/115, `tsc` + ESLint clean.

## Live-verified against Keycloak 26

Fresh Keycloak 26 container, realm + confidential client (with registered post-logout redirect URI) + user, PicPeak dev stack configured through the settings API, driven via Playwright:

1. SSO login → Keycloak credentials → PicPeak dashboard.
2. PicPeak logout → browser bounced through Keycloak's end-session **without a confirmation prompt** (id_token_hint accepted) → returned to `/admin/login`.
3. Clicking the SSO button again → Keycloak asks for credentials again — the IdP session is genuinely gone.

Screenshots in the comment below.

Part of #798.

### DIFF
--- a/backend/__tests__/integration/helpers/mockOidcProvider.js
+++ b/backend/__tests__/integration/helpers/mockOidcProvider.js
@@ -34,6 +34,7 @@ class MockOidcProvider {
     // Test hooks:
     this.tamperNonce = false; // sign the ID token with a WRONG nonce
     this.emailViaUserinfoOnly = false; // omit email from the ID token; serve it on /userinfo
+    this.advertiseEndSession = true; // include end_session_endpoint in discovery (#798 phase 3)
     this.accessTokens = new Map(); // access_token -> user (for /userinfo)
     this.server = null;
     this.issuer = null;
@@ -85,6 +86,7 @@ class MockOidcProvider {
         token_endpoint: `${this.issuer}/token`,
         userinfo_endpoint: `${this.issuer}/userinfo`,
         jwks_uri: `${this.issuer}/jwks`,
+        ...(this.advertiseEndSession ? { end_session_endpoint: `${this.issuer}/logout` } : {}),
         response_types_supported: ['code'],
         subject_types_supported: ['public'],
         id_token_signing_alg_values_supported: ['RS256'],
--- a/backend/__tests__/integration/oidcLogout.test.js
+++ b/backend/__tests__/integration/oidcLogout.test.js
@@ -1,0 +1,188 @@
+/**
+ * OIDC logout-to-IdP integration tests (#798 phase 3).
+ *
+ * Same full-stack shape as oidcSso.test.js: real routes over a mock
+ * in-process IdP, genuine discovery/JWKS/PKCE via openid-client. Pins:
+ *
+ *   - the SSO callback stores the raw ID token in the oidc_id_token cookie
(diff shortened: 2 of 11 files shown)


the-luap:
Screenshots — settings UI (Vite dev build) and the live Keycloak 26 verification run.

**Login policy card with the new toggle** — post-logout redirect URI shown for copy-paste registration at the IdP:

Logout toggle in Login policy

**After PicPeak logout** — the browser went through Keycloak's end-session (no confirmation prompt, id_token_hint accepted) and Keycloak returned to the registered `/admin/login`:

Back on admin login after IdP logout

**Proof the IdP session ended** — clicking the SSO button again lands on Keycloak's credential form instead of silently signing back in:

Keycloak asks for credentials again

PR #890, opened 2026-07-28
fix(gallery): don't close the lightbox when clicking beside the photo (#883)

### LINKED ISSUE #883: [FEATURE] Disable closing gallery when clicking elsewhere than the arrow
I shared picpeak galleries to differents model of mine and they reported me some frustrations. 
On a positive note all of them appreciate Picpeak and saw the potential of this project :)
When browsing in the gallery fullscreen, when they miss-click on the black bars instead of the arrows it close the gallery. It's a bit inconvenient

Disable the fact that when clicking on the black bars close the browsing.

At the moment none

Use Case
- When would they use it? in someway, always and never
- Why is it important? This avoids the frustration of the gallery closing and having to find the spot where it closed.

Similar Features
- In Picdrop, clicking the black bars do nothing and it's pretty convenient

Mockups or Examples:

Image

the-luap:
Closes #883.

## The problem

While paging through a gallery in the lightbox, a missed click on the black bars beside the photo (instead of the prev/next arrows) closed the lightbox and dropped the guest back into the grid — they then had to scroll to find where they left off. Reported by a photographer whose clients hit this constantly.

## The fix

`PhotoLightbox.tsx`: remove the close-on-backdrop-click handler (`handleImageClick`) and its two attachment points (the track container and the current-slide wrapper). Clicking the black area now does nothing — the lightbox only closes via the **X button** or **Escape**, matching Picdrop's behavior referenced in the issue.

Pure deletion, no new state or props. Zoom/pan drag, swipe navigation, and the arrow/toolbar buttons are unaffected (they live on their own elements and never depended on the backdrop handler).

## Verification

- Dev server against a local gallery: clicked the black bars left/right of a portrait photo repeatedly — lightbox stays open; X and Escape still close it; arrows still navigate.

_Screenshot of the lightbox staying open after a backdrop click follows in a comment (CONTRIBUTING screenshot policy)._

### DIFF
--- a/frontend/src/components/gallery/PhotoLightbox.tsx
+++ b/frontend/src/components/gallery/PhotoLightbox.tsx
@@ -407,13 +407,6 @@ export const PhotoLightbox: React.FC<PhotoLightboxProps> = ({
     setIsDragging(false);
   };
 
-  const handleImageClick = (e: React.MouseEvent) => {
-    // Only close if clicking the background, not the image
-    if (e.target === e.currentTarget) {
-      onClose();
-    }
-  };
-
   // Touch event handlers: pinch-to-zoom (2 fingers) + single-finger
   // carousel-style swipe nav. Swipe is suppressed while zoomed in so the
   // user can pan instead. The carousel is also disabled mid-animation.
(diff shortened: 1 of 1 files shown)


the-luap:
Screenshot — Vite dev build against the local backend, Playwright, desktop viewport.

**Lightbox open after clicking the black bar** left of a portrait photo (the exact miss-click from the issue) — it stays open; only the X button and Escape close it:

Lightbox stays open after backdrop click

PR #891, opened 2026-07-28
feat(gallery): quick return from zoomed to fit-to-screen in the lightbox (#886)

### LINKED ISSUE #886: [FEATURE] Easily switch back from zoomed mode to full-screen mode.
Is your feature request related to a problem? No
When I have a zoomed-in image in my gallery, I'd like to be able to easily return to full-screen mode.

Having two simultaneous functions would be good.
The first would be a clickable button to return to the full image.
The second would involve double-clicking (or perhaps just a single click?) the image to return to the full image.

At the moment none

Use Case
- Why is it important? This feature would allow for improved maneuverability.

Similar Features
This feature is available in picdrop.

Mockups or Examples:

Image

the-luap:
Closes #886.

## What this adds

Both mechanisms the issue asked for, in `PhotoLightbox.tsx`:

1. **Fit-to-screen toolbar button** (`Minimize2` icon, between zoom-in and download) — disabled at 100%, one click returns a zoomed photo to fit and re-centres it (`resetZoom()` also clears the pan offset).
2. **Double-click on the image** resets zoom the same way. At 100% the handler is a no-op, so it cannot fight the existing backdrop click-to-close; on videos it isn't attached at all, mirroring the other pointer handlers.

## Verification

- Dev server against a local gallery: zoomed to 200% → double-click snapped back to 100% with the lightbox still open; the toolbar button does the same and correctly disables at 100%. Pan-while-zoomed and swipe nav unaffected.

_Screenshot (zoomed lightbox with the new fit-to-screen button enabled) follows in a comment._

### DIFF
--- a/frontend/src/components/gallery/PhotoLightbox.tsx
+++ b/frontend/src/components/gallery/PhotoLightbox.tsx
@@ -1,6 +1,6 @@
 import React, { useState, useEffect, useRef } from 'react';
 import { useDevToolsProtection } from '../../hooks/useDevToolsProtection';
-import { X, ChevronLeft, ChevronRight, Download, ZoomIn, ZoomOut, MessageSquare, Heart, Star } from 'lucide-react';
+import { X, ChevronLeft, ChevronRight, Download, ZoomIn, ZoomOut, Minimize2, MessageSquare, Heart, Star } from 'lucide-react';
 import type { Photo } from '../../types';
 import { useSavePhotoToDevice } from '../../hooks/useGallery';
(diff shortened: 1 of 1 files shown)


the-luap:
Screenshot — Vite dev build against the local backend, Playwright, desktop viewport.

**Zoomed to 200%** with the new fit-to-screen button enabled in the toolbar (arrows-inward icon between zoom-in and download). One click — or a double-click on the image — snaps back to 100% and re-centres; at 100% the button disables:

Zoomed lightbox with fit-to-screen button

Issue #508, opened 2026-05-16
[BUG] showing original names in lightbox .


Rekoo-PS:
Showing original names in lightbox is usefull in many cases . It could be great if we have it .

as shown in the example here : 
<img width="869" height="1012" alt="Image" src="https://github.com/user-attachments/assets/11e91189-eed7-44c3-96c5-e2d29f4d2747" />

the-luap:
Added in #513. Tied to the same toggle that drives original-name downloads (#493) — one switch controls both surfaces. Off by default; existing galleries keep showing only the position counter.

- Backend (\`gallery.js\`) now serialises \`photos[].original_filename\` and surfaces the toggle as \`event.use_original_filenames\` on the gallery response.
- The bespoke \`PhotoLightbox\` renders the original filename (falling back to \`filename\` only for pre-migration-062 uploads) in a muted truncated line under the position counter.
- \`GalleryStoryLayout\` mounts the same lightbox, so it follows along.
- \`GalleryPremiumLayout\` uses \`yet-another-react-lightbox\` instead; added the Captions plugin and a \`title\` field on the slides so the name appears as a caption when the toggle is on.

Other layouts feed through \`PhotoLightbox\` via \`PhotoGridWithLayouts\`, so the prop reaches them through the layout-props bag.

PR #892, opened 2026-07-28
fix(gallery): keep the lightbox toolbar from masking the photo (#888)

### LINKED ISSUE #888: [FEATURE] Bottom of the galery mask a bit of the pictures
One of my clients informed me that the info bar at the bottom of the gallery is obscuring part of the photo.

One solution I have in mind:
the info bar at the bottom is opaque, and the photo is visible in its entirety above it.

At the moment none

Use Case
- Why is it important? It allows us to see the photo in its entirety.

Similar Features
In picdrop, yes

Mockups or Examples:

Image
Image

the-luap:
Closes #888.

## The problem

The lightbox's bottom info/action bar is an `absolute` translucent-gradient overlay on top of the image area, so the lower edge of every photo sat hidden underneath it — the reporter's client noticed part of the photo being obscured.

## The fix

`PhotoLightbox.tsx`, following the solution proposed in the issue (opaque bar, photo fully visible above it — Picdrop behavior):

- The toolbar background changes from `bg-gradient-to-t from-black/80 to-transparent` to solid `bg-black`.
- The image/video track container now stops **above** the toolbar instead of running underneath it: the toolbar's height is measured with a `ResizeObserver` and applied as the container's `bottom` offset. Measurement (rather than a fixed offset) because the bar's height is genuinely variable — `flex-wrap` on narrow screens, the optional original-filename line (#508), the desktop feedback panel changing the wrap width, and iOS safe-area padding.

`object-contain` then scales the photo into the reduced area, so nothing is ever cropped or covered.

## Verification

- Dev server, portrait + landscape photos: image bottom edge lands exactly on the toolbar's top edge (measured 719/719 on desktop, 537/537 at 375px mobile width where the toolbar wraps to two rows). Zoom, swipe nav, and the desktop feedback panel unaffected.

_Screenshots (desktop + mobile, photo ending flush above the opaque bar) follow in a comment._

### DIFF
--- a/frontend/src/components/gallery/PhotoLightbox.tsx
+++ b/frontend/src/components/gallery/PhotoLightbox.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect, useRef } from 'react';
+import React, { useState, useEffect, useLayoutEffect, useRef } from 'react';
 import { useDevToolsProtection } from '../../hooks/useDevToolsProtection';
 import { X, ChevronLeft, ChevronRight, Download, ZoomIn, ZoomOut, Minimize2, MessageSquare, Heart, Star } from 'lucide-react';
 import type { Photo } from '../../types';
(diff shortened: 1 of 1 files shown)


the-luap:
Screenshots — Vite dev build against the local backend, Playwright.

**Desktop**: the photo's bottom edge lands exactly on the opaque toolbar's top edge (measured 719px/719px) — nothing is masked anymore:

Desktop: photo fully above the toolbar

**Mobile (375px)** — the toolbar wraps to two rows here; the ResizeObserver picks up the taller bar and the photo still ends flush above it (537px/537px):

Mobile: photo fully above the two-row toolbar

Issue #655, opened 2026-06-22
[FEATURE] Limit for how many images can be liked by users.


Duecki1:
**Is your feature request related to a problem? Please describe.**
Not really, its just annoying to tell people to only select 5 images. 

**Describe the solution you'd like**
I'd like a input field which limits the amount of images all/individual users can heart.

**Describe alternatives you've considered**
Just tell the people how many they are allowed to select

**Use Case**
Please describe how this feature would be used:
- Who would use it? Photographers
- When would they use it? Asking the client to heart images
- Why is it important? It would be practical

**Similar Features**
I don't know :D

the-luap:
Shipped in #658. Per-event cap, enforced server-side, with an explicit popup when the guest clicks past the limit.

**Where it lives**: Event → Feedback settings → new "Per-guest limits" section. Two numeric inputs:

- **Max favorites per guest** (0 = unlimited)
- **Max likes per guest** (0 = unlimited)

Saving `max_favorites_per_guest = 10` enforces it on the next favorite click.

**The popup** — when the guest clicks the heart on the 11th photo (with `max_favorites_per_guest = 10`):

- **Mobile**: slides up from the bottom (native action-sheet feel), full-width with safe-area padding for the iOS home indicator + Android gesture bar.
- **Desktop**: centered card, ~420px wide.
- Shows the icon, title ("Favorite limit reached"), body ("You can favorite up to 10 photos in this gallery. Remove one to add a new one."), a "10 of 10 used" counter pill, and a single "Got it" button. Escape and backdrop click both dismiss.
- Above the lightbox layer — works whether the guest clicks the heart inside the photo lightbox or on the gallery grid.

**Toggle still works**: a guest at 10/10 can un-favorite an existing photo to free a slot, then favorite a different one.

**Limit reduction is non-destructive**: if you save `max_favorites_per_guest = 5` while a guest already has 15 favorites, their existing 15 stay. New adds are blocked until they remove some. No bulk-delete on save.

**Per-guest, not per-gallery**: every guest at the wedding picks their own top 10 independently. Guest A hitting their cap doesn't affect guest B.

**i18n**: EN + DE, both for the admin field labels and for the popup itself (`{{limit}}` is the actual cap value you set).

15 unit/integration tests cover the contract (8 backend on the cap logic + 7 frontend on the error parsing). Pre-push smoke 13/13.

Closing once #658 merges. Ships to the next beta via release-please, then to main via the standard beta → main merge.

[...]

PR #893, opened 2026-07-28
feat(feedback): let guests remove their star rating (#884)

### LINKED ISSUE #884: [FEATURE] Possibility to remove a star during the selection process
When me and my clients are selecting pictures, once a pictures is rated with stars we can't undo it.

Two possibility: 
	- A sixth star, but crossed out
	- Clicking the same star twice, or clicking star 1 twice, cancels the rating.

At the moment none

Use Case
- Why is it important? This avoids too mess the gallery during the selection process. Sometimes, client can be really uncertain !

Similar Features
Not to my knowledge. I only used PicPeak and Picdrop

Mockups or Examples:

Image

the-luap:
Closes #884.

Implements option 2 from the issue: clicking the star you already gave clears your rating.

## What was actually broken

`PhotoRating.tsx` (feedback panel) has had click-same-star-to-clear since day one — it sends `rating: 0`. But the backend rejected it end-to-end, so removal never worked anywhere:

- `feedbackValidation.js:143` — `isInt({ min: 1, max: 5 })` → `rating: 0` got a 400.
- `feedbackService.js` — even past validation, an existing rating would have been **updated to 0**, not deleted: the 0 keeps counting in `total_ratings` and drags `AVG(rating)` down.

And the lightbox's inline stars never sent 0 at all — clicking your current star just resubmitted it.

## The fix

**Backend**
- Validator accepts `0–5`; `0` means "clear my rating".
- `submitFeedback`: existing rating + `rating: 0` → **delete** the guest-scoped row and recalculate photo stats (mirrors the like/favorite toggle-off path). `rating: 0` with no existing row is a no-op that never inserts.

**Frontend (`PhotoLightbox.tsx`)**
- Clicking your current star submits `0`; aria-label/title flip to "Remove rating".
- Identity-modal path: `pendingAction.rating` truthiness check → explicit number check so a pending clear isn't silently dropped.

`PhotoRating.tsx` needed no change — its existing 0 now works.

## Tests

New `__tests__/utils/feedbackRatingRemoval.test.js` (4 tests, modeled on the #655 suite): delete-not-zero + average reset, no-op on 0 without existing row, guest-scoping (guest B's rating and the average survive guest A's clear), and re-rating still updates in place. Full feedback test set (22 tests across 3 suites) green.

## Verification

Dev server + patched backend against a real gallery: rated 3 stars (summary "3.0 (1)") → clicked the third star again → stars cleared, summary "0.0 (0)", and the `photo_feedback` row is gone from the DB.

_Screenshots (rated state with "Remove rating" affordance, and cleared state) follow in a comment._

### DIFF
--- a/backend/__tests__/utils/feedbackRatingRemoval.test.js
+++ b/backend/__tests__/utils/feedbackRatingRemoval.test.js
@@ -1,0 +1,139 @@
+/**
+ * Unit tests for rating removal (#884).
+ *
+ * Pins the contract of `feedbackService.submitFeedback` for
+ * `feedback_type: 'rating'` with `rating: 0` ("clear my rating"):
+ *  - An existing rating row is DELETED (not updated to 0 — a stored 0
+ *    would drag the photo's average down and still count in totals).
+ *  - Photo stats (average_rating) are recalculated after the delete.
+ *  - Rating 0 with no existing rating is a no-op that never inserts a row.
+ *  - Removal is guest-scoped: clearing guest A's rating leaves guest B's
+ *    rating (and the resulting average) intact.
+ *  - Regular re-rating (3 → 5) still updates in place.
+ */
+
+const path = require('path');
+const fs = require('fs');
+const os = require('os');
+
+process.env.NODE_ENV = 'test';
+process.env.TEST_DATABASE_PATH = path.join(
+  fs.mkdtempSync(path.join(os.tmpdir(), 'picpeak-rating-removal-')), 'db.sqlite',
+);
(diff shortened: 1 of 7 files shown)


the-luap:
Screenshots — Vite dev build + patched backend against a local gallery, Playwright, simple identity mode.

**Rated 3 stars** — summary shows "3.0 (1)", and the third star's affordance flips to "Remove rating":

Rated 3 stars

**After clicking the third star again** — stars cleared, summary "0.0 (0)", and the `photo_feedback` row is deleted from the DB (verified with a direct query):

Rating removed

Issue #866, opened 2026-07-23
Accounting/Incoming invoices: attach source PDF as proof on re-bill + surface re-bill/passthrough status under CRM → Customer


Luca-Timo:
## Context

When a captured supplier invoice (Accounting → Incoming invoices) is re-billed to a client, we already store the original inbound PDF. Two gaps today:

1. That stored PDF never leaves picpeak — the client gets our re-bill line item with no underlying proof.
2. There's no per-customer view of what's been re-billed/passed through and where each item stands.

Covers both modes: **Passthrough** (cost forwarded 1:1, no markup) and **Re-bill** (cost forwarded with the contract-driven markup).

## Feature 1 — Attach the stored inbound PDF as proof

When a supplier invoice is re-billed/passed through, attach the original stored PDF **as a separate attachment on the client-invoice email** (not merged into the invoice PDF).

**Config**
- **Global default: off.** A tenant-wide toggle enables it.
- **Per-customer override** on the customer record, so individual clients can opt in/out regardless of the global default.

**Requirements**
- Separate email attachment (keep the invoice PDF itself untouched).
- Respect invoice immutability — proof attached at issue time; issued invoices never mutated (Storno + reissue).
- If the stored PDF is missing at issue time, persist a visible failure marker on the re-bill row — don't silently drop it.
- Behind the `incomingInvoices` flag, backend route + frontend both gated.

## Feature 2 — Re-bill / passthrough list under CRM → Customer

Add a panel in `CustomerCrmPanels` (gated on `incomingInvoices`) that mirrors the **Hours section** pattern: compact, history-only, with a **"Create invoice from re-bills" button up top**.

List every re-bill/passthrough item for the customer, grouped by status:
- **Open** — allocated to the client, not yet invoiced.
- **Sent** — rolled into an issued client invoice, awaiting payment.
- **Paid** — the covering client invoice is settled.

**Requirements**
- Each row: source supplier, date, cost vs re-billed amount (incl. markup), mode (passthrough / re-bill), linked event, and a link to the client invoice once issued.
- Storno rows filtered from every status aggregate (`kind='storno'` / negative totals).
- Money via `utils/money`; dates via `useLocalizedDate()`.
- Status derives from the linked invoice lifecycle, not a duplicated field (avoids drift).

## Feature 3 — Cross-add dialog on invoice creation (hours ↔ re-bills)

[...]

Issue #969, opened 2026-08-02
CRM document email controls always 404 for non-super-admins in the project cockpit


the-luap:
Follow-up from the GHSA-93x4 fix in #960 / #966.

`requireOwnedQueuedEmail` (`backend/src/routes/adminProjects.js`) scopes the queued-mail routes through `email_queue.event_id` → the caller's owned events. CRM document mail (quote / contract / invoice sends) carries **no** `event_id`, so for a non-super-admin those rows always take the `!row.event_id` branch and 404.

That is the right security posture — those rows have no ownable parent to check, and guessing would be worse. The problem is purely that the cockpit does not know it:

`getProjectOverview` still returns the CRM document emails for a project the caller legitimately owns, and the frontend renders preview / resend / cancel / retry controls for each one. For any admin who is not a super_admin, **every one of those buttons 404s**. The feature looks available and is not.

Two ways out:

1. Authorize those rows through their document → project association (the project *is* ownership-checked at that point), instead of through `event_id`. More permissive, but consistent with the fact that the caller already owns the project the document hangs off.
2. Have `getProjectOverview` omit or flag the rows a scoped caller cannot act on, so the UI stops offering dead controls.

(1) is the better product answer; (2) is the smaller change. Either way it is a behaviour decision rather than a security fix, which is why it did not ride along in the advisory PR.

PR #979, opened 2026-08-03
feat(accounting): re-bill proof attachment, CRM panel & hours↔re-bills cross-add (#866)

Luca-Timo:
Implements #866, plus two follow-up fixes to the incoming-invoice flow. Bundled as one diff for review. Everything is behind the `incomingInvoices` feature flag (backend routes + frontend surfaces).

## Feature 1 — Attach the stored supplier proof to the client-invoice email
When a captured supplier invoice is re-billed/passed through, its stored PDF can ride the client-invoice email as a **separate** attachment (the invoice PDF is never touched → invoice immutability).
- **Global default (off)** in Settings → Accounting, **per-customer tri-state override** (inherit/on/off), and **per-file selection** in a new Send dialog (select-all + individual, pre-checked per the resolved default).
- Auto-sends (scheduler / monthly flush) fall back to the resolved default.
- Missing proof at issue time → persisted `proof_attach_error` marker (surfaced in the CRM panel), never silently dropped, never blocks the send.
- Proof filename is a **configurable template** (`crm_rebill_proof_filename_format`) like the invoice/quote/contract number formats — tokens `{INVOICE} {SUPPLIER} {YEAR} {MONTH} {SEQ}/{SEQ:0Nd}`, default `Beleg-{INVOICE}`; multi-proof invoices auto-index; output is filesystem-safe.

## Feature 2 — Re-bills & passthrough panel under CRM → Customer
Mirrors the Hours section: grouped by **Open / Sent / Paid**, status **derived from the linked invoice lifecycle** (no duplicated field; a cancelled/Storno'd cover drops back to Open). Rows show supplier, date, cost vs re-billed (incl. markup), mode, linked event + client invoice, and any proof-attach failure. Money via `utils/money`, dates via `useLocalizedDate`.

## Feature 3 — Cross-add dialog (hours ↔ re-bills)
Billing one category for a customer who also has open items in the other offers to combine them into one invoice, symmetric from both entry points. **Hours and re-bills stay distinct, contiguous line groups — never merged into shared line items.**

## Follow-up fixes
- **Value-less re-bills no longer dead-end:** a customer-billed doc now requires a non-null amount at categorise/re-bill time (`0` is allowed; only empty is blocked), caught early with a clear message instead of failing the whole pending bundle at billing.

## Implementation notes
- **Migration 169** (additive, `hasColumn`-guarded, idempotent, sane `down()`): `inbound_documents.proof_attach_error`, `customer_accounts.rebill_attach_proof` (nullable tri-state), global `accounting_rebill_attach_proof` (default off, no seed row).
- Shared build/stamp helpers extracted from `billUnbilledEntries` / `billPendingRebills` (no duplication); new `invoice/rebillProofs.js` + `combinedBillingService.js` kept separate to avoid the `invoiceService ↔ expenseService` require cycle.
- New routes: `GET /invoices/:id/rebill-proofs`, `GET /expenses/inbound/by-customer/:id`, `POST /customers/:id/bill-combined`; `POST /invoices/:id/send` accepts `proofInboundIds`. All new routes carry the `incomingInvoices` flag guard + sibling permission checks.
- Path containment via `assertPathInside([business-docs])` at the proof-file boundary.
- i18n: **EN + DE** added (no other locales touched).

## Verification
- Frontend `npm run build` passes; migration 169 up/down + idempotent on SQLite.
- New tests: `renderProofName` unit (tokens/padding/index/sanitise) + `rebillProofAndPanel` integration (status derivation, proof selection/marker/default resolution, flag-off) + existing rebill/hours/customer suites green.
- **Coverage caveat:** the invoice-minting paths (`billCombinedForCustomer` / `billPendingRebills`) call `createInvoice` inside a transaction, which deadlocks on the SQLite test harness (documented in `incomingInvoiceRebill.test.js`); those rely on the existing suites, and the new tests cover the createInvoice-free logic.
- **UI screenshots**
Screenshot 2026-08-03 at 19 42 08
Screenshot 2026-08-03 at 19 46 01

### DIFF
--- a/backend/__tests__/integration/rebillProofAndPanel.test.js
+++ b/backend/__tests__/integration/rebillProofAndPanel.test.js
@@ -1,0 +1,215 @@
+/**
+ * Issue #866 — the createInvoice-free halves of the re-bill proof + CRM panel
+ * feature, against a real SQLite schema:
+ *
+ *   • listCustomerRebills — status DERIVED from the linked invoice lifecycle
+ *     (open / sent / paid; a cancelled/Storno'd cover drops back to open) plus
+ *     cost-vs-rebilled math and mode.
+ *   • collectRebillProofAttachments — the Send-dialog per-file selection, the
+ *     all-or-none default resolution (per-customer override else global), the
+ *     Beleg-<inv#> filename (suffix only when >1), and the missing-file marker.
+ *
+ * The invoice-MINTING paths (billCombinedForCustomer / billPendingRebills) call
+ * createInvoice inside a db.transaction, which deadlocks on the SQLite harness
+ * (global-db sequence write vs. held write lock) — same limitation the sibling
+ * incomingInvoiceRebill.test.js documents. They're covered by the existing
+ * billPendingRebills / billUnbilledEntries suites; here we hand-craft billed
+ * state instead.
+ */
+const fs = require('fs');
+const path = require('path');
+const { bootCrmDb, seedMinimal } = require('./helpers/crmDb');
+
+jest.setTimeout(120000);
+
+describe('#866 re-bill proof attachment + CRM panel', () => {
+  let db;
+  let cleanup;
+  let adminId;
+  let expenseService;
+  let rebillProofs;
+  let flagCache;
+
+  beforeAll(async () => {
+    ({ db, cleanup } = await bootCrmDb());
+    const dbModule = require('../../src/database/db');
+    dbModule.logActivity = async () => {};
+    ({ adminId } = await seedMinimal(db));
+    expenseService = require('../../src/services/expenseService');
+    rebillProofs = require('../../src/services/invoice/rebillProofs');
+    flagCache = require('../../src/middleware/requireFeatureFlag');
+  }, 120000);
+
+  afterAll(async () => {
+    if (cleanup) await cleanup();
+  });
+
+  const unwrapId = (ins) => (typeof ins[0] === 'object' ? ins[0].id : ins[0]);
+  let seq = 0;
+
+  async function makeCustomer(overrides = {}) {
+    seq += 1;
+    const ins = await db('customer_accounts').insert({
+      email: `c866-${seq}@example.com`,
+      display_name: `C866 ${seq}`,
+      password_hash: 'x',
+      preferred_language: 'de',
+      is_active: 1,
+      billing_cadence: 'per_event',
+      created_at: new Date(),
+      ...overrides,
+    }).returning('id');
+    return unwrapId(ins);
+  }
+
+  async function makeDoc(customerId, overrides = {}) {
+    const ins = await db('inbound_documents').insert({
+      source: 'upload', status: 'categorized', parse_status: 'parsed', parse_method: 'none',
+      supplier_name: 'ACME AG', currency: 'CHF', total_amount_minor: 10000,
+      invoice_date: '2026-06-01', disposition: 'rebill', customer_account_id: customerId,
+      created_at: new Date(), updated_at: new Date(),
+      ...overrides,
+    }).returning('id');
+    return unwrapId(ins);
+  }
+
+  async function makeInvoice(customerId, status, number) {
+    const ins = await db('invoices').insert({
+      invoice_number: number,
+      customer_account_id: customerId,
+      status,
+      currency: 'CHF',
+      issue_date: '2026-06-01', due_date: '2026-07-01',
+      vat_rate: 0, net_amount_minor: 10000, vat_amount_minor: 0, total_amount_minor: 10000,
+      created_at: new Date(), updated_at: new Date(),
+    }).returning('id');
+    return unwrapId(ins);
+  }
+
+  describe('listCustomerRebills', () => {
+    it('derives open / sent / paid and open→cost==rebilled for passthrough, +markup for rebill', async () => {
+      const customerId = await makeCustomer();
+
+      // Open re-bill (10% markup): rebilled = 11000.
+      await makeDoc(customerId, { total_amount_minor: 10000, markup_type: 'percent', markup_percent: 10 });
+      // Open passthrough: no markup, rebilled == cost.
+      await makeDoc(customerId, { disposition: 'durchlaufend', total_amount_minor: 5000, markup_type: 'none' });
+      // Sent (on a 'sent' invoice).
+      const sentInv = await makeInvoice(customerId, 'sent', 'R-2026-0001');
+      await makeDoc(customerId, { total_amount_minor: 8000, markup_type: 'none', billed_invoice_id: sentInv });
+      // Paid.
+      const paidInv = await makeInvoice(customerId, 'paid', 'R-2026-0002');
+      await makeDoc(customerId, { total_amount_minor: 8000, markup_type: 'none', billed_invoice_id: paidInv });
+      // Cancelled cover → drops back to 'open', no invoice link surfaced.
+      const cancInv = await makeInvoice(customerId, 'cancelled', 'R-2026-0003');
+      await makeDoc(customerId, { total_amount_minor: 8000, markup_type: 'none', billed_invoice_id: cancInv });
+
+      const items = await expenseService.listCustomerRebills(customerId);
+      const byStatus = (s) => items.filter((r) => r.status === s);
+
+      expect(items).toHaveLength(5);
+      expect(byStatus('open')).toHaveLength(3); // 2 genuinely-open + 1 cancelled-cover
+      expect(byStatus('sent')).toHaveLength(1);
+      expect(byStatus('paid')).toHaveLength(1);
+
+      const rebill = items.find((r) => r.mode === 'rebill' && r.costMinor === 10000);
+      expect(rebill.rebilledMinor).toBe(11000);
+      const passthrough = items.find((r) => r.mode === 'passthrough');
+      expect(passthrough.rebilledMinor).toBe(passthrough.costMinor);
+
+      const sent = byStatus('sent')[0];
+      expect(sent.invoiceNumber).toBe('R-2026-0001');
+      expect(sent.invoiceId).toBe(sentInv);
+
+      const cancelledCover = items.find((r) => r.status === 'open' && r.invoiceNumber === null && r.costMinor === 8000);
+      expect(cancelledCover).toBeDefined(); // cancelled cover isn't shown as a live invoice link
+    });
+  });
+
+  describe('collectRebillProofAttachments', () => {
+    const businessDocs = () => path.join(process.env.STORAGE_PATH, 'business-docs', 'inbound', '2026');
+
+    async function enableIncoming() {
+      const existing = await db('feature_flags').where({ key: 'incomingInvoices' }).first();
+      if (existing) await db('feature_flags').where({ key: 'incomingInvoices' }).update({ value: 1 });
+      else await db('feature_flags').insert({ key: 'incomingInvoices', value: 1 });
+      flagCache.invalidateFeatureFlagCache();
+    }
+
+    function writeProof(name) {
+      fs.mkdirSync(businessDocs(), { recursive: true });
+      const p = path.join(businessDocs(), name);
+      fs.writeFileSync(p, '%PDF-1.4\n% test proof\n');
+      return p;
+    }
+
+    it('honours explicit selection, names Beleg-<inv#>, and marks a missing file', async () => {
+      await enableIncoming();
+      const customerId = await makeCustomer();
+      const invId = await makeInvoice(customerId, 'scheduled', 'R-2026-1000');
+      const invoice = await db('invoices').where({ id: invId }).first();
+
+      const good1 = await makeDoc(customerId, { billed_invoice_id: invId, file_path: writeProof('p1.pdf') });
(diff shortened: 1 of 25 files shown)


the-luap:
Reviewed the backend in full. Good shape — the migration is clean, path handling is correct, and the new routes match the existing gating posture. One functional defect to close, one nit.

## 🟡 Concerns

### 1. A storno'd re-bill is stranded — shows as Open, can never be billed again

`expenseService.js:657-664` deliberately drops a cancelled cover back to `open`:

```js
if (invoiceStatus === 'cancelled') return 'open';
```

The PR describes this as the item becoming open again. It doesn't become *billable* again. Nothing clears `inbound_documents.billed_invoice_id` when the covering invoice is cancelled — that field is only reset at categorise time (`expenseService.js:463`). Every billing path filters on it being NULL:

- `buildPendingRebillLineItems` — `expenseService.js:587` (feeds both `billPendingRebills` and the new `billCombinedForCustomer`)
- `listPendingRebillSummary` — `expenseService.js:538` (the pending badge + the cross-add offer)

So after a Storno the row renders under **Open** in the new panel, with `invoiceId: null` / `invoiceNumber: null` (`expenseService.js:718-719`), i.e. indistinguishable from a genuinely unbilled item — but no billing path will ever pick it up.

Repro: re-bill €500 → cancel the covering invoice → panel shows "Open €500" → "Bill pending re-bills" reports nothing to bill, permanently. The supplier cost is silently stranded.

Fix shape: clear the linkage when the cover is cancelled, mirroring the categorise-time reset —

```js
billed_invoice_id: null,
billed_invoice_line_item_id: null,
```

— or teach the two pending queries to treat "linked invoice cancelled" as billable. The first is closer to the existing convention.

## 🟢 Nits

- Setting the filename template to exactly `{INVOICE}` renders the proof as `<invoice_number>.pdf` — the same name as the invoice PDF attachment (`invoice/sending.js:134`). Two identically-named attachments on one mail; harmless, but the proof can be mistaken for the invoice. Worth either reserving that case or documenting it next to the token list.

- **Migration 169** — `hasTable` + `hasColumn` guarded, additive, symmetric `down()`, idempotent on re-run.
- **Path containment** — `assertPathInside` (`utils/safePath.js:79-104`) is realpath-based with separator-appended roots and refuses when no root resolves; correctly applied at the proof-file boundary in `rebillProofs.js:150`.
- **Filename sanitisation** — traced traversal through `renderProofName`. `../../etc/passwd` in a supplier name collapses to `etc-passwd` (separators → `-`, leading `[-.]` runs trimmed, empty → `Beleg`). It is only an attachment display name regardless; `contentPath` carries the real path, consistent with the invoice PDF.
- **Authz + flag gating** — all four new routes carry `requireIncoming` plus a permission check. `/:id/bill-combined` mirrors `/:id/hour-entries/bill` exactly (`adminCustomers.js:685-716`).
- **Foreign-id injection** — `proofInboundIds` is validated as positive ints, then intersected against rows already scoped by `billed_invoice_id: invoice.id` (`rebillProofs.js:129-133`), so foreign ids are inert. Right shape.
- **Settings writes** — explicit per-key `hasOwnProperty` allowlist, boolean coerced via `!!`, format trimmed and capped at 120 chars.
- **The untested minting path.** Since the SQLite harness deadlock means `billCombinedForCustomer` can't be covered, I checked the position→line-id mapping by hand. It holds: `buildUnbilledHourLineItems` (`customerHoursService.js:459`) and `buildPendingRebillLineItems` (`expenseService.js:591-600`) are both strictly 1:1 with their source rows, `hoursCount` offsets the second group correctly, and the `insertedLines` / `lineByPos` pattern is identical to the already-shipped `billPendingRebills` (`expenseService.js:645-647`). No misalignment.

## Not verified

The frontend half (~600 lines across 12 files — dialogs, panels, i18n) I reviewed at diff-stat and PR-description level only, not line by line. Calling that out rather than implying full coverage.

## CI

`bypass-size-gate` is failing as designed — 1645 lines exceeds the 300-line self-merge limit and needs a maintainer approval, not a code change. Everything else green.

## Verdict

Close the storno stranding and I'm comfortable merging.

the-luap:
Went through the frontend half line by line now (the part I flagged as unreviewed above). One concern, two nits. The i18n and flag work is clean.

## 🟡 Concerns

### 2. New controls render on data presence alone — no permission gating

There are zero uses of `usePermissions` / `hasPermission` / `PermissionGate` anywhere in the diff, but the new endpoints need three different permissions:

| Control | Endpoint | Required |
|---|---|---|
| Panel contents | `GET /expenses/inbound/by-customer/:id` | `accounting.view` |
| "Create invoice from re-bills" | `POST /expenses/inbound/bill-pending` | `accounting.manage` |
| Cross-add → "Add both" | `POST /customers/:id/bill-combined` | `customers.edit` |

`CustomerCrmPanels.tsx:286-290` renders the create button purely on `openItems.length > 0`. So an admin with `accounting.view` but not `accounting.manage` sees the button, clicks it, and gets a 403 toast. Same on the hours side (`HoursSection.tsx:206-210`).

The sharper case is inside the cross-add dialog itself: "Just the re-bills" needs `accounting.manage` while "Add both" needs `customers.edit`. Two buttons, two different permissions, no indication — an admin holding one but not the other finds half the dialog dead.

This is the same class as #969, which we just shipped a fix for on `main`/`stable` — the cockpit was offering email controls the API rejected. Worth closing here rather than re-discovering it as a bug report.

Minor variant of the same: without `accounting.view` the panel still renders, just permanently empty (the query 403s and `items` defaults to `[]`), so the customer gets an empty "Re-bills & passthrough" card rather than no card.

## 🟢 Nits

- `CrossAddInvoiceDialog.tsx:52-55` has no explicit Cancel. Both buttons proceed to billing ("Just the …" / "Add both"); the only way out is a backdrop click. An admin who opens it by mistake has no obvious escape — worth a third button or at least an Escape handler.
- `BillDetailPage.tsx:561` uses `formatMoney(p.amountMinor / 100, …)`. `utils/money` exports `formatMoneyMinor` for exactly this (its doc comment even warns about the 100× trap on `*_amount_minor` values). Same output, but the hand-rolled `/100` is the pattern that helper exists to remove.

- **Feature-flag gating is correct and symmetric.** Panel behind `flags.incomingInvoices` (`CustomerCrmPanels.tsx:43`); the customer detail tri-state select behind the same. Both cross-add counters are `enabled:` on the *other* feature's flag — `HoursSection.tsx:171` (`incomingInvoices`) and `CustomerCrmPanels.tsx:262` (`hoursLogging`) — so no request fires at a flag-disabled endpoint.
- **i18n parity is exact.** 42 new keys in `en.json`, 42 in `de.json`, no key present in one and missing from the other. No hardcoded user-facing strings spotted.
- **Plural handling.** `bills.send.sendWithCount` and `crossAdd.body` pass `{{count}}` without `_one`/`_other` variants, but both strings are deliberately plural-agnostic ("proof(s)") and i18next falls back to the base key. Fine as written.
- **Send-dialog semantics.** Defaults pre-check only proofs that actually exist (`attachDefault ? proofs.filter(p => p.hasProof) : []`); missing-proof rows are checkbox-disabled with an explicit label; the send payload re-filters on `hasProof`. Crucially, `bills.service.ts:281` omits the body entirely when `proofInboundIds` is `undefined`, so "attach none" (`[]`) stays distinguishable from "use the default" — that distinction survives the whole client→route→service path.
- **Tri-state select** maps `null`/`true`/`false` ↔ inherit/on/off correctly and round-trips through `transformCustomer` (`adminCustomers.js:90`).
- **Cache invalidation is symmetric.** Both entry points invalidate the other side's keys, and every invalidated key matches a real query key (`admin-customer-hour-entries`, `customer-rebills`, `customer-open-rebills-count`, `customer-open-hours-count`). Both paths navigate to the new invoice afterwards, so the un-invalidated `customer-invoices` is moot.
- **Settings client coercion** — `accounting_rebill_attach_proof === true`, format falls back to `''` on a non-string.

## Verdict

Unchanged from my first pass, plus this: close the storno stranding and gate the new controls on their permissions, and I'm comfortable merging. The permission one is a small diff — `PermissionGate` already exists for it.

Luca-Timo:
Both blockers closed + the three nits, in `085e76de`.

## Concerns — closed

**1. Storno stranding.** New `releaseRebillsForCancelledInvoice(conn, invoiceId)` in `invoice/sending.js` clears `billed_invoice_id` + `billed_invoice_line_item_id` on the linked inbound docs whenever a cover is cancelled — called after the status flip in `createStorno` and in the draft-cancel branch of `cancelInvoice`. Mirrors the categorise-time reset, so a Storno'd re-bill drops back to a genuinely billable Open item (`billed_invoice_id IS NULL`), and `buildPendingRebillLineItems` / `listPendingRebillSummary` pick it up again. Regression test on the issued-cancel path in `rebillProofAndPanel.test.js` asserts the linkage is cleared and the row re-enters the pending pool.

**2. Permission gating.** Wired `usePermission` into the new controls, matched to the endpoints:
- RebillsPanel renders only with `accounting.view` (the `by-customer` query is `enabled: canView`, so no empty card and no 403 fetch).
- "Create invoice from re-bills" button requires `accounting.manage`.
- Cross-add "Add both" (`billCombined`) requires `customers.edit` — the dialog is only offered when the admin holds it, so there's no dead half-dialog.
- Hours "Create draft invoice" gated on `customers.edit`.

## Nits — done

- `CrossAddInvoiceDialog` now has an explicit **Cancel** button + an Escape handler (no longer backdrop-only).
- `BillDetailPage` uses `formatMoneyMinor(p.amountMinor, …)` instead of the hand-rolled `/100`.
- The proof-filename hint documents that an `{INVOICE}`-only template collides with the invoice PDF's own name — recommends keeping a `Beleg-` prefix.

Build + 39 backend tests green. `bypass-size-gate` still needs your approval — no code change there.

the-luap:
Re-verified `085e76de` at file:line. Both concerns closed, all three nits done. One small residual from the permission work, not a blocker.

## Concerns — closed

**1. Storno stranding ✅**

`releaseRebillsForCancelledInvoice` (`invoice/sending.js:217-227`) clears both columns and is wired into both call sites: `createStorno` at `:349` (inside the trx, right after the status flip on `originalId`) and the draft branch of `cancelInvoice` at `:643`. I checked those are the *only* two paths that write `status: 'cancelled'` to `invoices` — `git grep` over `backend/src` returns exactly `sending.js:344` and `sending.js:640`, the other hits being contracts and `email_queue`. So no cancel path is left uncovered.

Schema-guarded and best-effort, so it no-ops on non-accounting installs rather than breaking cancellation. Correct call for a cleanup step.

The regression test drives the real `createStorno` and asserts the full invariant — both columns null, panel row back to `open` with `invoiceId: null`, and the doc re-entering the pending pool. Nice touch seeding `document_sequences` past the hand-written invoice numbers with a comment explaining it's a test artifact.

Small note, not worth changing now: the pending-pool assertion re-implements the filter inline rather than calling `buildPendingRebillLineItems`, so it won't catch drift if that query's predicate changes later. The columns-cleared assertions are the real invariant either way.

**2. Permission gating ✅**

Matched to the endpoints, and gated at both layers rather than just hiding buttons:

- `CustomerCrmPanels.tsx:249-251` resolves all three permissions; `:262` sets `enabled: canView` on the panel query and `:311` returns `null` outright without `accounting.view` — so no 403 fetch *and* no empty card. That was the better of the two fixes I suggested.
- Create button gated on `canManage` (`:320`).
- Cross-add offered only when `canCombine` (`:305`), so the split-permission dead half-dialog is gone.
- `HoursSection.tsx:57` + `:424` gate the hours billing button on `customers.edit`.

## 🟢 Residual nit (new, from the fix itself)

Two of the cross-add counter queries are now `enabled:` on a permission their endpoint doesn't check:

- `HoursSection.tsx:174` — `enabled: flags.incomingInvoices && canBill` where `canBill` is `customers.edit`, but `GET /expenses/inbound/by-customer/:id` requires **`accounting.view`** (`adminExpenses.js:118`).
- `CustomerCrmPanels.tsx:270` — `enabled: flags.hoursLogging && canCombine` (`customers.edit`), but `listHourEntries` requires **`customers.view`** (`adminCustomers.js:126`).

An admin with `customers.edit` but not `accounting.view` — a plausible "manages customers and hours, not the books" role — fires a guaranteed 403 on every customer-detail render. It degrades safely (the count stays `0`, so the cross-add is simply never offered, which is the right outcome for that role) so this is noise rather than breakage. Adding the endpoint's own permission to each `enabled` would close it.

## CI

12 pass. The only red is `bypass-size-gate`, which is the 300-line self-merge policy and needs my approval, not a code change.

## Verdict

Merge cleared from my side. The residual above is a one-line-each follow-up — take it here or in a separate PR, your call; I won't hold the merge for it.

the-luap (approved):
Approving — both concerns from the first pass are closed and re-verified at file:line in `085e76de`.

Covered in review: migration 169, the proof path/filename boundary, authz + flag gating on all four new routes, the position→line-id mapping in the untested combined-billing path, the full frontend surface, and the storno release across every invoice-cancel path.

This also clears the `bypass-size-gate` check, which was only ever waiting on a maintainer approval for the 1645-line diff.

The one residual noted above (two counter queries gated on a permission their endpoint doesn't check — `HoursSection.tsx:174`, `CustomerCrmPanels.tsx:270`) is a one-liner each and not a merge blocker. Fine either here or as a follow-up.

Issue #868, opened 2026-07-24
Admin preview: skip the gallery password on published, password-protected galleries (needs maintainer sign-off)


Luca-Timo:
## Summary

Extend the existing **admin preview-token** mechanism so that a logged-in admin opening a **published, password-protected** gallery via "Galerie ansehen" is let straight in — no password prompt. Today the admin session token only bypasses draft-visibility, not the password gate.

@the-luap — **would this be acceptable to you?** It changes the security posture of password-protected galleries (a valid admin JWT would skip the guest password), so I'd like your sign-off on the approach before building it. Details and safeguards below.

## Current behaviour

The admin session token is already reused as a preview token, but its scope is narrow:

- **Frontend** — the "Galerie ansehen" link only appends `?preview=<admin_token>` when the event **is a draft** (`EventDetailsHeader.tsx`, ~L191–201). Published galleries get the plain share link with no token.
- **Backend** — `isAdminPreview()` (`middleware/gallery.js`) validates the admin JWT and its **only** effect is relaxing the `is_draft = false` filter so an unpublished gallery is viewable (`routes/gallery.js` ~L185, plus `verifyGalleryAccess`).
- The password requirement is independent: the info endpoint returns `requires_password` straight from `event.require_password` with no preview check (`routes/gallery.js` ~L88, ~L197), and `GalleryPage.tsx` renders the password form whenever `requiresPassword` is true. Nothing consumes the preview param to auto-authenticate.

Result:
- Draft, no password → opens directly ✅
- Published + password → admin still gets the password prompt ❌
- Draft + password → viewable as draft, but still prompts ❌

## Proposed change

When a **valid admin token** is present, let the admin past the password gate on **published** galleries too:

1. `isAdminPreview` (or the gallery info/access path) short-circuits the password requirement when a valid admin JWT is present — return `requires_password: false` for that request, or auto-mint the gallery access token server-side.
2. Frontend attaches `?preview=<admin_token>` for **published** galleries as well (not only drafts), or auto-logs-in when the preview token validates, so the admin never sees the password form.
3. Keep the existing draft `?preview=` path intact.

## Safeguards / constraints

- **Admin JWT only.** The bypass must key off a verified admin token (`decoded.type === 'admin'`, issuer `picpeak-auth`) — a guest must never be able to spoof it. Fail closed on any verification error.
[...]

Issue #838, opened 2026-07-18
[FEATURE] Reveal mode: hide gallery from guests until manual/scheduled reveal


the-luap:
## Motivation

SayCheese offers a "reveal" flow: the album stays hidden during the event — guests can upload but see nothing — and the host later reveals everything, optionally as a curated live slideshow ("magical reveal"). This is popular for weddings (surprise slideshow at the party, curated gallery the morning after).

PicPeak has the building blocks (per-photo visibility control, guest-upload moderation, live slideshow) but no explicit "hidden until reveal" switch — approximating it today means abusing moderation for every single photo.

## Scope

- New per-event toggle: **Reveal mode** (works only in combination with guest uploads, off by default).
  - While active: guests can upload via the gallery link, but the gallery grid/lightbox is hidden for guests (upload-only view with a friendly "photos will be revealed later" message). Host/admin sees everything as usual.
  - Reveal: manual toggle in admin ("Reveal now") and/or scheduled reveal at a configurable date/time (reuse the pattern of the expiration scheduler).
- Live slideshow keeps working for the host during reveal mode (slideshow token access is independent of guest gallery visibility) — that's the "surprise slideshow on the beamer" case.
- Optional: notification email to the host when reveal fires (existing email template infrastructure).

## Non-goals

- Per-guest visibility rules (already covered by photo visibility / client access controls).
- Curation UI beyond what moderation already provides.

## Implementation notes

- Event flag + reveal timestamp (migration), enforcement in `gallery.js` (photo listing endpoints return empty/hidden state for guests while active — not just frontend hiding).
- Guest upload endpoint (`adminPhotos.js` guest-upload path) stays functional while gallery listing is gated.
- Scheduled reveal can piggyback on the existing background scheduler used by the expiration checker.

PR #981, opened 2026-08-03
feat(gallery): admin preview skips the password on protected galleries (#868)

Luca-Timo:
Implements #868 (you signed off in the issue). A logged-in admin opening a **published, password-protected** gallery via "Galerie ansehen" is let straight in — no password prompt — mirroring the existing draft-visibility bypass.

Followed your spec from the issue: **server-side detection + explicit `?admin_preview=1` flag** (no token in the URL), **no setting**, **views + downloads both**, and the two implementation cautions (per-request, no gallery JWT; never touch the lockout buckets).

## Mechanism
- **`isAdminPreview`** now requires the explicit `?admin_preview=1` flag **AND** a verified admin session from the httpOnly `admin_token` cookie (or `Authorization: Bearer`) — `type: 'admin'`, issuer `picpeak-auth`, **fail-closed**. Never reads a token from the URL.
- **Retires the old `?preview=<raw-admin-JWT>`** scheme (the address-bar token leak from the safeguards section). The draft "Galerie ansehen" link now uses the same `?admin_preview=1` flag for **both** draft and published galleries.
- **Per-request bypass only** — `verifyGalleryAccess` lets an admin preview past the no-token/password branch, and `/info` returns `requires_password: false`. **No gallery JWT is minted**, and the password endpoint / `login_attempts` lockout buckets are never touched (the password form is simply skipped).
- **Excluded from guest analytics** — the `view` + all six `download*` `access_logs` inserts, `download_count`, `bumpEventDownloadCounts`, and the "gallery opened"/download notification bells are all skipped when `req.isAdminPreview`.
- **Images/downloads covered** — photo image URLs carry the flag so each cookie-authenticated ``/download re-asserts the admin session; the frontend forwards the flag on every gallery API call via the shared axios interceptor, plus the two browser-native download hrefs.

## Backward compatibility (no env change, no user action)
- **Code-only**: no migration, no new env var, no setting.
- Every new path is guarded by `req.isAdminPreview` / the URL flag, so **guest traffic is byte-identical** (image URLs, logs, password + draft gates all unchanged for guests).
- The `admin_token` cookie is `path: '/'`, `SameSite: Lax`, `secure: auto` — it reaches `/api/gallery/*` same-origin over HTTP and HTTPS, so the feature works on update with **zero admin action**.
- Fails closed: multi-origin installs (gallery on a different host) simply degrade to today's password prompt; stale `?preview=` bookmarks (ephemeral 24h links) stop bypassing but the gallery still loads.

## Verification
- New unit test `galleryAdminPreview.test.js` (8 cases) pins the fail-closed gate: flag-without-token, token-without-flag, guest/gallery token, tampered token, wrong issuer, non-`"1"` flag all return false; only flag + valid admin cookie/Bearer returns true.
- Frontend `npm run build` passes; existing gallery route/integration suites green (the one pre-existing `galleryPasswordSanitize` harness failure is unrelated — fails identically on `main`).

## Open items
- **Screenshots**: happy to add a before/after (password form → gallery opens directly) if you'd like them in the description before merge.
- Left `POST /gallery/verify` (`auth.js`) untouched — the admin never reaches it (form skipped), so the lockout buckets stay clean without a code change there. Shout if you'd prefer an explicit guard anyway.

### DIFF
--- a/backend/__tests__/integration/revealMode.test.js
+++ b/backend/__tests__/integration/revealMode.test.js
@@ -152,9 +152,13 @@ describe('Reveal mode (#838)', () => {
       expect(res.body.photos).toHaveLength(2);
     });
 
-    it('/photos serves the admin preview everything', async () => {
+    it('/photos serves the admin preview everything (new transport: ?admin_preview=1 + admin cookie, even with a coexisting gallery session)', async () => {
+      // #868/#981: reveal-mode hiding is bypassed for an admin preview via the
+      // new transport (explicit flag + httpOnly admin_token cookie), NOT the
+      // retired ?preview=<jwt>. The coexisting gallery Bearer must not shadow it.
       const res = await request(app)
-        .get(`/api/gallery/${SLUG}/photos?preview=${encodeURIComponent(adminToken)}`)
+        .get(`/api/gallery/${SLUG}/photos?admin_preview=1`)
+        .set('Cookie', [`admin_token=${adminToken}`])
         .set('Authorization', `Bearer ${galleryToken()}`);
       expect(res.status).toBe(200);
       expect(res.body.hidden_until_reveal).toBe(false);
--- a/backend/__tests__/middleware/galleryAdminPreview.test.js
+++ b/backend/__tests__/middleware/galleryAdminPreview.test.js
@@ -1,0 +1,59 @@
+/**
+ * #868 — the admin gallery-preview gate. isAdminPreview must fail CLOSED: it
+ * grants the draft/password bypass only for an explicit `?admin_preview=1` flag
+ * AND a verified admin JWT (type 'admin', issuer 'picpeak-auth') read from the
+ * httpOnly admin_token cookie or a Bearer header — never from the URL, never for
+ * a guest/gallery token.
+ */
+process.env.JWT_SECRET = process.env.JWT_SECRET || 'admin-preview-test-secret';
+const jwt = require('jsonwebtoken');
+const { isAdminPreview } = require('../../src/middleware/gallery');
+
+// Read the secret at call time — a jest setup file can set JWT_SECRET after this
+// module loads, and isAdminPreview verifies against the live value.
+const adminToken = () => jwt.sign({ type: 'admin', id: 1 }, process.env.JWT_SECRET, { issuer: 'picpeak-auth' });
+const galleryToken = () => jwt.sign({ type: 'gallery', eventId: 1 }, process.env.JWT_SECRET, { issuer: 'picpeak-auth' });
+
+function req({ flag, cookie, bearer } = {}) {
+  return {
+    query: flag === undefined ? {} : { admin_preview: flag },
+    cookies: cookie ? { admin_token: cookie } : {},
+    headers: bearer ? { authorization: `Bearer ${bearer}` } : {},
+  };
+}
+
+describe('isAdminPreview (#868) fails closed', () => {
+  it('false without the explicit flag, even with a valid admin cookie (plain link stays guest-identical)', () => {
+    expect(isAdminPreview(req({ cookie: adminToken() }))).toBe(false);
+  });
+
+  it('false with the flag but no session token', () => {
+    expect(isAdminPreview(req({ flag: '1' }))).toBe(false);
+  });
+
+  it('true with the flag + a valid admin cookie', () => {
+    expect(isAdminPreview(req({ flag: '1', cookie: adminToken() }))).toBe(true);
+  });
+
+  it('true with the flag + a valid admin Bearer header', () => {
+    expect(isAdminPreview(req({ flag: '1', bearer: adminToken() }))).toBe(true);
+  });
+
+  it('false for a gallery (guest) token — must be type admin', () => {
+    expect(isAdminPreview(req({ flag: '1', cookie: galleryToken() }))).toBe(false);
+  });
+
+  it('false on a tampered token', () => {
+    expect(isAdminPreview(req({ flag: '1', cookie: `${adminToken()}x` }))).toBe(false);
+  });
+
+  it('false on the wrong issuer', () => {
+    const t = jwt.sign({ type: 'admin' }, process.env.JWT_SECRET, { issuer: 'not-picpeak' });
+    expect(isAdminPreview(req({ flag: '1', cookie: t }))).toBe(false);
+  });
+
+  it('false when the flag is anything other than exactly "1"', () => {
+    expect(isAdminPreview(req({ flag: 'true', cookie: adminToken() }))).toBe(false);
+    expect(isAdminPreview(req({ flag: '0', cookie: adminToken() }))).toBe(false);
+  });
+});
--- a/backend/__tests__/middleware/galleryAdminPreview.test.js
+++ b/backend/__tests__/middleware/galleryAdminPreview.test.js
@@ -43,6 +43,14 @@ describe('isAdminPreview (#868) fails closed', () => {
     expect(isAdminPreview(req({ flag: '1', cookie: galleryToken() }))).toBe(false);
   });
 
+  it('true from the admin cookie even when a gallery Bearer is also present (#981 coexisting session)', () => {
+    expect(isAdminPreview(req({ flag: '1', cookie: adminToken(), bearer: galleryToken() }))).toBe(true);
+  });
+
+  it('false when only a gallery Bearer is present — a gallery header can never satisfy it (#981)', () => {
+    expect(isAdminPreview(req({ flag: '1', bearer: galleryToken() }))).toBe(false);
+  });
+
   it('false on a tampered token', () => {
     expect(isAdminPreview(req({ flag: '1', cookie: `${adminToken()}x` }))).toBe(false);
   });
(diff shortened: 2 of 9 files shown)


the-luap:
Reviewed backend in full, frontend partially. The direction is right — retiring `?preview=<raw-admin-JWT>` removes a 24h admin token from the address bar, referrers and proxy logs, and that was worth doing on its own. The transport that replaces it doesn't hold up yet.

## 🔴 Blockers

### 1. The `Authorization: Bearer` transport 403s on every gallery route

`verifyGalleryAccess` resolves the gallery credential *before* any preview check:

```js
const token = getGalleryTokenFromRequest(req, requestedSlug);
```

`getGalleryTokenFromRequest` is header-first and type-blind (`utils/tokenUtils.js:177-180`) — it returns whatever is in `Authorization: Bearer`, including an admin JWT. Execution then takes the token branch and hits:

```js
if (decoded.type !== 'gallery') {
  return res.status(403).json({ error: 'Invalid token type for gallery access' });
}
```

`middleware/gallery.js:108-110`. So the Bearer path documented in the PR body never reaches the admin-preview branch — it 403s first, for `/photos`, images and downloads alike.

### 2. A coexisting gallery session silently disables the cookie transport

`isAdminPreview` reads the admin session via `getAdminTokenFromRequest`, which is also header-first (`utils/tokenUtils.js:158-163`). The frontend interceptor attaches a gallery token to gallery endpoints whenever one exists for that slug, so for an admin who has previously opened this gallery as a guest, that gallery JWT is what `getAdminTokenFromRequest` returns. It verifies fine, `decoded.type === 'gallery'`, and `isAdminPreview` returns false — the httpOnly `admin_token` cookie is never consulted because a header was present.

Net effect: the preview works only for an admin with no lingering gallery session for that gallery. The PR body states it tolerates that case; it doesn't.

Both blockers are the same root cause — the two helpers read `Authorization` without discriminating token type, and the cookie is only reached when no header exists. Resolving admin preview *before* consuming the gallery credential, and having `isAdminPreview` prefer the `admin_token` cookie over a type-mismatched header, closes both.

### 3. CI is red, and the failing test is this PR's

The description says the only failure is a pre-existing `galleryPasswordSanitize` harness issue that "fails identically on `main`". That isn't the failure. The `backend` job fails on:

```
● Reveal mode (#838) › gallery routes while hidden › /photos serves the admin preview everything
  Expected: false
  Received: true
  at __tests__/integration/revealMode.test.js:160
```

I ran that suite against `origin/main` — it passes there (✓ 4 ms). The test requests `?preview=<adminToken>` (`revealMode.test.js:155-162`), the scheme this PR retires, so it now gets the hidden shell.

The test being stale is fixable, but it surfaces something the PR doesn't mention: **reveal mode is a second consumer of `isAdminPreview`** (`utils/revealMode.js:53-54`). #868 is scoped to the password gate, but this changes the admin-preview transport for #838's hidden-gallery bypass too. That needs a deliberate call and a deliberate test update — please state the intended reveal-mode behaviour explicitly rather than letting the test change carry it.

## 🟡 Concerns

1. `routes/gallery.js:1636` — `increment('view_count', 1)` is unguarded, so admin previews still inflate per-photo view counts. The PR body says photo analytics are excluded; the gallery-level `access_logs` insert is, this one isn't.
2. `routes/gallery.js:1995, 2004, 2019, 2058, 2095, 2104, 2113` — each redirect builds a bare `/api/gallery/:slug/photo/:photoId` with no query string, dropping `admin_preview=1`. Since no gallery JWT is minted, the redirected request has nothing left to authorise it and falls back to the draft/password gate. Video heroes and missing/failed derivatives are the paths that hit this.

## Not verified

One more was raised that I could not confirm without tracing the frontend: with `requires_password: false`, `GalleryPage` attempts the public auto-login, but `GalleryAuthContext` only accepts a `type: 'gallery'` session, so the empty-password login may 401 and leave the page on `GallerySkeleton`. Flagging it as unconfirmed rather than asserting it — worth checking on your side, since it would mean the happy path doesn't complete even once the transport is fixed.

## ✅ Verified

- `isAdminPreview` itself is correctly fail-closed on the flag: `!== '1'` short-circuits, unverifiable/tampered/wrong-issuer/non-admin tokens all return false. The new `galleryAdminPreview.test.js` covers those eight cases properly.
- No token is read from the URL anywhere in the new path.
- No gallery JWT is minted for the preview, and `login_attempts` / lockout buckets are untouched — the password endpoint is simply never reached.
- Guest traffic is unchanged: every new branch is behind `req.isAdminPreview` or the URL flag.

## Verdict

Close the two transport blockers and the CI failure (with an explicit decision on reveal mode) and I'll re-review. The security direction is right and I'd like this to land — it just can't merge while the documented transports don't work and the suite is red.

Screenshots: yes please, once the flow completes end to end — before/after on a published password-protected gallery.

Luca-Timo:
All three blockers + both concerns closed, and the happy path now completes. `e14184fb`.

## Blockers — closed

**1 & 2 (transport).** Same root cause, fixed at the root: `verifyGalleryAccess` now resolves admin preview **before** `getGalleryTokenFromRequest`, so a coexisting gallery token/Bearer can't shadow it and the admin session never reaches the `type !== 'gallery'` reject. `isAdminPreview` reads the httpOnly `admin_token` cookie **first** and accepts a Bearer only when it is itself an admin token — a gallery Bearer on a gallery endpoint can no longer disable the preview. `galleryAdminPreview.test.js` now asserts both: cookie wins over a coexisting gallery Bearer, and a gallery Bearer alone never satisfies the gate.

**3 (CI + reveal mode).** You're right — that was this PR's failure, not the pre-existing one; thanks for running it down. **Intended reveal-mode behaviour: admin preview continues to bypass reveal-mode hiding** (unchanged from #838), only the transport moves to `?admin_preview=1` + cookie. `revealMode.test.js` is updated off the retired `?preview=<jwt>` and now also carries a coexisting gallery Bearer to prove it doesn't shadow the preview. Suite is green.

## Concerns — closed

1. `view_count` increment is now guarded by `!req.isAdminPreview`.
2. The internal photo redirects (video hero / missing derivatives) preserve `admin_preview=1` via a `withPreview(req, url)` helper, so the redirected request still authorises instead of falling back to the gate.

## The "not verified" note — confirmed real, fixed

You were right to flag it. With `requires_password: false` the public empty-password auto-login POSTs `/gallery/verify` and 401s against the real password, leaving the page on the skeleton. `GalleryPage` now detects admin preview and renders `GalleryView` directly — **no gallery session, no `/gallery/verify` call** (so the lockout buckets stay untouched, per your caution), photos load per-request via the flag + admin cookie. `GalleryView` already reads its live event from its own `/photos` fetch, so the `/info` payload just seeds the header.

Build green, gallery + reveal suites green. Screenshots (before/after on a published password-protected gallery) coming next now that the flow completes end to end.

the-luap (approved):
Re-verified `e14184fb` at file:line. All three blockers and both concerns are closed, and the item I could only flag as unconfirmed turned out to be real and is fixed.

## Blockers — closed

**1 & 2 (transport).** `middleware/gallery.js:53-73` — admin preview is resolved and `return next()`s before `getGalleryTokenFromRequest` is reached, so the admin session can no longer fall into the `type !== 'gallery'` reject. `middleware/gallery.js:27-41` — candidates are tried cookie-first and each is type-checked, so a coexisting gallery Bearer is tried, fails `type === 'admin'`, and the loop continues to the cookie rather than shadowing it. Both shapes are pinned by the two new `galleryAdminPreview` cases.

**3 (reveal mode).** Intent is now stated rather than carried implicitly by a test edit, which was the ask. `revealMode.test.js` moved off the retired `?preview=<jwt>` and carries a coexisting gallery Bearer, so it proves the new transport survives the exact shape that broke the old one.

## Concerns — closed

1. `routes/gallery.js:1641` — `view_count` increment guarded by `!req.isAdminPreview`.
2. `withPreview()` at `routes/gallery.js:26`, applied to **8** redirect targets. I listed 7; you caught one more at `:2161`.

## The unconfirmed item

Good catch confirming it rather than waving it through. `GalleryPage` now guards the auto-login on `!isAdminPreview` and renders `GalleryView` directly, so `/gallery/verify` is never POSTed and the `login_attempts` buckets stay clean — which was the original caution in #868.

## One thing still outstanding — CI

The `backend` job on `e14184fb` was **cancelled at exactly 10m00s** against `timeout-minutes: 10` (`.github/workflows/tests.yml:31`), so "suite is green" does not hold for CI even though it does locally.

For calibration: `backend` completes in **~2m10s** on `main`, across six consecutive runs. I ran the full suite on `e14184fb` and on `origin/main` side by side and got 19s and identical results both times — the five failing suites are exactly the ones CI excludes as known infra debt. So it does not reproduce locally and I am not attributing a cause.

I have re-run the job in case it is a runner flake. If it times out again, the reproducer is the CI command itself (`tests.yml:63-66`), which differs from a plain local run in both the ignore patterns and `--ci`.

Approving on the code — this is the only thing between it and merge.

the-luap:
Re-ran the `backend` job — **2m02s, green**, in line with `main`'s ~2m10s baseline. So the 10-minute timeout was a runner flake, not a hang in your change. Correcting my earlier read: I'd taken the 5× delta against a stable baseline as evidence something in the commit hung, and that inference was wrong. Your "suite is green" was right; the job just needed re-running.

Merged as `f0066151`. Thanks for the two rounds — particularly for confirming the auto-login/skeleton item I could only flag as unverified, and for finding the eighth redirect site.

Screenshots are still worth adding when you get to them — the before/after on a published password-protected gallery. Nothing blocking now, just for the record on #868.

Issue #997, opened 2026-08-05
feat: PicTransfer — WeTransfer/SwissTransfer-style time-limited download links


Luca-Timo:
> **Working title: PicTransfer** (a.k.a. "PicPeak Transfer").
> **Delivery intent:** built as **one PR covering the whole feature set** — developed in iterations, but landed together, not shipped as separate phase PRs. The "iterations" below are the build order within that single PR, not independent releases.
> **Download-only, no preview:** the transfer page shows **no image preview** — no thumbnails, no lightbox, no in-browser viewing. The recipient sees a file list and downloads. This is explicit product intent, not an oversight.

## Summary

Add a **PicTransfer** surface: WeTransfer / SwissTransfer-style **time-limited, password-less, preview-less download links** for sending one or a handful of photos to someone, without opening the whole gallery or minting a share/password.

The gallery is the wrong tool for "a client asked for these three shots" or "send the venue this set": it's the full branded, previewable experience, it's password-gated, and the link is the event. A PicTransfer is a **throwaway, expiring, download-only page** — pick some photos, get a link, it dies in 14 days.

## Use cases

- A client requests a few specific images from their event — fulfil it with a link instead of re-sharing the gallery.
- Send a subset of an event's photos to a third party (venue, vendor, press) quickly.
- Ad-hoc file send that isn't tied to any event at all.

## Current state — why this isn't the gallery

- Galleries are event-scoped, branded-full-experience, previewable, and (usually) password-protected; the share link *is* the gallery.
- There's no way to say "just these 3 photos, no password, no preview, expires soon, not the whole gallery."
- The building blocks already exist and can be reused: the **token-only access model** (the Live Slideshow share, migration 138, where the URL is the secret and there's no password), the **ZIP/download-all pipeline** (`downloadZipService`), and the **email queue/templates**.

## Proposed feature

A **PicTransfer** = a curated set of items + a **token-only, password-less, preview-less download page** with an expiry.

**Creating one (entry points):**
- From an event's photo grid: select photos → **Create transfer**.
- Standalone: **New transfer** → cross-event photo picker and/or drag-in ad-hoc uploads.

**The download page (public, token-only — no login, no password, NO image preview):**
- The URL is the secret (high-entropy token), exactly like the slideshow share / SwissTransfer.
- **Download-only** — a plain list of the files (filename + size), total count + total size. **No thumbnails, no lightbox, no in-browser image rendering**; the recipient downloads the file(s) to view them.
[...]

PR #998, opened 2026-08-06
feat(transfers): add PicTransfer — cross-event file transfers (#997)

Luca-Timo:
## PicTransfer — cross-event file transfers (#997)

Send original files from any event(s) as a token-protected download link, with an optional client-upload channel. **Strictly opt-in** behind a new `transfers` feature flag (default **OFF**) — nothing changes for existing installs until an admin enables it under Settings → Features.

### Backend
- **Migrations 170–172** — `transfers`, `transfer_files` (cross-event photo refs), `transfer_extra_files` (admin-uploaded deliverables), `transfer_uploads` (client uploads), `transfer_recipients`, `transfer_downloads` (audit); default settings; `transfer_link_expired` + `transfer_ready` email templates (EN/DE). All `hasTable`/`hasColumn`-guarded, idempotent, sane `down()`.
- **transferService** — CRUD, 256-bit download token + 6-char upload token, cross-event ZIP streaming of **originals** (reuses the gallery download-selected loop — `resolvePhotoStorageKey → storage.get → archiver`, no watermarking). Max-downloads cap is evaluated against the freshly-incremented persisted count (no concurrent-download race).
- **Routes** — `/api/admin/transfers` (CRUD, multipart create with picked photos + the operator's own uploaded files, delivery by link or email, extra-file management, admin ZIP); public token routes `/api/public/transfer/:token` (thumbnail-free recipient view) and `/api/public/transfer-upload/:token` (6-char code, IP lockout + rate limit, dynamic multer, storage-backend agnostic). All three routers fail closed via `requireFeatureFlag('transfers')`.
- **transferCleanupService** — hourly sweep: disables expired links, emails the admin once, hard-deletes client uploads + extra files after the grace window. Source-event photos are never touched.

### Frontend
- Admin PicTransfer page: list + create (cross-event image picker, drag/drop of your own files, **Link / Email** delivery toggle with multi-recipient input) + detail panel (files, uploaded deliverables, recipient/upload link management, retention). Sidebar entry + route both flag-gated; Features-tab toggle.
- Public recipient download page + client-upload page — both read the branding theme CSS variables, so they follow the instance's colours and light/dark (public pages never get the admin `.dark` class).
- Full EN + DE i18n; displayed dates go through `useLocalizedDate`.

Reuses the existing original-file resolution, storage abstraction, download-filename service, public-token IP lockout, and email + cron patterns. One squashed commit; 23 files, feature is self-contained (builds + gating tests green on a clean `main`).

### Screenshots
_Admin list_
Screenshot 2026-08-06 at 13 07 46

_Create flow — upload your own files + Link/Email delivery toggle_
Screenshot 2026-08-06 at 13 08 54
Screenshot 2026-08-06 at 13 08 46

_Recipie
Screenshot 2026-08-06 at 13 11 19
nt download page (branding theme, dark)_

_Recipient email_
Screenshot 2026-08-06 at 13 09 34

### Notes for review
- Behind `transfers` flag end-to-end (route mounts, `RequireFeature`, sidebar, and `requireFeatureFlag` on all three routers).
- DE strings are author-provided (native de/en); worth a light native review of the email copy.

### DIFF
--- a/backend/__tests__/services/transferService.gating.test.js
+++ b/backend/__tests__/services/transferService.gating.test.js
@@ -1,0 +1,78 @@
+/**
+ * Unit tests for the pure gating logic in transferService (PicTransfer, #997).
+ * These exercise the download/upload eligibility rules without touching the DB.
+ */
+const transferService = require('../../src/services/transferService');
+
+const HOUR = 60 * 60 * 1000;
+
+function make(overrides = {}) {
+  return {
+    id: 1,
+    title: 'T',
+    is_active: true,
+    deleted_at: null,
+    expires_at: new Date(Date.now() + 24 * HOUR),
+    max_downloads: null,
+    download_count: 0,
+    allow_uploads: false,
+    upload_expires_at: null,
+    ...overrides,
+  };
+}
+
+describe('transferService.downloadsRemaining', () => {
+  it('returns null (unlimited) when no cap or zero cap', () => {
+    expect(transferService.downloadsRemaining(make({ max_downloads: null }))).toBeNull();
+    expect(transferService.downloadsRemaining(make({ max_downloads: 0 }))).toBeNull();
+  });
+
+  it('returns the remaining count and never goes negative', () => {
+    expect(transferService.downloadsRemaining(make({ max_downloads: 5, download_count: 2 }))).toBe(3);
+    expect(transferService.downloadsRemaining(make({ max_downloads: 5, download_count: 9 }))).toBe(0);
+  });
+});
+
+describe('transferService.computeStatus', () => {
+  it('is deleted when deleted_at set, regardless of activity', () => {
+    expect(transferService.computeStatus(make({ deleted_at: new Date(), is_active: true }))).toBe('deleted');
+  });
+  it('is expired when inactive or past expiry', () => {
+    expect(transferService.computeStatus(make({ is_active: false }))).toBe('expired');
+    expect(transferService.computeStatus(make({ expires_at: new Date(Date.now() - HOUR) }))).toBe('expired');
+  });
+  it('is active within the window', () => {
+    expect(transferService.computeStatus(make())).toBe('active');
+  });
+});
+
+describe('transferService.assertDownloadable', () => {
+  it('allows a live, in-window, uncapped transfer', () => {
+    expect(transferService.assertDownloadable(make()).ok).toBe(true);
+  });
+  it('404s a missing/deleted transfer', () => {
+    expect(transferService.assertDownloadable(null)).toMatchObject({ ok: false, status: 404 });
+    expect(transferService.assertDownloadable(make({ deleted_at: new Date() }))).toMatchObject({ ok: false, status: 404 });
+  });
+  it('410s when disabled or expired', () => {
+    expect(transferService.assertDownloadable(make({ is_active: false }))).toMatchObject({ ok: false, code: 'TRANSFER_DISABLED', status: 410 });
+    expect(transferService.assertDownloadable(make({ expires_at: new Date(Date.now() - HOUR) }))).toMatchObject({ ok: false, code: 'TRANSFER_EXPIRED', status: 410 });
+  });
+  it('410s when the download cap is reached', () => {
+    expect(transferService.assertDownloadable(make({ max_downloads: 2, download_count: 2 })))
+      .toMatchObject({ ok: false, code: 'DOWNLOAD_LIMIT_REACHED', status: 410 });
+  });
+});
+
+describe('transferService.assertUploadable', () => {
+  it('403s when uploads are disabled', () => {
+    expect(transferService.assertUploadable(make({ allow_uploads: false }))).toMatchObject({ ok: false, code: 'UPLOADS_DISABLED', status: 403 });
+  });
+  it('allows when uploads enabled and not expired', () => {
+    expect(transferService.assertUploadable(make({ allow_uploads: true })).ok).toBe(true);
+  });
+  it('410s when the upload window has passed', () => {
+    expect(transferService.assertUploadable(make({ allow_uploads: true, upload_expires_at: new Date(Date.now() - HOUR) })))
+      .toMatchObject({ ok: false, code: 'UPLOAD_EXPIRED', status: 410 });
+  });
+});
--- a/backend/migrations/core/170_add_transfers.js
+++ b/backend/migrations/core/170_add_transfers.js
@@ -1,0 +1,244 @@
+/**
+ * Migration 170: PicTransfer — cross-event file transfers (#997).
+ *
+ * Adds the tables that back the "send these files to someone" feature:
+ *
+ *   transfers          One share link. Bundles photos picked from ANY event,
+ *                      protected by a 64-hex recipient token. Optionally opens
+ *                      a 6-char upload token so the client can send files back
+ *                      (logos etc.). Disabled after `expires_at`; files are
+ *                      kept `grace_days` days past disable, then hard-deleted.
+ *   transfer_files     Join rows: which photos are in a transfer (cross-event).
+ *                      photo_id → photos CASCADE, so removing the underlying
+ *                      photo just drops it from the transfer; the reverse
+ *                      (deleting a transfer) never touches the source photos.
+ *   transfer_uploads   Files the client uploaded through the upload token.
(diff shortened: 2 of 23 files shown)


the-luap:
Reviewed the backend in full, frontend at diff level. Nice piece of work — the upload path in particular is more carefully hardened than most of what comes through here. Two blockers though, and they share a root cause: **permissions are used in the places that need ownership**.

Flagging up front that the `transfers` flag defaults **OFF** and the gating is genuinely thorough (all three routers, both mounts, sidebar, route), so no existing install is exposed. This is fixable before merge, not an incident.

## 🔴 Blockers

### 1. A scoped admin can exfiltrate any event's originals

`photoIds` arrives straight from the request body (`routes/adminTransfers.js:146-147`) and `addFiles` validates only that the photos exist:

```js
// services/transferService.js:357-359
// Only real, existing photos.
const existing = await db('photos').whereIn('id', ids).select('id');
```

There is no event-ownership filter — `filterOwnedEventIds` is not imported in either file, and `created_by` is only ever written (`transferService.js:137`) and serialized (`:209`), never used to scope anything.

So an admin holding `events.edit` posts arbitrary photo ids (up to the 5000 cap), receives a `download_url`, and `GET /:token/download` streams a ZIP of **unwatermarked originals** with no authentication at all. That is every ownership control in the product bypassed, with the originals as the payload and a public URL as the delivery.

Same class as GHSA-wrg5 / #991. The established fix applies: run `photoIds` through `filterOwnedEventIds(req.admin, …)` at creation (and on `POST /:id/files`), dropping anything the caller does not own.

### 2. The transfer list hands every admin everyone else's download tokens

`listTransfers` (`services/transferService.js:165-168`) filters on `deleted_at` alone, and `serializeTransfer` returns `token`, `upload_token` and `download_url` (`:206-224`). `GET /` requires only `events.view`.

That means blocker 1 is not even necessary to reach the same outcome: list the transfers, read another admin's 64-hex token, open `/transfer/<token>`, download their originals.

The gap runs through the whole router — detail, patch, delete, add/remove files, upload-link create/revoke and extra-file download all take an `:id` with no ownership predicate, so a scoped admin can also mutate or delete another admin's transfer.

Scoping reads and writes by `created_by` for non-super_admins closes it. Worth considering whether `token` needs to be in the list payload at all, or only on detail.

## 🟢 Nit

`generateUploadTokenCandidate` uses `bytes[i] % 31` (`transferService.js:54`). 256 % 31 = 8, so the first eight alphabet characters come up ~3% more often. Immaterial at 31^6 behind the lockout, but `crypto.randomInt(0, 31)` removes the bias for free.

- **Upload hardening** — per-IP bad-attempt lockout, separate info/upload rate limiters, `fileSize` + `files` caps, and the token guard runs **before multer** so bytes are never accepted ahead of validation. That ordering is the easy thing to get wrong and you got it right.
- **Migrations 170–172** — 6 / 4 / 1 `hasTable`/`hasColumn` guards, all three carry a `down()`, and every destructive statement (`dropTableIfExists`, the flag and template deletes) is confined to `down()`.
- **Download token** — 256-bit from `crypto.randomBytes(32)`, format-validated on the public route, rate-limited.
- **Feature-flag gating** — `requireFeatureFlag('transfers')` on all three routers, plus the `server.js` mounts and cleanup start. Fails closed.
- **Max-downloads cap** — evaluated against the freshly incremented persisted count, so the concurrent-download race you mention in the description is genuinely closed.

## Not verified

Frontend (~1600 lines across the picker, admin page, both public pages and i18n) reviewed at diff-stat and description level only, not line by line. Saying so rather than implying coverage. Note that once the backend scoping lands, the admin photo picker will also need to stop offering events the caller does not own — otherwise it just moves the dead-control problem from #969 into a new surface.

## CI

12 green. `bypass-size-gate` is red as designed for a 4106-line PR from a bypass user — that needs a maintainer approval, not a code change.

## Verdict

Close the two ownership gaps and I am happy to merge. Everything else here is solid, and the feature is well contained behind its flag.

Luca-Timo:
Thanks for the thorough pass — especially calling out that both blockers shared the one root cause. Pushed two commits (9e809e21, 80dc4be6); both ownership gaps closed with filterOwnedEventIds, so the model matches the rest of the events surface.

🔴 Blocker 1 — event exfiltration via photoIds
transferService.filterOwnedPhotoIds (transferService.js:68) resolves the ids to their events and runs them through filterOwnedEventIds, dropping anything foreign or non-existent. addFiles now takes the admin and gates on it (transferService.js:400), so both entry points are covered — POST /:id/files (adminTransfers.js:264) and create, which routes through addFiles internally.

🔴 Blocker 2 — token leak + unguarded router
listTransfers is scoped by created_by for non-super_admins (transferService.js:193), and the list payload drops token/upload_token/download_url/upload_url (transferService.js:227) — recipient/upload links live on the detail response only. Removed the list's copy shortcut accordingly.
One router.use('/:id', requireTransferOwnership) (adminTransfers.js:197, guard at :132) covers detail/patch/delete/files/upload-link/extra-file/download in one place — 404 on foreign and missing, so it's not an existence oracle.
🟢 Nit
generateUploadTokenCandidate uses crypto.randomInt(0, 31) (transferService.js:56) — bias gone.

Frontend follow-on
Caught the dead-control point — closed it in the same round. The picker now filters its event list to the same rule as the backend gate (TransferPhotoPicker.tsx:64): super_admin unrestricted, everyone else their own events plus ownerless legacy ones. Backend stays the enforcer; this just stops it offering picks the API would discard.

@the-luap ready for another look.

the-luap (approved):
Re-verified `9e809e21` + `80dc4be6` at file:line. Both blockers and the nit are closed, and the fix is wired everywhere it needs to be — which was the part most likely to go wrong.

## Blockers — closed

**1. Photo exfiltration.** `filterOwnedPhotoIds` (`transferService.js:68`) resolves ids to their events and runs them through `filterOwnedEventIds`, dropping foreign and non-existent alike. Wired on both entry points — `createTransfer:181` → `addFiles:400`, and `POST /:id/files` (`adminTransfers.js:264`). Both routes pass `req.admin` rather than a bare id, so `roleName` is present and super_admin isn't wrongly scoped by its own guard.

**2. Token leak + unguarded router.** Three parts, all correct:
- `listTransfers:193` scopes by `created_by`, and the route actually passes `admin: req.admin` (`:150`). That wiring is what would have quietly made the scoping inert.
- The list payload deletes `token`, `upload_token`, `download_url` and `upload_url` — defence in depth on top of the scoping, which is the right call.
- `router.use('/:id', requireTransferOwnership)` sits at `:197`, ahead of all twelve `/:id` routes (200–373). Ordering is the trap there and you avoided it. The guard 404s foreign and missing alike, with a super_admin bypass.

**Nit.** `crypto.randomInt(0, len)` with a comment explaining the bias it removes.

## Your frontend fix does more than you claimed

Worth saying: the picker filter is load-bearing, not just cosmetic. The events list only scopes for a hardcoded `roleName === 'editor'` (`adminEvents/crud.js:657`), so any *other* non-super_admin role — `admin`, or anything built in the Role editor — receives every event from that endpoint. Your client-side filter is what stops those users being offered picks the API would discard. I also confirmed it isn't inert: `created_by` survives to the client, since `mapEventForApi` (`helpers.js:168-190`) strips only the host/customer/password fields and spreads the rest.

## CI

The 12 green checks on this PR were stale — all seven workflow runs were against `f2418bab`, the pre-fix head. Fork PRs need "Approve and run" per push and it had not been granted, so nothing had run on `80dc4be6` for three days.

Rather than merge on stale signal I pushed the head to a temporary branch and dispatched the real `tests.yml` against it: **backend green (3m11s), frontend green (39s)**. Locally, `transferService.gating` 12/12 and ESLint clean on both changed backend files. Temp branch removed.

## One follow-up, not blocking

Neither guard has a regression test. `transferService.gating.test.js` covers `downloadsRemaining` / `computeStatus` / `assertDownloadable` — pure helpers, untouched by these commits. `filterOwnedPhotoIds` is already exported (`:989`), and `projectDealLineageOwnership.test.js` is the established harness for exactly this shape (`bootCrmDb` + `seedMinimal`).

Worth adding, because the whole PR turned on these two gaps and an untested guard is one refactor away from silently reopening — #999 just produced a live example, where the same fix landed in one place and was left stale in another. Tracking separately rather than holding the merge.

Approving.

PR #999, opened 2026-08-06
fix(branding): hide "Powered by PicPeak" on every page, not only the gallery

lbossuyt:
## Description

The `branding_hide_powered_by` setting only hid the attribution on the main gallery footer; it stayed visible on the gallery password screen, client access page, Premium layout, admin/customer login, accept-invite and CMS pages. This routes all of those through one small `<PoweredBy />` component in `components/common` that reads the public setting itself (same approach as `DynamicFavicon`/`RobotsMetaTags`) and renders nothing when white-labeling is on. It also collapses three duplicate translation keys (`gallery.poweredBy`, `adminLogin.poweredBy`, `customer.login.poweredBy`) into a single `common.poweredBy`, and a few pages that had "Powered by" hardcoded in English are now translated in all 8 locales.

No linked issue.

## Type of change

- [x] Code refactoring

## How Has This Been Tested?

Added `PoweredBy.test.tsx` (5 cases: default render, loading, hidden when the flag is set, explicit `false`, className/style forwarding); full frontend suite green on Node 20 (124 tests), `tsc` and ESLint clean. Verified both toggle states in the browser (local dev and a Docker env) on the login pages: the attribution shows when the setting is off and disappears when it's on.

- [x] Manual testing completed (both toggle states)

**Test Configuration**:
* PicPeak Version: main (3.100.1-beta.0)
* Node.js Version: 20.x
* Database: PostgreSQL
* Browser: Chrome, Safari

## Checklist:

- [ ] I have made corresponding changes to the documentation (N/A, no doc changes)
- [ ] Any dependent changes have been merged and published (N/A, no dependency changes)
- [ ] I have updated the CHANGELOG.md file (N/A, handled by release-please)

Setting OFF shows the attribution, setting ON removes it:

| Page | Setting OFF (shown) | Setting ON (hidden) |
|------|---------------------|---------------------|
| Admin login | image | image |
| Customer login | image | image |

## Additional Notes:

- Deliberately a small fix to get familiar with the contribution flow before proposing anything larger.
- Left `GalleryLayout.tsx` as-is: its "Powered by" is inline with the copyright line (different markup) and already respected the setting.
- "PicPeak" now renders semibold everywhere for consistency; the login pages showed it plain before.
- `<PoweredBy>` takes className/style per call site on purpose: the gallery/legal footers and the login screens already used different styles (fixed neutral gray vs themed text at 50% opacity), kept as-is to avoid visual changes. Could be folded into a `variant` prop if preferred, let me know
- Possible follow-up: "PicPeak" is a hardcoded literal in ~64 spots; a shared `PRODUCT_NAME` constant (separate from the customer's `branding_company_name` / `BRAND_TITLE`) could centralize it. Left out to keep this focused.

### DIFF
--- a/frontend/src/components/common/CMSContentBlock.tsx
+++ b/frontend/src/components/common/CMSContentBlock.tsx
@@ -5,6 +5,7 @@ import { useTranslation } from 'react-i18next';
 import DOMPurify from 'dompurify';
 import { Card } from './Card';
 import { Loading } from './Loading';
+import { PoweredBy } from './PoweredBy';
 import { cmsService } from '../../services/cms.service';
 import { usePublicSettings } from '../../hooks/usePublicSettings';
 import { buildResourceUrl } from '../../utils/url';
@@ -132,11 +133,7 @@ export const CMSContentBlock: React.FC<CMSContentBlockProps> = ({ slug, fallback
             {lang === 'de' ? 'Datenschutz' : 'Privacy Policy'}
           </Link>
         </div>
-        {!settings?.branding_hide_powered_by && (
-          <p className="mt-2">
-            Powered by <span className="font-semibold">PicPeak</span>
-          </p>
-        )}
+        <PoweredBy className="mt-2" />
       </footer>
     </div>
   );
--- a/frontend/src/components/common/PoweredBy.tsx
+++ b/frontend/src/components/common/PoweredBy.tsx
@@ -3,32 +3,17 @@ import { useTranslation } from 'react-i18next';
 import { usePublicSettings } from '../../hooks/usePublicSettings';
 
 interface PoweredByProps {
-  /** Extra classes applied to the wrapping `<p>`, for per-context styling. */
   className?: string;
-  /** Inline styles applied to the wrapping `<p>`. */
   style?: React.CSSProperties;
 }
 
-/**
- * Renders the "Powered by PicPeak" attribution shown in public footers and
- * login screens.
- *
- * The component reads the public `branding_hide_powered_by` setting itself —
- * like `DynamicFavicon` / `RobotsMetaTags` — and renders nothing when the admin
- * has enabled white-labeling. Keeping the visibility check inside the component
- * lets every call site drop in `<PoweredBy />` instead of repeating the guard
- * or drilling the flag down through layout props.
- *
- * Only the "Powered by" prefix is translated (`common.poweredBy`); the brand
- * name is rendered verbatim and never localized.
- */
+// "Powered by PicPeak" footer/login attribution. Reads the setting itself (like
+// DynamicFavicon) so callers don't repeat the guard; hidden for white-label.
 export const PoweredBy: React.FC<PoweredByProps> = ({ className, style }) => {
   const { t } = useTranslation();
   const { data: settings } = usePublicSettings();
 
-  if (settings?.branding_hide_powered_by) {
-    return null;
-  }
+  if (settings?.branding_hide_powered_by) return null;
 
   return (
     <p className={className} style={style}>
--- a/frontend/src/components/common/PoweredBy.tsx
+++ b/frontend/src/components/common/PoweredBy.tsx
@@ -1,0 +1,38 @@
+import React from 'react';
+import { useTranslation } from 'react-i18next';
+import { usePublicSettings } from '../../hooks/usePublicSettings';
+
+interface PoweredByProps {
+  /** Extra classes applied to the wrapping `<p>`, for per-context styling. */
+  className?: string;
+  /** Inline styles applied to the wrapping `<p>`. */
+  style?: React.CSSProperties;
+}
+
+/**
+ * Renders the "Powered by PicPeak" attribution shown in public footers and
+ * login screens.
+ *
+ * The component reads the public `branding_hide_powered_by` setting itself —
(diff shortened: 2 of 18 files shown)


the-luap:
Thanks — the diagnosis is right and the shape of the fix is the one I'd have asked for.

I confirmed the premise rather than taking it on trust: `AdminLoginPage` rendered `t('adminLogin.poweredBy')` unconditionally, with no guard at all, so `branding_hide_powered_by` genuinely did not apply there. An operator toggling "hide powered by" and still seeing it on six surfaces is a broken promise, so this is a bug fix rather than a feature request. Routing it through one self-reading component (the `DynamicFavicon` pattern you cite) is the right call, and collapsing the three duplicate keys into `common.poweredBy` is a nice bonus.

## 🟡 One thing to change before merge

`PoweredBy.tsx:17` renders the attribution while the settings are still loading:

```tsx
if (settings?.branding_hide_powered_by) return null;
```

`undefined` is falsy, so a white-labelled instance briefly **flashes "Powered by PicPeak"** before it disappears. That is exactly what the setting exists to prevent, and the worst case is a login page — often the first thing loaded, before `usePublicSettings` (react-query, `staleTime: 60_000`) has resolved.

I can see it is deliberate rather than an oversight, since `PoweredBy.test.tsx:36` pins it (`still renders while settings are loading`), so I want to argue the case rather than just ask for a change. The tradeoff is asymmetric:

- Render-while-unknown: a white-label customer sees the attribution flash on first paint. That is the failure the feature exists to prevent, and it is visible to the operator's own client.
- Hide-while-unknown: an ordinary instance briefly lacks attribution that appears a fraction of a second later. Nobody notices.

So `if (!settings) return null` is the safer default, with the test inverted to match. Happy to hear the counter-argument if you had a specific reason for the current choice.

## 🔵 CI has not run yet — that is on us, not you

This is a fork PR from a first-time contributor, so the workflows need a maintainer to click "Approve and run". Right now only GitGuardian has reported (1 check instead of the usual 13), so none of the build, lint or test signal exists on our side yet. I will approve the run — nothing for you to do.

For what it is worth, your numbers corroborate: our frontend suite was 119 before this, and 119 + your 5 new `PoweredBy` cases = the 124 you report.

## ✅ Verified

- All 8 locales carry `common.poweredBy` with real translations rather than English placeholders (`de` "Bereitgestellt von", `ru` "Работает на", `sl` "Poganja", …). Easy to leave half-done; you did not.
- No dangling references to the three removed keys anywhere in the tree.
- Component matches the existing self-reading pattern, so callers do not repeat the guard.
- Frontend-only; no backend or security surface.

## One question, not a blocker

The change also hides the attribution on the **admin login**, which only the operator ever sees — so that surface gets no white-label benefit, just consistency. I lean towards keeping it consistent, but flagging it in case you deliberately scoped it that way.

## Verdict

Resolve the loading flash and get a green CI run, and I am happy to merge this.

lbossuyt:
Thanks for your review! Fair call on the flash, no counter-argument. Changed it to `if (!settings || settings.branding_hide_powered_by) return null;` and inverted the loading test.

Two quick calls for you:

- Admin login: hiding it there was deliberate, for consistency (an operator can screenshot or hand off that screen, so a lone "Powered by PicPeak" would look like a bug). I lean the same way you do. If you'd rather, I can add a prop to override the setting per placement. Your call.
- Styling: I added a note on this in the description, probably after you'd reviewed, so flagging it here too. it's passed per call site (className / style) because the gallery footers and login screens already used different styles, kept as-is to avoid a visual change. Three ways to go: leave it, fold the two recipes into a named variant, or unify to one style everywhere (that last one slightly changes the footers' look). Variant example:
```tsx
// now
<PoweredBy className="text-xs mt-2 text-neutral-500" />
<PoweredBy className="text-xs mt-2" style={{ color: 'var(--color-text)', opacity: 0.5 }} />

// with a variant
<PoweredBy variant="footer" />
<PoweredBy variant="login" />
```

the-luap (approved):
`cf243b44` does exactly the right thing, and the test is properly inverted rather than loosely flipped — it asserts an empty DOM now, not just the absence of one string. Thanks for turning that round quickly.

One correction on my own advice: I suggested `!settings` bluntly, and afterwards realised `isLoading` would be more precise, since on a *failed* settings fetch `data` stays undefined forever and the attribution would disappear everywhere rather than just while loading. Having thought it through it does not matter here — if the public-settings request fails, the page has already lost its theme, logo and favicon, because they all read the same endpoint. A missing footer line is not noticeable on an already-unbranded page. No change wanted.

## On CI

Our fork-PR workflows never got approved, so there is still no run on this PR. Rather than merge on faith I verified locally against `cf243b44`:

- `tsc --noEmit` — clean
- ESLint on both changed files — clean
- `vitest run` — **124 passed across 24 files**, matching your number exactly
- `npm run build` — succeeds

That covers the whole relevant surface for a frontend-only diff; the backend job would have had nothing to say about it.

## Follow-up, not for this PR

`GalleryLayout.tsx:705` keeps its own inline `!brandingSettings?.hide_powered_by` guard, so the main gallery footer is not routed through `<PoweredBy />` and still has the flash your commit just fixed everywhere else. Not a blocker — that footer already honours the setting, which is why it was out of scope — but I am filing it separately so the second implementation does not quietly drift from the first. Exactly the pattern that produced this bug in the first place.

Approving. Thanks for a clean first contribution — the locale coverage in particular was more thorough than it needed to be.

Issue #1029, opened 2026-08-11
[BUG] Expiration date should be deleteable when not required by setting


Apokalypser:
The expiration date of an event should be deletable when the setting to require it is switched off. 

**To Reproduce**
1. Disable "Require expiration date"
2. Try to delete expiration date
3. Error: Failed to update data

**Expected behavior**
Expiration date should be deletable

**Environment (please complete the following information):**
 - OS: Talos OS
 - Browser: Chrome 151
 - PicPeak Version: v3.102.1-beta0
 - Deployment Method: Kubernetes
 - Database: PostgreSQL 15

the-luap:
Clearing the expiration on edit has been explicitly allowed since `e5445613` ("admins can clear `expires_at` on edit even when Require expiration is ON", #426), which shipped in v3.44.1-beta.0:

- `crud.js:1498` — an empty/missing `expires_at` normalizes to `NULL` ("never expires")
- `crud.js:1183` — the validator is `optional({ nullable: true, checkFalsy: true })`, so `null` and `""` both pass
- migration `061` drops the `NOT NULL` on `events.expires_at` for Postgres
- the frontend has no expiration validation left on edit either (`EventDetailsPage.tsx:454`)

I re-tested on `main` (3.102.1-beta.0) against PG 15 with both a minimal body and the exact payload the edit form sends — `PUT /api/admin/events/<id>` with `expires_at: null` returns 200 and the column goes `NULL`.

So whatever fails on your instance is not this gate. Can you grab:

1. the failing request in DevTools → Network: status code and response body for the `PUT /api/admin/events/<id>`, and
2. the matching backend log line.

A 400 will name the field that's being rejected; a 500 points at your DB (worth checking that migrations completed on that Kubernetes deploy — `061` in particular, if the install was ever restored from a dump).

Apokalypser:
[...]

Issue #1028, opened 2026-08-11
[BUG] Download button in gallery header not hidden when photo download not allowed


Apokalypser:
The download button is not hidden, even when the photo download is not allowed.

**To Reproduce**
1. Disallow photo download
2. Go to gallery root

**Expected behavior**
Download button should be hidden

**Environment (please complete the following information):**
 - OS: Talos OS
 - Browser: Chrome 151
 - PicPeak Version: v3.102.1-beta0
 - Deployment Method: Kubernetes
 - Database: PostgreSQL 15

<img width="1287" height="57" alt="Image" src="https://github.com/user-attachments/assets/5fb80ff9-a9c8-475d-92ba-35104b945e26" />

the-luap:
Can't reproduce the event-level path on `main` (3.102.1-beta.0). Traced it end to end against a local PG instance:

- `events.allow_downloads = false` → `GET /api/gallery/:slug/info` returns `"allow_downloads": false` (`gallery.js:314`)
- same for the payload the gallery actually renders from, `GET /api/gallery/:slug/photos` (`gallery.js:909`)
- `GalleryView.tsx:613` computes `allowDownloads = !isExpired && data?.event?.allow_downloads === true` — strict — and that is what gates `showHeaderDownload` (`GalleryView.tsx:991`). `GalleryLayout` is the only component that renders the header button, and all three header styles are behind that flag.

`PUT /api/admin/events/:id {"allow_downloads": false}` also persists correctly, so the admin save path is fine too.

Two things I need to place your case:

1. Which toggle did you switch off — **Allow photo downloads** on the event (Event → Edit), or the per-category download toggle under **Categories**? The header Download-all is only gated on the event-level flag; per-category blocking (#640) is not considered there. If you turned downloads off per category, the header button stays and clicking it returns `404 No photos found` (`gallery.js:1392`) — that's a real gap on our side and I'll fix it either way.
2. If it was the event-level toggle: reload the gallery with DevTools open and paste what `GET /api/gallery/<slug>/photos` returns for `event.allow_downloads`. If it says `true` while the admin form shows the toggle off, the write isn't landing on your instance and the backend log for the `PUT /api/admin/events/<id>` is the next thing to look at.

Apokalypser:
I used the event settings to disable downloads. It is persisted correctly
<img width="778" height="578" alt="Image" src="https://github.com/user-attachments/assets/7920c34a-e11f-4bfe-bf0e-2ce357303e6e" />

[...]

PR #1039, opened 2026-08-12
fix(docker): default NODE_ENV=production so non-compose deploys don't fall back to SQLite (#1038)

### LINKED ISSUE #1038: [BUG] NODE_ENV unset in the Docker image makes non-compose deployments silently run on SQLite
## Summary

`backend/Dockerfile` never sets `NODE_ENV`. `knexfile.js` selects its config block by `NODE_ENV`, and the `development` block defaults to `sqlite3`. So **every deployment that doesn't go through our compose files — Kubernetes, Helm, Nomad, plain `docker run` — silently runs on SQLite and ignores `DB_HOST` / `DB_USER` / `DB_PASSWORD` entirely.**

Nothing logs which engine is in use, so affected operators have no way to notice. Surfaced while triaging #1028 and #1029, both of which turned out to be SQLite-only bugs hit by a reporter who had correctly provisioned PostgreSQL 15.

## Why it's invisible

`wait-for-db.sh` is a shell script that reads `DB_HOST` directly and never consults `NODE_ENV`. It connects to Postgres, creates the database, and reports success — in the same container, in the same log, where the Node process then goes to SQLite:

```
Waiting for PostgreSQL at localhost:5432...
PostgreSQL is up - checking if database "picpeak" needs to be created...
Target database "picpeak" is ready.
Ensuring storage directories exist...
Running database migrations...
Starting database migrations...          ← run-migrations.js, i.e. NODE_ENV != production
Existing deployment detected - checking all migrations
No pending migrations
All migrations completed successfully    ← ran against the SQLite file
```

That log reads as a healthy Postgres install. It isn't one.

## Mechanism

1. `backend/Dockerfile` — no `ENV NODE_ENV`. Only `docker-compose.yml:42` (`NODE_ENV=${NODE_ENV:-production}`) and `docker-compose.production.yml:97` supply it.
2. `knexfile.js:134` — `const env = process.env.NODE_ENV || 'development'`.
3. `knexfile.js:54` — the `development` block is `client: process.env.DATABASE_CLIENT || 'sqlite3'`, and its connection block ignores every `DB_*` variable.
4. `wait-for-db.sh:150` picks the migration runner off the same variable, which is how the log above identifies the state precisely: `Starting database migrations...` and `No pending migrations` exist **only** in `run-migrations.js`, never in `run-migrations-safe.js`.
5. `run-migrations.js` imports `src/database/db.js` → the same knexfile → so migrations run against SQLite too. The Postgres database that `wait-for-db.sh` created stays empty.

## Impact

- Kubernetes/Helm operators are running on SQLite without knowing: no concurrent writers, no second backend replica, and the SQLite file may sit somewhere that isn't backed up.
- The Postgres database they provisioned (and back up) is empty.
- Every SQLite-only bug lands on them — #1028 (boolean flags read backwards; disabling downloads was completely inert) and #1029 (expiration could never be cleared) were both found by one such operator in a single day.

## The trap in the obvious fix

Adding `ENV NODE_ENV=production` alone is not safe on its own: an affected install would flip to Postgres on its next image pull and come up against an **empty** database — indistinguishable from total data loss. The fix needs a guard that refuses to start in exactly that state and explains the migration path (`.picpeak` export is engine-neutral, so it carries SQLite → Postgres).

## Fix

1. `ENV NODE_ENV=production` in `backend/Dockerfile` (compose still overrides it, so nothing changes for compose users).
2. Log the resolved engine and target at boot.
3. Refuse to start when pointed at a virgin Postgres while a populated SQLite file exists, with instructions and an explicit `PICPEAK_ALLOW_EMPTY_PG=true` escape hatch.
4. Warn (but boot) when Postgres settings are present yet SQLite is in use.

## Affected operators — how to check

```sh
kubectl exec <backend-pod> -- printenv NODE_ENV DATABASE_CLIENT
kubectl exec <backend-pod> -- ls -la /app/data/photo_sharing.db
```

A `photo_sharing.db` with a recent mtime, or an empty `\dt` in the Postgres database, means the app was on SQLite.

the-luap:
Closes #1038.

`backend/Dockerfile` never set `NODE_ENV`. `knexfile.js:134` picks its config block off that variable, and the `development` block defaults to `sqlite3` (`:54`) while ignoring every `DB_*` setting. So **every deployment that doesn't go through our compose files — Kubernetes, Helm, Nomad, plain `docker run` — silently ran on SQLite**, no matter how correctly Postgres was configured.

It stayed invisible because `wait-for-db.sh` is a shell script: it reads `DB_HOST` directly, connects to Postgres, creates the database, and logs `PostgreSQL is up` in the same container where the Node process then writes to a SQLite file. `run-migrations.js` imports `src/database/db.js` → the same knexfile, so the migrations also ran against SQLite and the provisioned Postgres database stayed empty.

Found while triaging #1028 and #1029 — both SQLite-only bugs, reported by an operator running Kubernetes with PostgreSQL 15 correctly provisioned. Their own startup log identifies the state precisely: it contains `Starting database migrations...` and `No pending migrations`, strings that exist only in `run-migrations.js`, which `wait-for-db.sh:150` runs **only** when `NODE_ENV != production`.

## Why the one-line fix isn't enough

Setting `ENV NODE_ENV=production` alone would flip every affected install to Postgres on its next image pull — landing on an **empty** database, which reads as total data loss. So the default comes with a guard.

## Changes

- **`backend/Dockerfile`** — `ENV NODE_ENV=production`. Both compose files set it explicitly already, so compose users see no change.
- **`backend/src/utils/databaseEngine.js`** (new) — engine description + the guard decision, kept as a pure function with thin IO probes so the interesting part is testable without a live Postgres.
- **`backend/server.js`** — the boot line now names the engine and target. Nothing logged this before, which is the reason this went unnoticed for so long. Credentials are never included (pinned by a test).
- **both migration runners** — the guard runs before any schema work:
  - **blocks** when pointed at a virgin Postgres while a populated SQLite file exists, naming the file and walking through the `.picpeak` export path (deliberately engine-neutral NDJSON, so it carries SQLite → Postgres). Escape hatch: `PICPEAK_ALLOW_EMPTY_PG=true`.
  - **warns but boots** when Postgres settings are present yet SQLite is in use.

Blocked-boot output, exercised against a real populated SQLite file:

```
[info] Database engine: postgres (pg:5432/picpeak)
REFUSING TO START — this instance is configured for postgres (pg:5432/picpeak), but that
database is empty while a populated SQLite database exists at:

  /app/data/photo_sharing.db
  …
```

## Tests

`backend/__tests__/utils/databaseEngine.test.js` — 13 tests:

- unset `NODE_ENV` resolves to sqlite3; `NODE_ENV=production` resolves to pg. Both resolve `knexfile` **in a child process with a clean cwd** — `dotenv.config()` would otherwise let the developer's `backend/.env` decide the answer instead of the defaults under test (it does set `DATABASE_CLIENT=sqlite3` locally, which is how I caught it).
- the Dockerfile pins `NODE_ENV=production`
- the boot line names host/port/database and never leaks the password
- the guard blocks exactly one state (empty PG + populated SQLite) and leaves the three neighbouring states alone

Full backend suite: failure set byte-identical to `main`'s baseline. Lint clean on both new files; the 3 errors reported in the migration runners are pre-existing and unchanged.

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -27,6 +27,15 @@ FROM node:22-alpine
 
 WORKDIR /app
 
+# knexfile.js picks its config block by NODE_ENV, and the `development` block
+# defaults to sqlite3. Leaving NODE_ENV unset here meant every deployment that
+# doesn't go through our compose files — Kubernetes, Helm, plain `docker run` —
+# silently ran on SQLite and ignored DB_HOST/DB_USER/DB_PASSWORD, while
+# wait-for-db.sh (shell, reads DB_HOST directly) reported "PostgreSQL is up" in
+# the same log. The compose files still override this, so nothing changes for
+# compose users. See #1038.
+ENV NODE_ENV=production
+
 # Redeclare CACHEBUST — ARGs don't cross stage boundaries, so the builder
 # stage's declaration never reached this stage. Consuming it in the RUN below
 # busts that layer's cache every CI run (CACHEBUST=github.run_number), so the
--- a/backend/__tests__/utils/databaseEngine.test.js
+++ b/backend/__tests__/utils/databaseEngine.test.js
@@ -28,10 +28,10 @@ const {
   migrationInProgressPath,
   hasMigrationInProgress,
   isUntouchedBootstrapRow,
+  adminsIndicateUse,
 } = require('../../src/utils/databaseEngine');
 const {
   epochToIso,
-  parseJsonText,
   coerceForTargetEngine,
 } = require('../../src/services/picpeakImportService');
 
@@ -224,54 +224,6 @@ describe('cross-engine row coercion (#1038)', () => {
   });
 });
 
-describe('SQLite JSON text is decoded before Postgres serialisation (#1038 review)', () => {
-  // SQLite has no json type: its json columns are TEXT holding JSON. The export
-  // dumps that as a STRING, and serialiseJsonColumns would then stringify it a
-  // SECOND time, so Postgres stored `true` as the scalar string "true".
-  // app_settings.setting_value is json on every install, so this reshaped every
-  // migrated setting.
-  test('a JSON object in text form is decoded', () => {
-    expect(parseJsonText('{"primaryColor":"#123456"}')).toEqual({ primaryColor: '#123456' });
-  });
-
-  test('JSON scalars are decoded to their real type', () => {
-    expect(parseJsonText('true')).toBe(true);
-    expect(parseJsonText('42')).toBe(42);
-    expect(parseJsonText('[1,2]')).toEqual([1, 2]);
-  });
-
-  test('a plain word stays a plain word — text columns are not reinterpreted', () => {
-    expect(parseJsonText('PicPeak')).toBe('PicPeak');
-    expect(parseJsonText('hero')).toBe('hero');
-    expect(parseJsonText('')).toBe('');
-  });
-
-  test('malformed JSON is left untouched rather than thrown away', () => {
-    expect(parseJsonText('{not json')).toBe('{not json');
-  });
-
-  test('non-strings pass through', () => {
-    expect(parseJsonText(null)).toBeNull();
-    expect(parseJsonText(7)).toBe(7);
-    expect(parseJsonText({ a: 1 })).toEqual({ a: 1 });
-  });
-
-  test('json columns are decoded alongside the timestamp/boolean coercion', () => {
-    const [out] = coerceForTargetEngine(
-      [{ setting_value: 'true', created_at: 1786548038763, feedback_enabled: 1, name: 'plain' }],
-      {
-        timestamps: ['created_at'],
-        booleans: ['feedback_enabled'],
-        jsonCols: new Set(['setting_value']),
-      },
-    );
-    expect(out.setting_value).toBe(true);
-    expect(out.created_at).toBe('2026-08-12T15:20:38.763Z');
-    expect(out.feedback_enabled).toBe(true);
-    expect(out.name).toBe('plain');
-  });
-});
-
 describe('probeSqliteData fails closed (#1038 review)', () => {
   function tmpDb(contents) {
     const dir = fs.mkdtempSync(path.join(os.tmpdir(), 'picpeak-probe-'));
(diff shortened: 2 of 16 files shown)


the-luap:
## Updated: the guard no longer blocks the boot, and there is a migration path

Walking through what an existing install experiences on its next image pull changed the design.

**The block was the wrong trade.** An operator who had unknowingly been running on SQLite would have pulled this fix and got a `CrashLoopBackOff` — data safe, galleries offline, for something they didn't do. The boot now resolves the engine *before* migrations run and stays on whichever one holds the data.

### What each install actually gets

| Install | Before the pull | After the pull |
|---|---|---|
| Compose + Postgres (most users) | Postgres | Postgres — unchanged, `NODE_ENV` already set by compose |
| Kubernetes/Helm, Postgres provisioned but empty (this issue) | SQLite, silently | **SQLite, loudly** — same data, same uptime, plus a boot banner explaining it and naming the migration command |
| Same install, after running the migration script | — | Postgres, automatically, on the next restart |
| Deliberate SQLite (`DATABASE_CLIENT=sqlite3`) | SQLite | SQLite — an explicit setting is always honoured |
| Fresh install, no SQLite file | Postgres | Postgres |

The check is keyed on Postgres holding **data**, not on it having tables: a stray `run-migrations` against the empty database creates every table, which would otherwise blind the check and strand the operator on an empty install. Manual migration runs now refuse rather than build a schema in the wrong database.

### Migrating SQLite → Postgres

```
node scripts/migrate-sqlite-to-postgres.js [--force] [--keep-archive]
```

It reuses the `.picpeak` export/import services rather than hand-rolling a cross-engine copy — they already handle FK suspension, JSON columns and Postgres sequence resync. Photo files are never touched (only rows move) and the SQLite file is left in place as a rollback.

Two additions were needed for the SQLite → Postgres direction, both opt-in and CLI-only so the upload/restore UI is untouched:

- `allowEngineSwitch` relaxes the importer's same-engine guard
- **cross-engine row coercion.** SQLite has no real date or boolean types, so its rows carry epoch numbers where Postgres wants a timestamp and `0`/`1` where it wants a boolean. Postgres rejects both outright — the first real run failed with `date/time field value out of range: "1786548038763"`. Coercion is driven by the **target** schema, never guessed from the value.

### Verified against a real PostgreSQL 15

Not just unit tests — a seeded SQLite install was migrated end to end:

```
 id |          slug          | allow_downloads |         created_at
  1 | migration-test-gallery | f               | 2026-08-12 15:20:47.624+00
```

- booleans landed as real booleans (`0` → `f`, `1` → `t`)
- epoch timestamps landed as real Postgres timestamps
- foreign keys intact (3 photos → event 1)
- **sequences advanced**: the next `INSERT` returned id 2 and photo id 4, not a primary-key collision — the failure mode this class of migration usually ships with

All five engine-resolution scenarios were exercised against a live Postgres as well (stranded → sqlite, post-migration → pg, fresh → pg, forced → pg, deliberate sqlite → sqlite).

20 tests. Full backend suite unchanged from `main`'s baseline.

the-luap:
## External review: 6 rounds, 21 findings, all fixed

Ran this branch through repeated adversarial review (OpenAI Codex CLI) with a data-loss focus, verifying every finding against the code and every fix against a real PostgreSQL 15 — not just unit tests. **Every finding was valid; none were false positives.** Two of them were regressions introduced by fixes from an earlier round, which is the main argument for having kept going.

### Round 1 — 4 findings

| # | Finding | Why it mattered |
|---|---|---|
| P1 | Resolver only reachable via `wait-for-db.sh` | A K8s manifest with `command:`/`args:`, or `docker run … node server.js`, bypasses the entrypoint — precisely this population. With `NODE_ENV` now baked in, such an install would have booted an empty Postgres with its SQLite data unseen. `server.js` resolves the engine itself now. |
| P1 | Concurrent writes during migration | Rows written after the export never reach Postgres and vanish once the engine switches. |
| P2 | Double-encoded JSON on cross-engine load | SQLite json columns are TEXT holding JSON; `serialiseJsonColumns` stringified them again. `app_settings.setting_value` is `json` on **every** install, so this silently reshaped every migrated setting. |
| P2 | Child logs shared stdout with the payload protocol | With `LOG_TO_CONSOLE=true` the archive path came back concatenated with a log line and the migration died on a bogus filename. |

### Round 2 — 6 findings

Engine choice pinned by a durable marker (a Postgres that merely *looks* empty later — every gallery deleted — no longer sends the install back to a stale SQLite file); target/source occupancy no longer judged by galleries alone (admins, customers, invoices count); drift detection widened from 4 tables to every exported table plus `max(updated_at)`; `probeSqliteData` fails **closed** on an unreadable file; the "leaving SQLite behind" warning made reachable.

### Round 3 — 6 findings

Both probes were still judging occupancy by galleries only; the target was checked *after* schema creation, so the bootstrap admin seeded by `ADMIN_PASSWORD` made the script demand `--force` against an empty database; the probe's warning went through the app logger (stdout) and could have become `DATABASE_CLIENT`; the `.picpeak` archive — plaintext password hashes, SMTP credentials, API keys — survived every failure path; a database-only migration still hauled every business doc through `/tmp`.

### Round 4 — 1 finding

A migration that dies after touching Postgres leaves rows behind, which read as "occupied" and would flip the engine on the next boot. The script now writes a pin file **before** its first Postgres write and clears it only on success.

### Round 5 — 4 findings

The bootstrap admin counted as occupancy (regression from round 3 — the two sides are deliberately asymmetric now); the guard resolved `DATABASE_PATH` with its own logic while knexfile trimmed/collapsed it, so it could probe a file nobody uses; retrying after a partial run demanded `--force`; `wait-for-db.sh` verified readiness against `postgres` while knexfile's production default is `db`.

### Verified against real PostgreSQL 15

- SQLite → Postgres migration: booleans (`0` → `f`), epoch timestamps → real timestamps, FKs intact, **sequences advanced** (next `INSERT` got id 2, not a collision), `json_typeof(setting_value)` = `boolean`, matching a native install exactly
- `node server.js` with the entrypoint bypassed → resolves to SQLite and logs the banner
- A live backend writing during the copy → caught **before** Postgres is touched; Postgres left untouched; secret-bearing archive removed
- Failed migration leaving a bootstrap admin in Postgres → next boot stays on SQLite (without the pin it goes to Postgres — the failure this closes)
- Marker lifecycle: pinned after success even with every gallery deleted; remove the marker + restore the file to roll back
- `DATABASE_PATH` with surrounding whitespace resolves identically in knexfile and the guard

34 unit tests. Full backend suite failure set byte-identical to `main`'s baseline. Lint parity with `main` on every touched file.

### Note on #1043

@Luca-Timo's #1043 overlaps the importer changes here (same-engine gate + coercion). Per discussion there, this merges first and #1043 rebases onto it, keeping its direction-policy version of the gate.

PR #1040, opened 2026-08-12
fix(docker): default NODE_ENV=production so non-compose deploys don't fall back to SQLite (#1038)

the-luap:
Stable backport of #1039 (#1038). Clean cherry-pick — the mechanism is identical on `stable`:

- `backend/Dockerfile` sets no `NODE_ENV`
- `knexfile.js:134` — `process.env.NODE_ENV || 'development'`
- `knexfile.js:56` — the `development` block is `DATABASE_CLIENT || 'sqlite3'`, ignoring every `DB_*` variable
- `wait-for-db.sh:128-130` picks the migration runner off the same variable

So every Kubernetes / Helm / plain `docker run` deployment on stable is silently on SQLite too, with the provisioned Postgres database left empty.

Same three parts as #1039: `ENV NODE_ENV=production`, a boot line naming the resolved engine, and the guard that refuses to start against a virgin Postgres while a populated SQLite file exists (with `PICPEAK_ALLOW_EMPTY_PG=true` as the escape hatch). Compose users are unaffected — both compose files set `NODE_ENV` explicitly.

13 tests pass on this branch. Full backend suite: failure set byte-identical to pristine `stable`. Lint clean on both new files.

See #1039 for the full write-up.

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -27,6 +27,15 @@ FROM node:22-alpine
 
 WORKDIR /app
 
+# knexfile.js picks its config block by NODE_ENV, and the `development` block
+# defaults to sqlite3. Leaving NODE_ENV unset here meant every deployment that
+# doesn't go through our compose files — Kubernetes, Helm, plain `docker run` —
+# silently ran on SQLite and ignored DB_HOST/DB_USER/DB_PASSWORD, while
+# wait-for-db.sh (shell, reads DB_HOST directly) reported "PostgreSQL is up" in
+# the same log. The compose files still override this, so nothing changes for
+# compose users. See #1038.
+ENV NODE_ENV=production
+
 # Redeclare CACHEBUST — ARGs don't cross stage boundaries, so the builder
 # stage's declaration never reached this stage. Consuming it in the RUN below
 # busts that layer's cache every CI run (CACHEBUST=github.run_number), so the
--- a/backend/__tests__/utils/databaseEngine.test.js
+++ b/backend/__tests__/utils/databaseEngine.test.js
@@ -28,10 +28,10 @@ const {
   migrationInProgressPath,
   hasMigrationInProgress,
   isUntouchedBootstrapRow,
+  adminsIndicateUse,
 } = require('../../src/utils/databaseEngine');
 const {
   epochToIso,
-  parseJsonText,
   coerceForTargetEngine,
 } = require('../../src/services/picpeakImportService');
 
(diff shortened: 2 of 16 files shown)


the-luap:
## Updated to match #1039 — no boot block, plus a migration path

Same rework as the beta PR: the boot resolves the engine before migrations run and **stays on whichever engine holds the data**, instead of refusing to start. An affected operator keeps their uptime and their data, gets a boot banner explaining what happened, and moves when they choose.

Adds `scripts/migrate-sqlite-to-postgres.js`. See #1039 for the full write-up and the behaviour matrix.

### One extra delta on this branch

**`resyncSequences()` does not exist in this branch's import service** — it landed on `main` only. Without it, a cross-engine load leaves Postgres identity sequences at 1 and the very next insert collides on the primary key, which would have made the migration script quietly produce a broken install here.

So it is backported in this PR and called **only** on the cross-engine path:

```js
if (allowEngineSwitch) await resyncSequences(tables);
```

Same-engine restores through the UI keep their current behaviour exactly — this branch's restore path is not otherwise touched.

### Verified on this branch against a real PostgreSQL 15

A seeded SQLite install migrated across: `allow_downloads` `0` → `f`, epoch timestamps → real Postgres timestamps, 2 photos with their foreign key intact, and the next `INSERT` returned id 2 rather than colliding — i.e. the backported sequence resync does its job.

20 tests pass. Full backend suite: failure set byte-identical to pristine `stable`. Lint: the 3 errors in `picpeakImportService.js` are pre-existing and unchanged.

the-luap:
## Brought up to parity with #1039 — all 25 review findings

This branch was cut before the external review of #1039 started, so it carried the original implementation and **none** of the fixes that review produced. That is corrected: all six review-fix commits are cherry-picked here, so stable gets at minimum everything main gets.

Rounds 1–6 fixed on this branch too (full write-up in #1039):

- the resolver was reachable only via `wait-for-db.sh`, so a Kubernetes `command:`/`args:` override or `docker run … node server.js` bypassed it and booted an empty Postgres
- SQLite json columns were double-encoded on the cross-engine load — `app_settings.setting_value` is `json` on every install, so every migrated setting was reshaped
- the engine choice was not pinned, so deleting every gallery later sent an install back to a stale SQLite file
- an unreadable SQLite file was read as "no data" and switched the install to an empty Postgres
- a failed migration left rows in Postgres that flipped the engine on the next boot
- `DATABASE_CLIENT=pg` bypassed that pin; with `=sqlite3`, a *successful* migration served a freshly-created empty database
- occupancy judged by galleries only; a seeded bootstrap admin counted as real data
- `DATABASE_PATH` resolved differently than knexfile, so the guard could probe a file nobody uses
- the `.picpeak` archive — plaintext hashes, SMTP credentials, API keys — survived every failure path
- drift detection covered four tables and missed in-place updates

### Verified independently on THIS branch, not assumed from main

Stable's import service differs in two ways that matter here, so the whole battery was re-run against a real PostgreSQL 15:

- it has **no `setSessionsValidAfter`**, so the post-import row-count comparison sees a different shape than main — the shortfall-only check handles both
- it had **no `resyncSequences`**, which this PR backports and calls on the cross-engine path only

Results: 1 event + 3 photos migrated with `allow_downloads` `0` → `f`, epoch timestamps → real timestamps, `json_typeof(setting_value)` = `boolean`, and the next `INSERT` returned photo id **4** rather than colliding — i.e. the backported sequence resync works. The resolver stays on Postgres after the migration, including after every gallery is deleted.

38 unit tests. Full backend suite failure set byte-identical to pristine `stable`. Lint parity on every touched file (`knexfile.js` keeps its 29 pre-existing errors, unchanged).

Issue #1042, opened 2026-08-12
[FEATURE] Official single-container (all-in-one) image — one Node process, SQLite default


Luca-Timo:
## Summary

Ship an official **single-container (all-in-one) image**: one `docker run`, one volume, working PicPeak. Target audience is #705's — photographers on Synology/UGREEN/QNAP NAS boxes or a small VPS who compare PicPeak's install story against SaaS onboarding. Today the smallest supported deployment is a 4-service compose stack (frontend nginx, backend, postgres, redis); people already *try* single-container deployments (see #1038) and silently land on an unsupported, unlogged SQLite fallback.

## Design decisions

**One Node process — no supervisor, no bundled nginx/Postgres/Redis.**

- The backend serves the built frontend (`express.static` + SPA fallback); nginx's proxy rules and security headers move into Express middleware. One process means clean PID-1 signal handling, one log stream, one healthcheck.
- **SQLite as the explicit, documented default.** SQLite is a library inside the Node process writing one file on the volume — no second daemon, no credentials, no startup ordering. The single-photographer/NAS workload (one admin writing, guests reading, WAL mode) fits it comfortably. Bundling Postgres would mean a process supervisor, `initdb`, data-dir permissions, and the `pg_upgrade` dance on image bumps — exactly the ops burden this image exists to remove.
- The image can still point at an **external** Postgres via the standard `DB_*` env vars for users who have one.
- Redis is not included (backend code has no runtime dependency on it).

## Depends on

- **#1038** — engine selection must be explicit, logged at boot, and guarded (refuse virgin-Postgres-next-to-populated-SQLite) *before* an image whose default is SQLite ships. This issue inherits all four fixes proposed there.
- **#1041** — cross-engine `.picpeak` restore (`sqlite → pg`), the upgrade path. A single-container SQLite install must never be a dead end: export `.picpeak` → stand up the compose stack → import. That issue is the first of the two PRs; this one is the second.

## Scope

- [ ] `Dockerfile.aio` (or build stage): backend + built frontend assets in one image, `ENV NODE_ENV=production`, explicit engine default logged at boot
- [ ] Backend serves frontend statics + SPA fallback; nginx security headers replicated in Express middleware (parity checked against `nginx` config)
- [ ] Single data layout on one volume: database file, storage, logs under one mountable root; documented
- [ ] `HEALTHCHECK` wired to `/api/health`
[...]

PR #1043, opened 2026-08-12
feat(backup): cross-engine .picpeak restore — SQLite backup onto PostgreSQL

### LINKED ISSUE #1041: [FEATURE] Cross-engine .picpeak restore: SQLite → PostgreSQL upgrade path
## Summary

The `.picpeak` backup format is engine-neutral **by design** — `picpeakExportService.js` deliberately avoids `pg_dump`/`.backup` and writes every table as NDJSON so rows "re-import cleanly on both engines", and the target instance rebuilds its schema from its own migrations before loading data. But the v1 importer enforces **same-engine** restore: any manifest engine mismatch is a hard validation error (`picpeakImportService.js`, `validateManifest`).

That check was a v1 scoping decision, not a format limitation. Relaxing it in the **upgrade direction only** (`sqlite → pg`) turns `.picpeak` into the official migration path from a small SQLite install (single-container / NAS, see #705 and the companion single-container issue) to the full compose stack with PostgreSQL:

1. Old instance: Settings → Backup → export `.picpeak` with photos included
2. Stand up the compose stack with Postgres, run the setup wizard
3. Import the `.picpeak` — schema from migrations, data + files from the archive; the existing forward-only guard and wizard-admin preservation already handle version skew and operator lockout

## Scope decision: `sqlite → pg` only

`pg → sqlite` stays blocked. Postgres archives carry ISO `T`/`Z` timestamps that would land as-is in SQLite text columns, resurrecting the exact date-drift bug class behind #1028/#1029 — and downgrading engines is rarely intentional. The mismatch error message should say the supported direction explicitly.

## Changes

1. **Relax the engine gate** in `validateManifest` (`picpeakImportService.js` ~L56–59): allow manifest `sqlite` → instance `pg`; keep `pg` → `sqlite` as a hard error with updated wording.
2. **Target-schema-driven boolean coercion on insert.** SQLite archives carry booleans as `0`/`1` integers (the export writes raw rows). Postgres rejects integers in parameterized inserts into `boolean` columns. Add a `booleanColumnsFor(trx, table)` lookup (mirroring the existing `jsonColumnsFor`) and map `0/1 → false/true` before `batchInsert`.
3. **Fix a latent double-encoding bug in `serialiseJsonColumns`.** It `JSON.stringify`s every non-null value in a json/jsonb column. From a pg archive the values are parsed objects — correct. From a SQLite archive, json columns are already JSON **text**, so stringifying again double-encodes. Only stringify non-string values.
4. **Date round-trip is verified by test, not assumed.** SQLite's `YYYY-MM-DD HH:MM:SS` strings should insert fine into pg `timestamp` columns, but this exact drift class has bitten before — the round-trip test must assert stored values, not just row counts.
5. **No export-format change.** Import-side coercion driven by the *target* schema handles every already-downloaded v1 archive; canonicalizing on export would orphan existing backups.
6. **Restore-confirmation UI** states the cross-engine case explicitly ("SQLite backup → PostgreSQL instance") so the operator knows what is happening.

Already correct as-is (no changes needed): pg sequence resync after restore, wizard-admin re-injection + MFA preservation, superuser check, forward-only migration guard, files/ tree restore.

## Acceptance criteria

- [ ] Importing a SQLite-created `.picpeak` into a PostgreSQL instance succeeds end-to-end (data + files)
- [ ] Booleans, dates, and json/jsonb columns hold correct **values** after the cross-engine restore (asserted, not assumed)
- [ ] `pg → sqlite` import is still rejected, with an error naming the supported direction
- [ ] Existing same-engine restores (`pg↔pg`, `sqlite↔sqlite`) are unaffected, including archives exported before this change
- [ ] Round-trip test: seed SQLite → export → import into Postgres → assert values + working id sequences (next insert does not collide)
- [ ] Forward-migration cross-engine test: older-schema SQLite backup into newer pg instance
- [ ] Restore dialog shows the cross-engine notice

## Relationship

Part of #705 (deployment simplification). Prerequisite for the upgrade story of the single-container image issue — an all-in-one SQLite install is a dead end without an official path to the full stack.

Luca-Timo:
Closes #1041. Part of #705 (first of the two PRs planned there — the single-container image #1042 builds on this).

**Reshaped onto current `main` per review** — the coercion engine from #1039 (`typedColumnsFor` / `epochToIso` / `coerceForTargetEngine`) is kept as-is; my earlier reimplementation (`pgColumnTypes` / `prepareRowsForPg`, including its epoch-seconds bug) is gone. What remains is the policy delta, the UI copy, and the tests. Branch is rebased clean on `main`.

## Policy change (explicit, per review)

This PR deliberately reverses #1039's decision to keep cross-engine restore CLI-only:

> **`sqlite → pg` restore is now allowed for anyone holding `backup.restore`, from the upload UI and the CLI alike, gated by the manifest-direction rule in `validateManifest`. `pg → sqlite` stays refused, with an error naming the supported direction.**

Rationale: the upgrade path (#1041) is aimed at operators moving a small SQLite install to the full stack **through the browser** — the same population #1038's fix pushes toward explicit engine choice. The dangerous direction (pg ISO timestamps landing in SQLite text columns, the #1028/#1029 drift class) stays blocked, and the coercion layer this relies on is the reviewed, landed one, now exercised by real-Postgres stored-value tests. The `allowEngineSwitch` flag is removed rather than kept alongside: one gate, one policy, no way to drive the refused direction.

## Changes

**`picpeakImportService.js`** (~30 lines against `main`)
- `validateManifest`: direction rule replaces the `allowEngineSwitch` opt-in; refusal message names the supported direction.
- `importFromPicpeak`: derives `crossEngine` from the manifest's engine vs the target (absent engine field ⇒ target engine, i.e. exact pre-change behavior for old archives), logs cross-engine restores, returns the flag. `resyncSequences()` keeps running post-restore — covered by a test that does a natural insert after an explicit-id restore.
- `typedColumnsFor` added to the test exports next to `epochToIso`/`coerceForTargetEngine`.

**`scripts/migrate-sqlite-to-postgres.js`** — relies on the shared gate; flag dropped.

**`adminBackup.js`** — `crossEngine` included in the import response.

**`PicpeakBackupCard.tsx` + locales** — intro copy states the supported direction; success panel shows a cross-engine notice. Both strings live in `en.json` + `de.json`. Also removed the orphaned `settings.backup.picpeak` node from both locale files (unreferenced, carried the stale "same engine only" copy — flagged in review).

## Screenshots

| Before | After |
|---|---|
| before | after |

(The green cross-engine notice renders in the success panel after a converting restore — driven by the new `crossEngine` response flag.)

## Tests

`__tests__/integration/picpeakCrossEngine.test.js`, repointed at the landed helpers:
- **Ungated**: `validateManifest` direction rules on a mocked pg target; `epochToIso` units — epoch **milliseconds**, epoch **seconds** (the review's 54-years-off case, now asserting the same instant), numeric strings, non-numeric passthrough; `coerceForTargetEngine` on sqlite-shaped rows. The reverse-direction block stays pinned by `picpeakRoundtrip.test.js` on the real sqlite harness.
- **Gated on `PICPEAK_PG_TEST_URL`** (same contract as `picpeakRestorePg.test.js`): `typedColumnsFor` classification via `columnInfo()`; sqlite-shaped NDJSON rows landing in real Postgres with correct **stored values** (`is_active 1 → true`, `allow_downloads 0 → false`, epoch → same instant, json TEXT → intact jsonb, no double encoding); sequence resync then a natural insert.

Verified locally: all four picpeak suites (`picpeakCrossEngine`, `picpeakRoundtrip`, `picpeakRestorePg`, `picpeakExport`) — 24 tests, 0 failures — against SQLite **and** a throwaway Postgres 15 with the gated cases running. Frontend `npm run build` passes. Remaining lint findings in touched files are pre-existing (`quotes` on the double-quoted SQL literals, `no-console` in the CLI script).

### DIFF
--- a/backend/__tests__/integration/picpeakCrossEngine.test.js
+++ b/backend/__tests__/integration/picpeakCrossEngine.test.js
@@ -1,0 +1,197 @@
+'use strict';
+
+/**
+ * Cross-engine .picpeak restore policy (#1041): a SQLite archive restored onto
+ * a PostgreSQL instance — the official small-install → full-stack upgrade
+ * path — now allowed by validateManifest's direction rule instead of the
+ * former CLI-only allowEngineSwitch flag. The coercion engine itself
+ * (typedColumnsFor / epochToIso / coerceForTargetEngine) landed with #1039;
+ * these tests pin the direction policy and the coercion's cross-engine
+ * value-correctness.
+ *
+ * Ungated: validateManifest direction rules and the pure coercion units.
+ * The reverse direction (pg backup onto a sqlite instance) staying blocked is
+ * pinned by picpeakRoundtrip.test.js, which runs on the real sqlite harness.
+ *
+ * Gated on PICPEAK_PG_TEST_URL (same contract as picpeakRestorePg.test.js):
+ * sqlite-shaped NDJSON rows land in real Postgres with correct stored VALUES,
+ * not just row counts, e.g.
+ *   PICPEAK_PG_TEST_URL="postgres://picpeak:pw@127.0.0.1:7102/picpeak_xengine_test" \
+ *     npx jest __tests__/integration/picpeakCrossEngine.test.js
+ */
+const knexLib = require('knex');
+
+describe('validateManifest cross-engine direction (pg target)', () => {
+  let validateManifest;
+
+  beforeAll(() => {
+    jest.resetModules();
+    jest.doMock('../../knexfile', () => ({ client: 'pg' }));
+    // validateManifest wraps its knex_migrations lookup in try/catch — a
+    // throwing stub simply skips the forward-only check, which is not under
+    // test here.
+    jest.doMock('../../src/database/db', () => ({ db: () => { throw new Error('stub'); } }));
+    ({ validateManifest } = require('../../src/services/picpeakImportService'));
+  });
+
+  afterAll(() => {
+    jest.dontMock('../../src/database/db');
+    jest.dontMock('../../knexfile');
+    jest.resetModules();
+  });
+
+  it('allows a sqlite backup onto a pg instance (upgrade direction)', async () => {
+    const blockers = await validateManifest({
+      kind: 'picpeak-backup', format: 1, database: { engine: 'sqlite' }, tables: {},
+    });
+    expect(blockers.filter((b) => /engine/i.test(b))).toHaveLength(0);
+  });
+
+  it('still allows same-engine pg → pg', async () => {
+    const blockers = await validateManifest({
+      kind: 'picpeak-backup', format: 1, database: { engine: 'pg' }, tables: {},
+    });
+    expect(blockers.filter((b) => /engine/i.test(b))).toHaveLength(0);
+  });
+});
+
+describe('epochToIso (landed with #1039)', () => {
+  let epochToIso;
+
+  beforeAll(() => {
+    jest.resetModules();
+    ({ epochToIso } = require('../../src/services/picpeakImportService'));
+  });
+
+  it('converts epoch milliseconds', () => {
+    expect(epochToIso(1723400000000)).toBe('2024-08-11T18:13:20.000Z');
+  });
+
+  it('converts epoch SECONDS to the same instant, not January 1970', () => {
+    expect(epochToIso(1723400000)).toBe('2024-08-11T18:13:20.000Z');
+  });
+
+  it('converts numeric strings', () => {
+    expect(epochToIso('1723400000000')).toBe('2024-08-11T18:13:20.000Z');
+  });
+
+  it('passes non-numeric values through untouched', () => {
+    expect(epochToIso('2026-08-12 10:00:00')).toBe('2026-08-12 10:00:00');
+  });
+});
+
+describe('coerceForTargetEngine on sqlite-shaped rows', () => {
+  let coerceForTargetEngine;
+
+  beforeAll(() => {
+    jest.resetModules();
+    ({ coerceForTargetEngine } = require('../../src/services/picpeakImportService'));
+  });
+
+  const types = { timestamps: ['created_at', 'expires_at'], booleans: ['is_active'] };
+
+  it('coerces 0/1 booleans and epoch timestamps, leaves date strings alone', () => {
+    const [row] = coerceForTargetEngine(
+      [{ id: 1, is_active: 1, created_at: 1723400000000, expires_at: '2026-09-01 12:00:00' }],
+      types
+    );
+    expect(row.is_active).toBe(true);
+    expect(row.created_at).toBe('2024-08-11T18:13:20.000Z');
+    expect(row.expires_at).toBe('2026-09-01 12:00:00'); // pg parses this natively
+  });
+
+  it('coerces falsy variants and passes null/empty through', () => {
+    const [row] = coerceForTargetEngine(
+      [{ is_active: 0, created_at: null, expires_at: '' }],
+      types
+    );
+    expect(row.is_active).toBe(false);
+    expect(row.created_at).toBeNull();
+    expect(row.expires_at).toBe('');
+  });
+});
+
+// ── Real-Postgres integration (gated) ────────────────────────────────────────
+const PG_URL = process.env.PICPEAK_PG_TEST_URL;
+const maybe = PG_URL ? describe : describe.skip;
+
+maybe('sqlite-shaped rows land correctly in real Postgres', () => {
+  let pgDb;
+  let svc;
+
+  beforeAll(async () => {
+    pgDb = knexLib({ client: 'pg', connection: PG_URL });
+    await pgDb.raw('DROP TABLE IF EXISTS xengine_events, xengine_settings CASCADE');
+    await pgDb.schema.createTable('xengine_events', (t) => {
+      t.increments('id');
+      t.string('slug');
+      t.boolean('is_active').defaultTo(true);
+      t.boolean('allow_downloads').defaultTo(true);
+      t.timestamp('created_at');
+      t.timestamp('expires_at');
+    });
+    await pgDb.schema.createTable('xengine_settings', (t) => {
+      t.increments('id');
+      t.string('setting_key').notNullable().unique();
+      t.jsonb('setting_value');
+    });
+
+    jest.resetModules();
+    jest.doMock('../../knexfile', () => ({ client: 'pg' }));
+    jest.doMock('../../src/database/db', () => ({ db: pgDb }));
+    svc = require('../../src/services/picpeakImportService');
+  });
+
+  afterAll(async () => {
+    jest.dontMock('../../src/database/db');
+    jest.dontMock('../../knexfile');
+    if (pgDb) {
+      await pgDb.raw('DROP TABLE IF EXISTS xengine_events, xengine_settings CASCADE');
+      await pgDb.destroy();
+    }
+  });
+
+  it('typedColumnsFor classifies boolean and timestamp columns via columnInfo()', async () => {
+    const types = await svc.typedColumnsFor(pgDb, 'xengine_events');
+    expect(types.booleans.sort()).toEqual(['allow_downloads', 'is_active']);
+    expect(types.timestamps.sort()).toEqual(['created_at', 'expires_at']);
+  });
+
+  it('inserts a sqlite archive row (0/1 booleans, epoch dates, json text) with correct stored values', async () => {
+    // Exactly what a sqlite-created .picpeak carries: integers for booleans,
+    // epoch numbers for #485-shape timestamps (ms here, seconds covered by the
+    // epochToIso unit), a "YYYY-MM-DD HH:MM:SS" string for clean ones, and
+    // json columns as TEXT (the crossEngine path skips serialiseJsonColumns —
+    // the text is already what pg wants).
+    const epoch = 1723400000000;
+    const eventRows = [
+      { id: 1, slug: 'wedding', is_active: 1, allow_downloads: 0, created_at: epoch, expires_at: '2026-09-01 12:00:00' },
+    ];
+    const settingRows = [{ id: 1, setting_key: 'brand', setting_value: '{"name":"PicPeak","dark":true}' }];
+
+    await pgDb.transaction(async (trx) => {
+      const evTypes = await svc.typedColumnsFor(trx, 'xengine_events');
+      await trx.batchInsert('xengine_events', svc.coerceForTargetEngine(eventRows, evTypes), 100);
+      const stTypes = await svc.typedColumnsFor(trx, 'xengine_settings');
+      await trx.batchInsert('xengine_settings', svc.coerceForTargetEngine(settingRows, stTypes), 100);
+    });
+
+    const ev = await pgDb('xengine_events').where({ id: 1 }).first();
+    expect(ev.is_active).toBe(true);          // 1 → true, not backwards (#1028 class)
+    expect(ev.allow_downloads).toBe(false);   // 0 → false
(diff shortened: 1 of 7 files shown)


the-luap:
Heads-up on a collision: #1039 (for #1038) touches the same importer code, and it needs to go in first — it fixes the reason installs end up on SQLite unintentionally in the first place, which is the population this upgrade path serves.

**The overlap is exactly your two changes:**

- `validateManifest`'s same-engine gate
- the value-shape coercion (`pgColumnTypes` + `prepareRowsForPg` here; `allowEngineSwitch` + `coerceForTargetEngine` there)

**Your version is the better one for the UI path**, and I'd like to keep it: the explicit direction policy (`sqlite → pg` allowed, `pg → sqlite` refused with a reason) is a cleaner contract than my CLI-only opt-in flag, and your `pgColumnTypes` classification with real-Postgres insert tests is more thorough than what I have.

**What I'd ask when you rebase onto #1039:**

1. Drop `allowEngineSwitch` and `coerceForTargetEngine`/`typedColumnsFor`/`parseJsonText` from `picpeakImportService.js` — replace them with your `pgColumnTypes`/`prepareRowsForPg`.
2. `scripts/migrate-sqlite-to-postgres.js` (new in #1039) calls `importFromPicpeak({ allowEngineSwitch: true })`. Point it at your gate instead — it is the same operation, just driven from the CLI.
3. Keep `resyncSequences()` running for the cross-engine case. Rows arrive with explicit ids, so without it the next insert collides on the primary key. Already covered on `main`; just don't lose it in the merge.

One finding from the review of #1039 that is worth cross-checking against yours, since it bit me: SQLite json columns are TEXT holding JSON, and `serialiseJsonColumns` would stringify them a second time — `app_settings.setting_value` is `json` on every install, so a migration silently turned every setting into a JSON string scalar. Your header comment says "json text NOT double-encoded" and you have a test for it, so I think you already have this; flagging it only because it is invisible without checking `json_typeof()` on the target.

I'll ping you here when #1039 merges.

the-luap:
Reviewed against current `main`. #1039 merged in the meantime and landed the same coercion layer, so this needs a reshape rather than a rebase — details below.

## 🔴 Blockers

### 1. Superseded by #1039

`backend/src/services/picpeakImportService.js` — this branch was cut before #1039 and reimplements the SQLite→Postgres coercion that already landed (`typedColumnsFor` / `epochToIso` / `coerceForTargetEngine`). Both versions now exist in the same file and the PR is `DIRTY`.

### 2. The reimplementation loses data the landed one keeps

Ran both against the same inputs:

```
                     this PR                    main (#1039)
epoch ms      "2024-08-11T18:13:20.000Z"   "2024-08-11T18:13:20.000Z"   agree
epoch secs    "1970-01-20T22:43:20.000Z"   "2024-08-11T18:13:20.000Z"   54 years off
numeric str   "1723400000000"              "2024-08-11T18:13:20.000Z"   pg rejects
```

`prepareRowsForPg` gates on `typeof v === 'number'` and treats every finite number as epoch milliseconds. `epochToIso` splits seconds from ms at the `1e11` boundary and also coerces `/^-?\d+$/` strings. On an install carrying epoch-seconds rows this rewrites every affected timestamp to January 1970 — inside the restore transaction, so it commits looking like a success. Restoring a client's gallery history into 1970 is not a failure mode we can ship.

### 3. No screenshot

`PicpeakBackupCard.tsx` changes the restore-card intro copy and adds the green cross-engine note. UI-touching PRs need a screenshot attached before review — before/after of the card is enough here.

## 🟡 Concerns

### 1. It reverses a deliberate decision from #1039

`picpeakImportService.js:48` on `main` gates cross-engine behind `allowEngineSwitch`, passed only by `scripts/migrate-sqlite-to-postgres.js`, explicitly because the path had never been exercised. This PR opens sqlite→pg to anyone holding `backup.restore` via the upload UI.

That may well be the right call — the upgrade path is real and #1041 wants it. But it should be a decision we make on purpose, not an artifact of two branches building the same thing in parallel. If we open it, the manifest-direction rule is the right shape; I just want it stated as a policy change in the PR body.

### 2. `pgColumnTypes` hardcodes the schema

`SELECT ... WHERE table_schema = 'public'`. The landed `typedColumnsFor` uses knex `columnInfo()`, which respects the configured search path. Pre-existing pattern, but #1039 spent four commits centralising how we resolve the Postgres target — this is the direction we're moving away from.

## Suggested reshape

The new contribution here is the **policy**, not the coercion engine. Rebased onto current `main` this should collapse to roughly:

- drop `pgColumnTypes` + `prepareRowsForPg`; keep the landed `coerceForTargetEngine` / `epochToIso`
- have the upload route pass the sqlite→pg-only rule into `validateManifest`
- keep the `crossEngine` response flag, the card note, and the locale keys
- keep `picpeakCrossEngine.test.js`, repointed at the landed helpers — the Postgres-gated cases in it are worth having either way

~40 lines instead of 285, and it keeps the stronger coercion. Worth doing before you resolve that conflict by hand.

## ✅ Verified

- `picpeakCrossEngine.test.js` passes on this branch — 6 passed, 3 skipped (the Postgres-gated ones need `PICPEAK_PG_TEST_URL`).
- Locale keys: the rendered key is `backup.picpeak.restoreIntro` and it's correctly updated in both en and de. There's a second `settings.backup.picpeak.restoreIntro` node still carrying the old "same database engine only" copy — grepped it, no references, so nothing renders it. Worth deleting while you're in there.
- Same-engine restores are untouched: `pgColumnTypes` returns empty sets off Postgres and `prepareRowsForPg` short-circuits, so sqlite→sqlite and pg→pg are byte-for-byte unchanged. Directly confirmed.

## Verdict

Hold the rebase. Strip this back to the policy delta on top of landed `main` and it's a small, clean PR.

the-luap:
Nudge on this one — no movement since the review and it's still `DIRTY` against `main`.

No rush from my side, but don't spend time resolving that `picpeakImportService.js` conflict by hand: the reshape in the review above makes most of it moot. Dropping `pgColumnTypes` + `prepareRowsForPg` in favour of the landed `coerceForTargetEngine` / `epochToIso` removes the conflicting hunks entirely, and what's left is the policy delta plus the UI copy — a much smaller rebase.

#1045 is nearly through, so no contention if you want to pick this up after it lands.

Luca-Timo:
Reshaped as suggested — reset onto current `main`, coercion reimplementation dropped, force-pushed as a single commit (`07bf303a`). Non-test delta is ~35 lines on top of #1039's engine. Ready for a recheck.

@the-luap ready to retest

## Blockers — closed

1. **Superseded by #1039** — `pgColumnTypes`/`prepareRowsForPg` are gone; the landed `typedColumnsFor`/`epochToIso`/`coerceForTargetEngine` run unmodified. Branch is rebased clean, `mergeStateStatus` no longer `DIRTY`.
2. **Epoch-seconds loss** — moot via adoption of `epochToIso`, and pinned so it can't regress: `picpeakCrossEngine.test.js:70` asserts `1723400000` (seconds) → `2024-08-11T18:13:20Z`, alongside ms and numeric-string cases from your table.
3. **Screenshot** — before/after of the restore card in the PR body.

## Concerns — closed

1. **Policy stated on purpose** — PR body now leads with it: sqlite → pg open to `backup.restore` holders from UI and CLI via the manifest-direction rule (`picpeakImportService.js:66`), pg → sqlite refused with the direction named. `allowEngineSwitch` is removed rather than kept alongside (`migrate-sqlite-to-postgres.js:206` now goes through the shared gate) — one gate, no way to drive the refused direction.
2. **`table_schema = 'public'`** — moot; nothing of mine queries `information_schema` anymore, `columnInfo()` path is what runs and its classification is covered by a gated test.

Your three merge asks from the first comment all landed: flag dropped, CLI pointed at the gate, `resyncSequences()` kept (`crossEngine` derived at `picpeakImportService.js:447`, sequences covered by a natural-insert-after-restore test).

Also took the ✅-section cleanup: the orphaned `settings.backup.picpeak` node is deleted from `en.json` + `de.json`; the two strings this PR touches live under the rendered `backup.picpeak.*` keys in both locales.

Verified: all four picpeak suites (24 tests) green on SQLite and on a throwaway Postgres 15 with the `PICPEAK_PG_TEST_URL`-gated cases executing; `npm run build` passes. Remaining lint findings in touched files are pre-existing (`quotes` on the double-quoted SQL literals, `no-console` in the CLI script).

the-luap (approved):
Re-verified the reshape at file:line rather than from the summary. This is exactly the shape I asked for — merge cleared.

## Blockers — closed

| | |
|---|---|
| **Superseded by #1039** | `pgColumnTypes` / `prepareRowsForPg` are gone (zero references repo-wide); `typedColumnsFor` / `epochToIso` / `coerceForTargetEngine` run unmodified. One clean commit on current `main`, merges clean. |
| **Epoch-seconds loss** | Moot by adoption, and pinned so it can't come back — all three rows from my table are now tests: ms, **seconds → the same instant not Jan 1970**, and numeric-string, plus a passthrough case. |
| **Screenshot** | Present and resolving; the card shows the corrected copy. |

## Concerns — closed

**Policy stated deliberately.** That was the whole ask, and the PR body now leads with it. I traced the gate as a truth table rather than reading the comment — `picpeakImportService.js:66` behaves correctly in all five cases: `sqlite→sqlite` and `pg→pg` pass, `sqlite→pg` passes via the exception, `pg→sqlite` errors with the direction named, and a manifest with no engine field falls through to same-engine behaviour. The refusal stays pinned by `picpeakRoundtrip.test.js:146`, which still passes under the loosened gate.

Removing `allowEngineSwitch` rather than keeping it alongside was the right call — one gate, and no code path that can drive the refused direction.

**`table_schema = 'public'`** — moot, nothing queries `information_schema` any more.

## One thing you fixed that nobody asked for

Deriving `crossEngine` from the manifest instead of the flag closes a latent bug on `main`: `crossEngine: allowEngineSwitch` meant a CLI-invoked *same-engine* sqlite→sqlite restore ran `coerceForTargetEngine` anyway and rewrote epoch timestamps to ISO. Now unreachable. Good catch.

## The gap I raised, and what I did about it

The real-Postgres cases in these suites gate behind `PICPEAK_PG_TEST_URL`, which was set in no workflow — so they had never run in CI. That mattered more once this PR opens the path to the upload UI: the tests proving a coerced row lands with the right *stored values* were exactly the ones skipping.

Rather than ask you to carry it, I wired it up in #1056 (merged): a `postgres:15-alpine` service on the backend job, same shape `schema-drift` already used. I then updated this branch so it picks up the workflow.

Result on this PR's latest run:

```
Test Suites: 179 passed, 179 total
Tests:       1770 passed, 1770 total     ← 0 skipped
```

`main` sits at 1759; your suite adds 11 and every one executed, including the three real-Postgres cases. Before #1056 that same run reported `4 skipped`. The coercion this PR depends on is now verified against a real database on every PR, not just on your laptop.

## Verified

- All four picpeak suites green on SQLite, and green again against a throwaway Postgres 15 locally before I wired CI (`picpeakRestorePg` 4/4, `picpeakCrossEngine` 11/11)
- Orphaned `settings.backup.picpeak` node gone from both locales; the two live strings sit under the rendered `backup.picpeak.*` keys in `en` and `de`
- CLI at `migrate-sqlite-to-postgres.js:206` goes through the shared gate with no flag
- `resyncSequences()` still runs post-restore, covered by the natural-insert-after-restore test

## Nit

The screenshot shows the intro copy but not the green cross-engine notice on the success panel — that only appears after an actual restore. Not worth a re-shoot.

## Verdict

Merge cleared. `bypass-size-gate` fails by design at 314 lines against its 300-line self-merge cap; merging with an admin override on the basis that the maintainer review it asks for is these three rounds.

Issue #747, opened 2026-07-04
[FEATURE] Team workspaces / creative-team roles


Luca-Timo:
Per-role access for studios with editors / digital ops / assistants — a broader generalization of the festival multi-photographer role idea into reusable team roles.

- [ ] Workspace-scoped roles & permissions beyond single-instance admin roles

Effort: L. Overlaps the festival multi-photographer upload idea.

_Parked idea — no timeline._

Luca-Timo:
Delivered by PicPeak/picpeak#1045 (granular permission gating + role editor).

**What landed**
- Reusable, granular team roles — a role editor (create / clone / edit / delete) over a 66-permission catalog grouped by area, replacing the fixed `super_admin`/`admin`/`editor`/`viewer` set with arbitrary custom roles.
- Shipped presets: **Solo Photographer** (full operator) and **Team Photographer** (contributor — view events + manage photos + read-only CRM; no settings/users/billing edits), alongside the existing three.
- Permission-gated UI end to end — nav, settings tabs, and in-page action buttons hide when the role lacks the perm; the backend enforces with 403 regardless of the UI.
- `roles.manage` is a delegation primitive, not root escalation: a non-super holder can only grant permissions its own role already holds, and can't edit its own role.

**Scope note** — "workspace-scoped" here means per-role access within the single studio instance (picpeak is single-instance). Roles scoped to *specific events/galleries* — a contributor who can only touch their assigned shoots — are tracked under #743 (event→photographer scoping), which is the phase-2 build.

Closing as delivered.

Issue #743, opened 2026-07-04
[FEATURE] Multi-photographer "festival" upload (scoped role + attribution + curation)


Luca-Timo:
One admin (event owner) + multiple photographers contributing to a single shared event/gallery. For festivals, weddings with 2nd/3rd shooters, sports, conferences.

- [ ] Scoped `photographer` role — upload only to *assigned* events; no admin/settings/CRM access
- [ ] Per-photo attribution — derived from **EXIF metadata (Artist/Creator) or the uploading photographer's account/role** (not a manual field); stored per photo, shown as credit + filter in the gallery and exports
- [ ] Event → photographer assignment (optional per-photographer upload link/token)
- [ ] Curation — admin approves/hides/reorders across all contributors before publish

Effort: L. Overlaps team workspaces.

_Parked idea — no timeline._

Luca-Timo:
The permission foundation landed in PicPeak/picpeak#1045; the festival-upload mechanics remain. Status per checkbox:

- [~] **Scoped `photographer` role** — the *capability* half is done: the **Team Photographer** preset has no admin/settings/CRM-edit access (view events + manage photos + read-only CRM). The *"upload only to assigned events"* half is **not** done — `ownership.js` still scopes by `created_by`, and there is no event→photographer assignment yet.
- [ ] **Per-photo attribution** — derived from EXIF (Artist/Creator) or the uploader's account, stored per photo, shown as credit + gallery filter + in exports. Not started.
- [ ] **Event → photographer assignment** (+ optional per-photographer upload link/token). Not started.
- [ ] **Curation** — admin approves / hides / reorders across all contributors before publish. Not started.

So #1045 is phase 1 (the role's global permissions). Phase 2 is the assignment + assigned-event scoping + attribution + curation. Keeping this open to track that work.

the-luap:
Scoping the second checkbox — per-photo attribution — as a slice that can ship on its own, ahead of the assignment and curation halves.

## Why now

The EXIF plumbing this needs landed for capture dates in #1172. `extractCaptureDate` is already an `exifr` call with a pick list (`imageProcessor.js:1223-1227`):

```js
const exif = await exifr.parse(imagePath, {
  pick: ['DateTimeOriginal', 'CreateDate', 'DateTimeDigitized', 'ModifyDate']
});
```

Adding `Artist` / `Creator` / `Copyright` to that array is the extraction half. Nothing new to reach for.

## What is actually there today

Less than the schema suggests:

[...]

PR #1045, opened 2026-08-13
feat(permissions): granular permission gating + role editor & presets (#747, phase 1 of #743)

Luca-Timo:
Phase 1 of the multi-photographer epic — **#747** (team/creative-team roles) and the permission foundation for **#743** (multi-photographer "festival" upload). Makes every admin feature permission-gateable so a multi-user studio can split capability across roles, adds a role editor, and ships two photographer presets.

Opened as a **draft** for early review — it's been running on a fork instance for testing.

## Screenshots

**1. Role editor — Roles tab** — the shipped system presets + any custom roles, with per-role permission and user counts.
Screenshot 2026-08-13 at 21 30 46

**2. Role editor — permission matrix** — create/edit a role; permissions grouped by category with a per-group "Select all".
Screenshot 2026-08-13 at 21 31 05

**3. Restricted role — navigation gating** — a non-owner role sees a trimmed sidebar (Settings / Users / CRM hidden).
Screenshot 2026-08-13 at 21 37 05

**4. Restricted role — event detail** — the event Overview with role-gated mutating actions hidden.
Screenshot 2026-08-13 at 21 41 44

## Permission model

Two axes per feature: everyday **use** vs **configure/administer**. The headline goal: a team member can do day-to-day work but can't break global config.

- **Split the catch-all `settings.edit`** into dedicated dangerous-config perms: `settings.banking` (IBAN/bank accounts/QR), `settings.domains` (site URL), `settings.security` (security policy + SSO), `settings.integrations` (webhooks + API tokens), `settings.features` (feature flags). Reads keep an `OR` with `settings.view` so existing roles don't lose visibility.
- **Dedicated perms** for admin surfaces previously borrowing `settings.*`: `whatsapp.*`, `event_types.*`, `image_security.*`, `notifications.*`, `system.*`, plus `roles.manage` and `vat_codes.view` (the VAT read was previously ungated).
- Migration `174_granular_permissions_and_presets.js` — idempotent, `hasTable`-guarded, `knex.fn.now()`, cross-DB (SQLite + PG both exercised).

## Backward compatibility

`settings.edit` was `super_admin`-only, so splitting it touches no other role. `admin` (41 perms), `editor` (8), `viewer` (6) are **unchanged**. Verified on a full SQLite migration run.

## Presets + boot self-heal

- `_permissionsBoot.js`: `super_admin` always holds every permission (tracks-all) so a new perm never needs a compensation migration; **all other roles stay frozen** (no silent capability gain on upgrade).
- Two shipped presets: **Solo Photographer** (full operator, one-person studio) and **Team Photographer** (second/festival shooter — view events + upload/manage photos + read-only CRM; no settings/users/billing edits, no `events.edit`).

## Role editor

- Backend `adminRoles` CRUD (create/edit/clone/delete + permission matrix), gated by `roles.manage`. System roles protected; `super_admin` immutable; reserved names blocked; delete refused while users are assigned.
- Frontend: a **Roles** tab in User Management with a category-grouped permission matrix and preset cloning.

## Frontend gating (hide, not disable)

- Settings-page tabs are permission-gated with snap-back for deep links.
- In-page mutating controls on the surfaces restricted roles reach — Events list, Archives, gallery photo grid, event detail, Photos tab — are wrapped in `PermissionGate`, so they're **hidden** when the user lacks the perm rather than shown-then-403.

## Security

A self-review caught and fixed an access-control bypass introduced by the split: the generic settings writers (`/general`, `/analytics`, `/seo`, `/security`) upsert arbitrary `setting_key`s, so a `settings.edit`-only role could write `general_site_url` / `security_*` / `accounting_*` via the wrong endpoint. Fixed with `stripUnauthorizedProtectedKeys()` — every generic upsert now drops protected keys the caller isn't permitted to write. Regression covered in `authzPermissionGaps.test.js`.

## Not in scope (phase 2)

Event-scoping — a Team Photographer seeing only their **assigned** events (the #743 event→photographer assignment). This PR delivers the global permission layer; `ownership.js` currently scopes by `created_by`.

## Testing

- Frontend Vitest green; backend Jest green except pre-existing environmental suites (storage/sharp/S3/backup/webhook-worker-timing) unrelated to this change.
- i18n: `en` + `de` added for the role editor. **Other locales (nl/pt/ru/fr/es/sl) fall back to English and need native review.**

### DIFF
--- a/backend/__tests__/routes/adminRolesGuards.test.js
+++ b/backend/__tests__/routes/adminRolesGuards.test.js
@@ -1,0 +1,86 @@
+/**
+ * Role-editor self-amplification guard (migration 175 / adminRoles).
+ *
+ * `roles.manage` must be a DELEGATION primitive, not root escalation: a
+ * non-super_admin holder can only grant permissions their OWN role already
+ * holds, and can't edit their own role. super_admin bypasses. Pins
+ * userManagementService.createRole / updateRole (assertActorMayGrant).
+ */
+const path = require('path');
+const fs = require('fs');
+const os = require('os');
+
+process.env.NODE_ENV = 'test';
+process.env.TEST_DATABASE_PATH = path.join(
+  fs.mkdtempSync(path.join(os.tmpdir(), 'picpeak-roleguard-')), 'db.sqlite',
+);
+process.env.JWT_SECRET = process.env.JWT_SECRET || 'roleguard-test-secret';
+process.env.STORAGE_PATH = fs.mkdtempSync(path.join(os.tmpdir(), 'picpeak-roleguard-storage-'));
+
+const { bootCrmDb, seedMinimal, assignAdminRole } = require('../integration/helpers/crmDb');
+const svc = require('../../src/services/userManagementService');
+const { clearPermissionCache } = require('../../src/middleware/permissions');
+
+describe('role editor — self-amplification guard', () => {
+  let db; let cleanup;
+  let superId; let mgrRoleId; let mgrId;
+
+  beforeAll(async () => {
+    ({ db, cleanup } = await bootCrmDb());
+    ({ adminId: superId } = await seedMinimal(db));
+    await assignAdminRole(db, superId, 'super_admin');
+
+    // A non-super role that CAN manage roles but only holds a couple of perms.
+    const mgrRole = await svc.createRole(
+      { name: 'limited_mgr', permissions: ['roles.manage', 'events.view'] },
+      superId,
+    );
+    mgrRoleId = mgrRole.id;
+    const ins = await db('admin_users').insert({
+      username: 'mgr', email: 'mgr@example.com', password_hash: 'x',
+      role_id: mgrRoleId, must_change_password: false, created_at: new Date(),
+    }).returning('id');
+    mgrId = ins[0]?.id ?? ins[0];
+    clearPermissionCache();
+  }, 120000);
+
+  afterAll(async () => { if (cleanup) await cleanup(); });
+
+  it('super_admin can grant any permission', async () => {
+    const r = await svc.createRole(
+      { name: 'power_role', permissions: ['settings.banking', 'users.delete'] },
+      superId,
+    );
+    expect(r.permissions).toEqual(expect.arrayContaining(['settings.banking', 'users.delete']));
+  });
+
+  it('non-super cannot grant a permission its own role lacks', async () => {
+    await expect(
+      svc.createRole({ name: 'sneaky', permissions: ['events.view', 'settings.banking'] }, mgrId),
+    ).rejects.toThrow(/only grant permissions your own role/i);
+  });
+
+  it('non-super can create a role within its own permissions', async () => {
+    const r = await svc.createRole({ name: 'viewer_lite', permissions: ['events.view'] }, mgrId);
+    expect(r.permissions).toEqual(['events.view']);
+  });
+
+  it('non-super cannot edit its own role', async () => {
+    await expect(
+      svc.updateRole(mgrRoleId, { permissions: ['roles.manage', 'events.view'] }, mgrId),
+    ).rejects.toThrow(/cannot edit your own role/i);
+  });
+
+  it('non-super cannot escalate another role beyond its own permissions', async () => {
+    const adminRole = await db('roles').where({ name: 'admin' }).first();
+    await expect(
+      svc.updateRole(adminRole.id, { permissions: ['settings.banking'] }, mgrId),
+    ).rejects.toThrow(/only grant permissions your own role/i);
+  });
+
+  it('the built-in team_photographer name is reserved', async () => {
+    await expect(
+      svc.createRole({ name: 'team_photographer', permissions: [] }, superId),
+    ).rejects.toThrow(/reserved/i);
+  });
+});
--- a/backend/__tests__/routes/adminSettingsProtectedKeys.test.js
+++ b/backend/__tests__/routes/adminSettingsProtectedKeys.test.js
@@ -1,0 +1,95 @@
+/**
+ * Protected-key boundary on the generic settings writers (migration 175).
+ *
+ * A role with settings.edit but NOT settings.domains (the "office manager" this
+ * PR enables) must be able to save the General tab — which re-posts
+ * general_site_url on every save — as long as the URL is UNCHANGED, and must be
+ * 403'd only when it actually tries to change a protected key. Regression pin for
+ * the change-detection fix (the presence-only check over-fired on every save).
+ */
+const path = require('path');
+const fs = require('fs');
+const os = require('os');
+
+process.env.NODE_ENV = 'test';
+process.env.TEST_DATABASE_PATH = path.join(
+  fs.mkdtempSync(path.join(os.tmpdir(), 'picpeak-setkeys-')), 'db.sqlite',
+);
+process.env.JWT_SECRET = process.env.JWT_SECRET || 'setkeys-test-secret';
+process.env.STORAGE_PATH = fs.mkdtempSync(path.join(os.tmpdir(), 'picpeak-setkeys-storage-'));
+
+const request = require('supertest');
+const express = require('express');
+const cookieParser = require('cookie-parser');
+const {
+  bootCrmDb, seedMinimal, assignAdminRole, mintAdminToken,
+} = require('../integration/helpers/crmDb');
+const svc = require('../../src/services/userManagementService');
+const { clearPermissionCache } = require('../../src/middleware/permissions');
+
+const STORED_URL = 'https://stored.example';
+
+describe('settings protected-key boundary (/general)', () => {
+  let db; let cleanup; let app;
+  let superTok; let mgrTok;
+
+  const auth = (req, tok) => req.set('Authorization', `Bearer ${tok}`);
+  const readSiteUrl = async () => {
+    const row = await db('app_settings').where({ setting_key: 'general_site_url' }).first();
+    return row ? JSON.parse(row.setting_value) : null;
+  };
+
+  beforeAll(async () => {
+    ({ db, cleanup } = await bootCrmDb());
+    const { adminId: superId } = await seedMinimal(db);
+    await assignAdminRole(db, superId, 'super_admin');
+    superTok = mintAdminToken(superId);
+
+    // Office-manager role: settings.view + settings.edit, NOT settings.domains.
+    const mgrRole = await svc.createRole(
+      { name: 'office_mgr', permissions: ['settings.view', 'settings.edit'] },
+      superId,
+    );
+    const ins = await db('admin_users').insert({
+      username: 'office', email: 'office@example.com', password_hash: 'x',
+      role_id: mgrRole.id, must_change_password: false, created_at: new Date(),
+    }).returning('id');
+    mgrTok = mintAdminToken(ins[0]?.id ?? ins[0]);
+
+    await db('app_settings').insert({
+      setting_key: 'general_site_url', setting_value: JSON.stringify(STORED_URL), setting_type: 'general',
+    });
+    clearPermissionCache();
+
+    app = express();
+    app.use(express.json());
+    app.use(cookieParser());
+    app.use('/api/admin/settings', require('../../src/routes/adminSettings'));
+  }, 120000);
+
+  afterAll(async () => { if (cleanup) await cleanup(); });
+
+  it('settings.edit role can save /general when general_site_url is unchanged', async () => {
+    const res = await auth(request(app).put('/api/admin/settings/general'), mgrTok)
+      .send({ general_site_url: STORED_URL, general_max_file_size_mb: 50 });
+    expect(res.status).not.toBe(403);
+    expect(res.status).toBe(200);
+    expect(await readSiteUrl()).toBe(STORED_URL);
+  });
+
+  it('settings.edit role is 403d when it actually changes general_site_url', async () => {
+    const res = await auth(request(app).put('/api/admin/settings/general'), mgrTok)
+      .send({ general_site_url: 'https://evil.example' });
+    expect(res.status).toBe(403);
+    expect(res.body.code).toBe('FORBIDDEN');
+    expect(res.body.keys.map((k) => k.key)).toContain('general_site_url');
+    expect(await readSiteUrl()).toBe(STORED_URL); // unchanged
+  });
+
+  it('super_admin can change general_site_url', async () => {
+    const res = await auth(request(app).put('/api/admin/settings/general'), superTok)
+      .send({ general_site_url: 'https://new.example' });
+    expect(res.status).toBe(200);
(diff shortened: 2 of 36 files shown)


the-luap:
Reviewed against current `main` (merges clean, 0 conflicts). Solid shape overall — the settings split is carved along the right lines and the forward-projection reasoning in the migration header is correct. Two blockers, both cheap.

## 🔴 Blockers

### 1. Migration number collision — and it breaks the restore guard

`backend/migrations/core/174_granular_permissions_and_presets.js` collides with `174_sqlite_nullable_event_dates.js`, which landed from #1035. Different filenames, so git won't conflict and CI stays green.

The damage isn't cosmetic. `picpeakImportService.js:75`:

```js
if (backupLatest && targetLatest && migrationOrder(backupLatest) > migrationOrder(targetLatest))
```

`migrationOrder` parses the numeric prefix, so both files return `174`. An instance carrying only `174_sqlite_nullable_event_dates` will accept a `.picpeak` from a newer instance carrying `174_granular_permissions_and_presets` — `174 > 174` is false, the forward-only check passes, and the restore lands on a schema without the permission rows.

Renumber to `175_`.

### 2. No screenshot

This adds a whole role-editor surface — `RoleManagementTab.tsx`, `RoleEditorModal.tsx`, the permission matrix — plus `PermissionGate` across 60 call sites. Needs the role editor itself and at least one gated page (before/after showing a hidden action) attached before review.

## 🟡 Concerns

### 1. `roles.manage` has no self-amplification guard

`adminRoles.js:92` → `userManagementService.js` `updateRole`. Only `super_admin` is protected. A non-super holder of `roles.manage` can `PUT /api/admin/roles/:id` against **their own role** and grant themselves the entire catalog — `settings.banking`, `users.delete`, everything.

The permission description ("Highly privileged — a holder can grant any capability, so keep it to owners") acknowledges it, but documentation isn't the enforcement layer, and the two standard checks are cheap:

```js
if (role.id === actorRole.id) throw new ValidationError('You cannot edit your own role.');
// and: reject granting any permission the actor does not already hold
```

The second one is what actually contains the blast radius — it makes `roles.manage` a delegation primitive instead of a root escalation.

### 2. `stripUnauthorizedProtectedKeys` silently discards

`adminSettings.js:81-88`. The in-handler guard it replaces returned a 403 with a message naming the required permission. Now an unauthorized `general_site_url` write is deleted from the payload, the upsert succeeds, and the UI says "saved" while nothing changed.

Silent no-op on a permission boundary is the worst of both — the user thinks they repointed the install and the audit log shows a successful settings write. Collect the stripped keys and 403, or at minimum return them so the client can surface it.

### 3. `team_photographer` missing from `RESERVED_ROLE_NAMES`

`userManagementService.js:576` lists `super_admin, admin, editor, viewer, solo_photographer` — the `adminRoles.js` header comment omits it too. If preset seeding is ever skipped (RBAC tables absent when 174 runs, or the swallowed boot error in 4 below), a custom role can take the name with `is_system: false`, and `ensurePreset` skips it forever after because it only checks existence.

### 4. Boot self-heal aborts the rest of its work on first error

`_permissionsBoot.js` `seedPermissionsAtBoot` wraps the whole sequence in one try/catch. On a multi-replica start the loser of the `role_permissions` insert race throws a PK violation inside `ensureSuperAdminHasAllPermissions`, which skips preset seeding entirely on that replica. The winning replica covers it so real impact is low, but per-step try/catch is one line and makes the self-heal actually idempotent under concurrency.

## 🟢 Nits

- Forward-projection is asymmetric: `SETTINGS_SPLIT_PERMISSIONS` project onto every role holding `settings.edit`, but the `FEATURE_PERMISSIONS` that also replaced `settings.edit` gates (`event_types.manage`, `image_security.manage`, `notifications.manage`, `system.manage`, `whatsapp.manage`) don't. Harmless today for the reason your header gives — verified below — but this PR ships the editor that makes custom `settings.edit` holders possible, so the same pattern in a phase-2 migration would silently strip capability. Worth projecting both lists now.
- `exports.down` guards on `hasTable('permissions')` then queries `roles` and `admin_users` unguarded.

## ✅ Verified — directly inspected

- **The split strips nobody.** Migration 056 grants `settings.edit` to `super_admin` only (`admin` gets `settings.view`), and no role-editing endpoint exists on `main` — `adminRoles.js`, `RoleManagementTab.tsx`, `RoleEditorModal.tsx` and `roles.service.ts` are all new files. So no custom role can hold `settings.edit` today and the compat claim in the migration header holds.
- **No bypass of the split via the generic settings writers.** Traced each new perm to its storage: `oidc_*` and `download_*` are already blanket-reserved by `stripReservedSettingKeys`, so no SSO bypass; feature flags live in `feature_flags`; banking lives in `business_profile`. Neither table is reachable from the `app_settings` upsert. The three-rule protected-key list covers what's actually in `app_settings`.
- **Migration** is idempotent throughout, batches at 50, and matches the real schemas — `permissions` has no `updated_at` and defaults `created_at`, `admin_users.role_id` is nullable with `ON DELETE SET NULL`.
- **`authzPermissionGaps.test.js` passes, 7/7** — ran it in a worktree, not taken from CI.
- **i18n** en/de parity complete for every role/permission key, both directions.
- `server.js` mount + boot-hook placement, `requirePermission` / `clearPermissionCache` contracts, and `createRole`'s `returning('id')` handling on both engines.

## Verdict

Renumber to `175_`, add the self-amplification guard, attach screenshots, and I'm comfortable merging. Concerns 2-4 I'd take in the same pass but won't block on.

Luca-Timo:
Thanks for the thorough pass. Rebased onto current `main` (`9b976386`) and addressed everything — force-pushed as `79d3507e`.

**🔴 Blockers**
- **Migration renumbered 174 → 175** (`175_granular_permissions_and_presets.js`). It now runs after `174_sqlite_nullable_event_dates`, so `migrationOrder` no longer ties them and the forward-only guard in `picpeakImportService.js` reads them correctly.
- **Screenshots** attached (role editor Roles tab + permission matrix + restricted-role nav + event detail).

**🟡 Concerns**
1. **`roles.manage` self-amplification** — `assertActorMayGrant()` in `userManagementService.js` (`createRole` + `updateRole`): a non-super_admin can't edit their own role, and can only grant permissions their own role already holds (`ForbiddenError`/403). super_admin bypasses. Both checks you suggested are in.
2. **Silent discard → 403** — replaced with `rejectUnauthorizedProtectedKeys()` in `adminSettings.js`: 403s naming the denied keys + required perm, across all four generic writers (`/general`, `/security`, `/analytics`, `/seo`). No more "saved" no-op on a permission boundary.
3. **`team_photographer` reserved** — added to `RESERVED_ROLE_NAMES` and the `adminRoles.js` header.
4. **Boot self-heal** — per-step try/catch in `_permissionsBoot.js` so a `role_permissions` insert race on one replica no longer skips preset seeding.

**🟢 Nits**
- Forward-projection now also covers the feature `.manage` perms that replaced `settings.edit` gates (`whatsapp`/`event_types`/`image_security`/`notifications`/`system`), matching the `settings.*` split.
- `exports.down` now `hasTable`-guards `roles`/`admin_users`.

Verified: migration up/down on SQLite; `authzPermissionGaps` 7/7; a targeted test for the self-amplification guard (grant-you-lack / own-role / escalate-another all blocked, super_admin allowed); frontend build clean. PR is `MERGEABLE`.

Ready for another look when you have a moment.

the-luap:
Re-verified `79d3507e` on top of the rebase. Spot-checked every item at file:line rather than reading the summary — all eight landed, and the shapes match what you described.

## Blockers — closed

| | Where | |
|---|---|---|
| Migration renumbered | `175_granular_permissions_and_presets.js`; 174 is solely `sqlite_nullable_event_dates`, so `migrationOrder` separates them again | ✅ |
| Screenshots | 4 attached — Roles tab correctly renders **View** (not Edit) on Super Admin, matrix is legible, restricted-role sidebar is Dashboard+Events only, event detail carries no edit/delete affordances | ✅ |

## Concerns — closed

| | Where | |
|---|---|---|
| `roles.manage` self-amplification | `assertActorMayGrant` at `userManagementService.js:606`, own-role check at `:741-743`, called from `createRole:688` and `updateRole:745`. super_admin bypass is explicit. `cloneRole` inherits it through `createRole`, so cloning a preset you can't fully hold is refused too — good catch on the transitive case | ✅ |
| Silent discard → 403 | `rejectUnauthorizedProtectedKeys` at `adminSettings.js:100`, wired into all four generic writers (`:1354`, `:1485`, `:1524`, `:1580`), each with the `if (await …) return;` early-out so there's no double-response | ✅ |
| `team_photographer` reserved | `userManagementService.js:578` + the `adminRoles.js:11` header | ✅ |
| Boot self-heal isolation | `_permissionsBoot.js:118-133` — per-step try/catch, presets loop independently | ✅ |
| Forward-projection symmetry | `175:161-165` — `SETTINGS_EDIT_REPLACEMENTS` now spans the `settings.*` split plus all five feature `.manage` perms | ✅ |
| `down()` guards | `:226` / `:231` | ✅ |

51 backend permission-related tests pass locally on this head, not just in CI.

## 🔴 One new blocker — the 403 fix over-fires

`rejectUnauthorizedProtectedKeys` triggers on key **presence**, not on value **change**, and the General tab posts `general_site_url` on every save.

`useSettingsState.ts:319-326` serialises the whole `generalSettings` object, and `site_url` is always in it (initialised at `:203` with `|| ''`). So a role holding `settings.edit` but not `settings.domains` — the office-manager role this PR exists to make possible — gets a hard 403 on every General-tab save and can't touch date format, upload limits or maintenance mode.

Reproduced against `79d3507e` with a `settings.view` + `settings.edit` role:

```
PUT /api/admin/settings/general
  { general_site_url: '', general_date_format: {...}, general_max_file_size_mb: 75 }

-> 403 {"error":"You don't have permission to change: general_site_url",
        "code":"FORBIDDEN","keys":[{"key":"general_site_url","perm":"settings.domains"}]}
```

`general_site_url` was unchanged — it round-tripped the stored value.

The 403 is right; it just needs the change-detection the in-handler guard had before commit `bfba413b`. That version compared stored vs incoming for exactly this reason:

> Enforced only when the value actually changes, so unrelated /general saves by a safe-settings editor aren't blocked.

Fold the two together — compare against the stored value, 403 only when it actually differs:

```js
const denied = [];
for (const key of Object.keys(settings)) {
  const rule = PROTECTED_SETTING_KEY_PERMS.find((r) => r.match(key));
  if (!rule) continue;
  if (await userHasAnyPermission(adminId, [rule.perm])) continue;
  const row = await db('app_settings').where({ setting_key: key }).first();
  let stored = null;
  if (row) { try { stored = JSON.parse(row.setting_value); } catch (_) { stored = row.setting_value; } }
  if (String(stored ?? '') === String(settings[key] ?? '')) { delete settings[key]; continue; } // no-op, let it through
  denied.push({ key, perm: rule.perm });
}
```

Scope-checked so you don't have to chase it: only `/general` is affected. `security_*` resolves to `settings.security`, which is also that route's gate, so the caller always holds it; the analytics and seo forms carry no protected keys.

## 🟡 The self-amplification test isn't on the branch

The Verified list mentions "a targeted test for the self-amplification guard (grant-you-lack / own-role / escalate-another all blocked, super_admin allowed)". `79d3507e` touches 5 files and none are tests, and nothing under `backend/__tests__` references `assertActorMayGrant` or either guard's error string — I grepped for both.

Guessing it exists locally and didn't get staged. Worth committing: it's the guard that turns `roles.manage` from a root escalation into delegation, and right now nothing pins it.

## 🟢 Nit

The role-editor matrix renders the full catalog regardless of what the actor can grant, so a non-super user ticking a permission they lack gets a 403 on submit instead of a disabled checkbox. `GET /roles/permissions` already knows the caller — returning a grantable flag would let the modal grey them out.

## Verdict

Change-detection on the protected-key 403 and the guard test committed, and this is merge-cleared. Everything else from round 1 is closed.

Luca-Timo:
Both round-2 items done — pushed as `e6dd4bbc`.

**🔴 403 over-fire** — restored change-detection in `collectUnauthorizedProtectedKeys` (`adminSettings.js`): for a protected key the caller can't write, it now compares the incoming value against the stored `app_settings` row and only 403s on an actual change; an unchanged round-trip is dropped so the rest of the save proceeds. A `settings.edit`-only role now saves the General tab fine and is blocked only when it truly changes `general_site_url`. Confirmed `/general` was the only affected route.

**🟡 Guard test committed** — `__tests__/routes/adminRolesGuards.test.js` pins `assertActorMayGrant`: a non-super can't grant a perm its own role lacks, can't edit its own role, can't escalate another role; super_admin bypasses; `team_photographer` name reserved. Also added `__tests__/routes/adminSettingsProtectedKeys.test.js` for the change-detection itself — unchanged `general_site_url` saves (200), an actual change 403s, super_admin changes it. Both green.

**🟢 Nit (grantable flag)** — agreed. I'll have `GET /roles/permissions` return a per-permission `grantable` flag so the matrix greys out what the caller can't grant instead of 403-on-submit. Glad to fold it into this PR before merge if you'd prefer, otherwise it lands as a fast follow.

@the-luap Ready for another look.

the-luap (approved):
Re-verified `e6dd4bbc`. Both round-2 items closed — merge cleared.

## Round-2 items — closed

**🔴 403 over-fire** — `adminSettings.js:91-107` now reads the stored `app_settings` row, JSON-parses it, and only pushes to `denied` on an actual change; an unauthorized no-op is dropped so the rest of the save proceeds. Re-ran the repro from my round-2 comment against this head:

```
unchanged site_url                 -> 200   (was 403 on 79d3507e)
general_max_file_size_mb persisted -> 75
actual change                      -> 403   general_site_url / settings.domains
stored site_url                    -> null  (not written)
```

Also probed four cases beyond the committed tests, all correct: a real stored URL round-tripped unchanged passes; changing it 403s with the stored value intact; `null` against a stored URL is treated as a change, not a no-op (the `String(x ?? '')` coercion doesn't open a hole); and `accounting_*` is still refused through the `/general` writer.

**🟡 Guard tests** — `adminRolesGuards.test.js` (6 cases) + `adminSettingsProtectedKeys.test.js` (3). 16 pass locally alongside `authzPermissionGaps`. The guard that makes `roles.manage` a delegation primitive rather than a root escalation is now pinned.

## Pre-merge state

- 0 commits behind `main`, merge simulation clean
- `175_granular_permissions_and_presets` sits correctly after `174_sqlite_nullable_event_dates` on `main` — no collision, `migrationOrder` separates them
- All required checks green except `bypass-size-gate`, which fails by design: 2546 lines against its 300-line self-merge cap. That gate exists to force a maintainer review on a large bypass-user PR, which is what these two rounds were. Merging with an admin override on that basis.

## Grantable flag

Take it as a fast follow rather than folding it in — this PR is already large, and the 403-on-submit enforces the boundary correctly today. The flag is a UX improvement on top, better reviewed on its own.

Nice work on the two rounds — the transitive `cloneRole` case and the reserved-name hole were both closed without being asked.

Issue #858, opened 2026-07-21
[FEATURE] Download Resolutions


eopo:
**Is your feature request related to a problem? Please describe.**
Some of my clients require Images to be in a certain resolution, smaller than what I usually export. Currently, this requires me to export images in this resolution from Lightroom manually.

**Describe the solution you'd like**
Clients can choose from multiple resolutions or original size when downloading images. After choosing download for one or multiple images, they are asked what resolution they want.

**Use Case**
- Clients can download smaller versions of their images, e.g. for Social Media usage
- The same image is available in different resolutions without manual intervention.

**Implementation Ideas**
These resolutions could be generated on the fly or pre-generated. In the case of pre-generation, these resolutions should be deleted on archiving of galleries to save space.

the-luap:
understood the requirement. which resultion do you expect to have. is it based on file size or real resolution and if yes, which resolutions would you prefer?

eopo:
I'd favor admin configurable file sizes, e.g. 3000x2000, 1500x1000, ... with a "fit in"-Approach (e.g. non 3:2-Images are slightly smaller, because the configured size is a "up to"-Definition). Ideally optionally "no upscaling" checkbox.

the-luap:
Implemented in #1022, shipping from `v3.102.0-beta.0`.

@eopo — your spec drove the design directly: sizes are admin-configurable, treated as a **"fit inside" upper bound** (`fit: 'inside'` + `withoutEnlargement`), so a non-matching aspect ratio comes out slightly smaller on one edge and an image already below the box is left alone rather than upscaled. No separate "no upscaling" checkbox — it's unconditional, since upscaling a delivered photo is never what you want.

Two settings, both **off by default**, so nothing changes until you turn them on:

**Standard resolution** — the size a gallery hands out for every ordinary download (single photo, selection, download-all). Global default in **Settings → Download resolutions**, overridable per gallery on the event's Overview tab with an explicit "Inherit" option. It applies to single-photo downloads too, otherwise a lowered standard is trivially bypassed one photo at a time.

**Resolution picker** — an opt-in modal letting guests choose a different size when they download. Defaults ship as Large 3000×2000 / Medium 1500×1000 / Small 800×600, and the list is editable.

Two behaviours worth knowing:

- The picker never offers a size **above** the gallery's standard. If you set the standard to 1500×1000, guests can't help themselves to full-res — "Original" reappears only if you explicitly enable it. Lowering the standard is meant to mean something.
[...]

PR #1048, opened 2026-08-14
fix: single-photo gallery downloads 404 on S3 storage backends

peipeimo:
## Problem

On deployments with `STORAGE_BACKEND=s3`, every single-photo download from a gallery (`GET /api/gallery/:slug/download/:photoId`) fails with a 404 and logs:

```
Error streaming gallery download ... ENOENT: no such file or directory,
stat '/app/storage/events/active/<slug>/<file>.jpg'
```

The route resolves a **local filesystem path** unconditionally (`resolvePhotoFilePath`) and streams it with `res.sendFile`, but in S3 mode managed photos never exist on local disk. Meanwhile `download-all` (zip) and the `secure-images` routes work fine because they already go through `getStorage()` — which is what makes this easy to miss: the gallery looks fully functional until a guest clicks the per-photo download button.

## Fix

Resolve the photo's storage key first (`resolvePhotoStorageKey`), same pattern as `secure-images`/`protected-images`:

- **watermark/resize branch**: materialize a tmp local copy via `withLocalCopy` in S3 mode and run the identical processing on it (the path-based watermark cache optimization still applies to the copy's path)
- **plain branch**: pipe the storage backend stream directly — keeps memory flat for as-is video downloads, which can be large
- **external-mode photos** (`resolvePhotoStorageKey` → null) keep the existing local-path `res.sendFile` fallback unchanged

## Tested

On a production-style deployment (v3.45.16 + this change cherry-picked, Cloudflare R2 as the S3 backend, postgres+redis, official images):

- before: per-photo download → 404, ENOENT in error.log
- after: HTTP 200, downloaded bytes byte-identical to the uploaded original (1,373,137 bytes), `file` confirms valid JPEG
- zip download and photo viewing unaffected

### DIFF
--- a/backend/__tests__/routes/galleryDownloadStorage.test.js
+++ b/backend/__tests__/routes/galleryDownloadStorage.test.js
@@ -237,4 +237,33 @@ describe('single-photo download through the storage backend (#1048)', () => {
     expect(res.headers['content-type']).toMatch(/json/);
     expect(res.headers['content-range']).toBeUndefined();
   });
+
+  it('answers HEAD from stat instead of draining the object out of S3', async () => {
+    const res = await request(app).head(`/api/gallery/${SLUG}/download/${photoId}`);
+
+    expect(res.status).toBe(200);
+    expect(res.headers['content-length']).toBe(String(mockObjectBody.length));
+    expect(res.headers['accept-ranges']).toBe('bytes');
+    // The whole point: no egress for a metadata probe.
+    expect(mockStorage.get).not.toHaveBeenCalled();
+    expect(mockStorage.getRange).not.toHaveBeenCalled();
+  });
+
+  it('returns a clean error when the range stream dies before its first chunk', async () => {
+    // Resolves, then errors — writeHead would already have committed the 206,
+    // leaving a connection reset as the only possible outcome.
+    const { Readable: R } = require('stream');
+    mockStorage.getRange.mockImplementationOnce(async () => {
+      const dead = new R({ read() { this.destroy(new Error('socket hang up')); } });
+      return dead;
+    });
+
+    const res = await request(app)
+      .get(`/api/gallery/${SLUG}/download/${photoId}`)
+      .set('Range', 'bytes=0-9');
+
+    expect(res.status).toBe(500);
+    expect(res.headers['content-type']).toMatch(/json/);
+    expect(res.headers['content-range']).toBeUndefined();
+  });
 });
--- a/backend/__tests__/routes/galleryDownloadStorage.test.js
+++ b/backend/__tests__/routes/galleryDownloadStorage.test.js
@@ -187,4 +187,54 @@ describe('single-photo download through the storage backend (#1048)', () => {
 
     fs.rmSync(abs, { force: true });
   });
+
+  it('does not serve a partial body when the If-Range validator is stale', async () => {
+    // The object was replaced since the client's last attempt. Answering 206
+    // from the new bytes would let it splice two versions into one file.
+    const res = await request(app)
+      .get(`/api/gallery/${SLUG}/download/${photoId}`)
+      .set('Range', 'bytes=0-9')
+      .set('If-Range', new Date('2020-01-01T00:00:00Z').toUTCString());
+
+    expect(res.status).toBe(200);
+    expect(res.headers['content-range']).toBeUndefined();
+    expect(res.headers['content-length']).toBe(String(mockObjectBody.length));
+  });
+
+  it('still serves 206 when the If-Range validator matches', async () => {
+    const res = await request(app)
+      .get(`/api/gallery/${SLUG}/download/${photoId}`)
+      .set('Range', 'bytes=0-9')
+      .set('If-Range', new Date('2026-08-20T10:00:00Z').toUTCString());
+
+    expect(res.status).toBe(206);
+    expect(res.headers['content-range']).toBe(`bytes 0-9/${mockObjectBody.length}`);
+  });
+
+  it('errors cleanly when the object vanishes between stat and get', async () => {
+    // HeadObject succeeding does not mean GetObject will — a concurrent
+    // delete lands here. The staged image headers must not escape with it.
+    const gone = new Error('NoSuchKey');
+    gone.name = 'NoSuchKey';
+    mockStorage.get.mockRejectedValueOnce(gone);
+
+    const res = await request(app).get(`/api/gallery/${SLUG}/download/${photoId}`);
+
+    expect(res.status).toBe(404);
+    expect(res.headers['content-type']).toMatch(/json/);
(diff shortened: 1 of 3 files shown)


the-luap:
Thanks for this — the diagnosis is exactly right, and the reason it went unnoticed (download-all and secure-images already go through `getStorage()`, so the gallery looks healthy until a guest clicks the per-photo button) is a good catch.

One blocker before it can go in, though: the fix is correct for S3 but changes behaviour for **local-disk** installs too.

## 🔴 Blocker — local installs lose `sendFile` semantics

`resolvePhotoStorageKey` returns a key for every managed photo regardless of backend, so `storageKey` is truthy on local-disk installs as well. That means the majority of deployments stop using `res.sendFile` and get a raw `stream.pipe(res)`.

Measured on a 512 KB file, `sendFile` vs a bare pipe:

```
             Content-Length   Accept-Ranges   ETag   Last-Modified
sendFile     524288           bytes           yes    yes
pipe         —                —               —      —

Request with `Range: bytes=0-99`
sendFile  -> 206 Partial Content, 100 bytes
pipe      -> 200 OK, 524288 bytes        ← Range ignored entirely
```

Three consequences on local disk, all regressions from `main`:

1. **No `Content-Length`** — browsers lose the download progress indicator, which matters most on exactly the large files this route serves.
2. **Range requests break.** A client resuming an interrupted download sends `Range:` and gets back `200` with the *whole* body. Appended onto the partial file, that corrupts it. This hits the as-is video case the PR cites as its reason for piping.
3. **No `ETag`/`Last-Modified`** — conditional requests stop working, so every re-download transfers the full body.

Both adapters already expose `kind()`, `stat()` and `getRange()`, so this is cheap to fix. Simplest version — keep the existing path when the backend is local:

```js
const storage = getStorage();
if (storageKey && storage.kind() !== 'local') {
  // stream from the backend (S3/R2)
} else {
  res.sendFile(filePath ?? resolvePhotoFilePath(req.event, photo), ...);
}
```

If you'd rather keep one code path, setting `Content-Length` from `stat()` and honouring `Range` via `getRange()` also works and is arguably nicer — your call, the branch is lower-risk.

## 🟡 Concerns

**1. The 404 responses inherit image headers.** Both new error paths run after:

```js
res.set({ 'Content-Type': photo.mime_type || 'image/jpeg', 'Content-Disposition': contentDisposition });
```

Express's `res.json()` doesn't override an already-set `Content-Type`, so the error goes out as:

```
status: 404   content-type: image/jpeg   content-disposition: attachment; filename="x.jpg"
body: {"error":"Photo file not found"}
```

The browser saves a `.jpg` containing JSON. Clear the two headers (`res.removeHeader`) before responding, or set them only once the stream is known good.

**2. `LocalFsStorage.get()` never throws for a missing file** — it returns `fs.createReadStream(abs)`, which surfaces `ENOENT` asynchronously. So on local disk the `try/catch` around `getStorage().get()` is dead code and the miss always lands in the `stream.on('error')` handler. Not wrong (the handler is there), just worth knowing the catch isn't what saves you.

**3. No test.** The only file changed is `gallery.js`. A test that stubs `getStorage()` and asserts the route streams from the backend rather than the filesystem would pin the fix — `__tests__/routes/adminEventsCascadeStorage.test.js` from your #1051 is a good model for stubbing the storage layer.

## ✅ Verified

- Imports are all present and correct (`withLocalCopy`, `getStorage`, `resolvePhotoStorageKey` at `gallery.js:36/43/61`).
- The watermark/resize branch is right: `withLocalCopy` mirrors what `secure-images`/`protected-images` do, the path-based watermark cache still applies to the tmp copy, and `res.send(buffer)` sets `Content-Length` itself, so that branch has none of the problems above.
- External/reference photos correctly keep the local-path fallback — `resolvePhotoStorageKey` returns `null` for them, and the mixed-source case (`reference` event holding managed uploads) falls through to `resolvePhotoFilePath`, which handles it.
- Merges clean against `main` and against #1049/#1050/#1051.

## Note for the backport

This bug is present on `stable` too — `gallery.js:908` resolves the local path unconditionally there as well. Since you're running v3.45.16, that's presumably where you actually hit it. Once this lands on `main` we'll want a stable twin.

Close the blocker and I'm happy to take it.

the-luap:
Rewritten on top of current `main` and pushed to your branch. The diagnosis was right; the implementation had to change because `main` moved.

## Why not a rebase

#858 landed `services/downloadRendition.js` — `renderPhotoForDownload()` does exactly the storage-key + `withLocalCopy` + resize-then-watermark dance your patch hand-rolled. It's **already imported in `gallery.js:47`** and used by the zip builders at `:1866` and `:2027`. The single-photo route was the last download path in that file still doing local-only I/O, duplicating logic the service owns.

Resolving the conflict would have preserved the duplicate. The route now delegates, which makes the change mostly a deletion.

## The blocker is closed by branching, not by unifying

`resolvePhotoStorageKey` returns a key for every managed photo regardless of backend, so your version routed local-disk installs — the majority — through a bare `stream.pipe(res)`. Measured on a 512 KB file:

```
             Content-Length   Accept-Ranges   ETag   Last-Modified
sendFile     524288           bytes           yes    yes
pipe         —                —               —      —

Request with `Range: bytes=0-99`
sendFile  -> 206 Partial Content, 100 bytes
pipe      -> 200 OK, 524288 bytes        ← Range ignored entirely
```

Local disk keeps `sendFile`. On S3 the two that matter for a download are reproduced via `stat()` and `getRange()`: `Content-Length` (progress indicator, matters most on the large files this route serves) and `Range` (a resumed download otherwise appends a second full body onto the partial file — which hits the as-is video case your PR cited as its reason for piping). Conditional requests aren't reproduced; there's no ETag on that path, so a revalidating client gets the whole body, same as today. That's stated in the comment rather than left implicit.

Ranges are parsed defensively — an unchecked parse yields `NaN` bounds and a `206` with a nonsense `Content-Range`, which corrupts a resumed download rather than merely failing it.

The other two review items are in: the pre-stream 404s now run before any image header is staged, so the error goes out as JSON instead of a `.jpg` containing JSON; and `pipeStreamToResponse` handles the mid-stream case, which clears staged headers itself.

## One correction to the problem statement

It doesn't 404. `res.sendFile` is called **with a callback**, so Express doesn't send a response on error and the route's callback only logs — the request hangs until the client gives up. The tests pin this: all five backend-path cases *time out* against the previous implementation rather than failing fast. Whatever produced the 404 you saw was downstream, probably your proxy.

## Tests

Six in `backend/__tests__/routes/galleryDownloadStorage.test.js`, the item the review asked for. The mock object's bytes are never written to disk, so a filesystem read can't produce them. Includes the local-backend case, which passes both before and after — that's the regression guard.

## Stable

#1246, written against stable's shape — `renderPhotoForDownload` doesn't exist there, so the watermark half uses `withLocalCopy` directly, matching stable's own zip builders.

All checks green.

PR #1049, opened 2026-08-14
fix(storage): add S3 client timeouts so a dropped connection can't wedge uploads

peipeimo:
## The problem

Under sustained upload load to an S3-compatible backend, the backend eventually hangs: an upload request is accepted, the handler never completes, CPU goes idle, and no socket to the storage endpoint remains open. Every subsequent storage operation queues behind it forever — process-wide. Only a backend restart recovers.

Observed on v3.45.16 against Cloudflare R2 during a bulk migration (~19k photos): the wedge hit roughly every 40 minutes with serial uploads and every ~15–20 minutes with 4 parallel uploaders.

## Root cause

`S3StorageAdapter` constructs its `S3Client` with no `requestHandler` timeouts, and the SDK's defaults wait indefinitely. When a connection is dropped silently (no FIN/RST delivered — NAT/LB idle reaps, transient network faults), the in-flight request waits forever. The adapter's own `_retryOperation` exponential-backoff logic never gets a chance to run, because the promise it wraps never settles.

## The fix

Configure the client's `requestHandler` with `connectionTimeout` (default 120 000 ms) and `socketTimeout` (default 60 000 ms), overridable via config or the new `STORAGE_S3_CONNECTION_TIMEOUT` / `STORAGE_S3_SOCKET_TIMEOUT` env vars, and add `'TimeoutError'` to `_retryOperation`'s retryable list so a timed-out request is retried with the existing backoff.

**Two things I got wrong first, both found in production — worth recording so reviewers don't repeat them:**

**1. `requestTimeout` is the wrong knob.**

| | `requestTimeout` | `socketTimeout` |
|---|---|---|
| measures | total request duration | socket inactivity |
| large upload | aborts a legitimate slow/large transfer | unaffected — any byte resets the timer |
| on expiry | **only logs a warning** unless `throwOnRequestTimeout: true` | destroys the request, rejects with `TimeoutError` |

With `requestTimeout: 10000` the backend logged hundreds of `a request has exceeded the configured 10000 ms requestTimeout` warnings while continuing to hang — detection without action.

**2. The values must be generous, not tight.**

`connectionTimeout`'s timer starts when the request object is created and is only cleared once a socket is both *assigned and connected* (`set-connection-timeout.js`). Time spent **queuing for a free socket** from the agent pool (`maxSockets: 50`) therefore counts against it. A 10 s value looks reasonable and is not: under concurrent upload load it expires while merely waiting in line, and **every read fails** — admin photo downloads, thumbnail serving, and background thumbnail generation all returned 500 with `TimeoutError` until the value was raised.

These timeouts exist to convert an *infinite* hang into a *bounded* failure that the retry path can act on. They are not latency targets, and tightening them re-breaks the app in a new way.

No new dependencies: passing a plain config object as `requestHandler` is supported by the bundled SDK (resolves to a configured `NodeHttpHandler` — verified against `@aws-sdk/client-s3@^3.850`, `@smithy/node-http-handler@4.4.12`).

## Testing

- 5 new unit tests: defaults reach the client config; overrides work; `requestTimeout` is deliberately *not* set; `connectionTimeout`/`socketTimeout` are held above sane floors so a future tightening trips a test; `TimeoutError` is retried. Full `s3Storage.test.js` suite passes (23/23).
- Deployed on our production v3.45.16 instance (R2, Hetzner VPS): uploads run clean, and photo/thumbnail reads verified returning 200 with correct byte counts after the timeout values were corrected.

### DIFF
--- a/backend/src/services/storage/__tests__/s3Storage.test.js
+++ b/backend/src/services/storage/__tests__/s3Storage.test.js
@@ -71,24 +71,32 @@ describe('S3StorageAdapter', () => {
         expect.objectContaining({
           requestHandler: {
             connectionTimeout: 10000,
-            requestTimeout: 10000
+            socketTimeout: 10000
           }
         })
       );
+    });
+
+    it('should not set requestTimeout, which caps total duration and only warns', () => {
+      // requestTimeout would abort legitimate large uploads (it is a
+      // total-duration cap, not inactivity) and by default only logs a
+      // warning — it needs throwOnRequestTimeout to abort at all.
+      const [[config]] = S3Client.mock.calls;
+      expect(config.requestHandler).not.toHaveProperty('requestTimeout');
     });
 
     it('should allow overriding timeouts via config', () => {
       new S3StorageAdapter({
         bucket: 'test-bucket',
         connectionTimeout: 5000,
-        requestTimeout: 30000
+        socketTimeout: 30000
       });
 
       expect(S3Client).toHaveBeenLastCalledWith(
         expect.objectContaining({
           requestHandler: {
             connectionTimeout: 5000,
-            requestTimeout: 30000
+            socketTimeout: 30000
           }
         })
       );
--- a/backend/src/services/storage/__tests__/s3Storage.test.js
+++ b/backend/src/services/storage/__tests__/s3Storage.test.js
@@ -70,11 +70,22 @@ describe('S3StorageAdapter', () => {
       expect(S3Client).toHaveBeenCalledWith(
         expect.objectContaining({
           requestHandler: {
-            connectionTimeout: 10000,
-            socketTimeout: 10000
+            connectionTimeout: 120000,
+            socketTimeout: 60000
           }
         })
       );
+    });
+
+    it('should keep connectionTimeout generous enough to survive socket-pool queuing', () => {
+      // connectionTimeout starts at request creation and only clears once a
+      // socket is assigned AND connected, so waiting for a free socket from
+      // the agent pool counts against it. A short value (e.g. 10s) fails
+      // every read under concurrent upload load. These timeouts bound an
+      // infinite hang; they are not latency targets.
+      const [[config]] = S3Client.mock.calls;
+      expect(config.requestHandler.connectionTimeout).toBeGreaterThanOrEqual(60000);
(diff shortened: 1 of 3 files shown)


the-luap (approved):
Merging this. The debugging behind it is solid and the commit history shows the two false starts being caught — `requestTimeout` → `socketTimeout` in particular is the distinction most people get wrong, and the code comment explaining why (`requestTimeout` is a total-duration cap that would kill legitimate large uploads, and by default only warns without `throwOnRequestTimeout`) is worth having in the file.

## ✅ Verified

- **The object-form `requestHandler` actually applies** on the SDK we ship (`@aws-sdk/client-s3` 3.1000.0). Resolved it directly:
  ```
  handler: NodeHttpHandler
  config:  { connectionTimeout: 120000, socketTimeout: 60000, httpsAgent: { maxSockets: 50, ... } }
  ```
  That `maxSockets: 50` also confirms your reasoning about the generous `connectionTimeout` — the agent pool is exactly where a short value would expire while merely queuing for a socket, and it would take out reads (thumbnails, downloads) rather than the uploads it was aimed at.
- **Config reaches the client.** `S3StorageBackend` forwards its whole config to `S3StorageAdapter`, so the two new `index.js` values land on the real client rather than being silently dropped.
- **`TimeoutError` is in the retryable list**, so `_retryOperation` now actually engages on a tripped timeout instead of failing straight out. That's the half that makes the timeouts useful rather than just converting a hang into an error.
- 23 tests pass locally, including the new "retry when the request handler times out a dead connection" case.
- Defaults preserve current behaviour for anyone not on S3, and both values are env-overridable.

## 🟢 Nit (not blocking, taking it as-is)

`STORAGE_S3_CONNECTION_TIMEOUT` and `STORAGE_S3_SOCKET_TIMEOUT` aren't in `.env.example`, where the rest of the `STORAGE_S3_*` family is documented (lines 164-172). Happy to add them in a follow-up — no need to spin this PR for it.

## Backport

This is present on `stable` too — `s3Storage.js` there has no `requestHandler` at all, and since you reported the wedge against v3.45.16 that's the branch you're actually running. A stable twin is warranted; we'll sort that out separately.

Thanks for the careful writeup — the reproduction detail (every ~40 min serial, ~15-20 min at 4 parallel uploaders) made this straightforward to evaluate.

Issue #608, opened 2026-06-08
[BUG] Delete on admin page doesn't delete on my raspberry Pi


jodrmx:
Hello,
When i'm on the admin page. I want to delete event, right after writing "DELETE" on the admin page it does disappear but when i go to storage/events/active on my raspberry the directories are still there.

**To Reproduce**
1. Go to events
2. Click on "delete selected"
3. Event does disappear 
4. See error in storage/events/active. Files are still present

When i delete an event on the admin page it should also be remove from my storage ?

**Screenshots**

<img width="1232" height="401" alt="Image" src="https://github.com/user-attachments/assets/7394781a-42ee-4ea3-a975-387f0e691500" />

<img width="1304" height="60" alt="Image" src="https://github.com/user-attachments/assets/19e74470-a76a-44f2-8abb-7c13c6d80316" />

<img width="1162" height="382" alt="Image" src="https://github.com/user-attachments/assets/86212d25-3fd4-48b6-b1ef-73797e70dad2" />

**Environment (please complete the following information):**
 - OS Pi Lite
 - Browser: Firefox 151.0.3 (64 bit)
 - PicPeak Version: Frontend: v3.44.0 / Backend: v3.44.0
 - Deployment Method: Docker compose
 - Database: [e.g. PostgreSQL 15, SQLite]

**Logs**
I can't find relevant logs to presents, that's why i'm here to get some help.

Best regards

the-luap:
Confirmed bug, tracked it down. Thanks for the precise repro @jodrmx.

The delete cascade in `adminEvents.js` was reading `event.folder_path` to figure out which folder to remove — but that column is never written anywhere in the codebase. Two reads in this one function, zero writes anywhere. So it's always `undefined`, the `if (event.folder_path)` branch silently no-op'd, and every event delete since this cascade landed orphaned its photos on disk. DB-side cascade worked fine, which is why the row vanished from the admin UI — only the filesystem cleanup was skipped.

Files actually live at `events/active/{slug}/` (and `events/archived/{slug}/` after the archive flow). Every other code path in the repo uses `event.slug` for the folder name. The delete was the only one looking at the wrong column.

Fix is up in #611 — drop the `event.folder_path` read, use `event.slug`, hit both `events/active/{slug}` and `events/archived/{slug}` so archived events also clean up. Goes into the next beta and then onto stable through release-please.

[...]

PR #1051, opened 2026-08-14
fix(events): delete stored objects when cascading an event delete

peipeimo:
## The problem

Deleting an event removes its database rows but leaves every stored file behind. On a local-disk install the folder cleanup added in #608 handles this. On an **S3-compatible backend (S3, R2, MinIO, Spaces) nothing is removed at all** — the objects stay in the bucket, unreferenced by any row, invisible in the UI, and billed every month forever.

`deleteEventCascade()` cleans up with `fs.rm(path.join(storagePath, 'events', sub, event.slug))`. When `STORAGE_BACKEND=s3` those paths don't exist locally, so the call succeeds against nothing and the real objects are never touched. The function never consults `getStorage()`.

## Measured, not inferred

Self-hosted v3.45.16 against Cloudflare R2, deleting one 403-photo event:

| | bucket objects | referenced by DB |
|---|---|---|
| before `DELETE /api/admin/events/:id` | 5,400 | 3,425 |
| after (HTTP 200, rows gone) | **5,400** | 2,746 |

The object count does not move. Those 403 photos and their thumbnails simply became unreachable.

This compounds badly during bulk imports, because loading a gallery is not idempotent — a failed load has to be deleted and redone, so the documented recovery path is exactly the path that leaks. Four failed attempts at one collection left us 1,158 orphaned originals (32 GiB); **58% of everything in our bucket was unreachable junk** before we swept it by hand.

## The fix

Collect the event's storage keys *before* the transaction removes the photo rows, then delete them through the storage backend after it commits.

Both halves of that ordering matter:

- **Collect first** — once the rows are gone, nothing records which objects belonged to the event. Only a full-bucket audit against the whole database could find them again.
- **Delete after commit** — a rolled-back transaction must never leave files destroyed for an event that still exists.

This mirrors what single-photo delete already does correctly in `routes/adminPhotos.js`, including the derived tiers (`thumbnail_path`, `hero_path`, `preview_path`), and reuses `resolvePhotoStorageKey()` so reference/external photos — which live outside the managed backend — are correctly left alone.

Failures are logged rather than thrown, matching the existing philosophy in this function: the database is the source of truth, an orphaned object is recoverable noise, a half-deleted event is not.

Local-disk installs are unaffected in behaviour (the existing `fs.rm` still runs; the storage adapter deletes the same files through its own path).

## Testing

3 new tests in `__tests__/routes/adminEventsCascadeStorage.test.js`:
- originals **and** every derived tier are deleted through the backend — this test fails on `main` and passes with the fix
- reference/external photos are left untouched
- a throwing storage backend does not break the delete

The mock deliberately returns an empty photo list once the rows are deleted, so collecting the keys too late fails the test rather than passing silently.

## Note for existing installs

Anyone who has deleted a gallery on an S3 backend is already carrying orphans that this fix cannot retroactively remove. A `storage:reconcile` maintenance command would be a natural follow-up — happy to contribute one if you'd like it.

### DIFF
--- a/backend/__tests__/routes/adminEventsCascadeStorage.test.js
+++ b/backend/__tests__/routes/adminEventsCascadeStorage.test.js
@@ -31,7 +31,16 @@ const mockEvent = {
   // Written through the backend by archiveService, so it is a bucket object
   // and the fs.unlink in the cascade never touched it on S3.
   archive_path: 'archives/other-demo-2026-01-01.zip',
+  // The pre-built "Download All" zip. Lives under the event prefix, so the
+  // recursive fs.rm covers it on local disk and nothing covers it on S3.
+  download_zip_path: 'events/active/other-demo-2026-01-01/.download-cache/all.zip',
 };
+
+// One zip per custom-resolution download job, same prefix. The rows are
+// ON DELETE CASCADE, so they must be read before the transaction.
+const mockDownloadJobs = [
+  { zip_path: 'events/active/other-demo-2026-01-01/.download-cache/job-abc123.zip' },
+];
 const mockPhotos = [
   {
     id: 1,
@@ -64,13 +73,18 @@ const mockPhotos = [
 ];
 
 let mockPhotoRowsDeleted = false;
+let mockJobRowsDeleted = false;
 
 function mockMakeDb() {
   const table = (name) => {
     const chain = {
       where: () => chain,
       first: async () => (name === 'events' ? mockEvent : undefined),
+      whereNotNull: () => chain,
       select: async () => {
+        if (name === 'download_jobs') {
+          return mockJobRowsDeleted ? [] : mockDownloadJobs;
+        }
         if (name === 'photos') {
           // The whole point: if this runs after the transaction, the rows
           // are gone and we would collect nothing.
@@ -80,13 +94,14 @@ function mockMakeDb() {
       },
       del: async () => {
         if (name === 'photos') mockPhotoRowsDeleted = true;
+        if (name === 'download_jobs') mockJobRowsDeleted = true;
         return 1;
       },
     };
     return chain;
   };
   // #1132 guards the merge-dismissals delete behind a hasTable check.
-  table.schema = { hasTable: async () => false };
+  table.schema = { hasTable: async (t) => t === 'download_jobs' };
   table.transaction = async (cb) => cb(table);
   return table;
 }
@@ -106,6 +121,7 @@ describe('deleteEventCascade — storage cleanup', () => {
   beforeEach(() => {
     mockStorage.delete.mockClear();
     mockPhotoRowsDeleted = false;
+    mockJobRowsDeleted = false;
   });
 
   it('deletes originals and every derived tier from the storage backend', async () => {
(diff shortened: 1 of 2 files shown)


the-luap:
Good fix, and the before/after object counts are the right way to report it — "5,400 → 5,400 while referenced rows dropped by 679" is much more convincing than a description of the code path.

The shape is right throughout: keys collected **before** the transaction (after it, nothing knows what belonged to the event), deletion **after** commit (so a rolled-back delete can't destroy referenced files), external/reference photos skipped, failures logged rather than thrown. It mirrors the single-photo precedent at `adminPhotos.js:654-679` closely, including the "derived tiers are canonical keys, pass through verbatim" semantics.

## ✅ Verified

- `event` is in scope at the collection site (fetched at the top of `deleteEventCascade`).
- `resolvePhotoStorageKey` returns `null` for reference/external, so those are correctly left alone — and your second test pins it.
- Post-commit ordering is genuinely better than the existing filesystem cleanup, which runs *inside* the transaction.
- 3 tests pass locally, including the storage-backend-throws case.

## 🟡 Three things still leak on S3

All in the same function, all the same class of bug you're fixing — worth either folding in or making an explicit scope call:

1. **`photo.watermark_path`** — not collected. Single-photo delete handles it via `watermarkGeneratorService.deleteForPhoto()`, which bottoms out in `watermarkService.deleteWatermarkFile()` → `getStorage().delete()`. So pre-generated watermarks are storage-backed and leak on cascade exactly like the photos did. It's not in your `select()` list either, so it'd need adding there too.
2. **`event.archive_path`** — still `fs.unlink` only. The archive zip is typically the largest single object an event owns.
3. **`event.hero_logo_path`** — same, `fs.unlink` only.

None of these make the PR worse than the status quo, so I won't block on them. But a partial fix to a silent-billing-leak bug is the kind of thing that gets marked "fixed" and then quietly keeps costing money, so I'd rather they go in now than become a follow-up nobody opens. If you'd prefer to keep this PR tight, say so and I'll open the issue myself.

## 🟢 Nit

On the local backend this now does N individual `storage.delete()` calls *and* the recursive folder `fs.rm`. Harmless and idempotent, just redundant work proportional to photo count. Skipping the per-key loop when `getStorage().kind() === 'local'` would avoid it, though the folder rm already covers that case entirely.

## Note for the backport

Present on `stable` — its `deleteEventCascade` never calls `getStorage()` at all. Given this one costs real money every month it goes unfixed, I'd treat the stable twin as the higher priority of the two.

the-luap:
Rebased onto `main` and pushed to your branch, plus a follow-up commit. Sorry for the wait on this one.

## Why it conflicted — and why the resolution isn't a plain merge

`main` moved underneath it. #1095 independently adopted the **exact** collect-before-transaction / sweep-after-commit pattern you proposed, for responsive tiers: `helpers.js` now reads a `tieredPhotos` list before the transaction and sweeps via `deleteThumbnailTiers`/`deletePreviewTiers` after the commit, both storage-backed. Your `storageKeys` collection landed on precisely those lines.

Two queries doing the same thing over different columns is the wrong shape, so the resolution folds them into one `select()` that serves both purposes. Your commit keeps its authorship.

Worth stating plainly: your design was right and `main` arrived at it independently a week later. What's still missing on `main` after #1095 is the part that matters most — the **originals**. Tiers are cleaned; `photo.path` objects are not.

## Follow-up commit — two of the three leaks from the review

- **`photo.watermark_path`** — added. Storage-backed via `watermarkService.deleteWatermarkFile`, and `archiveService.js:227` already deletes it through `getStorage()` on archive. Leaked here exactly like the originals.
- **`event.archive_path`** — added. `archiveService.js:160` writes it with `storage.putFromFile`, so `fs.unlink` is a no-op on S3, and it's typically the largest single object an event owns.
- **`event.hero_logo_path`** — **not** added, and my review was wrong to list it. Multer writes logos to local disk with `diskStorage` regardless of backend (`adminEvents/logo.js:19-28`), so they're never bucket objects and the existing `fs.unlink` is correct.

I also dropped the nit about skipping the per-key loop on the local backend. It doesn't hold: only the originals live under the event folder that `fs.rm` removes — tiers, watermarks and the archive don't, so skipping would leak them on local disk too. #1095's tier sweep already established deleting through the backend on both engines.

Collection is a `Set` now: an unresized gallery can carry one object in both `hero_path` and `preview_path`, and the second delete would log a spurious failure.

## Tests

Your three plus two. Your db mock needed `trx.schema.hasTable` (added by #1132) to run against current `main`, and I pointed `STORAGE_PATH` at a temp dir — the cascade runs a real recursive `fs.rm` and the default resolves into the working tree.

Verified both directions rather than assuming: the two substantive tests fail on `main` and pass with the fix.

## Stable

#1245 is open, written against stable's own shape (no `tieredPhotos` there). That's the higher priority of the pair — stable's `deleteEventCascade` never calls `getStorage()` at all.

All checks green.

PR #1055, opened 2026-08-16
fix(pdf): RFC 6266-encode Content-Disposition on quote/invoice PDFs (#1024)

### LINKED ISSUE #1024: Quote/Invoice PDF preview returns 500 for customers with non-ASCII names (umlauts) — Content-Disposition ERR_INVALID_CHAR
## Summary
Opening the PDF preview/download of a quote or invoice returns **HTTP 500** whenever the associated customer's name or company contains a non-ASCII character (e.g. German umlauts `ä ö ü ß`). The admin UI shows only a generic toast: `Request failed with status code 500`.

## Severity
**High** — the feature is completely broken for any customer with an umlaut in their name, which is the common case on German/DACH instances. Affects both admin and customer-facing PDF endpoints.

## Environment
- Reproduced on a production instance (Docker Compose, PostgreSQL). All migrations applied — not a schema/migration problem.
- The backend logs the error to `/app/logs/error.log` (winston), **not** stdout, so `docker compose logs backend` looks empty — worth knowing when diagnosing.

## Steps to reproduce
1. Create a customer whose company or name contains an umlaut, e.g. `Müller Fotografie`.
2. Create a quote (or invoice) for that customer.
3. Open the quote detail page and click **Preview / PDF** → `GET /api/admin/quotes/:id/pdf`.

## Expected
The PDF renders inline in the browser.

## Actual
- Request returns **500**; response body (production): `{"error":"An unexpected error occurred","code":"INTERNAL_ERROR"}`.
- The admin routes 500 with the generic body; the customer-facing routes 500 with `Failed to render quote PDF` (their own catch) — same root cause.

## Root cause
The PDF routes set `Content-Disposition` by inlining the filename directly into a template string:

```js
// backend/src/routes/adminQuotes.js
res.set('Content-Disposition', `inline; filename="${filename}"`);
```

`filename` comes from `buildPdfFilename()` (`backend/src/utils/pdfFilename.js`), which **deliberately preserves non-ASCII characters** (per its own docstring: *"non-ASCII letters preserved"*). For `Müller Fotografie` the filename becomes `Q-2026-xxxx_Müller-Fotografie.pdf`.

HTTP header values must be Latin1/ASCII. Node rejects the raw `ü` (U+00FC) with:

```
TypeError [ERR_INVALID_CHAR]: Invalid character in header content ["Content-Disposition"]
    at ServerResponse.setHeader (node:_http_outgoing:703:3)
    at ServerResponse.header (/app/node_modules/express/lib/response.js:794:10)
    at /app/src/routes/adminQuotes.js:548:9
```

The throw happens **after** the PDF buffer is generated successfully — purely at header serialization — so the whole request fails with an uncaught 500.

## Affected call sites
All inline `filename="${filename}"` where `filename` comes from `buildPdfFilename()` (i.e. carries a customer label):

- `backend/src/routes/adminQuotes.js` — persisted quote PDF + preview payload
- `backend/src/routes/adminInvoices.js` — persisted invoice PDF + preview payload
- `backend/src/routes/customer.js` — customer-facing quote + invoice PDF

## Suggested fix
The codebase already ships an RFC 6266 helper — `buildContentDisposition(name, disposition)` in `backend/src/utils/filenameSanitizer.js` (already used by `secureImages.js`). It emits an ASCII-safe `filename="…"` **and** an RFC 5987 `filename*=UTF-8''…`, so modern browsers keep the unicode filename while the header stays valid:

```js
res.set('Content-Disposition',
  require('../utils/filenameSanitizer').buildContentDisposition(filename, 'inline'));
```

Applied to all six call sites above. I have this change ready and will open a PR.

## Related notes
- `buildPdfFilename`'s docstring claiming "non-ASCII letters preserved … used by Content-Disposition headers" is misleading — a preserved umlaut is exactly what breaks the header. Worth updating to point callers at `buildContentDisposition`.
- Worth a sweep for any other raw `filename="${…}"` header sets that carry user-supplied strings.

the-luap:
Closes #1024.

## Root cause — with a correction to the issue

The six quote/invoice PDF endpoints built the header by interpolating `buildPdfFilename()`'s result directly:

```js
res.set('Content-Disposition', `inline; filename="${filename}"`);
```

`buildPdfFilename()` deliberately preserves non-ASCII (it doubles as the PDF's internal `Title` metadata, which Chrome uses when saving from a blob URL). HTTP header values are latin1, so a customer label reaching a header directly breaks — but **in two different ways, and the issue attributed the wrong one to umlauts**:

| input | what actually happens |
|---|---|
| **U+0080–U+00FF** — `ä ö ü ß`, i.e. every German umlaut | **No throw.** Node allows `\x80-\xff` in headers. The raw byte is sent and the client reads back a mangled name — `Müller` arrives as `Mýller`. Silent corruption, HTTP 200. |
| **above U+00FF** — Polish `ł`, Czech `ř`, Turkish `ş`, `€`, Cyrillic, CJK, emoji | **`ERR_INVALID_CHAR`.** The throw lands *after* the PDF buffer is rendered, so the whole request fails as an unhandled 500. |

Verified on Node v22.20.0 (what the image ships):

```
OK      ü  U+00FC (German)     -> "Müller"
OK      ß  U+00DF (German)     -> "Straße"
THROWS  ł  U+0142 (Polish)     -> ERR_INVALID_CHAR
THROWS  ř  U+0159 (Czech)      -> ERR_INVALID_CHAR
THROWS  ş  U+015F (Turkish)    -> ERR_INVALID_CHAR
THROWS  Cyrillic / CJK / emoji -> ERR_INVALID_CHAR
```

So `Müller Fotografie` — the issue's stated reproduction — returns 200 with a corrupted filename rather than the reported 500. The 500 in the pasted stack trace is real, but it came from a customer whose name carries a character outside latin1. Both symptoms share this one root cause and both are fixed here, so the fix stands as proposed; the diagnosis just needed narrowing.

## The fix

Route all six sites through `buildContentDisposition()` (`utils/filenameSanitizer`, already used by `secureImages.js`), which emits an ASCII fallback **and** the RFC 5987 `filename*=UTF-8''…` form:

```
inline; filename="Q-2026-0042_M_ller-Fotografie.pdf"; filename*=UTF-8''Q-2026-0042_M%C3%BCller-Fotografie.pdf
```

Modern browsers prefer `filename*` and show the correct unicode name; older ones fall back to the ASCII form. The header is always legal.

Sites changed:
- `adminQuotes.js` — persisted PDF + preview
- `adminInvoices.js` — persisted PDF + preview
- `customer.js` — customer-facing quote + invoice

Also corrected `buildPdfFilename`'s docstring, which advertised its preserved non-ASCII as suitable for `Content-Disposition` headers — the exact misreading that produced these six call sites.

## Tests

`__tests__/utils/pdfContentDisposition.test.js` — 10 cases driving a real express server through supertest, so they exercise Node's actual header validator rather than testing the helper in isolation.

Covers: the umlaut case keeps its name intact; six non-latin1 scripts no longer 500; plain ASCII still gets the familiar `filename="…"` form; the null-customer preview path; and one test pinning the raw interpolation's throw directly, so nobody "simplifies" the helper back out.

**Verified the tests actually catch the bug** — reverting the header line to the old interpolation fails 7 of 10; the 3 that still pass are the ASCII, fallback and root-cause-pin cases, which should be green either way.

Also ran the 24 surrounding suites (quotes, invoices, customer, pdf, filename): **308 passed**. Lint clean on every file touched.

## Scope note

The issue's closing note suggests sweeping for other raw `filename="${…}"` headers. I ran that sweep — there are ~30 more across gallery, backup, contracts, transfers and exports. Most interpolate slugs (ASCII by construction) or timestamps, and a few already `encodeURIComponent`. None are in this PR: they're a separate audit with a different risk profile, and I'd rather this stayed a tight fix for the reported break. Happy to open a follow-up issue for that sweep.

## Backport

Present on `stable` too. Worth a twin once this lands.

### DIFF
--- a/backend/__tests__/utils/pdfContentDisposition.test.js
+++ b/backend/__tests__/utils/pdfContentDisposition.test.js
@@ -27,7 +27,7 @@
 const express = require('express');
 const request = require('supertest');
 
-const { buildPdfFilename } = require('../../src/utils/pdfFilename');
+const { buildPdfFilename, sanitiseSegment } = require('../../src/utils/pdfFilename');
 const { buildContentDisposition } = require('../../src/utils/filenameSanitizer');
 
 // The RFC 5987 parameter prefix, i.e. filename*=UTF-8'' — the two trailing
@@ -103,6 +103,31 @@ describe('#1024 — PDF Content-Disposition with non-ASCII customer names', () =
 
     expect(res.status).toBe(200);
     expect(res.headers['content-disposition']).toContain('quote-preview_customer.pdf');
+  });
+
+  // sanitiseSegment caps each segment at 80 UTF-16 code units. A cap landing
+  // inside an astral character used to leave a dangling high surrogate, which
+  // makes encodeURIComponent throw URIError inside buildContentDisposition —
+  // a 500 on the very endpoint this PR fixes, reached a different way.
+  it.each([
+    ['emoji on the 80-char boundary',      `${'a'.repeat(79)}🎉`],
+    ['astral CJK on the boundary',         `${'a'.repeat(79)}𠜎`],
+    ['a label that is entirely astral',    '🎉'.repeat(60)],
+  ])('does not 500 when truncation splits a surrogate pair — %s', async (_label, company) => {
+    const res = await request(buildApp({ company_name: company })).get('/pdf');
+
+    expect(res.status).toBe(200);
+    expect(res.headers['content-disposition']).toContain(RFC5987_PREFIX);
+  });
+
+  it('drops the orphaned surrogate rather than widening the length cap', () => {
+    const seg = sanitiseSegment(`${'a'.repeat(79)}🎉`);
+
+    // 79 'a's + a half-emoji would be 80; the orphan is dropped, not kept.
+    expect(seg).toHaveLength(79);
+    expect(seg).toBe('a'.repeat(79));
+    // Nothing in the result may be an unpaired surrogate.
+    expect(seg).toBe(seg.replace(/[\uD800-\uDBFF](?![\uDC00-\uDFFF])/g, ''));
   });
 
   it('the raw interpolation these routes used to do really does throw', () => {
--- a/backend/__tests__/utils/pdfContentDisposition.test.js
+++ b/backend/__tests__/utils/pdfContentDisposition.test.js
@@ -76,7 +76,18 @@ describe('#1024 — PDF Content-Disposition with non-ASCII customer names', () =
     const res = await request(buildApp({ company_name: company })).get('/pdf');
 
     expect(res.status).toBe(200);
-    expect(res.headers['content-disposition']).toContain(RFC5987_PREFIX);
+    const cd = res.headers['content-disposition'];
+    expect(cd).toContain(RFC5987_PREFIX);
+    // The legacy filename= token drops non-ASCII, so a name written entirely
+    // in another script degrades to just the document number
+    // (`Q-2026-0042_.pdf`). That's the intended trade — filename* carries the
+    // real name — but the fallback must still be a legal, non-empty,
+    // ASCII-only token, since that is what a client without RFC 5987 support
+    // ends up saving.
+    const fallback = /filename="([^"]*)"/.exec(cd)[1];
+    expect(fallback.length).toBeGreaterThan(0);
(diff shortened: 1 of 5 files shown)


Luca-Timo:
Thank you for the work on this.

Just a bit of background info, the bug arrised on a customer name that is written with a '

PR #1063, opened 2026-08-16
feat(gallery): info banner above the photo grid (#932)

### LINKED ISSUE #932: [FEATURE Enhancement] Global inforessage/banner at the TOP of gallery pages (above the photo grid), separate from the existing promotional banner
**Is your feature request related to a problem? Please describe.**

PicPeak currently has a "gallery promotional banner" feature, but it renders below the gallery content. This works well for promotional/marketing messaging, but doesn't fit a different use case: a short, persistent, informational message that needs to be visible at the TOP of the gallery — before or above the photo grid — so guests see it immediately when the page loads.

Specific use case: informing first-time guests how to use the sidebar/filter menu (e.g. "Use the menu button in the top-left corner to access content filter options"). If this message is placed below the gallery (like the current promotional banner), guests would have to scroll through the entire gallery first — defeating the purpose of an onboarding/usage hint.

**Describe the solution you'd like**

This would be distinct from the existing promotional banner, which could remain below the gallery for marketing/CTA purposes.

Ideally this would be:
- A plain text or lightly-formatted rich text field (not raw CSS injection)
- Configurable as a global default in Event Defaults, so it applies automatically to every new gallery without having to set it per-event
- Optionally dismissible/collapsible by the guest (nice-to-have, not required)

A real informational-banner-at-top feature (or alternatively, documenting the exact theme CSS variables/class names per template so custom CSS is reliable) would solve this more robustly.

the-luap:
Closes #932.

A short informational note rendered at the **top** of a gallery, above the photos. Distinct from the promotional banner (#440), which stays by the footer for marketing copy — the reporter's case is an onboarding hint ("use the menu button to filter"), which is useless below a gallery the guest has to scroll past first.

## Shape

Mirrors the promo feature rather than inventing a second idiom:

- **Global default** — Settings → Branding → *Gallery Info Banner* (`branding_info_markdown`). Applies to every gallery automatically, which is what the issue asked for.
- **Per-event override** — `inherit` / `custom` / `off` on the Edit Event form.
- **Markdown** through the existing `MarkdownContent` sanitiser — no raw HTML, no CSS injection.
- **Empty global default = nothing renders**, so upgrading an existing install changes nothing visible.

Migration `176` is idempotent (`hasColumn` / existing-key guarded) with a working `down()`.

## Deliberately not included

- **Alignment knob.** The promo banner has one because it's marketing layout; this is short helper copy and stays centred with the grid.
- **Guest dismissal.** The issue lists it as a nice-to-have. It needs per-guest persistence (which key? cleared when the copy changes?) — that's its own decision, not something to guess at here.

## A plumbing trap worth knowing about

The per-event fields have to travel in the **`/photos`** response, not only `/info`.

`GalleryAuthContext` seeds its cached `event` from the gallery **login** response, which carries only a small identity subset. Anything absent there is `undefined` immediately after a guest signs in. `/photos` is the payload that refreshes on every gallery load, so that's where the fields belong and why `GalleryView` reads them from `data.event`.

I hit this in the browser: with the fields only on `/info`, the banner silently fell back to the global text for an event explicitly set to `custom`. Every mode "worked" in unit tests and failed in a real session.

## Screenshots

| Before — no banner (also the upgrade state) | After — global default |
|---|---|
| before | after |

| Per-event `custom` override | Admin settings card |
|---|---|
| custom | admin |

Mobile (390×844):

mobile

## Verified in a real browser, not just tests

Ran the stack against a scratch SQLite DB and drove it end to end:

- migration applied cleanly (`176` → `added events.info_mode`, `info_markdown`, `branding_info_markdown`)
- set the global default through the admin UI → persisted → rendered above the grid
- **`inherit`** → global text ✅
- **`custom`** → the event's own text, global suppressed ✅
- **`off`** → banner absent entirely while a global default is still set ✅
- markdown renders (`**bold**`), live preview in admin matches the gallery
- promo banner unaffected throughout — the two slots are independent
- mobile wraps cleanly at 390px

Tests/lint: **29 backend** (new suite + adjacent gallery/event suites), **128 frontend** (24 files), `tsc --noEmit` clean, production build clean. The two `no-unused-vars` errors ESLint reports in `adminEvents/crud.js` are pre-existing on `main` — confirmed by running ESLint against `main` — and untouched here.

## Found while testing: the per-event promo override is broken on `main`

Not fixed here, flagging it.

`GalleryView` reads `data?.event?.promo_mode`, but `/photos` never returned `promo_mode` — verified live:

```
GET /api/gallery/:slug/photos → { event: { info_mode: "custom", promo_mode: ABSENT } }
```

So `promo_mode` is always `undefined` and every per-event promo override collapses to `inherit`. Concretely: an admin who sets a gallery's promo banner to **Off** still gets the global promo banner. Same root cause my feature would have shipped with.

The fix is one line next to mine (`promo_mode: req.event.promo_mode || 'inherit'` in the `/photos` payload), but it changes the behaviour of a shipped feature — galleries currently set to `custom` or `off` would start honouring it — so it deserves its own PR and its own review rather than riding along here. Happy to open the issue.

### DIFF
--- a/backend/__tests__/utils/galleryInfoBanner.test.js
+++ b/backend/__tests__/utils/galleryInfoBanner.test.js
@@ -1,0 +1,177 @@
+/**
+ * Gallery info banner (#932).
+ *
+ * A short note rendered ABOVE the photo grid — the reporter's case is an
+ * onboarding hint ("use the menu button to filter"), which is useless in the
+ * promo slot down by the footer because the guest has to scroll the whole
+ * gallery to reach it.
+ *
+ * Covers what the migration actually produces on a real engine (the harness
+ * runs SQLite) and the inherit/custom/off resolution the gallery render
+ * depends on. The resolution is duplicated here rather than imported because
+ * it lives in the React layer; the point is to pin the CONTRACT — which
+ * source wins for each mode — so a change on either side has to update this
+ * file deliberately.
+ */
+
+const path = require('path');
+const fs = require('fs');
+const os = require('os');
+
+process.env.NODE_ENV = 'test';
+process.env.TEST_DATABASE_PATH = path.join(
+  fs.mkdtempSync(path.join(os.tmpdir(), 'picpeak-info-banner-')), 'db.sqlite',
+);
+process.env.JWT_SECRET = process.env.JWT_SECRET || 'info-banner-test-secret';
+
+const { bootCrmDb, seedMinimal } = require('../integration/helpers/crmDb');
+
+let db;
+let cleanup;
+
+beforeAll(async () => {
+  ({ db, cleanup } = await bootCrmDb());
+  await seedMinimal(db);
+});
+
+afterAll(async () => {
+  if (cleanup) await cleanup();
+});
+
+describe('migration 176 — schema', () => {
+  it('adds events.info_mode defaulting to inherit', async () => {
+    expect(await db.schema.hasColumn('events', 'info_mode')).toBe(true);
+
+    const [id] = await db('events').insert({
+      slug: 'info-default-test',
+      event_type: 'wedding',
+      event_name: 'Info Default Test',
+      event_date: '2026-06-22',
+      host_email: 'host@example.com',
+      admin_email: 'admin@example.com',
+      password_hash: 'x',
+      share_link: '/gallery/info-default-test/share',
+      share_token: 'info-default-test-share',
+      expires_at: new Date(Date.now() + 7 * 24 * 3600 * 1000).toISOString(),
+      is_active: 1,
+      is_archived: 0,
+      is_draft: 0,
+      created_at: new Date().toISOString(),
+    }).returning('id');
+    const eventId = id?.id ?? id;
+
+    const row = await db('events').where({ id: eventId }).first();
+    // A gallery created before anyone configures the feature must inherit,
+    // so switching the global default on lights up every existing gallery.
+    expect(row.info_mode).toBe('inherit');
+    expect(row.info_markdown ?? null).toBeNull();
+  });
+
+  it('adds events.info_markdown as nullable text', async () => {
+    expect(await db.schema.hasColumn('events', 'info_markdown')).toBe(true);
+  });
+
+  it('seeds branding_info_markdown empty, so upgrading shows no banner', async () => {
+    const row = await db('app_settings').where({ setting_key: 'branding_info_markdown' }).first();
+    expect(row).toBeTruthy();
+    expect(JSON.parse(row.setting_value)).toBe('');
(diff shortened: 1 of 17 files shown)


the-luap:
## External review (Codex, 3 rounds) — two real defects fixed

**1. Data loss on the global banner** — `settings.service.ts`

`BrandingSettings` never declared `info_markdown` and `formatBrandingSettings()` never mapped `branding_info_markdown`. `BrandingPage` hydrates with `setBrandingSettings(prev => ({ ...prev, ...formatted }))`, so the unmapped key kept the component's empty-string initializer: the form loaded blank and the **next Save posted `''` back, wiping a configured banner** — silently, because the gallery keeps rendering the old copy until that save lands.

This is the same bug the footer/promo fields hit in #441 + #440 / #460 — the read mapper still carries the comment about it. My browser testing missed it because I set the value and saved in one session; the wipe needs a *reload* then save.

Fixed, plus `settings.brandingRoundTrip.test.ts` pinning the round-trip for the whole editable branding set, so the next field added is caught by a test rather than by someone losing copy. Verified it fails 3/4 with the mapper line removed.

**2. Reveal-hidden view ignored the override** — `GalleryView.tsx`

The hidden-until-reveal branch renders `GalleryLayout` with the context `event`, which is seeded from the gallery **login** response and carries no banner fields. While a gallery was hidden, a per-event **Off** resolved to `inherit` and the global banner appeared on a gallery the admin had muted. Now resolved the same way as the main path.

## Held, with reasoning

Three findings all reduce to *"the info banner faithfully mirrors gaps the promo banner already has"*, and fixing only `info_*` would make the two banners behave differently — a worse inconsistency than the current uniform one. When I put the promo precedent to Codex directly it **withdrew the duplication finding** and agreed a separate issue covering both banners is the more defensible call:

- `POST /admin/events` accepts `info_mode`/`info_markdown` in validators but the insert omits them — **exactly as it does for `promo_mode`/`promo_markdown`**.
- `POST /admin/events/:id/duplicate` copies neither banner's columns.
- The partial-PUT normalisation reads `updates.info_mode`, which is `undefined` when a request sends only the markdown — the identical flaw sits in the promo block directly above. Unreachable from the admin UI, which always sends both fields (`EventDetailsPage.tsx:498-499`).

Together with the pre-existing `/photos` gap already noted in this PR (per-event **promo** overrides never reach the client at all), that's a coherent follow-up: *per-event banner overrides are not honoured on create, duplicate, or partial update, for either banner*.

## Verification

Backend 13, frontend 132 (25 files), `tsc --noEmit` clean, production build clean, ESLint clean on every touched file. All 14 required checks green.

PR #1064, opened 2026-08-16
fix(gallery): make per-event banner overrides actually work, both banners (#440, #932)

the-luap:
Stacked on #1063 — the `info_*` fixes need its migration. **Review and merge #1063 first**; this PR's diff collapses to just these four fixes once that lands.

## The headline: "Off" has never worked for the promo banner

The promotional banner (#440) shipped with a per-event **inherit / custom / off** override. It has never reached a guest.

`GalleryView` reads `promo_mode` from the `/photos` payload — and `/photos` never sent it. So the value was always `undefined`, every gallery resolved to `inherit`, and **a gallery whose promo banner an admin set to "Off" kept showing the global banner anyway.** Same for "Custom": the per-event copy was ignored and the global rendered instead.

Confirmed live before the fix:

```
GET /api/gallery/:slug/photos → { event: { info_mode: "custom", promo_mode: ABSENT } }
```

The info banner (#932) mirrored the promo shape deliberately and so inherited the same gaps. Fixing only one would have left the two banners behaving differently, which is why this is a single PR covering both.

## Four places dropped the fields

| # | Where | Was |
|---|---|---|
| 1 | `GET /gallery/:slug/photos` | Never sent `promo_mode`/`promo_markdown` — the bug above |
| 2 | `POST /admin/events` | Validators accepted both banners, the insert discarded them: POST `info_mode:'off'` → 201 → row still `inherit` |
| 3 | `POST /admin/events/:id/duplicate` | Copied neither banner, so a muted gallery un-muted itself on duplication |
| 4 | `PUT /admin/events/:id` | A partial update sending only the markdown left `updates.promo_mode` undefined, parking hidden text on an `inherit`/`off` gallery that resurfaced when someone later switched it to `custom` |

On (4) the effective mode is now resolved from the **stored** row, looked up lazily — one extra query, only on that path. The two normalisation blocks became one loop over both banners so the pair can't drift apart again.

## Behaviour change — read this before merging

This changes what live guest galleries display. Any gallery an admin set to promo **Off** or **Custom** has been showing the global banner; after this it will honour the setting. That is the fix, but it is visible to guests immediately, which is why it is deliberately **not** folded into the #932 feature PR — it deserves its own review and its own revert handle.

## Verified in a browser, both directions

Against the same global promo banner, one gallery:

- `promo_mode = 'off'` → banner **not** rendered ✅
- `promo_mode = 'inherit'` → banner **rendered** ✅

Both directions matter: "not rendered" alone would also be explained by the settings simply not loading, which is exactly the false negative I hit mid-testing and had to rule out.

promo off honoured

*Gallery with a global promo banner configured and the event set to Off — no banner. Before this PR the global banner appeared here.*

## Tests

`bannerOverridePlumbing.test.js` — 24 passing, parameterised over **both** banners so neither can regress alone: partial-update resolution against a stored `inherit`/`off`/`custom` row, switching away from custom clearing the copy, requests touching neither banner left untouched, and the duplicate carrying both across.

Backend suites green; ESLint clean apart from two `no-unused-vars` in `crud.js` that are pre-existing on `main`.

### DIFF
--- a/backend/__tests__/utils/bannerOverridePlumbing.test.js
+++ b/backend/__tests__/utils/bannerOverridePlumbing.test.js
@@ -1,0 +1,173 @@
+/**
+ * Per-event banner overrides — end-to-end plumbing for BOTH banners.
+ *
+ * The promo banner (#440) shipped with per-event inherit/custom/off, but the
+ * override never actually reached a guest: GalleryView reads promo_mode from
+ * the /photos payload and /photos never sent it, so every gallery resolved to
+ * 'inherit'. Setting a gallery's promo banner to "Off" did nothing. The info
+ * banner (#932) mirrored that shape and inherited the same gaps.
+ *
+ * Four places dropped the fields. This pins all of them for both banners so
+ * the two stay in step:
+ *
+ *   1. GET  /gallery/:slug/photos   — must carry the columns
+ *   2. POST /admin/events           — validators accepted them, insert dropped
+ *   3. POST /admin/events/:id/duplicate — copy promised, not delivered
+ *   4. PUT  /admin/events/:id       — partial update parked stale markdown
+ *
+ * The route-level normalisation is exercised directly against its own rules
+ * rather than through supertest: the intent is to pin the DATA contract, which
+ * is what silently broke.
+ */
+
+const path = require('path');
+const fs = require('fs');
+const os = require('os');
+
+process.env.NODE_ENV = 'test';
+process.env.TEST_DATABASE_PATH = path.join(
+  fs.mkdtempSync(path.join(os.tmpdir(), 'picpeak-banner-plumbing-')), 'db.sqlite',
+);
+process.env.JWT_SECRET = process.env.JWT_SECRET || 'banner-plumbing-secret';
+
+const { bootCrmDb, seedMinimal } = require('../integration/helpers/crmDb');
+
+let db;
+let cleanup;
+
+const baseEvent = (slug, extra = {}) => ({
+  slug,
+  event_type: 'wedding',
+  event_name: slug,
+  event_date: '2026-06-22',
+  host_email: 'host@example.com',
+  admin_email: 'admin@example.com',
+  password_hash: 'x',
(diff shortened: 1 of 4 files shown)


the-luap:
## External review (Codex, 2 rounds) — converged

Reviewed against the stack parent (#1063) so only this PR's own delta was in scope. Round 1 found two real gaps in the plumbing this PR introduces; round 2 came back clean.

**1. Reveal-hidden view ignored the promo override** — `GalleryView.tsx`

#1063 taught that branch to resolve the *info* banner from `/photos`. Once this PR made `/photos` carry promo too, the branch was still copying only the info fields — so a reveal-hidden gallery with `promo_mode: 'off'` fell back to `inherit` and showed the global banner on the first load after login. Both banners are now threaded there.

A neat illustration of why this PR exists: the promo override was broken in *one more place* than the original audit found, and it only became visible once the data started flowing.

**2. Duplicate carried stale markdown** — `adminEvents/crud.js`

The copy took `promo_markdown`/`info_markdown` verbatim. Rows written *before* this PR's PUT normalisation can legitimately hold text while their mode is `inherit`/`off` — that's exactly the state the normalisation fix exists to stop creating. Copying it verbatim smuggled hidden text into the duplicate, where it would resurface the moment someone switched the copy to `custom`, violating the invariant this PR establishes.

Markdown now rides along only when the source mode is `custom`, with a test pinning a stale-markdown source explicitly.

## Verification

24 → 12 focused tests in `bannerOverridePlumbing.test.js` (parameterised over both banners), frontend 132 across 25 files, `tsc --noEmit` clean, ESLint clean apart from two `no-unused-vars` in `crud.js` that are pre-existing on `main`. All 14 required checks green.

**Verification depth, stated honestly:** the promo `off`/`inherit` behaviour was confirmed in a real browser earlier in this PR (both directions, against the same global banner). The reveal-hidden threading added in this round is verified by type-check, tests and code symmetry with the already-browser-verified main path — not by a fresh browser run, since reaching that state needs a gallery in reveal-hidden mode.

PR #1067, opened 2026-08-16
feat(docker): official single-container (all-in-one) image (#1042)

### LINKED ISSUE #1042: [FEATURE] Official single-container (all-in-one) image — one Node process, SQLite default

the-luap:
Closes #1042. Second of the two PRs planned in #705 — #1041/#1043 shipped the upgrade path *out* of a small install first, so this one is never a dead end.

```bash
docker run -d --name picpeak -p 3000:3000 -v picpeak:/data \
  -e JWT_SECRET="$(openssl rand -base64 48)" \
  ghcr.io/picpeak/picpeak/aio:stable
```

One Node process. No supervisor, no nginx, no Postgres, no Redis. SQLite is the explicit, documented default; `DB_*` still points at an external Postgres and the engine is resolved and logged at boot (#1038). Everything lives under one mount: `/data/db` (database + `SETUP_TOKEN`), `/data/storage`, `/data/logs`.

PicPeak admin from a single container

## Four backend fixes this needed — all of which also fix the existing `SERVE_FRONTEND` path

These weren't anticipated in the issue. Each was found by running the image, not by reading code.

**1. `wait-for-db.sh` hung forever on SQLite.** It waited for PostgreSQL *unconditionally* and resolved the engine only afterwards, so a SQLite container blocked on a host that will never answer. Resolution now runs first and the wait is skipped for `sqlite3`. Safe to reorder: `resolveBootEngine` reads the filesystem and env, never a live connection.

**2. The writable-root checks hard-coded `/app/storage`, `/app/data`, `/app/logs`** — so a single `/data` volume couldn't be guarded at all. They now follow the same `STORAGE_PATH` / `DATA_DIR` / `LOG_DIR` the app itself reads.

**3. `express.static` sent no `Cache-Control`.** nginx marks hashed assets `immutable` and `index.html` `no-store`. Without the latter, a stale `index.html` after an upgrade names chunks that no longer exist and the app won't boot until a hard refresh.

**4. The SPA history fallback only listed `/admin/*` and `/gallery/*`.** nginx does `try_files $uri $uri/ /index.html`, so behind compose every client-side route survived a reload — masking this completely. Direct hits all 404'd:

```
/setup 404   /impressum 404   /quote/x 404      /transfer/x 404
/customer 404   /datenschutz 404   /invite/x 404   /contract/x 404
/payment-check 404   /transfer-upload/x 404   /some-short-url 404
```

**`/setup` is the first URL a new install visits.** All 200 now. The catch-all is registered *after* the API 404 handler, so `/api/*` still answers JSON — verified.

## Header parity — checked against a running container, not assumed

CSP, `X-Frame-Options`, `X-Content-Type-Options`, `Referrer-Policy` and `Permissions-Policy` all match `frontend/nginx.conf` exactly. helmet already emitted them — nginx even `proxy_hide_header`s its own copies to avoid duplicates — so dropping nginx costs no header coverage. Cache tiers confirmed live: assets `public, max-age=31536000, immutable`; `index.html` `no-cache, no-store, must-revalidate`.

## Verified by actually running it

- **SQLite (default):** healthy in ~6s from a cold `docker run`; engine logged as `sqlite (/data/db/picpeak.db)`; PostgreSQL wait skipped; migrations applied.
- **Full first-run flow in a browser:** setup token read from `/data/db/SETUP_TOKEN`, all three wizard steps completed, admin dashboard live (screenshot above), System Health "Healthy".
- **Persistence:** wrote a marker, destroyed the container, recreated it — marker intact, admin account preserved (`needsAdmin: false`), no fresh-install re-detection.
- **External Postgres:** healthy in ~6s against a real Postgres 15 with `DB_HOST`/`DB_*`.

That last one caught a bug in my own Dockerfile: I'd dropped `postgresql-client` reasoning the readiness probe would "degrade gracefully". It doesn't — the wait loop *is* `psql`, so external-Postgres deployments failed outright. Restored, retested.

## CI

Publishes `ghcr.io/…/aio` using the same per-arch build → manifest-merge shape as the other two images, so it inherits the same Trivy-by-digest scanning rather than a weaker one-off publish. `summary` gates on it.

**Honest limitation:** the publish workflow only fires on release tags, so it cannot run on this PR. The YAML is validated and the job graph checked (`build-aio` → `merge-aio` → `summary`), but the first real execution will be the next tagged release.

## Docs

`docs/single-container.md` — the `docker run` one-liner, volume layout, minimal env reference, external-Postgres variant, TLS guidance, Synology/QNAP notes, the `.picpeak` upgrade path to the full stack, and the limits (one SQLite writer, no bundled TLS, no horizontal scaling). Linked from the README Quick Start.

## Scope

Out of scope per the issue and untouched: removing Redis from the compose stacks, NAS-native SPK/QPKG packages, bundled reverse proxy / TLS.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -18,3 +18,18 @@ storage/events/archived/*
 storage/thumbnails/*
 data/*.db
 logs/*
+
+# The all-in-one image (#1042) builds from the REPOSITORY ROOT, not ./backend,
+# and Docker only reads this file — backend/.dockerignore is never consulted.
+# The unprefixed rules above therefore miss backend/data, backend/storage and
+# backend/logs, so a checkout that has been used to run PicPeak would bake its
+# database, photos and logs into a published image layer.
+backend/node_modules
+backend/data/*.db
+backend/data/*.db-wal
+backend/data/*.db-shm
+backend/data/SETUP_TOKEN
+backend/logs
+backend/storage
+frontend/node_modules
+frontend/dist
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -345,6 +345,289 @@ jobs:
   # -----------------------------------------------------------------------------
   # Frontend: per-arch build, then merge into a multi-arch manifest
   # -----------------------------------------------------------------------------
+  # -----------------------------------------------------------------------------
+  # All-in-one (#1042): backend + built frontend in a single image, SQLite default.
+  # Same per-arch build -> manifest merge shape as the other two, so it inherits
+  # the same Trivy-by-digest scanning rather than a weaker one-off publish.
+  # -----------------------------------------------------------------------------
+  build-aio:
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - platform: linux/amd64
+            runner: ubuntu-latest
+          - platform: linux/arm64
+            runner: ubuntu-24.04-arm
+    runs-on: ${{ matrix.runner }}
+    permissions:
+      contents: read
+      packages: write
+      # Trivy uploads its SARIF to the Security tab from this job — see
+      # the "Run Trivy" step below. Scanning per-arch by digest (#476)
+      # is reliable; scanning the multi-arch index by tag from the
+      # merge-* job was not.
+      security-events: write
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v4
+
+    - name: Compute image names (lowercase for GHCR)
+      run: |
+        repo_lc="${GITHUB_REPOSITORY,,}"
+        echo "AIO_IMAGE_NAME=${repo_lc}/aio" >> "$GITHUB_ENV"
+        echo "FRONTEND_IMAGE_NAME=${repo_lc}/frontend" >> "$GITHUB_ENV"
+        # Mirror manifests to Docker Hub (picpeak/{backend,frontend}) only on the
+        # canonical org repo, where the DOCKERHUB_* secrets live. Forks (and any
+        # other owner) fall back to GHCR-only — the Docker Hub image line and login
+        # are gated on this flag so their builds keep working unchanged.
+        if [[ "$GITHUB_REPOSITORY" == "PicPeak/picpeak" ]]; then
+          echo "DOCKERHUB_ENABLED=true" >> "$GITHUB_ENV"
+        else
+          echo "DOCKERHUB_ENABLED=false" >> "$GITHUB_ENV"
+        fi
+
+    - name: Prepare platform pair
+      run: |
+        platform="${{ matrix.platform }}"
+        echo "PLATFORM_PAIR=${platform//\//-}" >> "$GITHUB_ENV"
+
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v3
+
+    - name: Log in to Container Registry
+      if: github.event_name != 'pull_request' || github.event.inputs.push == 'true'
+      id: login-ghcr
+      continue-on-error: true
+      uses: docker/login-action@v3
+      with:
+        registry: ${{ env.REGISTRY }}
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+
+    - name: Determine if pushing
+      id: push-decision
+      run: |
+        if [[ "${{ github.event_name }}" == "pull_request" && "${{ github.event.inputs.push }}" != "true" ]]; then
+          echo "push=false" >> "$GITHUB_OUTPUT"
+        elif [[ "${{ steps.login-ghcr.outcome }}" != "success" ]]; then
+          echo "push=false" >> "$GITHUB_OUTPUT"
+        else
+          echo "push=true" >> "$GITHUB_OUTPUT"
+        fi
+
+    - name: Extract metadata for Backend (labels only)
+      id: meta-backend
+      uses: docker/metadata-action@v5
+      with:
+        images: ${{ env.REGISTRY }}/${{ env.AIO_IMAGE_NAME }}
+        labels: |
+          org.opencontainers.image.title=PicPeak All-in-One
+          org.opencontainers.image.description=PicPeak photo sharing platform backend service
+          org.opencontainers.image.vendor=PicPeak
+          maintainer=${{ github.repository_owner }}
+
+    - name: Build Backend image (push by digest)
+      id: build
+      uses: docker/build-push-action@v5
+      with:
+        context: .
+        file: ./Dockerfile.aio
+        platforms: ${{ matrix.platform }}
+        labels: ${{ steps.meta-backend.outputs.labels }}
+        cache-from: type=gha,scope=backend-${{ env.PLATFORM_PAIR }}
+        # ignore-error: a flaky GitHub Actions cache write ("error writing
+        # layer blob: not_found") must not fail an otherwise-successful build
+        # that already pushed the image.
+        cache-to: type=gha,mode=max,scope=backend-${{ env.PLATFORM_PAIR }},ignore-error=true
+        outputs: ${{ steps.push-decision.outputs.push == 'true' && format('type=image,name={0}/{1},push-by-digest=true,name-canonical=true,push=true', env.REGISTRY, env.AIO_IMAGE_NAME) || 'type=cacheonly' }}
+        build-args: |
+          CACHEBUST=${{ github.run_number }}
+          BUILD_DATE=${{ github.event.head_commit.timestamp }}
+          VCS_REF=${{ github.sha }}
+          VERSION=${{ steps.meta-backend.outputs.version }}
+
+    - name: Export digest
+      if: steps.push-decision.outputs.push == 'true'
+      run: |
+        mkdir -p /tmp/digests
+        digest="${{ steps.build.outputs.digest }}"
+        touch "/tmp/digests/${digest#sha256:}"
+
+    - name: Upload digest artifact
+      if: steps.push-decision.outputs.push == 'true'
+      uses: actions/upload-artifact@v4
+      with:
+        name: digests-aio-${{ env.PLATFORM_PAIR }}
+        path: /tmp/digests/*
+        if-no-files-found: error
+        retention-days: 1
+
+    # Per-arch vulnerability scan (#476). Scanning the multi-arch
+    # manifest from the merge-* job by tag is unreliable — Trivy's
+    # remote resolver crashes intermittently with "no child with
+    # platform linux/amd64 in index". The fix is to scan each leg
+    # by its single-platform digest right here, where it just landed
+    # in GHCR. Tag pinned (was @master) so the action + bundled
+    # Trivy binary don't float between runs.
+    #
+    # exit-code is left unset (=0) for now: Trivy reports findings
+    # to the Security tab but doesn't fail the build. Flipping that
+    # to '1' to actually gate CI is a deliberate follow-up — needs an
+    # audit pass first so the next beta build doesn't surprise red.
+    - name: Run Trivy vulnerability scanner (per-arch, by digest)
+      if: steps.push-decision.outputs.push == 'true'
+      uses: aquasecurity/trivy-action@v0.36.0
+      env:
+        # docker/build-push-action wraps every push in an OCI index
+        # (carries the SLSA provenance attestation alongside the
+        # actual image). Trivy's remote backend defaults to
+        # linux/amd64 regardless of host arch when resolving an
+        # index, which makes the arm64 leg crash with "no child
+        # with platform linux/amd64". Telling Trivy which child to
+        # scan keeps the provenance attestation intact and fixes
+        # the resolver crash. Pin to matrix.platform so each leg
+        # scans its own arch.
+        TRIVY_PLATFORM: ${{ matrix.platform }}
+      with:
+        image-ref: ${{ env.REGISTRY }}/${{ env.AIO_IMAGE_NAME }}@${{ steps.build.outputs.digest }}
+        format: 'sarif'
+        output: 'trivy-aio-${{ env.PLATFORM_PAIR }}.sarif'
+        severity: 'CRITICAL,HIGH'
+        timeout: '10m'
+
+    - name: Upload Trivy scan results to GitHub Security tab
+      if: steps.push-decision.outputs.push == 'true'
+      uses: github/codeql-action/upload-sarif@v4
+      with:
+        sarif_file: 'trivy-aio-${{ env.PLATFORM_PAIR }}.sarif'
+        # Distinct category per arch so the Security tab surfaces
+        # per-platform findings independently — an amd64-only CVE in
+        # a base layer doesn't get masked by the arm64 scan.
+        category: 'backend-vulnerabilities-${{ env.PLATFORM_PAIR }}'
+
+  merge-aio:
+    needs: build-aio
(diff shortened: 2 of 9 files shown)


the-luap:
## Deeper testing found two bugs my first pass missed

Asked whether I'd tested the *full* image, I hadn't — I'd verified boot, serve and persistence, but not the photo workload, not a bind mount, and not a clean build. Closing those gaps turned up two real failures.

### 1. Bind mounts failed outright

The image creates `/data/db`, `/data/storage`, `/data/logs` at build time. A **named volume inherits that tree**; a **bind-mounted host directory hides it**. So `chown` ran against paths that didn't exist and reported:

```
ERROR: failed to chown /data/storage /data/db /data/logs to nodejs (UID 1001).
  This usually means the host filesystem rejects chown …
```

Wrong diagnosis, and a dead end for exactly the NAS users this image targets — the docs tell them to point the volume at a folder on their data pool. Fixed by `mkdir -p` before `chown`. The compose stack was never affected: each of its three paths is its own mount point.

### 2. `logger.js` ignored `LOG_DIR`

It hard-coded `<backend>/logs`, so the variable the image sets was **inert** and the process died with `EACCES: mkdir '/app/logs'` — a directory `backend/Dockerfile` creates but this image doesn't. It now honours `LOG_DIR`, defaulting to the old path, so compose and native installs are unchanged and the image's logs land on the volume as documented.

### Why the first pass missed both

The earlier image was built from a worktree where I'd run `wait-for-db.sh` directly during testing. That left a stray `backend/logs`, which `COPY` baked into the image and which masked bug 2 entirely. A clean checkout — CI, and every user — has no such directory. **My "verified" runs were passing on an artifact of my own testing.** Re-verified from a fresh worktree.

## Now covered

| | |
|---|---|
| Named volume | healthy in 6s |
| **Bind mount** | healthy in 1s, host dir populated, logs written host-side |
| **sharp / libvips 8.18.3** | real 1200×800 → 300×200 resize |
| **ffmpeg / ffprobe** | 8.1.2, both present |
| **pdftoppm / exiftool** | present and responding |
| **Fonts** | 86 registered; SVG with live `<text>` rasterises (the tofu-box risk) |
| Setup wizard | completed again on the clean image |
| External Postgres | healthy against Postgres 15 |

## Still not covered — stated plainly

- **`.picpeak` export/restore**, which I document as the upgrade path. Untested from this image.
- **The amd64 image has never been *run*.** CI builds it `type=cacheonly`; my local builds are arm64.
- **The GHCR publish job**, which only fires on release tags.
- **Email/SMTP.**

the-luap:
## Remaining gaps closed

Both of the things I'd flagged as untested are now tested. No new code changes — the image as it stands passes both.

### amd64 image actually run

Previously only ever *built* (`type=cacheonly` in CI, arm64 locally). Built for `linux/amd64` and ran it under emulation:

```
container arch : x86_64
health         : HEALTHY after 11s
sharp          : 1000x700 -> 250x175 jpeg, libvips 8.18.3
/health /api/public/settings /setup /admin/login  -> 200 200 200 200
engine         : sqlite (/data/db/picpeak.db), PostgreSQL wait skipped
```

The native-module risk (sharp/libvips against musl on a second architecture) is the reason this mattered, and it's clear.

### `.picpeak` upgrade path — the documented promise

`docs/single-container.md` tells people a single-container install is never a dead end: export, stand up the compose stack, restore. That claim was untested. Full round trip now run:

1. **Source:** AIO container on SQLite, seeded with 1 event, 4 photos (real sharp-generated JPEGs + thumbnails), an admin, and a branding setting.
2. **Export:** `createPicpeak({ includePhotos: true })` → 90.5 KB archive, manifest engine `sqlite`, 95 tables.
3. **Target:** second AIO container pointed at a real PostgreSQL 15 — healthy in 6s, engine logged as `postgres`.
4. **Import:** `restored: true, tables: 94, filesRestored: 4, crossEngine: true`.

Verified in Postgres afterwards, values not just row counts:

```
events: 1     photos: 4     admins: 1
name=Upgrade Demo Gallery | active=true | downloads=true | expires=2026-11-14
branding_company_name = "Upgrade Test Studio"
```

Photo files present on the target volume (`shot-0..3.jpg`), and the migrated gallery serves from the new instance: `/gallery/upgrade-demo` → 200, `event_name: 'Upgrade Demo Gallery'`.

Note the booleans came across as real Postgres `true` and the timestamp as a proper date — that's the #1039 coercion layer doing its job on a SQLite-sourced archive, exercised here through the actual image rather than a unit test.

## Now genuinely untested

Only two things left, both structural rather than skippable:

- **The GHCR publish job** — fires on release tags only, so its first real run is the next tagged release.
- **Email/SMTP** — no mail server involved in any of this.

the-luap:
## Final round: cold build + SMTP

Two more things covered, no code changes needed.

### Truly cold build

An unrelated Docker cleanup removed every cached base image, which made a genuinely CI-equivalent build possible — nothing warm, `--no-cache`, `node:22-alpine` pulled from scratch:

```
build: 32s
run  : HEALTHY after 6s
engine: sqlite (/data/db/picpeak.db), PostgreSQL wait skipped
```

This is the check that would have caught the earlier contamination — the first image only worked because a stray `backend/logs` from my own local testing had been baked in. From zero cache, on a clean checkout, it builds and boots.

### SMTP

The last functional gap. Ran a MailHog sidecar on a user network and sent through the image's own `nodemailer`:

```
SMTP verified -> aio-mail:1025
accepted: <fb75980e-…@example.com>
mailbox total: 1
from: picpeak@example.com | to: client@example.com | subject: PicPeak AIO smoke test
```

Connection verify and delivery both succeed from inside the container.

## Coverage now

| Area | Status |
|---|---|
| Cold build, no cache, clean checkout | ✅ |
| SQLite default (arm64 + amd64) | ✅ |
| External PostgreSQL | ✅ |
| Named volume + bind mount | ✅ |
| Setup wizard → live admin dashboard | ✅ |
| Persistence across container replacement | ✅ |
| sharp/libvips, ffmpeg, pdftoppm, exiftool, fonts | ✅ |
| Header + cache parity vs nginx | ✅ |
| All 13 top-level routes; API 404s stay JSON | ✅ |
| `.picpeak` sqlite → Postgres upgrade path | ✅ |
| SMTP | ✅ |
| **GHCR publish job** | ❌ tags only — first run is the next release |

That last one is structural: the workflow triggers on `v*.*.*` tags, so it cannot execute on a PR. Everything else about it is verified — YAML valid, job graph `build-aio` → `merge-aio` → `summary`, and the build itself ran green on both architectures in this PR's checks.

the-luap:
## External review (Codex, 3 rounds) — 15 findings, 14 fixed, 1 retracted

Stopped at the skill's 3-round limit. Two findings were regressions an *earlier round of this same review* introduced, which is worth saying out loud.

### Round 1 — 7 findings, all valid

| Finding | Reality |
|---|---|
| Built-in backups wrote nowhere | Migrations 029/030 seed `/backup/{picpeak,database}`; the image neither created nor mounted `/backup`. `/backup` is now a symlink into `/data/backup`. |
| `sqlite3` CLI missing | `DatabaseBackupService` **spawns** it for `.backup` and `PRAGMA integrity_check`; the npm module doesn't provide the binary. The compose image omits it because it always runs Postgres — this one defaults to SQLite, so backups died with ENOENT. |
| Every install served literal `${BRAND_TITLE}` | `index.html` carries placeholders that `frontend/docker-entrypoint.sh` renders; this image runs no nginx and never invoked it. **My browser testing missed this** because the SPA rewrites `document.title` at runtime — view-source and every og:/twitter: card showed the raw token. |
| SPA fallback swallowed backend 404s | nginx gives `/api`, `/photos`, `/thumbnails`, `/health` their own `location` blocks, so `try_files` never applies to them. Excluding only `/api/` made my fallback strictly broader than the behaviour I claimed parity with — a missing photo became a 200 HTML body under an image URL. |
| Docs wrong on engine switching | `DB_*` alone does **not** switch engines; the image declares `sqlite3` and the resolver treats a declared client as explicit. `DATABASE_CLIENT=pg` is required. |
| Trivy SARIF category collided | AIO reused `backend-vulnerabilities-*`, so the two scans replaced each other. |
| buildx cache scope collided | AIO reused `scope=backend-*`, two concurrent jobs writing one cache object. |

### Round 2 — 5 findings, 4 valid, 1 retracted

The important one: **my engine-resolution reordering broke a real upgrade path.** `probePgData` deliberately reports "occupied" when Postgres is unreachable, and the target database may not exist until the wait loop creates it — so an install declaring no `DATABASE_CLIENT` with a populated legacy SQLite file would resolve to `ambiguous-both-populated` and refuse to boot. The resolver is back in its original position; the wait is skipped only when `sqlite3` is *explicitly declared*. Installs that declare nothing keep today's ordering byte for byte.

Also fixed: backup subdirectories missing on bind mounts; `chown -R` acting on the `/backup` symlink rather than its target; branding moved from a build-time `sed` to a startup entrypoint so `BRAND_TITLE`/`BRAND_DESCRIPTION` actually work; `/fonts/` added to the fallback exclusions.

**Retracted:** the `workflow_dispatch` push opt-out. The `push-decision` block is byte-identical in `build-backend` (129), `build-aio` (412) and `build-frontend` (688) — pre-existing behaviour, not something this job introduces. Shown that evidence, Codex agreed: *"matching the established manual-dispatch behavior is correct for this PR."*

### Round 3 — 5 findings, 4 fixed, 1 deferred

**Two were regressions round 2 introduced:**

- Adding the backup root to the *mandatory* preflight broke non-root deployments — `docker-compose.production.yml` doesn't mount `/backup`, so a hardened install would fail `mkdir -p /backup` against a root-owned `/` and refuse to boot over a directory it never needed. Now adopted only when `BACKUP_DIR` is explicitly set.
- **Runtime data could leak into published images.** `Dockerfile.aio` builds from the repo root, and Docker reads only the root `.dockerignore` — `backend/.dockerignore` is never consulted — so the unprefixed `data/*.db`, `logs/*`, `storage/*` rules missed `backend/`-prefixed paths entirely. A checkout used to run PicPeak would bake its database, photos, logs and `SETUP_TOKEN` into a layer. Not hypothetical: an earlier build in this PR silently picked up a stray `backend/logs` exactly this way.

Plus: `HEALTHCHECK` now follows `$PORT` (hard-coded 3000 marked a working container unhealthy forever if anyone changed it), and `--max-http-header-size=32768` matches nginx's `large_client_header_buffers 4 32k` (Node's 16 KiB default would reject a guest carrying several per-gallery JWT cookies).

**Deferred:** response compression. nginx gzips the JS/CSS bundles and Express doesn't, which matters on a remote NAS. `compression` is not currently a backend dependency, and adding one in the last round of a review loop — affecting the compose image too — is the wrong moment. Worth its own PR.

## Verification

Every fix confirmed on a rebuilt image, not just by reading the diff:

- clean image carries no `/app/data/*.db`, `/app/logs` or `/app/storage`
- named volume **and** bind mount both healthy
- `<title>PicPeak</title>` by default; `BRAND_TITLE="Studio Nord Galerie"` renders in `<title>` and `og:title`
- `/backup/{picpeak,database}` present and writable host-side
- `PORT=8080` container healthy with the probe following it
- pid 1 shows `--max-http-header-size=32768`
- `/photos`, `/thumbnails`, `/api` back to 404; `/setup`, `/impressum`, `/gallery/x`, `/admin/login` still 200
- `sqlite3 3.53.2` present
- backend suites green; all 16 required checks pass

the-luap:
Closing in favour of #1068 — everything here has been folded into that branch, so there's one PR to review instead of two.

Luca's branch was the better base. Where the two designs disagreed his approach won, and one of those differences was a real bug here: `express.static`'s default `index` option shadows `handlePublicSiteRequest`, so this PR's `/` returned the SPA shell (200) instead of `302 -> /admin/login`. His `index: false` fixes it, and his `smoke-aio` job catches it — my testing was manual and didn't.

He also had the in-process brand render (simpler than the entrypoint script here), `compression`, correct `aio-` SARIF/cache scoping from the first commit, and a 12-line `wait-for-db.sh` change that avoided the legacy-upgrade regression this PR had to be corrected out of.

Carried across to #1068: the single `/data` volume layout, the `sqlite` CLI the backup service shells out to, `/backup` wiring, `LOG_DIR` support, the `.dockerignore` runtime-data exclusions, the `PORT`-aware healthcheck, the 32k header allowance, and `docs/single-container.md` — plus four new smoke assertions covering them.

No review comments here are lost: the Codex findings that produced those items are all represented in #1068.

PR #1075, opened 2026-08-18
feat(faces): People in this gallery — face recognition via an optional ML sidecar (#1074)

### LINKED ISSUE #1074: Face recognition per gallery ("People in this gallery") + rule-based auto-categories, via an optional ML sidecar
Face recognition ("People in this gallery") + rule-based auto-categories, via an optional ML sidecar container.

Comes out of discussion #1069 (@bluecow94). Scope agreed there:

- **Per gallery, not cross-gallery.** People are clustered inside one event. No global person identity across events.
- **Filter by face, exactly like filtering by category.** See the faces in a gallery → click one → see the photos they're in → download just those.
- **Runs in a separate, optional container.** Nothing about the base deployment changes if you don't opt in.
- **Categories on top:** rule-driven auto-classification (`#people == 1` → Portrait, `> 5` → Group, …), because face detection already gives us the inputs for free.

---

## 1. Which face engine

@bluecow94 suggested [Exadel CompreFace](https://github.com/exadel-inc/CompreFace). It is the well-known answer, but it does not fit this deployment:

- Its reference `docker-compose.yml` is **5 containers** — `compreface-admin` (Java), `compreface-api` (Java), `compreface-core` (Python), `compreface-fe` (nginx), plus **its own PostgreSQL**. We would be shipping a second database next to ours.
- It owns state we already own: CompreFace stores *its* face collections and subjects, so person identity would live in two places and drift.
- Its recognition models are the InsightFace family anyway — so we inherit the license question below without gaining anything.
- We would use maybe 5% of it (detect + embed). The rest is a UI and a multi-tenant API-key system we don't need.

**Proposal: a first-party single-container sidecar, `picpeak-ml`** — FastAPI + `onnxruntime` (CPU), two small ONNX models baked into the image, three endpoints. That is the whole thing; the interesting code stays in the backend where the data already lives.

**Default pair: [YuNet](https://github.com/opencv/opencv_zoo/tree/main/models/face_detection_yunet) (detection, MIT, ~340 KB) + [FaceNet-512](https://github.com/serengil/deepface) (embedding, MIT, 512-d, ~90 MB ONNX).**

Both criteria — accuracy and license — are covered below, because for this feature they pull in different directions and the naive answer gets one of them wrong.

#### Accuracy

Headline "LFW 99.x%" numbers are useless for choosing here: every model quotes one, on its own pipeline, on a benchmark that saturated years ago. The only comparison worth anything is a matched one — same dataset, same alignment, same distance metric, models swapped. DeepFace publishes exactly that ([benchmarks](https://github.com/serengil/deepface/tree/master/benchmarks), LFW, aligned, cosine):

| Embedding model | with RetinaFace | with YuNet | Distributor's license |
|---|---|---|---|
| **FaceNet-512** | **98.4%** | **97.9%** | MIT |
| ArcFace | 96.6% | 96.7% | MIT code / see below |
| SFace | 92.4% | 91.0% | Apache-2.0 |

SFace advertises 0.9940 on the OpenCV Zoo's own protocol and lands at **91–92%** in a matched test — roughly 7 points behind FaceNet-512. For verification that's a bad day; for *clustering* it's fatal, because errors compound: every false split invents a duplicate person in the strip and every false merge puts a stranger in someone's "download my photos".

The second useful column is that YuNet costs only ~0.5 points against RetinaFace while being ~340 KB and orders of magnitude faster. Detection is where faces are lost entirely, so this was the finding worth having.

#### License

Essentially **every** open face-recognition weight set traces back to a scraped, research-only dataset — CASIA-WebFace, VGGFace2, MS1M, Glint360K, WebFace260M all carry academic-use-only agreements. So "trained on clean data" is not an option that exists. What differs, and what actually binds us, is **what the distributor grants on the artifact we would redistribute inside our image**:

| | Distributor's grant |
|---|---|
| YuNet, SFace (OpenCV Zoo) | MIT / Apache-2.0 — redistributable |
| FaceNet-512 (deepface, facenet-pytorch) | MIT — redistributable |
| Intel OMZ `face-reidentification-retail-0095` | Apache-2.0, and Intel trained it themselves — cleanest provenance of the lot |
| InsightFace `buffalo_*`, `antelopev2` | **Non-commercial research only**; commercial use needs a license from insightface.ai ([model zoo README](https://github.com/deepinsight/insightface/blob/master/model_zoo/README.md), [#2587](https://github.com/deepinsight/insightface/issues/2587)) |
| Idiap EdgeFace | **CC BY-NC-SA 4.0** — explicitly non-commercial |
| AdaFace WebFace4M weights | MIT code, non-commercial weights |

PicPeak's users are working photographers running this commercially. So the rule is simple: **we never bake a non-commercial artifact into a PicPeak image.** `buffalo_l` stays available as `FACE_MODEL=buffalo_l`, downloaded at runtime by the operator, behind an explicit acknowledgement of the license — never shipped by us.

That rule is what costs us, at most, half a point of accuracy. It would have cost seven if we'd read the headline numbers instead of the matched ones.

Every `photo_faces` row stores `model_version`; switching models invalidates and re-derives clusters rather than silently mixing embedding spaces.

#### What comparable projects do

- **PhotoPrism** — FaceNet (512-d, TensorFlow) + Pigo or ONNX SCRFD detection, DBSCAN clustering. A commercially sold product that landed on the same FaceNet route.
- **Immich** — InsightFace `buffalo_l`. Also the cautionary tale: even with the strongest model they ship user-facing tuning knobs (`max recognition distance`, `min recognized faces`) and a whole [guide](https://docs.immich.app/guides/better-facial-clusters/) on getting usable clusters. Clustering quality is a tuning problem, not only a model problem — noted for §2.
- **Ente** — YOLOv5Face + MobileFaceNet, on-device, incremental clustering.
- **Nextcloud Recognize** — deliberately restricted to Apache-licensed models.

#### Rejected

- **`serengil/deepface` container as-is** — MIT, one container, `/represent` gives us embeddings with zero ML code on our side. Genuinely tempting for a first spike, and worth using as the harness to validate the model choice. Not for shipping: TensorFlow-based (multi-GB image), downloads weights at first run (breaks airgapped installs, needs a volume), and a general-purpose API surface we'd be pinning our contract to.
- **`onnxruntime-node` in the backend process** — fattens the base image and competes with Sharp for the same constrained RAM (see the auto-concurrency logic at `backend/src/services/backgroundProcessor.js:44`).
- **`ghcr.io/immich-app/immich-machine-learning`** — Immich-specific contract, non-commercial default model.
- **EdgeFace** — the accuracy/size sweet spot (1.77M params, IJB-C 94.85%), but CC BY-NC-SA rules it out.
- **`face_recognition`/dlib** — CPU-slow and weak on non-frontal faces, which is most of an event gallery.

Resource shape: image ≈ 400 MB, RSS ≈ 300–500 MB idle-to-busy, ~80–150 ms per photo on one CPU core at 1600 px. For comparison Immich asks for 6 GB RAM minimum largely because of its ML container; we should land far below that, which matters for the 2 GB VPS crowd we already accommodate in `backgroundProcessor.js`.

### Sidecar contract

```
GET  /health            → {status}
GET  /info              → {detector, embedder, model_version, dim}
POST /faces             ← image bytes (multipart)
                        → {faces: [{bbox:[x,y,w,h], score, landmarks:[[x,y]×5],
                                    yaw, pitch, blur, embedding:[512 floats, L2-normalized]}]}
```

Stateless. No database, no volume, no model download at runtime (weights baked into the image → airgapped installs work). Reachable only on `picpeak-network`, and gated by a shared `FACE_ML_TOKEN` header so an accidentally published port isn't a free face-detection API.

```yaml
# docker-compose.yml — new, opt-in
  ml:
    build: ./ml            # or ghcr.io/picpeak/picpeak-ml:<tag>
    container_name: picpeak-ml
    restart: unless-stopped
    profiles: [ml]         # `docker compose --profile ml up -d` to enable
    environment:
      - FACE_MODEL=${FACE_MODEL:-facenet512}   # buffalo_l = opt-in, runtime download, NC license
      - FACE_ML_TOKEN=${FACE_ML_TOKEN:-}
    networks: [picpeak-network]
```

Backend reads `FACE_ML_URL`. Unset → the feature is invisible everywhere: no admin toggle, no gallery UI, no queue rows. Same posture as the `dev` profile mailhog service.

---

### Schema (one migration, `hasTable`/`hasColumn` guarded per house style)

```
photo_faces
  id, photo_id → photos(id) ON DELETE CASCADE, event_id (denormalized, indexed)
  bbox_x, bbox_y, bbox_w, bbox_h, det_score, yaw, pitch, blur
  embedding      BLOB/bytea   -- 512 × float32 = 2 KB, no pgvector needed
  model_version  string
  person_id      → event_people(id) ON DELETE SET NULL
  created_at

event_people
  id, event_id → events(id) ON DELETE CASCADE
  label            (null until named)
  cover_face_id    → photo_faces(id)
  centroid         BLOB/bytea
  face_count, model_version
  is_hidden        (photographer-only; guests never see it)
  is_ignored       (bystanders/false positives — excluded from the strip)
  created_at, updated_at

photos.face_status   'pending' | 'processing' | 'done' | 'failed' | 'skipped'
photos.face_count    int, denormalized for the category rules
events.face_recognition_enabled   boolean, default false
events.faces_visible_to_guests    boolean, default true (when the above is on)
```

Embeddings are the sensitive part: 2 KB/face keeps them in a plain BLOB, which means **no PostgreSQL extension requirement** and identical behaviour on the SQLite path. A 2,000-photo wedding at ~3 faces/photo is ~12 MB of embeddings — nothing.

### Queue

Mirror `backgroundProcessor.js` exactly rather than inventing a second pattern: a worker loop that claims `photos.face_status = 'pending'` with `FOR UPDATE SKIP LOCKED` on PG / status-guarded UPDATE on SQLite, plus the same stuck-row janitor. Tunables `FACE_PROCESSOR_CONCURRENCY` (default 1), `FACE_PROCESSOR_POLL_MS`, `FACE_PROCESSOR_DISABLED`.

Details that matter:
- **Detect on the preview rendition, not the original.** The `/preview` tier (#492) already exists and is generated lazily; a 1600 px JPEG is ~40× cheaper to decode than a 45 MP original and costs nothing in recall at typical event-photo face sizes.
- Sidecar unreachable → row goes back to `pending` with backoff, never `failed`. Turning the container off for a week must not require a manual re-scan.
- Videos are `skipped` in phase 1.
- Enabling the toggle on an event with existing photos enqueues a backfill; disabling it stops the workers and offers "delete all face data for this gallery".

### Clustering

Plain JS in the backend — a gallery is hundreds to low-thousands of faces, this does not need a vector database.

- Greedy incremental assignment: cosine-similarity a new face against existing person centroids, join above threshold (start from FaceNet-512's published cosine threshold, then tune on real galleries), else open a new person. Update the centroid as a running mean.
- **Expose the two knobs Immich learned it needed** — match distance and "minimum faces before a person appears" — as admin settings with sane defaults, not hardcoded constants. Their [clustering guide](https://docs.immich.app/guides/better-facial-clusters/) exists because no single threshold survives contact with every library, and they run a stronger model than we will.
- Faces below a quality floor (`det_score`, blur, bbox < ~40 px) are stored but left unassigned — they still show in "this photo contains", they just don't spawn junk people.
- A periodic per-event consolidation pass merges centroid pairs that drifted together, so upload order doesn't determine the outcome.
- `POST /api/admin/events/:id/faces/recluster` re-runs from stored embeddings without re-invoking the sidecar (cheap; needed after threshold tuning).

### API

```
GET   /api/gallery/:slug/people
        → [{id, label, face_count, cover: {photo_id, bbox}}]   (is_hidden/is_ignored filtered out)
GET   /api/gallery/:slug                                        (existing)
        → photos[].person_ids: number[]                         (new field)
POST  /api/gallery/:slug/download            { person_id }      (existing zip job, new selector)

GET   /api/admin/events/:id/people
PATCH /api/admin/events/:id/people/:pid      { label, is_hidden, is_ignored, cover_face_id }
POST  /api/admin/events/:id/people/merge     { source_ids[], target_id }
POST  /api/admin/events/:id/people/:pid/split { face_ids[] }
POST  /api/admin/events/:id/faces/rescan | /recluster
DELETE /api/admin/events/:id/faces
```

`person_ids` rides along on the existing gallery payload (`backend/src/routes/gallery.js:946`), so filtering stays **client-side and instant**, exactly like the like/rating filters at `GalleryView.tsx:498`. An int array per photo is negligible next to the URLs already in that response.

---

## 3. Gallery UI

The design goal: a guest who opens the gallery on their phone should find themselves in **two taps** and never learn the words "face recognition".

### 3.1 People strip

Sits between the filter bar and the grid. Circular crops from each person's cover face bbox, ordered by photo count, horizontally scrollable.

```
┌──────────────────────────────────────────────────────────────────────────┐
│  [🔍 Search photos            ]   [⇅ Sort by Date ▾]   [▦ All ▾]         │
│  ♡ Liked (12)   🔖 Saved (3)   ★ Rated (8)                               │
├──────────────────────────────────────────────────────────────────────────┤
│  People in this gallery                             Show all 14  ›       │
│   ╭───╮   ╭───╮   ╭───╮   ╭───╮   ╭───╮   ╭───╮   ╭───╮                  │
│   │ 😊│   │ 😊│   │ 😊│   │ 😊│   │ 😊│   │ 😊│   │ 😊│   →              │
│   ╰───╯   ╰───╯   ╰───╯   ╰───╯   ╰───╯   ╰───╯   ╰───╯                  │
│    Anna    Ben     Chris    ·84     ·61     ·47     ·31                  │
│    ·142    ·118    ·97                                                    │
├──────────────────────────────────────────────────────────────────────────┤
│                          ▦ photo grid ▦                                   │
└──────────────────────────────────────────────────────────────────────────┘
```

- Unnamed people show the photo count only — never "Person 7". A number is honest; a fake name is not.
- The strip collapses to a single line ("14 people found · Show") when the guest dismisses it, persisted in `localStorage` per slug.
- While the backfill runs: `Finding people… 240/1200 photos` with a thin progress bar, strip appears as soon as the first cluster crosses ~3 faces. No spinner-blocked gallery.
- Fewer than 2 people found → the strip does not render at all.

### 3.2 Active face filter

Selecting a face composes with the existing filters (`selectedCategoryId`, `activeFilters`, search, media type) rather than replacing them — same multi-select semantics as #889.

```
├──────────────────────────────────────────────────────────────────────────┤
│  Showing  (😊) Anna ✕   ·   ♡ Liked ✕            97 of 1,200 photos      │
│                                                   [⤓ Download these 97]   │
├──────────────────────────────────────────────────────────────────────────┤
```

- The selected avatar gets a ring in the strip; the chip row is the single place all active filters live, so "why am I seeing 97 photos" is always answerable at a glance.
- **"Download these 97"** is the payoff @bluecow94 asked for: it hands `person_id` to the existing zip job service, honours per-category download permissions (#640), and is hidden when downloads are off.
- Multi-select two people → AND ("photos with both Anna and Ben"). A small `∩ both / ∪ either` toggle appears only once a second person is picked.

### 3.3 "Show all" sheet

```
┌─ People in this gallery ──────────────────────────────── ✕ ─┐
│  [🔍 Find a person                                        ]  │
│                                                              │
│   ╭────╮   ╭────╮   ╭────╮   ╭────╮   ╭────╮                │
│   │ 😊 │   │ 😊 │   │ 😊 │   │ 😊 │   │ 😊 │                │
│   ╰────╯   ╰────╯   ╰────╯   ╰────╯   ╰────╯                │
│    Anna     Ben      Chris     ·84      ·61                  │
│    142      118       97                                     │
│      ⋮        ⋮        ⋮         ⋮        ⋮                   │
│                                                              │
│   ⓘ People are detected automatically inside this gallery.   │
│     Nothing is sent to any external service.                 │
└──────────────────────────────────────────────────────────────┘
```

The search box matches names once naming exists. The footnote is not decoration — it is the answer to the first question every guest will have.

### 3.4 Lightbox

The people in the open photo render as small avatar chips in the info panel, next to the existing category/filename block:

```
   In this photo:  (😊) Anna   (😊) Ben   (😊) ·47
```

Tapping a chip closes the lightbox and applies that person's filter. A `[⌐]` toggle draws the detection boxes over the image — off by default, useful for the photographer, and the fastest way to explain a bad cluster.

### 3.5 Mobile

```
┌───────────────────────┐
│ [🔍 Search      ] [⇅] │
├───────────────────────┤
│ People          14 ›  │
│  ╭──╮ ╭──╮ ╭──╮ ╭──╮ │
│  │😊│ │😊│ │😊│ │😊│→│
│  ╰──╯ ╰──╯ ╰──╯ ╰──╯ │
│  Anna  Ben  ·97  ·84  │
├───────────────────────┤
│ (😊) Anna ✕   97      │
│ [⤓ Download these 97] │
├───────────────────────┤
│  ▦▦▦   ▦▦▦   ▦▦▦     │
└───────────────────────┘
```

56 px avatars, 44 px minimum touch target, strip scrolls with momentum and snaps. "Show all" opens as a bottom sheet, not a modal.

### 3.6 Guest self-identification (phase 2)

Long-press / hover an avatar → **"This is me"**. When the gallery already runs in guest-identity mode (the existing `FeedbackIdentityModal` flow), the guest's name is proposed as the label, pending photographer approval. Guests never rename someone *else* — that's how a shared family gallery turns into an argument.

---

## 4. Admin UI

**Event → Settings → People** (only rendered when `FACE_ML_URL` is configured):

- `Detect people in this gallery` — off by default, with a one-paragraph plain-language explanation of what gets stored and the consent obligation.
- `Show the people bar to guests` — on by default once the above is on; off means the photographer gets the tool and guests see an unchanged gallery.
- Status line: `1,200 / 1,200 photos scanned · 14 people · last scan 2 h ago` + `Re-scan` / `Re-cluster` / `Delete all face data`.

**Event → People** management grid: rename, merge (multi-select → Merge), split (pick faces out of a cluster), set cover, hide, mark as ignored. Merge and split are the two operations that make automatic clustering survive contact with reality — a well-behaved wedding gallery still produces "Anna in daylight" and "Anna at the party" as separate clusters.

Global settings → Advanced: sidecar URL, model, health indicator, connection test.

---

## 5. Auto-categories (phase 3)

Once faces are detected, `photos.face_count` plus landmark yaw is free input for exactly what @bluecow94 described. A small rule list, global defaults, overridable per event:

| Condition | Category |
|---|---|
| `faces == 0` | Details |
| `faces == 1` and face area ≥ 8% of frame | Portraits |
| `2 ≤ faces ≤ 5` | Small groups |
| `faces > 5` | Groups |
| majority of faces with `abs(yaw) > 25°` | Candid *(experimental)* |

Non-negotiable rules for this engine:
- **Only ever fills an empty `category_id`.** It never overwrites a category the photographer set.
- Writes `photos.auto_categorized = true`, so "undo all automatic categories" is one query.
- Runs as a distinct step after face detection, and can be enabled without the People UI (some photographers will want sorting and not faces).

"In motion → candid" from the discussion needs blur/subject-motion analysis that the face pipeline doesn't produce; not in scope here.

---

## 6. Privacy & GDPR

Face embeddings are biometric data — **Art. 9 GDPR special category** in the EU, where a large share of this user base operates. This is a first-class requirement, not a disclaimer at the bottom:

- Off by default, globally *and* per event. Two deliberate actions to turn on.
- Nothing leaves the deployment. The sidecar has no egress, downloads no models at runtime, and phones nowhere home.
- `DELETE /faces` for an event wipes `photo_faces` + `event_people`; deleting a photo cascades; **archiving/expiring a gallery purges its face data** with the rest.
- Face data is included in backups only when the event's toggle is on (`backupService` must be taught the new tables either way).
- Admin docs get a short section stating plainly that the photographer is the controller and needs a lawful basis for the people in their photos. We provide the switch; we do not provide the consent.
- Guest-facing copy never says "biometric" or "recognition" — it says "People in this gallery" and explains where the data lives.

---

## 7. Phases

**Phase 1 — the thing that was asked for**
`picpeak-ml` container + contract · migration · queue worker · clustering · `GET /people` + `person_ids` · People strip, filter chip, "Download these N" · per-event toggle · lightbox chips · privacy/purge wiring · docs.

**Phase 2 — making the clusters usable**
Admin People grid (rename/merge/split/hide/ignore) · guest "This is me" · consolidation pass · re-cluster endpoint.

**Phase 3 — categories**
Rule engine, per-event rule overrides, `auto_categorized` + bulk undo.

Phase 1 is independently shippable and the whole feature is one env var away from not existing.

## 8. Acceptance criteria

- [ ] `FACE_ML_URL` unset → zero behaviour change, zero new queries, no new UI anywhere.
- [ ] Sidecar down for an hour → photos stay `pending`, catch up automatically, backend logs a rate-limited warning, no user-visible error.
- [ ] 1,000-photo gallery scans on a 2-core / 4 GB host without OOM-killing the backend or delaying thumbnail generation.
- [ ] Face filter composes correctly with category + search + liked/saved/rated + media type.
- [ ] "Download these N" honours per-category download permissions (#640) and the master download toggle.
- [ ] Deleting a photo, archiving an event, or disabling the toggle removes the corresponding face rows — verified by test.
- [ ] Migration is idempotent on re-run and has a working `down()`; both PostgreSQL and SQLite paths tested.
- [ ] Guests of a gallery with `faces_visible_to_guests = false` receive no `person_ids` and no `/people` response.
- [ ] Screenshots of the People strip, active filter chip, "Show all" sheet, and mobile layout attached to the PR (per CONTRIBUTING).

---

Refs: discussion #1069. cc @bluecow94 — the shape above follows what you described (single gallery, filter-by-face like filter-by-category, own container, category rules). The one deviation is the engine: CompreFace's 5-container topology and the non-commercial license on the InsightFace weights it builds on don't work for us, so the plan is a much smaller purpose-built sidecar on permissively licensed models, with buffalo_l available as an opt-in for anyone who's cleared that license themselves.

Phase 1 starts with a throwaway spike against `serengil/deepface` on a real 1,000-photo gallery to confirm the YuNet + FaceNet-512 clustering quality before any of the schema above gets written. If it can't tell two dark-haired guests apart at a reception, that's worth knowing in an afternoon rather than after the migration.

the-luap:
Implements #1074 — "People in this gallery": per-gallery face grouping via an optional ML sidecar, plus rule-based auto-categories.

**Off by default, twice over.** The `faces` feature flag defaults to false, and detection is then enabled per event. With the flag off nothing renders anywhere, no photo is scanned, and the backend never contacts the sidecar.

## Screenshots

Attached in the first comment below — captured against a running stack with real faces, not mockups: the People strip, the active face filter with "Download these N", the "Show all" sheet, mobile, the lightbox chips, the admin card and the People management modal.

## What's here

**`picpeak-ml` sidecar** — single FastAPI + onnxruntime container, three endpoints, no database, no volumes, no egress, no runtime model download. 300 MB, ~186 MiB idle RSS. Behind a `faces` compose profile, so a plain `docker compose up` is unchanged. YuNet + FaceNet-512, both MIT, pinned by URL and SHA-256; the licence reasoning is in `ml/LICENSES.md` (the more accurate InsightFace weights are non-commercial only, and PicPeak's users run businesses on it).

**Backend** — migrations 177/178, a face queue mirroring `backgroundProcessor.js`, greedy centroid clustering with merge/split/re-cluster, and the gallery + admin APIs.

**Frontend** — People strip, face filter composing with every existing filter, "Show all" sheet, "Download these N", lightbox person chips, admin card and a People management modal. EN + DE.

**Auto-categories** — rule engine over face count and face-area ratio. Only ever fills an empty category, marks everything `auto_categorized` so undo is exact, and is separately enableable.

## Worth reviewing closely

**Visibility scoping.** Face rows have no concept of photo visibility, but guests are restricted to `photos.visibility = 'visible'`. A raw count leaks how many hidden photos someone appears in, and an unscoped cover face renders a crop of a photo the guest may not open — with the best-scoring face being the likeliest pick, so it would happen often. `facePeopleService` recomputes both per request against the caller's own scope. Six tests, including the case where all of a person's photos are hidden and they must vanish entirely.

**The clustering threshold is 0.60, and the obvious answer was wrong.** LFW's 1000-pair protocol through this pipeline suggests ~0.50 (1.0% false merge, 8.2% false split pairwise). On a real gallery, 0.50 contaminated three of six clusters. Pairwise error rates do not predict cluster purity — greedy assignment compounds, because one wrong face drags the centroid toward the midpoint between two identities. Full reasoning and the sweep are in migration 177; the benchmark is committed as `ml/tools/benchmark_threshold.py`.

**Face data is excluded from backups and exports.** It is derived, so a restore re-scans rather than carrying biometrics between operators. Three separate mechanisms because the engines can't be filtered alike: `EXCLUDED_TABLES` for export, `--exclude-table-data` for Postgres (not `--exclude-table` — the CREATE TABLE must survive or restore breaks on the first query), and DELETE + VACUUM on the temp copy for SQLite, which cannot exclude a table from a whole-file `.backup`. The VACUUM is not cosmetic: without it the pages stay in the file.

## Testing

45 backend tests and 140 frontend tests green. The full backend suite has 15 pre-existing failures (a missing `cron-parser` dep in the backup suites) — verified identical to `origin/main` by running the suite in a worktree at the base commit and diffing the failure set. `tsc` and `eslint` clean; pre-existing lint counts unchanged.

Verified end-to-end on a running stack: 61 photos scanned with 0 failures, 5 clean clusters against 5 real identities, merge/split/undo confirmed against the database.

## Notes for the reviewer

- CI's `build-ml` leg is gated on the `FACENET_ONNX_URL` repository variable (already set) and publishes to `ghcr.io/picpeak/picpeak/ml`. This PR is the first time that leg runs.
- `AuthenticatedImage` attaches whatever gallery token is in session storage, so an admin who has also opened one of their own galleries sends a `type: "gallery"` bearer to admin routes. I hit this on the new modal and used a plain `` there. `AdminPhotoGrid` has the same latent shape — not touched here.
- Docs are a separate PR: PicPeak/docs#9.

Closes #1074.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -223,6 +223,47 @@ LOGS=./logs
 #   attempts is exponential: 1m, 5m, 30m, 2h, 12h.
 # WEBHOOK_MAX_ATTEMPTS=5
 
+# -----------------------------------------------------------------------------
+# Face recognition — "People in this gallery" (#1074, optional)
+# -----------------------------------------------------------------------------
+# Requires the optional picpeak-ml sidecar container:
+#   docker compose --profile faces up -d
+#
+# NONE of these variables do anything until the `faces` feature flag is
+# enabled in Admin → Settings, AND the per-event "Detect people in this
+# gallery" toggle is switched on. Both default to OFF. With the flag off the
+# backend never contacts the sidecar, so leaving these at their defaults on an
+# install without the container is completely inert.
+#
+# Face embeddings are biometric data (GDPR Art. 9 special category in the EU).
+# The photographer is the controller and needs a lawful basis for the people
+# in their photos — see docs/feature-face-recognition.md before enabling.
+#
+# FACE_ML_TOKEN (no default — REQUIRED to run the sidecar)
+#   Shared secret between the backend and the sidecar. The sidecar refuses to
+#   start without it rather than serving anonymously, so an accidentally
+#   published port is never a free face-detection API. Generate with:
+#     openssl rand -hex 32
+# FACE_ML_TOKEN=
+#
+# FACE_ML_URL (default: http://picpeak-ml:8000)
+#   Defaults to the sidecar's compose service name, so the standard
+#   deployment needs no configuration here. Only change it if you run the
+#   sidecar outside the default compose network.
+# FACE_ML_URL=http://picpeak-ml:8000
+#
+# FACE_PROCESSOR_CONCURRENCY (default: 1)
+#   Face-detection workers in the backend. Defaults to 1 deliberately: face
+#   scanning shares a host with Sharp image processing, which is the real
+#   memory pressure (see UPLOAD_PROCESSOR_CONCURRENCY). Raise only on hosts
+#   with headroom to spare.
+# FACE_PROCESSOR_CONCURRENCY=1
+#
+# FACE_ORT_THREADS (default: 1)
+#   ONNX Runtime threads inside the sidecar. More threads mean faster
+#   per-photo inference and higher RSS.
+# FACE_ORT_THREADS=1
+
 # Note on FRONTEND_API_URL (documentation only):
 # When using pre-built frontend images, runtime env vars cannot override the built JS.
 # Do NOT rely on FRONTEND_API_URL in Compose. Instead, keep VITE_API_URL=/api and
--- a/.env.example
+++ b/.env.example
@@ -239,6 +239,14 @@ LOGS=./logs
 # The photographer is the controller and needs a lawful basis for the people
 # in their photos — see docs/feature-face-recognition.md before enabling.
 #
+# NOT AVAILABLE ON THE ALL-IN-ONE IMAGE. The single-container build sets
+# PICPEAK_SINGLE_CONTAINER=true and the backend refuses to enable face
+# recognition there regardless of these variables or the feature flag: that
+# image runs the backend, frontend, database and every worker in one
+# container, with no ML sidecar to talk to, and face detection would compete
+# with image processing for the same CPU and memory. Use the standard
+# multi-container deployment if you want this feature.
+#
 # FACE_ML_TOKEN (no default — REQUIRED to run the sidecar)
 #   Shared secret between the backend and the sidecar. The sidecar refuses to
 #   start without it rather than serving anonymously, so an accidentally
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -599,8 +599,245 @@ jobs:
       run: |
         docker buildx imagetools inspect docker.io/picpeak/frontend:${{ steps.meta-frontend.outputs.version }}
 
+  # -----------------------------------------------------------------------------
+  # ML sidecar (#1074): per-arch build, then merge into a multi-arch manifest
+  # -----------------------------------------------------------------------------
+  # Gated on the FACENET_ONNX_URL repository *variable* (Settings → Variables,
+  # not Secrets — it's a public release-asset URL). While it is unset, both ML
+  # jobs skip and the workflow behaves exactly as it did before this feature.
+  #
+  # Why a gate at all: deepface distributes FaceNet-512 as Keras .h5 only, so
+  # the ONNX has to be produced once by ml/tools/convert_facenet.py and
+  # published as a release asset before anything can build. Converting inside
+  # this workflow would drag TensorFlow (~600MB) through BOTH architecture legs
+  # of EVERY build to produce a file that is byte-identical each time.
+  #
+  # To activate, set two repository variables:
+  #   FACENET_ONNX_URL     https://github.com/PicPeak/picpeak/releases/download/<tag>/facenet512.onnx
+  #   FACENET_ONNX_SHA256  <sha256 of that file>
+  # See ml/README.md for producing them.
+  build-ml:
+    if: vars.FACENET_ONNX_URL != ''
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - platform: linux/amd64
+            runner: ubuntu-latest
+          - platform: linux/arm64
+            runner: ubuntu-24.04-arm
+    runs-on: ${{ matrix.runner }}
+    permissions:
+      contents: read
+      packages: write
+      security-events: write
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v4
+
+    - name: Compute image names (lowercase for GHCR)
+      run: |
+        repo_lc="${GITHUB_REPOSITORY,,}"
+        echo "ML_IMAGE_NAME=${repo_lc}/ml" >> "$GITHUB_ENV"
+
+    - name: Prepare platform pair
+      run: |
+        platform="${{ matrix.platform }}"
+        echo "PLATFORM_PAIR=${platform//\//-}" >> "$GITHUB_ENV"
+
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v3
+
+    - name: Log in to Container Registry
+      if: github.event_name != 'pull_request' || github.event.inputs.push == 'true'
+      id: login-ghcr
+      continue-on-error: true
+      uses: docker/login-action@v3
+      with:
+        registry: ${{ env.REGISTRY }}
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+
+    - name: Determine if pushing
+      id: push-decision
+      run: |
+        if [[ "${{ github.event_name }}" == "pull_request" && "${{ github.event.inputs.push }}" != "true" ]]; then
+          echo "push=false" >> "$GITHUB_OUTPUT"
+        elif [[ "${{ steps.login-ghcr.outcome }}" != "success" ]]; then
+          echo "push=false" >> "$GITHUB_OUTPUT"
+        else
+          echo "push=true" >> "$GITHUB_OUTPUT"
+        fi
+
+    - name: Extract metadata for ML (labels only)
+      id: meta-ml
+      uses: docker/metadata-action@v5
+      with:
+        images: ${{ env.REGISTRY }}/${{ env.ML_IMAGE_NAME }}
+        labels: |
+          org.opencontainers.image.title=PicPeak ML
+          org.opencontainers.image.description=PicPeak face detection and embedding sidecar
+          org.opencontainers.image.vendor=PicPeak
+          maintainer=${{ github.repository_owner }}
+
+    - name: Build ML image (push by digest)
+      id: build
+      uses: docker/build-push-action@v5
+      with:
+        context: ./ml
+        file: ./ml/Dockerfile
+        platforms: ${{ matrix.platform }}
+        labels: ${{ steps.meta-ml.outputs.labels }}
+        cache-from: type=gha,scope=ml-${{ env.PLATFORM_PAIR }}
+        cache-to: type=gha,mode=max,scope=ml-${{ env.PLATFORM_PAIR }},ignore-error=true
+        outputs: ${{ steps.push-decision.outputs.push == 'true' && format('type=image,name={0}/{1},push-by-digest=true,name-canonical=true,push=true', env.REGISTRY, env.ML_IMAGE_NAME) || 'type=cacheonly' }}
+        build-args: |
+          CACHEBUST=${{ github.run_number }}
+          BUILD_DATE=${{ github.event.head_commit.timestamp }}
+          VCS_REF=${{ github.sha }}
+          VERSION=${{ steps.meta-ml.outputs.version }}
+          FACENET_ONNX_URL=${{ vars.FACENET_ONNX_URL }}
+          FACENET_ONNX_SHA256=${{ vars.FACENET_ONNX_SHA256 }}
+
+    - name: Export digest
+      if: steps.push-decision.outputs.push == 'true'
+      run: |
+        mkdir -p /tmp/digests
+        digest="${{ steps.build.outputs.digest }}"
+        touch "/tmp/digests/${digest#sha256:}"
+
+    - name: Upload digest artifact
+      if: steps.push-decision.outputs.push == 'true'
+      uses: actions/upload-artifact@v4
+      with:
+        name: digests-ml-${{ env.PLATFORM_PAIR }}
+        path: /tmp/digests/*
+        if-no-files-found: error
+        retention-days: 1
+
+    # Per-arch scan by digest, same reasoning as the backend leg (#476).
+    # This image carries a Python/Debian dependency surface the other two
+    # don't, so it gets its own Security-tab category.
+    - name: Run Trivy vulnerability scanner (per-arch, by digest)
+      if: steps.push-decision.outputs.push == 'true'
+      uses: aquasecurity/trivy-action@v0.36.0
+      env:
+        TRIVY_PLATFORM: ${{ matrix.platform }}
+      with:
+        image-ref: ${{ env.REGISTRY }}/${{ env.ML_IMAGE_NAME }}@${{ steps.build.outputs.digest }}
+        format: 'sarif'
+        output: 'trivy-ml-${{ env.PLATFORM_PAIR }}.sarif'
+        severity: 'CRITICAL,HIGH'
+        timeout: '10m'
+
+    - name: Upload Trivy scan results to GitHub Security tab
+      if: steps.push-decision.outputs.push == 'true'
+      uses: github/codeql-action/upload-sarif@v4
+      with:
+        sarif_file: 'trivy-ml-${{ env.PLATFORM_PAIR }}.sarif'
+        category: 'ml-vulnerabilities-${{ env.PLATFORM_PAIR }}'
+
+  merge-ml:
+    needs: build-ml
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+    if: vars.FACENET_ONNX_URL != '' && (github.event_name != 'pull_request' || github.event.inputs.push == 'true')
+
+    steps:
+    - name: Compute image names (lowercase for GHCR)
+      run: |
+        repo_lc="${GITHUB_REPOSITORY,,}"
+        echo "ML_IMAGE_NAME=${repo_lc}/ml" >> "$GITHUB_ENV"
+        if [[ "$GITHUB_REPOSITORY" == "PicPeak/picpeak" ]]; then
+          echo "DOCKERHUB_ENABLED=true" >> "$GITHUB_ENV"
+        else
+          echo "DOCKERHUB_ENABLED=false" >> "$GITHUB_ENV"
+        fi
+
+    - name: Download digest artifacts
+      uses: actions/download-artifact@v4
+      with:
+        path: /tmp/digests
+        pattern: digests-ml-*
+        merge-multiple: true
+
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v3
+
+    - name: Log in to Container Registry
+      id: login-ghcr
+      continue-on-error: true
+      uses: docker/login-action@v3
+      with:
+        registry: ${{ env.REGISTRY }}
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+
+    - name: Determine build context
+      id: context
+      run: |
+        if [[ "${{ github.ref }}" == refs/tags/v*-beta* ]] || [[ "${{ github.ref }}" == refs/heads/main ]]; then
+          echo "channel=beta" >> $GITHUB_OUTPUT
+          echo "is_prerelease=true" >> $GITHUB_OUTPUT
+        else
+          echo "channel=stable" >> $GITHUB_OUTPUT
+          echo "is_prerelease=false" >> $GITHUB_OUTPUT
+        fi
+
+    - name: Log in to Docker Hub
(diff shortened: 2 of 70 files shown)


the-luap:
# Screenshots

All captured against a running stack with the real pipeline — 61 photos, 5 people, live sidecar. Not mockups.

### Guest — People strip
The bar sits between the filters and the grid. Named people show their name and count; unnamed ones show a count only, never an invented "Person 7".

issue-1074-1-people-strip

### Guest — face filter active + "Download these N"
Selecting a person filters the grid and surfaces the download. The chip row is the single place the active selection is stated, so "why am I seeing 13 photos" is always answerable.

issue-1074-2-face-filter-active

### Guest — "Show all" sheet
Search by name, and the footnote answering the first question every guest has. *(Captured with the minimum cluster size temporarily lowered to 1 so the sheet had enough entries to open — the 1-photo entries it shows are exactly what the default of 3 keeps out.)*

issue-1074-3-show-all-sheet

### Guest — mobile (390 px)
Strip scrolls horizontally; the chip row wraps so the count and Clear stay inside the viewport.

issue-1074-4-mobile

### Guest — lightbox chips
"In this photo:" — the second way into the face filter, for someone looking at a photo of themselves. Tapping a chip closes the lightbox and filters the grid behind it.

issue-1074-6-lightbox-chips

### Admin — People card
The consent obligation sits next to the toggle rather than behind a "learn more", and the preview-tier disk cost is stated before anyone starts a 2,000-photo backfill.

issue-1074-5-admin-card

### Admin — People management
Rename inline, multi-select to merge, split via a face picker, hide, ignore. Merge matters because the clustering deliberately errs toward splitting: a duplicate entry is an annoyance, a wrong merge puts a stranger into someone's download.

issue-1074-7-people-manager

PR #1087, opened 2026-08-19
fix(faces): restore the :beta image tag and surface sidecar health

the-luap:
Both halves of what @bluecow94 hit on #1069: the People card sat at `Scanning… 0 of 227` for 30 minutes with no explanation, because the sidecar container could never have started in the first place.

## 1. Republish `:beta` — the ML image has no resolvable tag

`docker-compose.production.yml:158` pins the sidecar to `ghcr.io/picpeak/picpeak/ml:${PICPEAK_CHANNEL:-stable}`. Queried GHCR directly — the only published tags for that image are:

```
main, sha-*, 3.107.0-beta.0, v3.107.0-beta.0, 3.107.1-beta.0, v3.107.1-beta.0
```

**No `beta`, no `stable`, no `latest`.** So neither the default channel nor the documented `beta` channel can resolve the image, the container never starts, and the queue never drains.

### The cause is wider than the sidecar

`:beta` used to be produced for free by `type=ref,event=branch`, back when the active development branch was literally named `beta`. The rename to `main` silently retired it — no job in the workflow produces a `beta` tag today. Confirmed against the registry:

```
backend:beta   created 2026-06-29  revision 448da950   ← frozen at the rename
backend:main   created 2026-08-19  revision 74a8f9bf   ← current
```

So **`PICPEAK_CHANNEL=beta` has been serving a seven-week-old build for every image**, not just the sidecar. The ML image was added after the rename and so never got a `:beta` at all.

Fix: `type=raw,value=beta,enable=${{ github.ref == 'refs/heads/main' }}` on all four merge jobs.

`:stable` deliberately stays absent for ml — it is gated on `refs/heads/stable`, the sidecar does not exist there, and stable's `docker-compose.production.yml` carries no `picpeak-ml` service, so nothing can reference the missing tag.

## 2. Surface sidecar health instead of an endless spinner

An unreachable sidecar is not a failure by design — `faceQueue.js:132-136` catches `SidecarUnavailableError`, releases the photo back to `pending` and retries forever so a restart doesn't burn the queue. That is right, but it made a stopped container indistinguishable from a slow scan, permanently. The only signal was a backend log line rate-limited to once per five minutes.

`/admin/events/faces/health` already existed (`adminEvents/faces.js:45`) and **nothing in the frontend ever called it**. It is now polled while a scan is in progress; a failing check replaces the spinner with the sidecar URL, the underlying error, and the command to start it.

Health is only polled while a scan is running — an idle card has no reason to care whether the sidecar is up.

### Sidecar unreachable — container stopped or never started

unreachable

### Token rejected (401) — red, because these photos are being **burnt**, not queued

token

### Wrong FACE_ML_URL (non-401 4xx) — same class, different cause

wrongurl

### Scan already finished — counts kept, diagnosis added underneath

afterscan

### Healthy scan — unchanged

healthy

## External review

Ran an iterative Codex review over four rounds. It found real problems in my first cut, all now fixed:

1. **401 is not retryable.** `classify()` turns it into `SidecarRejectedError`, which `faceQueue.js:132` does not retry — photos are marked `failed`. My original message promised the scan would resume on its own, which was flatly wrong for that case.
2. **A busy sidecar can look dead.** `/faces` runs inference synchronously inside an `async def`, blocking the event loop uvicorn needs to answer `/info`. Confirmed with an isolated uvicorn repro: a blocking call in an async handler stalled the sync `/info` endpoint to **5.01s**, tripping its 5s timeout.
3. **Every 4xx burns photos, not just 401** — a wrong `FACE_ML_URL` answering 404 fails photos identically.
4. **`FACE_ML_TOKEN` has no default** and the container refuses to start without it, so the most likely first-run failure arrives as a plain connection refusal that "just start it" does not fix.
5. **A 4xx empties the queue with no backoff**, so `in_progress` can be false before anyone opens the card — leaving only "227 failed" with no explanation.

On magnitude I pushed back and Codex accepted the correction: real inference here is ~0.33s for a 7000×7000 image and a single facenet512 embedding is 12ms, so even a 40-face photo is ~0.5s — nowhere near 5s on normal hardware. The guard is for slow self-hosted hosts, not the common case.

I also **declined** one suggestion: deriving finished-run diagnoses from stored `face_error` rows. It is correct that a live probe cannot attribute historical failures, but that needs new backend plumbing and is out of scope here. Instead the post-scan notice is worded as present-tense service state and says explicitly that some failures above may have another cause. Noted as a follow-up.

## Verification

- **10 tests** in `FaceRecognitionCard.sidecarHealth.test.tsx` covering all three failure modes, the streak threshold, the post-scan case and the healthy/idle paths
- Confirmed the tests are not vacuous: stashing the component change fails the unreachable case, and relaxing the streak threshold from 3 to 2 fails the "stays quiet at two failures" case
- Frontend suite: **150 passed (27 files)**, `tsc --noEmit` clean, eslint clean, production build green
- Backend: eslint clean, `jest --testPathPattern=face` **47 passed**
- Workflow YAML re-parsed; `beta` tag confirmed present in all four `merge-*` jobs and gated on `refs/heads/main`
- Registry facts above were read from GHCR and `buildx imagetools inspect`, not inferred

## Note

The `:beta` tags only start existing after this merges and the next `main` build runs. Until then the workaround for anyone on the faces feature is to pin `PICPEAK_CHANNEL=v3.107.1-beta.0` explicitly.

New i18n keys are EN + DE only, matching the existing `admin.faces.*` convention; other locales fall back to the inline default.

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -331,6 +331,15 @@ jobs:
           # so `is_default_branch` no longer maps to "stable" — be explicit.
           type=raw,value=latest,enable=${{ github.ref == 'refs/heads/stable' || (startsWith(github.ref, 'refs/tags/v') && steps.context.outputs.is_prerelease == 'false') }}
           type=raw,value=stable,enable=${{ github.ref == 'refs/heads/stable' || (startsWith(github.ref, 'refs/tags/v') && steps.context.outputs.is_prerelease == 'false') }}
+          # `:beta` follows the active development branch. This used to happen
+          # for free via `type=ref,event=branch` back when that branch was
+          # literally named `beta`; the rename to `main` silently retired the
+          # tag, so `PICPEAK_CHANNEL=beta` has been pinned to the last pre-rename
+          # build (backend:beta sat at 2026-06-29 / 448da950) while :main moved
+          # on. The ml sidecar was added after the rename and so never had a
+          # `:beta` at all, which left docker-compose.production.yml unable to
+          # resolve the image for any documented channel.
+          type=raw,value=beta,enable=${{ github.ref == 'refs/heads/main' }}
           # `:beta` is RETIRED post-rename (Option B / #669). Active-dev pulls
           # are `:main` (auto via type=ref,event=branch). The pre-rename `:beta`
           # tag remains frozen at its last build — operators should update.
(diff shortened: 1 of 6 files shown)


bluecow94 (approved):
This has got the face scanning working - confirmed working here. It made some duplicates, but these were quickly merged using the 'manage people > merge', and appears to be working well. 

To get it started, I downloaded this PR, then had to add a secret on FACE_ML_TOKEN (which was still commented out in .env), then force restart with docker compose --profile faces up -d --force-recreate picpeak-ml backend

Image shows progress started, with two 'people' who are the same but which merged successfully in admin settings:
Image

PR #1104, opened 2026-08-20
feat(setup): configure the public address and SMTP in the wizard, not .env

Luca-Timo:
Towards #705 — the "zero-touch `.env`" half of the deployment-onboarding issue.

## Problem

A fresh install cannot configure its own public address. The plumbing looks like it is already there — `general_site_url` is a real permission-gated setting, `getFrontendBaseUrl()` already falls back to it, and the wizard already writes SMTP into `email_configs` — but nothing can actually reach any of it.

**An injected default always wins.** `docker-compose.yml` set `FRONTEND_URL=${FRONTEND_URL:-http://localhost:3000}` and `Dockerfile.aio` baked in `ENV FRONTEND_URL=http://localhost:3000`. The resolver returns on its first truthy branch, so `general_site_url` was never read in any default deployment, and an admin who corrected the URL in Settings saw it silently do nothing. `.env.example` shipped the same trap as uncommented placeholders:

```
FRONTEND_URL=https://yourdomain.com
ADMIN_URL=https://yourdomain.com
API_URL=https://yourdomain.com/api
```

`cp .env.example .env` — which the quickstart instructs — therefore configured every install to believe it lives at `yourdomain.com`. It does not error; it just emits client links nobody can open.

**The wizard never asked.** It collects invoicing and SMTP, but not the address. And `continueAfterEventTypes()` skipped the entire config step unless `bills`, `reminderEmails`, `incomingMail` or `whatsapp` was selected — so a gallery-only install was never offered SMTP either, despite gallery links, guest invites and expiry warnings all being delivered through the same `email_configs` row. The most basic install was the one guaranteed not to be asked how to send mail.

**Eleven call sites bypassed the resolver**, reading `process.env.FRONTEND_URL` directly. Three defaulted to placeholder hosts that reach real recipients:

| Call site | Old default | Sent to |
|---|---|---|
| `invoice/payments.js` | `https://app.example.com` | customer payment-reminder emails |
| `userManagementService.js` | `http://localhost:3005` | admin invitation emails |
| `adminDev.js` | `https://app.example.com` | template previews |

(`app_frontend_url`, read at `payments.js`, is referenced nowhere else in the codebase and is never written.)

## Approach

Stop injecting a default anywhere, and resolve instead:

**`FRONTEND_URL` → `general_site_url` → the origin the request arrived on → whichever exists → `''`**

- A **loopback candidate is treated as unconfigured**, so the installs that already have `http://localhost:3000` baked into their environment self-heal rather than mailing every recipient a link to their own machine. This guard already existed inline in `routes/gallery.js` for the slideshow QR (#848) — it is now shared rather than duplicated.
- The **empty return is preserved**: `shareLinkService` and the SSO redirects in `routes/auth` rely on it to emit *relative* urls, which is the better answer for a same-origin redirect. Callers needing an absolute url use `getAbsoluteFrontendUrl()`, which still ends at `http://localhost:3000` — so a genuine localhost install is unchanged.
- Env is still highest precedence. Config-as-code deployments keep working; the field just goes read-only in the admin UI so nobody edits a setting an env var is overriding.
- The setting is cached (it now sits in per-request CORS paths and a synchronous accessor) and invalidated on write. Invalidation **refreshes rather than blanks** — blanking would leave `Access-Control-Allow-Origin` falling back to `*` until some async caller re-read.

**The wizard** persists `window.location.origin` immediately after the admin account is created — background jobs like reminder emails have no request to derive an origin from, so it has to be stored, not inferred — and offers it as an editable **Public address** field. The config step is now always visited, and SMTP is no longer feature-gated.

## Full change inventory

Two commits. The second corrects a precedence reversal introduced by the first — flagged rather than squashed, since it is exactly the kind of thing worth seeing separately in review.

**The resolver**
- `backend/src/utils/frontendUrl.js` — the precedence chain above, plus `getAbsoluteFrontendUrl()`, a synchronous `getFrontendBaseUrlSync()` for middleware that cannot await, `getApiBaseUrl()`, `isLoopbackBase()`, `isEnvPinned()`, and the cache with explicit invalidation (30s TTL as backstop for writes that bypass it, e.g. a restore).

**Stop injecting defaults**
- `docker-compose.yml` — `FRONTEND_URL` / `ADMIN_URL` no longer default to localhost.
- `Dockerfile.aio` — `ENV FRONTEND_URL` removed.
- `.env.example` — `FRONTEND_URL`, `ADMIN_URL`, `API_URL`, `SMTP_*`, `EMAIL_FROM` commented out with explanations.

**Call sites routed through the resolver**
- `galleryOgService.js`, `transferService.js` — internal helpers become async (both callers already were).
- `updateNotificationService.js`, `invoice/payments.js`, `adminDev.js`, `userManagementService.js`.
- `adminGuests.js` — was resolving at **module load**, so it could never see a runtime setting; moved into the handlers.
- `adminEvents/crud.js` — client-access link in the gallery-created email.
- `secureImageMiddleware.js` — CORS header, via the sync accessor.
- `server.js` — **both** CORS allowlists (the API `corsOptions.origin` callback and the static-file `setCorsHeaders`), the OG canonical for short urls, and `primeSiteUrlCache()` at boot so the sync accessor can see the setting.
- `gallery.js` — the local `QR_LOCAL_BASE_RE` regex now delegates to `isLoopbackBase()`.

**Admin surface**
- `adminSettings.js` — GET `/` and GET `/:type` both return derived `general_site_url_env_pinned` / `general_site_url_effective`; both keys are added to `isReservedSettingKey` so a round-trip of the GET payload cannot create phantom setting rows; `general_site_url` writes invalidate the cache.
- `GeneralTab.tsx` / `useSettingsState.ts` — the Site URL field goes read-only with an explanation when the env pins it.

**Wizard**
- `SetupPage.tsx` — persists the origin after account creation; the config step is always visited.
- `SetupConfigStep.tsx` — new Public address section, SMTP no longer feature-gated, saves on both Finish and Skip, intro copy updated (it no longer describes only "the features you picked").
- `en.json` / `de.json` — new strings.

**Also**
- `001_init.js` — `email_configs` no longer seeds `mailhog`.
- `backend/src/services/emailService.js` — deleted.
- `docs/single-container.md` — `FRONTEND_URL` guidance rewritten.
- `backend/src/__tests__/frontendUrl.test.js` — new.

## The mailhog seed

`001_init.js` fell back to `smtp_host: process.env.SMTP_HOST || 'mailhog'`. That host only exists in the `dev` compose profile, which does not start on a plain `docker compose up -d`. Combined with removing `SMTP_*` from `.env.example`, every fresh install came up with a **live config pointing at a nonexistent host** while the wizard showed empty fields — so blank read as "unconfigured" and mail failed silently. Now seeded only when `SMTP_HOST` is set; with no row, `emailProcessor` logs `No email configuration found` and blank fields are the truth. Developers set `SMTP_HOST=mailhog` explicitly (documented in `.env.example`). Existing installs are untouched — `001` has already run for them and the `if (!emailConfig)` guard is unchanged. **This is a scope decision, not a repair:** an install that already has the bogus `mailhog` row keeps it, so its wizard and Settings → Email still show a "configured" SMTP host that does not exist. Fixing that needs a migration that decides whether a given `mailhog` row is a real dev config or the seed artefact, which belongs in its own change.

## Testing

- New `frontendUrl.test.js` — 26 cases pinning the precedence chain, the loopback demotion, the load-bearing empty return, the `localhost:3000` terminal fallback, `API_URL` split-origin override, and the loopback matcher.
- Full backend suite: **1929 passing**, up from 1910. The 8 failing suites are identical to those failing on a clean `main` in the same environment (storage/PDF path issues) — no regressions.
- Frontend `tsc --noEmit` and `npm run build` clean.
- End-to-end against a locally built **AIO image with no URL or SMTP environment at all**: bootstrap admin → `general_site_url_env_pinned: false` → wrote `http://192.168.1.50:3000` simulating a NAS → the image's own resolver returned it, with `getApiBaseUrl()` deriving `/api`. Confirmed a fresh install now seeds no `email_configs` row.

## Notes for review

- **Locales:** `en` and `de` only. The other six (`es`, `fr`, `nl`, `pt`, `ru`, `sl`) fall back to the inline English defaults — nothing breaks, but they need native strings rather than machine translation.
- **`gallery.js` shim.** Rather than touching its three `.test()` call sites, `QR_LOCAL_BASE_RE` became an object with a `test()` method delegating to `isLoopbackBase()`. It keeps the diff small but is an odd shape; happy to convert the call sites properly if you would rather.
- **`secureImageMiddleware` CORS** now resolves the configured origin rather than whatever compose injected. Deliberately *not* derived from the request — reflecting the caller's `Origin` would defeat the allowlist. Worth noting this header gates nothing today: the frontend builds those urls as relative paths, so they are always same-origin.
- **Editing `001_init.js` in place** changes fresh installs only. Flagging it explicitly since it is a long-deployed migration.
- **Size.** This trips `bypass-size-gate` (719 lines vs the 300 limit for bypass users), which is the gate working as intended — it wants a maintainer review rather than a self-merge. Happy to split the resolver, the wizard, and the mailhog seed into separate PRs if you would prefer to review them apart.

---

## Review round 2

All five points from @the-luap's review are addressed in `fix(setup): unpin loopback FRONTEND_URL, keep ADMIN_URL/APP_URL reachable`. Branch rebased onto `main`.

**🔴 1 — `isEnvPinned()` locked out the operators this PR is for.** Fixed exactly as suggested: the predicate now mirrors the resolver, so a loopback `FRONTEND_URL` is not reported as authoritative. Added `envPinnedBase()` alongside it and used it for `general_site_url_effective` in both GET paths, replacing the inline `process.env.FRONTEND_URL.trim().replace(...)` so the displayed value and the pinned-ness decision can't drift apart again. Two new cases cover the upgrade path.

**🔴 2 — `APP_URL` / `ADMIN_URL` were unreachable.** You're right that the comment described an order the code no longer had. They're now passed *into* the resolver as an explicit `override` that ranks directly below `FRONTEND_URL` and above `general_site_url` / the request origin, so an operator's per-purpose intent wins over anything derived. Historic `FRONTEND_URL || ADMIN_URL` order is preserved; the override gets the same loopback demotion as every other candidate. Five new cases.

**🟡 1 — protected-key round-trip.** The General tab no longer sends `general_site_url` at all while it is env-pinned (nor the derived `site_url_env_pinned`). Kept the field showing the *effective* value, since that's what the read-only display is for — the two halves now agree because the key simply isn't in the payload.

**🟡 2 — wizard advancing on a failed SMTP save.** `finish()` validates before posting (From address required and well-formed when a host is set, port in range) and `onDone()` moved out of the `finally`, so a failed save keeps the user on the step with their input intact. Errors render inline per field, and the From placeholder reads "(required)" once a host is entered. `Skip for now` still always leaves.

**🟡 3 — no URL validation.** Fixed at both layers rather than deferred, since the value reaches `Access-Control-Allow-Origin`:
- Backend, `PUT /admin/settings/general`: `validator.isURL(..., { protocols: ['http','https'], require_protocol: true, require_tld: false })` → 400. `require_tld` is off deliberately so `http://nas:3000` and bare IPs still work for LAN/NAS installs. Trailing slashes are normalised on save.
- Frontend: shared `isAbsoluteHttpUrl()` in `utils/url.ts`, used by both surfaces that write the key — the wizard blocks submit, the General tab shows an inline error and disables Save. (`type="url"` is inert in both, as you noted, because neither input is inside a `<form>`.)

**🟢 Nits.** Rebased. The `001_init.js` scope decision is now spelled out in *The mailhog seed* above.

New strings are `en` + `de` only, same as the rest of the PR.


the-luap:
Reviewed. The design here is sound — the precedence chain, the deliberate loopback demotion, and keeping the synchronous resolver away from `req` so `Origin` is never reflected are all the right calls, and the test suite for the chain is better than most of what lands in this repo. Two things block, three want a look.

## 🔴 Blockers

### 1. `isEnvPinned()` locks out the operators this PR is for

`backend/src/utils/frontendUrl.js:87` — `isEnvPinned = () => !!envBase()` reports *any* `FRONTEND_URL` as authoritative, including the loopback values the resolver two lines below deliberately demotes.

Concretely: an existing install upgrading with `FRONTEND_URL=http://localhost:3000` (the old compose default, and plenty of people have it pinned in their `.env`) gets the resolver correctly preferring `general_site_url` — but `general_site_url_env_pinned` comes back `true`, so Settings → General shows the Site URL field read-only, and `SetupPage` skips seeding it because of `!existing?.general_site_url_env_pinned`. That operator cannot configure their public address through the UI at all, which is precisely the situation the PR exists to fix.

The predicate needs to mirror the resolver:

```js
const isEnvPinned = () => { const e = envBase(); return !!e && !isLoopbackBase(e); };
```

### 2. `APP_URL` and `ADMIN_URL` are now unreachable

`backend/src/routes/adminEvents/crud.js:581` and `backend/src/services/userManagementService.js:71`.

The comment says *"Same FRONTEND_URL-before-APP_URL order as before"*, but that is not what happens. On `main` it was:

```js
const frontendUrl = process.env.FRONTEND_URL || process.env.APP_URL || '';
```

so with `FRONTEND_URL` unset, `APP_URL` was used. Now `getFrontendBaseUrl()` resolves `FRONTEND_URL` → `general_site_url` → request origin internally, so it only returns falsy when *nothing* is configured — and `|| process.env.APP_URL` is dead whenever a site URL exists, which after this PR is the normal case.

The practical failure is split-origin deployments, which this codebase explicitly supports (`API_URL` exists for #798): an operator who sets `ADMIN_URL` for a separate admin host now gets admin invite links pointing at the public gallery origin. Explicit per-purpose overrides should win over the database/request-derived fallback, not sit behind it.

## 🟡 Concerns

1. `frontend/src/features/settings/hooks/useSettingsState.ts:208` — when pinned, the form is seeded with `general_site_url_effective` rather than the stored `general_site_url`. Saving any unrelated General setting reposts that as a genuine change to a protected key, so `collectUnauthorizedProtectedKeys` 403s an `settings.edit`-without-`settings.domains` role. `adminSettings.js:91` explicitly documents that a no-op round-trip must not 403 — the frontend now round-trips a different value than the one stored, so the two halves disagree. Either keep the stored value for round-tripping or omit the key from saves while pinned.

2. `frontend/src/components/admin/SetupConfigStep.tsx:99` — a blank From address is posted as `from_email: ''`, `/admin/email/config` rejects it, the `catch` shows the generic "some settings could not be saved" warning, and `onDone()` runs from the `finally` regardless. The wizard advances and every SMTP value the user typed, password included, is gone. The field also carries no required marker (`:173`). Validate before submitting, or do not call `onDone()` on a failed save.

3. **No URL validation, at either layer.** `SetupConfigStep.tsx:50` only trims and strips trailing slashes, and `type="url"` never fires because the input is not in a form — so `gallery.example.com` saves happily. The backend does not validate it either.

   That matters more after this PR than before it: `general_site_url` now feeds the CORS allowlist (`server.js:225`, `:511`) and the `Access-Control-Allow-Origin` header (`secureImageMiddleware.js:279`), not just email links. A schemeless value produces links that do not resolve and an allowlist entry that can never match a real browser origin. Client-side validation is not a control here — the precedent for the server-side fix is already in the same file at `adminSettings.js:688`, where `oidc_issuer_url` uses `isURL({ protocols: ['http','https'] })`.

## 🟢 Nits

- Branch is behind `main`; needs a rebase before merge.
- The `001_init.js` change only helps fresh installs — existing ones keep the bogus `mailhog` row, so their wizard still shows a "configured" SMTP host that does not exist. Fine as a scope decision, worth a line in the PR body so it is not mistaken for a repair.

## ✅ Verified

Directly inspected, not taken from the description:

- `001_init.js` seed change is inside `if (!emailConfig)` and cannot re-run on an existing install
- CORS posture is preserved — both `server.js` call sites and `secureImageMiddleware.js` use the synchronous resolver **without** `req`, so the caller's `Origin` is never reflected back
- Cache invalidation fires on both write paths, including the wizard's (it goes through the same settings PUT → `adminSettings.js:1493`)
- `app.set('trust proxy')` is configured at `server.js:130`, so the `X-Forwarded-Proto` behaviour the tests assert is real in production
- Upgrade path for emails holds: with no env and no setting, `getAbsoluteFrontendUrl()` still ends at `DEFAULT_ABSOLUTE_BASE`, so links stay absolute rather than degrading to relative
- `emailService.js` is deleted cleanly — no remaining `require` of it anywhere in `backend/src`
- The precedence tests cover loopback demotion, `X-Forwarded-Proto`, the empty case and trailing slashes

Not independently verified: the i18n additions and the `docs/single-container.md` wording, which I read but did not exercise.

## Verdict

Close blockers 1 and 2 and I am comfortable merging. Concern 3 I would rather see fixed here than deferred, given the setting now reaches a security header.

Luca-Timo:
@the-luap thanks — both blockers were real, and blocker 2 was my comment describing an order the code no longer had. All five points are addressed in `19c68c7`, and the branch is rebased onto `main`.

### 🔴 1 — `isEnvPinned()` locking out the operators this PR is for

Fixed as you wrote it — the predicate mirrors the resolver now:

```js
const envPinnedBase = () => {
  const env = envBase();
  return env && !isLoopbackBase(env) ? env : '';
};
const isEnvPinned = () => !!envPinnedBase();
```

I pulled the value out into `envPinnedBase()` rather than inlining the check, because `adminSettings.js` was computing the displayed `general_site_url_effective` with its own `process.env.FRONTEND_URL.trim().replace(/\/+$/, '')` in two places. Both GET paths now read the same helper, so "is it pinned" and "what is it pinned to" can't drift apart again.

Covered by two new cases: a loopback `FRONTEND_URL` reports `isEnvPinned() === false`, and `envPinnedBase()` normalises.

### 🔴 2 — `APP_URL` / `ADMIN_URL` unreachable

Agreed — `|| process.env.ADMIN_URL` after the resolver is dead whenever anything is configured. Rather than reordering at each call site, the override is passed *into* the resolver so the precedence lives in one place:

```js
const frontendUrl = await getAbsoluteFrontendUrl(null, { override: process.env.ADMIN_URL });
```

It resolves directly below `FRONTEND_URL` and above `general_site_url` / the request origin, which preserves the historic `FRONTEND_URL || ADMIN_URL` order while making explicit per-purpose config beat anything derived. It gets the same loopback demotion as every other candidate — worth flagging, since the old `ADMIN_URL` default in the docs was `http://localhost:3005`, and taking that literally is exactly the failure the demotion exists to prevent.

Five new cases: override beats the setting, beats the request origin, still loses to `FRONTEND_URL`, is demoted when loopback, and reaches `getAbsoluteFrontendUrl()`.

Also dropped the "`APP_URL` looks vestigial" line from the PR body — with a working fallback chain it isn't.

### 🟡 1 — protected-key round-trip

Took the second option: the General tab omits `general_site_url` from the payload entirely while it's pinned (and `site_url_env_pinned` with it). Kept the field *showing* the effective env value, since that's the point of the read-only display — the two halves agree now because the key simply isn't in the save.

### 🟡 2 — wizard advancing on a failed save

`finish()` validates before posting — From address required and well-formed when a host is set, port in range — and `onDone()` moved out of the `finally`, so a failed save keeps the user on the step with everything they typed. Errors render inline per field, and the From placeholder switches to "From address (required)" as soon as a host is entered. `Skip for now` still always leaves (it only blocks on a malformed address, since that one *is* posted).

### 🟡 3 — URL validation

Fixed here rather than deferred, agreed on the reasoning. Both layers:

- **Backend**, `PUT /admin/settings/general`: `validator.isURL(value, { protocols: ['http','https'], require_protocol: true, require_tld: false })` → 400, following the `oidc_issuer_url` precedent you pointed at. `require_tld: false` is deliberate — LAN and NAS installs legitimately run on `http://nas:3000` or a bare IP, and this is the field they use. Trailing slashes normalised on save.
- **Frontend**: shared `isAbsoluteHttpUrl()` in `utils/url.ts`, used by both surfaces that write the key — wizard blocks submit, General tab shows an inline error and disables Save. Not a control, just so nobody hits the 400 by accident.

### 🟢 Nits

Rebased (was 8 behind). The `001_init.js` scope decision is spelled out in the PR body now — existing installs keep the `mailhog` row and still show a "configured" host that doesn't exist; repairing that needs a migration that can tell a real dev config from the seed artefact, which I'd rather not guess at here.

### Verification

- `frontendUrl.test.js` 19 → 26 cases; full `backend/src/__tests__` 87/87.
- `npm run build` clean on the frontend.
- The upgrade path is covered by unit test rather than a live install this round: `FRONTEND_URL=http://localhost:3000` + a `general_site_url` set → resolver returns the setting and `isEnvPinned()` is `false`, so the field stays editable.

New strings are `en` + `de` only, same as the rest of the PR.

the-luap:
Re-verified all five round-2 fixes at their final file:line. Everything from round 1 is closed, and closed well — `envPinnedBase()` and the `override` parameter both went further than what I asked for, in the right direction.

One thing round 2 introduced, which I've pushed a fix for rather than sending back: `4d09527`.

## Blockers — closed

| | | |
|---|---|---|
| 1 | `isEnvPinned()` demoted loopback | ✅ `frontendUrl.js:87-94` — the predicate now derives from `envPinnedBase()`, and both GET paths (`adminSettings.js:255`, `:856`) use that same helper for `general_site_url_effective`. The displayed value and the pinned-ness decision can no longer drift, which is a better fix than the one I suggested. |
| 2 | `APP_URL` / `ADMIN_URL` reachable | ✅ `options.override` sits below `FRONTEND_URL`, above the setting and the request origin, and takes the same loopback demotion as every other candidate. Confirmed at `crud.js:584` and `userManagementService.js:74`. |

## Concerns — closed

| | | |
|---|---|---|
| 1 | Protected-key round-trip | ✅ `useSettingsState.ts:338-339` — neither key is sent while pinned. Cleaner than round-tripping the stored value. |
| 2 | Wizard advancing on a failed save | ✅ `SetupConfigStep.tsx:100` gates on `validate()`; `onDone()` is out of the `finally` and runs only on `!failed`. |
| 3 | URL validation | ✅ Both layers. `PUT /general` is the only writer of the key, and the strip, the protected-key check, the validation and the cache invalidation are all in that one handler. |

## What I pushed

`saveSiteUrl()` caught and discarded every error. Defensible before round 2 — but round 2 added a server-side check that can answer 400, and the two validators disagreed:

| value | client `isAbsoluteHttpUrl` | server `validator.isURL` |
|---|---|---|
| `http://nas:3000` | accepted | accepted |
| `http://my_nas.local` | **accepted** | **rejected** |
| `http://foo_bar:3000` | **accepted** | **rejected** |

`validate()` passed those, the 400 was swallowed, `failed` stayed false, `onDone()` ran. The operator finishes the wizard believing the address is stored when nothing was — the silent misconfiguration this PR exists to remove, landing on exactly the LAN/NAS installs it targets.

Three parts, all on the branch:

- **`saveSiteUrl()` throws.** `finish()` resolves it before anything else is posted and puts the message on the address field rather than the generic warning. `Skip for now` still always leaves, by contract, but warns instead of dropping the value silently.
- **`allow_underscores` on the server check**, for the same reason `require_tld` is off — browsers resolve `http://my_nas.local` and the client accepts it, so rejecting it server-side only created the mismatch. Both validators now agree across the LAN/NAS, IDN, bare-IP and schemeless cases; I checked that as a table rather than by eye.
- **`LOOPBACK_BASE_RE` anchors its host token.** Bare prefix matching also demoted `https://localhost-nas.example.com`. It was byte-identical to the old `QR_LOCAL_BASE_RE` so this isn't yours — but that regex only gated the slideshow QR before, and it now decides "is this configured" for the entire resolver, so a false positive means a configured address is silently ignored. `127.` stays a bare prefix on purpose: all of `127.0.0.0/8` is loopback.

Resolver suite 31 passing, up from 26. Mutation-checked — restoring the unanchored regex fails the three new host-boundary cases.

## Left deliberately

- **`secureImageMiddleware.js:279`** — `getFrontendBaseUrlSync() || '*'` is the same shape as `main`'s `process.env.FRONTEND_URL || '*'`, but compose used to inject a value, so the wildcard branch was unreachable on Docker installs and now is reachable until the wizard runs. Impact is near-nil — `*` forbids credentials and the gallery token travels in a header an attacker page can't obtain — so I'd rather you accept it consciously than have me change CORS behaviour for unconfigured installs inside this PR.
- **`GeneralTab.tsx:376`** — `disabled={!!siteUrlError}` blocks every General save if an install already stored a schemeless value before this validation existed. Recoverable, the field is editable, and arguably the correct prompt. Worth knowing about.

## ✅ Verified

Directly inspected, and re-checked after the rebase since a rebase can quietly reintroduce things:

- 26 resolver tests passed before my commit, 31 after
- `emailService.js` has zero remaining references in `backend/src`
- `001_init.js` seed is still inside `if (!emailConfig && process.env.SMTP_HOST)` — existing installs untouched
- No injected default remains; `FRONTEND_URL=${FRONTEND_URL:-}` yields `''`, which `normalise()` treats as unset
- Dev mailhog survives via `SMTP_HOST=${SMTP_HOST}` in compose plus the `.env.example` note
- `settingsService.updateSettings` routes `general_*` to `/admin/settings/general`, so the wizard's write does hit the new validation
- Both `server.js` allowlists fail **closed** on an unconfigured install (`http://localhost:3005`, a port nothing uses)
- `adminSettings.logo.test.js` fails identically on clean `main` in this environment — pre-existing, matches your claim
- Backend 227 passing across the touched suites, frontend 174, build clean, lint clean (the file now has zero warnings, down from two)

Not independently verified: the i18n strings and `docs/single-container.md`, same as round 1. My own two new strings are `en` + `de` only, matching the rest of the PR.

## Verdict

Merge cleared. `bypass-size-gate` is the only red check and it is the gate doing its job — this needs a maintainer merge rather than a self-merge, which is what just happened here.

On the split you offered: not worth it now. The resolver, the wizard and the mailhog seed are individually reviewable and the history reads cleanly commit by commit.

Luca-Timo:
Screenshot for completeness:
Screenshot 2026-08-20 at 18 44 26

the-luap:
Follow-up to the review above — I pushed the second of the two "left deliberately" items after looking at it properly: `923b0c4`.

**I under-rated it.** I filed it as a nit on the grounds that it was recoverable. It isn't, for one class of admin.

`general_site_url` was free-text until this PR — `main` has no validation on it, only the permission rule at `adminSettings.js:78`, and the field has been in `GeneralTab` all along. So an upgraded install can hold something schemeless that nobody typed today. The tab flagged it on load, and `disabled={!!siteUrlError}` then killed Save for *every* General setting.

The part I missed: an admin holding `settings.edit` but not `settings.domains` can't clear it either. Correcting the address is a change to a protected key, so the write 403s — and `GeneralTab` has no permission gating at all, so nothing tells them that. They're locked out of the tab with no way back.

That's the same role `adminSettings.js:85-95` documents the no-op round-trip allowance for. The allowance only helps if the request gets made; this blocked it in the browser first.

Validation now waits until the field is actually edited, and an unchanged value is dropped from the payload rather than reposted — which is what the env-pinned case one line above already does, for the same reason.

| | |
|---|---|
| stored value invalid, untouched | Save works, key not sent |
| edited to something unusable | Save blocked |
| edited to a usable absolute url | saved |

Four tests, first coverage for this feature. Mutation-checked: removing the dirty gate fails the untouched-value case. Frontend suite 178 passing, build clean, no new type errors.

**The other item stays deliberately unfixed**, and I can now say why precisely rather than hedging: `Access-Control-Allow-Credentials` is never set in `setSecurityHeaders`, and the route it guards (`secureImages.js:110`, `/:slug/secure/:photoId/:token`) runs `secureImageAccess` — token, rate limit, fingerprint — before any header is written. So no credentialed cross-origin read works either way, and an anonymous one still needs a valid token in the path. `*` grants nothing a plain `` wouldn't. It's hygiene, not exposure.

If you want it tightened later, the clean form is to omit the header entirely when the origin is unresolved — no `Access-Control-Allow-Origin` means the browser default of same-origin, which is stricter than `*`. Worth its own issue rather than more scope here.

Verdict unchanged: merge cleared.

the-luap (approved):
Approving to unblock the merge.

Both review rounds are closed and verified at file:line — the two blockers (`isEnvPinned()` demoting loopback, `ADMIN_URL`/`APP_URL` reachable via the resolver `override`) and all three concerns. Round 2's fixes went further than what I asked for in the right places, particularly `envPinnedBase()` being shared between the predicate and the displayed value so they can't drift.

Disclosure for the record: two of the commits here are mine — `4d09527` (rejected public address surfaced instead of swallowed, validators aligned, loopback matcher anchored) and `923b0c4` (General tab no longer locks on a stored site URL nobody typed). @Luca-Timo has not reviewed those, and I'm approving a PR I contributed to, which is worth knowing when reading this approval.

Merging with an admin override of `bypass-size-gate`. That check is failing by design — it wants a maintainer decision rather than a self-merge, and this is that decision. All nine other required checks pass.

Left unfixed deliberately, tracked in the thread above: the `|| '*'` fallback in `secureImageMiddleware.js:279`. Not exposure — `Access-Control-Allow-Credentials` is never set and the route is token-gated before headers are written — so it's hygiene, and belongs in its own issue.

Issue #1095, opened 2026-08-20
[FEATURE] Responsive image delivery for mobile — DPR-aware thumbnail + preview tiers with srcset


BraynArts:
**Is your feature request related to a problem? Please describe.**

The gallery *layout* on phones is already in good shape — touch overlays in `PhotoCard`, pinch-zoom/swipe/safe-area handling in `PhotoLightbox`, per-breakpoint column counts, a mobile sidebar drawer. What is missing is **responsive image delivery**: PicPeak serves exactly the same bytes to a 375 px phone as to a 4K desktop.

Two concrete symptoms on mobile:

1. **Grid thumbnails look soft on retina phones.** Thumbnails default to 300×300 (`imageProcessor.js`, `DEFAULT_THUMBNAIL_WIDTH/HEIGHT = 300`, `withoutEnlargement: true`). On a 375 px viewport with the default `gridColumns.mobile = 2` (mosaic caps at 2, masonry forces 2 below 640 px) a tile is roughly 175–180 CSS px, i.e. ~530 device px on a DPR-3 phone. The 300 px source is upscaled ~1.8×, so faces and fine detail visibly mush.

2. **The lightbox is expensive on cellular.** The preview tier is a single 1920 px long-edge / q85 JPEG (`DEFAULT_PREVIEW_LONG_EDGE = 1920`), served by `GET /api/gallery/:slug/preview/:photoId` with no size parameter. A phone can display at most ~1170 px wide, so every swipe pulls roughly twice the bytes it can use — and the lightbox prefetches neighbours, so a guest flicking through 50 photos at a wedding burns tens of MB of mobile data for pixels that never reach the screen. If `preview_url` is null the fallback is the untouched original, which on external-source installs can be a 10–20 MB file.

There is no `srcset`/`sizes` anywhere in `frontend/src` (zero matches), and no `navigator.connection` / `Save-Data` awareness.

**Describe the solution you'd like**

DPR- and viewport-aware image tiers, so phones fetch a picture matched to the screen they actually have:

- A width parameter on the existing derivative routes, e.g. `GET /api/gallery/:slug/preview/:photoId?w=1280` and `GET /api/gallery/:slug/thumbnail/:photoId?w=600`, restricted to a small whitelist of widths (thumbs 300/600/900; previews 640/1280/1920).
- `srcset` + `sizes` on the gallery `<img>` elements so the browser picks the right tier per device and per layout column count.
[...]

PR #1109, opened 2026-08-21
feat(gallery): responsive grid thumbnails (#1095)

the-luap:
Completes #1095. #1099 shipped the lightbox half and deliberately held this back; this is the rest.

## The symptom

Grid tiles are ~175 CSS px at the mobile 2-column default — about **530 device px** on a DPR-3 phone. The 300px thumbnail is upscaled ~1.8×, so faces mush. That was symptom 1 of the issue and the only part still outstanding.

## Backend

Mirrors the preview tiers exactly: `?w=` on the gallery thumbnail route, whitelisted to **300/600/900**, cached by width in storage, never written to `photos.thumbnail_path`.

Keys are scoped by photo id for every source type from the start — basenames are not unique across events, and a tier is served from a cache hit *without re-reading the source*, which is how the preview tiers nearly leaked one gallery's photo into another. The tier is in the ETag, or a client holding the 300px file gets a `304` for its 600px request. Cleanup (delete / bulk-delete / archive) and regenerate invalidation are wired the same way.

`generateThumbnail` now takes width/height overrides. It **keeps the configured `fit`** — the grid renders with `object-cover`, so tiers framed differently would visibly jump as the viewport changes.

## The srcset only promises what the file delivers

This is why it was held back from #1099, and the guard is the substance of this PR.

Thumbnails are generated `withoutEnlargement`, so a 400px original asked for 900 comes back at **400**. Advertising `900w` for it would have the browser pick that candidate and upscale — the reported softness made *worse*. The photo's own dimensions are now the guard, measured on the **short edge**, because thumbnails are square and a 4000×600 panorama can still only fill a 600 tile. A source that clears only one tier gets **no srcset at all** rather than a single pointless candidate.

## Two things this surfaced

**`npx tsc --noEmit` type-checks nothing in this project.** The root `tsconfig.json` is `"files": []` with project references, so it resolves zero files; the real command is `tsc -b`, which is what `build:check` runs. I had been reporting "tsc clean" on that basis in earlier PRs — it was vacuous, and I'd rather say so than let it stand.

**Under `tsc -b` the repo has 43 files with pre-existing type errors.** This branch adds none — same 43 on `main` and here, no new files. The one error in a file I touched (`PeopleManagerModal.tsx:91`) is on `main` already and unrelated to the line I changed (`:99`). Worth its own cleanup, not this PR.

## Verification

- Backend **16 tier tests** (up from 11): key scoping, per-width generation, canonical width resolving to the normal thumbnail, cleanup, and that a small source is *not* upscaled — the fact the srcset guard exists to respect.
- Frontend **18 tier tests**: the guard is exercised for a source that fills nothing, one tier, two tiers, and all three, plus the panorama short-edge case.
- Mutation-checked: removing the short-edge guard fails 5 tests; removing the key scoping fails the backend collision test.
- Full frontend suite 179 passed, build green; backend tier + face suites 85 passed.

### DIFF
--- a/backend/__tests__/integration/previewTiers.test.js
+++ b/backend/__tests__/integration/previewTiers.test.js
@@ -188,4 +188,69 @@ describe('preview tiers (#1095)', () => {
       expect(meta.height).toBe(Math.round(640 * (2000 / 3000)));
     });
   });
+
+  describe('thumbnail tiers', () => {
+    async function seedThumbPhoto(w = 3000, h = 2000) {
+      const [e] = await db('events').insert({
+        slug: `tt-${Math.random().toString(36).slice(2, 8)}`,
+        event_type: 'wedding', event_name: 'tt', event_date: '2026-01-01',
+        host_email: 'h@example.com', admin_email: 'a@example.com',
+        password_hash: 'x', share_link: `tt-${Math.random()}`,
+        expires_at: new Date().toISOString(),
+      }).returning('id');
+      const eventId = typeof e === 'object' ? e.id : e;
+      const rel = `events/active/tt/${Math.random().toString(36).slice(2, 8)}.jpg`;
+      const abs = path.join(process.env.STORAGE_PATH, rel);
+      await fs.promises.mkdir(path.dirname(abs), { recursive: true });
+      await sharp({ create: { width: w, height: h, channels: 3, background: { r: 5, g: 5, b: 5 } } })
+        .jpeg().toFile(abs);
+      const [p2] = await db('photos').insert({
+        event_id: eventId, filename: path.basename(rel),
+        path: rel.replace(/^events\/active\//, ''), type: 'individual',
+        width: w, height: h, processing_status: 'complete', source_origin: 'managed',
+      }).returning('id');
+      return db('photos').where({ id: typeof p2 === 'object' ? p2.id : p2 }).first();
+    }
+
+    it('scopes thumbnail tier keys by photo id', async () => {
+      // Same cross-gallery hazard the preview tiers had: a cache hit serves
+      // without re-reading the source, so a shared basename leaks across events.
+      const keys = imageProcessor.thumbnailTierKeys({ id: 42, path: 'a/IMG_0001.jpg', source_origin: 'managed' });
+      expect(keys.every((k) => k.includes('p42_'))).toBe(true);
+      expect(keys.some((k) => k.includes('w300'))).toBe(false); // canonical excluded
+    });
+
+    it('generates a tier at the requested size', async () => {
+      const photo = await seedThumbPhoto();
+      const key = await imageProcessor.ensureThumbnailAtWidth(photo, 600);
+      expect(key).toContain('thumb_w600_');
+      const meta = await sharp(path.join(process.env.STORAGE_PATH, key)).metadata();
+      expect(Math.max(meta.width, meta.height)).toBe(600);
+    });
+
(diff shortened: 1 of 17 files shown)


the-luap:
## Screenshots

Captured against this branch running the real stack — backend on the PR's `src`, gallery of 62 photos (750px short edge, so both the 600 and 900 tiers are reachable). Layout is Masonry, deliberately: it's one of the three that passes no `lazy` prop, so before the follow-up commit it got no measurement at all.

### The change, at 1:1 device pixels

issue-1095-crop-comparison


Tile renders at **171 CSS px → 513 device px** on a DPR-3 phone. That's the case #1095 opens with: the 300px thumbnail upscaled 1.7×.

| | thumbnail requests | per tile | total |
|---|---|---|---|
| `main` | 204 × canonical (300px) | 11 KB | 2.3 MB |
| this PR | 204 × `?w=600` | 27 KB | 5.4 MB |

Bytes go up, and that is the trade being made — the issue is that phones were being served an image they had to upscale, not that they were being served too much. Desktop is where the saving sits: tiles there measure ~292 CSS px at DPR 1 and stay on the canonical file.

### Full page, mobile 390×844 @ DPR 3

Before (`main`) / after (this PR) — same scroll position, same viewport:

issue-1095-mobile-before
issue-1095-mobile-after


### Desktop 1440×900 @ DPR 1

issue-1095-desktop-after

the-luap:
## Screenshots

Captured against this branch on the real stack — backend running the PR's `src`, gallery of 62 photos at 750px short edge, so both the 600 and 900 tiers are reachable. Layout is Masonry columns, deliberately: it's the mode that exercised the measurement path hardest.

All figures below come from **production builds with the HTTP cache disabled**, and every request is attributed to a photo id. Dev-server numbers are not comparable — StrictMode double-invokes effects and inflates request counts 2–3×.

### The change, at 1:1 device pixels

issue-1095-crop-comparison


Tile renders at **171 CSS px → 513 device px** on a DPR-3 phone. That is the case the issue opens with: the 300px thumbnail upscaled 1.7×.

| mobile, 390×844 @ DPR 3 | tier served | requests | bytes |
|---|---|---|---|
| `main` | canonical 300px | 102 | 296 KB |
| this PR | `?w=600` | 62 | 872 KB |

| desktop, 1440×900 @ DPR 1 | tier served | requests | bytes |
|---|---|---|---|
| `main` | canonical 300px | 107 | 433 KB |
| this PR | canonical 300px | 62 | 293 KB |

Two things to read out of that.

**Mobile costs bytes.** ~296 KB → ~872 KB for 62 tiles. That is the trade being made and it should be stated plainly: the issue is that phones were served an image they had to upscale, not that they were served too much. Desktop tiles measure 292 CSS px at DPR 1 and correctly stay on the canonical file.

**Request counts drop on both.** 102 → 62 and 107 → 62. Masonry columns mode was mounting cards into an unmeasured 3-column layout and remounting them when it settled; the last commit holds the cards back until the container is measured, the same gate the rows mode in that file already applied. Without tiering that remount was harmless — both mounts requested the same url and the cache absorbed it — so this only became a real double download once tiers existed.

### Full page, mobile 390×844 @ DPR 3

Before (`main`) / after (this PR), same scroll position:

issue-1095-mobile-before
issue-1095-mobile-after

### Desktop 1440×900 @ DPR 1

issue-1095-desktop-after

62 photos, 62 requests, canonical tier throughout — desktop pulls the small file, which is the half of #1095 that saves rather than spends.

PR #1113, opened 2026-08-21
fix(ui): stop iOS Safari zooming in on 14px form fields

### LINKED ISSUE #1105: [BUG] iOS Safari zooms in on the 14px gallery password field and stays zoomed after unlock
**Describe the bug**

On iOS Safari, entering the gallery password leaves the whole gallery zoomed in. The page can then be panned left/right, the header action icons are partly off-screen, and nothing lines up until the visitor manually pinch-zooms back out.

The cause is not a layout overflow. It is iOS Safari's automatic zoom on focused form controls whose computed `font-size` is below 16px:

- `frontend/src/pages/GalleryPage.tsx` sets `className="text-sm sm:text-base"` on the password `<Input>`. The breakpoint is inverted: `text-sm` (14px) applies **below** 640px — i.e. on every phone — and 16px only kicks in from tablet width upwards, where iOS never zooms anyway.
- The same holds globally: `.input` and `.input-themed` in `frontend/src/index.css` both `@apply ... text-sm`, so effectively every input in the app is 14px on mobile.
- Because unlocking a gallery is a client-side React transition and not a document navigation, the zoom level Safari applied on the password screen carries straight over into the gallery. A real page load would have reset it.
- The viewport meta (`width=device-width, initial-scale=1.0, viewport-fit=cover`) correctly does **not** set `maximum-scale=1`. Suppressing the zoom that way would be the wrong fix — it would break pinch-to-zoom for everyone. The font size is the right lever.

**To Reproduce**

1. Open a password-protected gallery on an iPhone in Safari.
2. Tap the "Gallery password" field — Safari zooms the page in.
3. Submit the password.
4. The gallery renders zoomed: the page pans horizontally and the header actions (favourites / download / logout) are cut off at the right edge.
5. Pinch-zoom out manually — the layout is correct.

**Expected behavior**

Entering the gallery password should not change the zoom level. The gallery should render at 1x with no horizontal panning.

**What I verified**

- Live production build (`assets/index-*.css`): `.input{ ... font-size:.875rem; ... }`, while `.sm\:text-base{font-size:1rem}` sits inside `@media(min-width:640px)`.
- Chrome at a 614px viewport: `getComputedStyle(passwordInput).fontSize === "14px"`. At 1512px the same element reports `"16px"`.
- No layout overflow anywhere: in the unlocked gallery at 604px client width with 79 photos rendered, `document.documentElement.scrollWidth === 604`, and a sweep over every element in the DOM found none whose bounding box extends past the viewport. The horizontal panning users report is purely the residual zoom.

**Environment**

- OS / Browser: iPhone, Safari (reproduced against a live instance; font-size behaviour confirmed in desktop Chrome at mobile widths)
- PicPeak Version: 3.108.1-beta.0
- Deployment Method: Docker Compose (production/GHCR images)
- Database: PostgreSQL

**Possible Solution**

Make form controls 16px on mobile and keep 14px from `sm` upwards — the inverse of what is there today.

`frontend/src/index.css`:

```diff
   .input {
-    @apply flex h-10 w-full rounded-lg border px-3 py-2 text-sm file:border-0 ...;
+    @apply flex h-10 w-full rounded-lg border px-3 py-2 text-base sm:text-sm file:border-0 ...;
   }
   .input-themed {
-    @apply flex h-10 w-full rounded-lg border px-3 py-2 text-sm file:border-0 ...;
+    @apply flex h-10 w-full rounded-lg border px-3 py-2 text-base sm:text-sm file:border-0 ...;
   }
```

`frontend/src/pages/GalleryPage.tsx` — drop the inverted override on the password input:

```diff
-                  className="text-sm sm:text-base"
+                  className="text-base sm:text-sm"
```

The same inverted pattern appears on other inputs, e.g. `frontend/src/components/gallery/UserPhotoUpload.tsx`.

If a single global guard is preferred over touching each call site:

```css
@media (max-width: 639px) {
  input:not([type="checkbox"]):not([type="radio"]),
  select,
  textarea {
    font-size: 16px;
  }
}
```

Worth noting the payoff is wider than the gallery gate: `.input` backs the admin login, the guest-name prompt, comments and every other form, so all of them currently trigger the same iOS zoom.

the-luap:
Closes #1105.

@BraynArts's report is accurate on every point — I verified each against the tree before writing this. The breakpoint really is inverted at `GalleryPage.tsx:549`, `.input`/`.input-themed` really are `text-sm` (`index.css:209`, `:219`), the viewport meta really does omit `maximum-scale` (so font size is the right lever, not the meta), and `GalleryPage.tsx:421` confirms the unlock is a conditional render in the same component — no document navigation, which is exactly why the zoom survives it.

## Why one media query instead of flipping the call sites

The issue's preferred fix — flip `.input` to `text-base sm:text-sm` and correct the call sites — fixes the `<Input>` path but leaves most of the app:

- `.input` backs **334** `<Input>` usages
- there are also **240** raw `<input>`, **163** `<select>` and **39** `<textarea>` that never pass through it
- Tailwind utilities sit in a later layer than `@layer components`, so any call site adding `text-sm` overrides the fix — and any *new* one silently reintroduces the bug

A single guard covers all of them and is self-enforcing.

## The `:not()` is load-bearing

The issue offers a global-guard variant as an alternative. It has a specificity hole, which I measured in a browser rather than reasoning about:

| selector | computed | why |
|---|---|---|
| `input.text-sm` | **16px** | `input:not([type=…]):not([type=…])` is (0,2,1), beats `.text-sm` (0,1,0) |
| `select.text-sm` | **14px** | bare `select` is (0,0,1) — loses |
| `textarea.text-sm` | **14px** | same |

**24 selects and textareas** in the tree carry `text-sm`, so the bare form would have left them zooming. Adding `:not([hidden])` lifts them to (0,1,1) and all three resolve to 16px. Checkbox and radio stay excluded so font-size never sizes their box.

## Verified

Production build, real pages, measured computed `font-size`:

| | mobile 390×844 @DPR3 | desktop 1280×800 @DPR1 |
|---|---|---|
| gallery password | 14px → **16px** | 16px → 16px |
| admin login | 14px → **16px** | 14px → 14px |

Desktop density is deliberately untouched — the rule only applies below `sm`, and iOS never zooms at those widths anyway.

Frontend suite 174 passing, build clean, no new type errors (43 pre-existing, same as `main`).

## Screenshot

Gallery password field at 390×844 DPR3, before and after:

issue-1105-comparison

One honest cosmetic note visible in that shot: the German placeholder truncates one character earlier at 16px (`…Passwort ei` rather than `…Passwort ein`). It's a placeholder rather than content, and the trade against a page-wide zoom that never resets is clearly worth it — but it's there, so I'm pointing at it rather than letting it be found.

## Stable

Stable carries the identical bug (`index.css:209`, `GalleryPage.tsx:446` on `origin/stable`, v3.46.1), and stable is what most self-hosters run — so this needs a twin. Filing it separately and verifying on the stable branch itself rather than assuming this branch's results transfer.

### DIFF
--- a/frontend/src/index.css
+++ b/frontend/src/index.css
@@ -716,3 +716,56 @@
     margin-top: 0;
   }
 }
+
+/*
+ * iOS Safari zooms the whole page in when a focused form control computes to
+ * less than 16px, and it does not zoom back out (#1105). Unlocking a gallery
+ * is a client-side transition rather than a document navigation, so the zoom
+ * the password field triggered carries straight into the gallery: the layout
+ * pans horizontally and the header actions sit off-screen until the visitor
+ * pinch-zooms out by hand.
+ *
+ * The lever is the font size, not the viewport meta — adding maximum-scale=1
+ * would suppress the zoom by disabling pinch-to-zoom for everyone, which is an
+ * accessibility regression, so index.html deliberately omits it.
+ *
+ * Keyed to the POINTER, not a width. The zoom depends on the computed font
+ * size and a touch device, never on how wide the viewport is — and a phone in
+ * landscape is 667–956 CSS px, above any width you could call "phone". A
+ * max-width query fixes portrait and leaves every landscape phone (and iPad)
+ * still zooming. `pointer: coarse` is the population that actually has the
+ * behaviour; a mouse-driven desktop reports `fine` and keeps its 14px density.
+ *
+ * Deliberately NOT inside @layer, and deliberately more specific than a single
+ * utility class: `.input` is 14px and ~440 raw controls carry their own
+ * `text-sm`, so a rule that loses to a utility fixes almost nothing. The
+ * `:not()` on each selector is what buys that specificity — without it,
+ * `select`/`textarea` (0,0,1) lose to `.text-sm` (0,1,0) and keep zooming,
+ * while `input` alone happens to win. Excluding checkbox and radio keeps
+ * font-size off controls that size their box from it.
+ *
+ * max(16px, 1em, 1rem) is a FLOOR, not a size. Writing a flat 16px would make
+ * controls that are already larger smaller: Typography -> Large sets
+ * --font-size-base to 18px on body, so anything inheriting it would be clamped
+ * down and the setting quietly ignored. Each term covers a case the others
+ * miss - 1em follows the theme's body size, 1rem follows a browser default the
+ * visitor raised themselves, 16px catches Small themes and .text-sm controls:
+ *
+ *   normal (body 16)      16px      Large theme (body 18)   18px
+ *   Small theme (body 14) 16px      browser default 20px    20px
+ *
+ * The specificity that beats a utility class also beats a gallery's custom CSS
+ * (Theme -> Custom CSS), so `.input-themed { font-size: 20px }` lands at 16px
+ * on touch. That is unavoidable here rather than an oversight: nothing in CSS
+ * distinguishes a class that sets 14px from one that sets 20px, so a rule that
+ * loses to the second also loses to the first and fixes nothing. Overriding
+ * DOWNWARD is the point; upward is the cost. `font-size: 20px !important`
+ * still wins for anyone who wants it.
+ */
+@media (pointer: coarse) {
+  input:not([type="checkbox"]):not([type="radio"]),
+  select:not([hidden]),
+  textarea:not([hidden]) {
+    font-size: max(16px, 1em, 1rem);
+  }
+}


the-luap:
Reworked after an external review pass. Two real problems in what I first pushed — both my own, both found by measuring rather than reading.

**1. It only fixed portrait.** `max-width: 639px` misses every landscape phone (667–956 CSS px) and every iPad. Safari's focus zoom keys off the computed font size on a touch device, never off viewport width. Measured on `/admin/login`, which has no `sm:` override:

| | `main` | first push | now |
|---|---|---|---|
| portrait 390×844 | 14px zooms | 16px | 16px |
| landscape 844×390 | 14px zooms | **14px zooms** | 16px |
| landscape 667×375 | 14px zooms | **14px zooms** | 16px |
| iPad 820×1180 | 14px zooms | **14px zooms** | 16px |
| desktop (mouse) | 14px | 14px | 14px unchanged |

Now keyed to `pointer: coarse` — the population that actually has the behaviour. Desktop reports `fine` and keeps its 14px density.

**2. A flat `16px` was a size, not a floor.** Typography → Large sets `--font-size-base: 18px` on `body`, so anything inheriting it got *clamped down* to 16px and the setting silently ignored. Same shape for a visitor who raised their browser's own default. `max(16px, 1em, 1rem)` — each term covers a case the others miss:

| | result |
|---|---|
| normal (body 16) | 16px |
| Large theme (body 18) | 18px |
| Small theme (body 14) | 16px |
| browser default 20px | 20px |

**Rejected, with a reason:** the review also flagged that this outranks a gallery's Custom CSS, so `.input-themed { font-size: 20px }` lands at 16px on touch. Confirmed — single-class, element+class and even two-class custom rules all lose; only `!important` wins. It's unavoidable rather than an oversight: nothing in CSS distinguishes a class that sets 14px from one that sets 20px, so a rule that loses to the second also loses to the first and fixes nothing. Overriding downward is the point; upward is the cost. Documented in the rule's comment along with the `!important` escape hatch.

Rebased onto `main` (it had moved on since #1104 merged). Frontend suite 178 passing, build clean, no new type errors. Stable twin #1114 carries the byte-identical rule.

PR #1117, opened 2026-08-21
fix(ui): force Android to show camera option on guest uploads

Zszywany:
## Description

Currently Android users cannot select the Camera input on Guest Upload modal. This change forces Android to show the legacy picker with camera option. No impact on iOS or desktop users.

This is a workaround for Androids sub-optimal behavior.

## Type of change

Please delete options that are not relevant.

## How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce.

- [x] Tested on production-like environment

**Test Configuration**:
* PicPeak Version: 3.46.1 and 3.108.1-beta.0
* Node.js Version: 18.19.1
* Database: PostgreSQL 15.19
* Browser: Firefox 154.0, Chromium 151.0.7922.137, Chrome 151.0.7922.169 (Android 16), Orion 1.5.1 (Webkit 8624.4.5.10.5, iOS26.6)

## Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my code
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works

## Additional Notes:

This is a workaround, might need reverting in the future.

### DIFF
--- a/frontend/src/components/gallery/UserPhotoUpload.tsx
+++ b/frontend/src/components/gallery/UserPhotoUpload.tsx
@@ -56,10 +56,7 @@ export const UserPhotoUpload: React.FC<UserPhotoUploadProps> = ({
     [publicSettings?.allowed_file_types]
   );
 
-  //const acceptString = useMemo(
-  //  () => extensionsToAcceptString(publicSettings?.allowed_file_types),
-  //  [publicSettings?.allowed_file_types]
-  //);
+  // Force newer Android phones to fallback to file picker with camera option
   const acceptString = useMemo(() => {
     const baseAccept = extensionsToAcceptString(publicSettings?.allowed_file_types);
     // Append .pdf to force Android's system chooser (showing the Camera option)
--- a/frontend/src/components/gallery/UserPhotoUpload.tsx
+++ b/frontend/src/components/gallery/UserPhotoUpload.tsx
@@ -56,10 +56,16 @@ export const UserPhotoUpload: React.FC<UserPhotoUploadProps> = ({
     [publicSettings?.allowed_file_types]
   );
 
-  const acceptString = useMemo(
-    () => extensionsToAcceptString(publicSettings?.allowed_file_types),
-    [publicSettings?.allowed_file_types]
-  );
+  //const acceptString = useMemo(
+  //  () => extensionsToAcceptString(publicSettings?.allowed_file_types),
+  //  [publicSettings?.allowed_file_types]
+  //);
+  const acceptString = useMemo(() => {
+    const baseAccept = extensionsToAcceptString(publicSettings?.allowed_file_types);
+    // Append .pdf to force Android's system chooser (showing the Camera option)
+    // instead of defaulting to the restricted photo picker
+    return baseAccept ? `${baseAccept}, .pdf` : 'image/*, .pdf';
+  }, [publicSettings?.allowed_file_types]);
 
   // #821 — the requirements hint used to hardcode "JPEG, PNG or WebP"; render
   // the actually-configured formats so it never contradicts what's accepted.
@@ -71,6 +77,11 @@ export const UserPhotoUpload: React.FC<UserPhotoUploadProps> = ({
   // Shared filter pipeline for both <input> change and drag-and-drop (#504).
   const addFiles = (incoming: File[]) => {
     const validFiles = incoming.filter((file) => {
+      // Reject non-image files explicitly (e.g. if selected via the .pdf chooser workaround)
+      if (!file.type.startsWith('image/')) {
+        toast.error(`Please select an image file: ${file.name}`);
+        return false;
+      }
       if (!allowedMimeTypes.includes(file.type)) {
         toast.error(`Invalid file type: ${file.name}`);
         return false;


the-luap:
Thanks for chasing this one down — the Android photo-picker behaviour is real and the accept-attribute trick is the accepted way around it. One blocker in the second hunk, and a couple of things worth tightening in the first.

## 🔴 Blockers

### 1. The image-only guard breaks video uploads

`frontend/src/components/gallery/UserPhotoUpload.tsx:78-80` rejects every non-image *before* the existing allowlist check runs:

```js
if (!file.type.startsWith('image/')) { toast.error(...); return false; }
if (!allowedMimeTypes.includes(file.type)) { ... }
```

`frontend/src/utils/fileTypes.ts:10-14` maps `mp4`, `m4v`, `webm`, `mov` and `avi`, and `general_allowed_file_types` is a free-text admin setting (`GeneralTab.tsx:200`). On any install configured for video, `acceptString` still **offers** videos in the chooser and this guard then refuses them with "Please select an image file" — an enabled feature broken in the guest path.

The guard also isn't needed for what it's there for. A selected PDF has `file.type === 'application/pdf'`, which can never be in `allowedMimeTypes` — `extensionsToMimeTypes` only ever emits MIME types it has a mapping for. So the pre-existing check at `:75` already rejects the synthetic PDF with "Invalid file type", and the whole second hunk can go:

```diff
-      // Reject non-image files explicitly (e.g. if selected via the .pdf chooser workaround)
-      if (!file.type.startsWith('image/')) {
-        toast.error(`Please select an image file: ${file.name}`);
-        return false;
-      }
```

## 🟡 Concerns

1. **"No impact on iOS or desktop users" isn't quite right.** The `.pdf` is appended unconditionally — the diff mentions Android only in comments, there's no actual gate. `acceptString` feeds the single `<input>` at `:252`, so desktop and iOS pickers now present PDFs as selectable and picking one just produces an error. Gating it makes the statement true and keeps the workaround where it belongs:

   ```js
   const isAndroid = /Android/i.test(navigator.userAgent);
   // ...
   return isAndroid ? `${baseAccept},.pdf` : baseAccept;
   ```

2. **The empty-string fallback widens the allowlist.** `baseAccept ? ... : 'image/*, .pdf'` — `extensionsToMimeTypes` already falls back to `DEFAULT_ALLOWED`, so the branch looks unreachable, but if it ever fires `image/*` is broader than the configured list. Worth dropping rather than carrying a second, looser default.

3. **The new toast isn't translated.** The file has `useTranslation` and uses `t('upload.fileTooLarge')` two lines below. The existing `:75` string is hardcoded too, so this is a pre-existing wart rather than something you introduced — but if the guard survives in any form, it should use `t()`.

## 🟢 Nits

- A test in `frontend/src/utils/__tests__/` — or an `addFiles` filter test — would have caught the video case. The checklist notes none were added.
- Branch is 2 commits behind `main` (based on `7583b4f6`; `main` is at `ba221ba2`); it'll need a rebase before merge.
- `'image/*, .pdf'` carries a space the existing joiner never produces. Cosmetic.

## Checklist item

This touches a UI surface, so it needs a screenshot of the Android chooser before/after attached to the PR — that's a required item here rather than a nit, and the Screenshots section is currently empty. It also happens to be the one thing in this PR nobody else can verify (see below).

## ✅ Verified

Directly inspected rather than taken from the description:

- `EXTENSION_TO_MIME` genuinely contains video types and the setting is admin-editable, so the regression is reachable rather than theoretical
- `application/pdf` can never appear in `allowedMimeTypes`, so `:75` already covers the PDF case on its own
- no Android gate exists anywhere in the diff
- the trailing-newline fix at EOF is harmless

**Not verified:** that the `.pdf` trick actually produces the camera option on the Android versions you list. I have no Android device to hand and took that at description level — it's the premise of the whole change, which is the other reason the screenshot matters.

Worth knowing: CI hasn't really run on this yet. Only GitGuardian, the two size gates and `lint-pr-title` have executed — the build and test jobs need a maintainer approval for a first-time contributor, so nothing has compiled or tested the change so far.

## Verdict

Drop the second hunk, gate the `.pdf` on Android, and add the screenshot, and I'm comfortable merging.

the-luap:
Picked this up — the Android photo-picker behaviour is real and the accept-attribute trick is the right shape. Continued in #1244 rather than pushed here, because your PR branch is your fork's `main` and force-pushing someone's default branch isn't something I'll do. You're credited as co-author.

Two changes from the review:

**The `.pdf` is gated on the Android UA.** It was appended unconditionally — there's no gate anywhere in the diff — so "no impact on iOS or desktop users" wasn't quite true: those pickers gained a selectable PDF that only errors when chosen.

**The image-only guard is dropped.** It rejected every non-image *before* the allowlist check, and `fileTypes.ts` maps `mp4`/`m4v`/`webm`/`mov`/`avi` with `general_allowed_file_types` admin-editable — so on any video-enabled install, `acceptString` still offered videos and this guard then refused them with "Please select an image file". It was also unnecessary: `extensionsToMimeTypes` only emits types it has a mapping for, so `application/pdf` can never be in `allowedMimeTypes` and the existing "Invalid file type" check already covers a picked PDF.

The `'image/*, .pdf'` empty-string fallback went too — `extensionsToMimeTypes` already falls back to the configured default, and `image/*` was broader than the admin's allowlist.

It lives in `fileTypes.ts` as a pure function now, so the UA behaviour is testable — four cases, including the video one that would have caught the regression. Stable twin is #1248; you reported on 3.46.1, so that's the branch you'd actually get it on.

**One thing still open, and it's your part:** I have no Android device, so the premise — that the `.pdf` actually surfaces the camera option on Android 16 — is still taken at description level. That's also why the screenshot matters, and it's the one thing nobody here can produce. If you can attach a before/after of the chooser on #1244, both PRs can merge.

Closing this once #1244 merges — which is gated on that screenshot.

the-luap:
Landed as #1244 (main) and #1248 (stable), with you as co-author. Closing this one — merging it too would double-apply the change.

Three things differ from what's here, all from the review above plus a later round:

**The token is `android/allowCamera`, not `.pdf`.** Same trick, same mechanism — but `.pdf` advertises PDFs as selectable, so a guest can pick one and get "Invalid file type". `android/allowCamera` matches no file, so it flips the picker without offering anything.

**Gated to Android minus Firefox.** Yours appended unconditionally; iOS and desktop pickers work fine and only gained a selectable PDF. Firefox for Android matches `/Android/` too but opens a chooser that already offers the camera.

**No image-only guard.** That hunk rejected every non-image before the allowlist check, which breaks video uploads on any install configured for them — `fileTypes.ts` maps `mp4`/`m4v`/`webm`/`mov`/`avi` and the setting is admin-editable. It was also redundant: `application/pdf` can never reach `allowedMimeTypes`, so the existing check already covered the case it was written for.

Your diagnosis was the hard part and it was right. I confirmed the mechanism independently — [addpipe documented it](https://blog.addpipe.com/html-file-input-accept-video-camera-option-is-missing-android-14-15/) for Chrome/Edge on Android 14/15 — and captured what's capturable without a handset:

accept comparison

One thing still unconfirmed, and it's yours if you're willing: that source says **Android 14/15**, you reported it on **Android 16**. If you can open a gallery's upload button on your Pixel and say whether the camera option now appears, that closes the last gap. If it turns out Android 16 was never affected, the token is inert there and costs nothing — but I'd rather know.

Thanks for chasing it down.

PR #1121, opened 2026-08-21
chore(stable): release 3.46.2

the-luap:
:robot: I have created a release *beep* *boop*
---

### Bug Fixes

* **ui:** stop iOS Safari zooming in on 14px form fields ([#1114](https://github.com/PicPeak/picpeak/issues/1114)) ([32db1c8](https://github.com/PicPeak/picpeak/commit/32db1c8052d324b09462a17859c7adb5ccfe56e3))

---
This PR was generated with [Release Please](https://github.com/googleapis/release-please). See [documentation](https://github.com/googleapis/release-please#release-please).

PR #1124, opened 2026-08-21
feat(deploy): make the all-in-one image installable without a shell, and repair the onboarding paths

Luca-Timo:
Found while auditing #705's acceptance criteria. Everything here is on a path we
tell a new user to walk, and every one of them fails quietly rather than loudly.

The largest is that **the all-in-one image cannot be installed from a GUI at
all** — which is the deployment it exists for.

## The blocker: `JWT_SECRET` needs a shell

`validateEnv.js:10` marks `JWT_SECRET` critical and exits when it is unset. The
documented run command supplies it with `-e JWT_SECRET="$(openssl rand -base64
48)"` — a shell command. A Synology Container Manager or QNAP Container Station
form has no shell, so the user either pastes nothing and the container exits on
boot, or invents a short string by hand.

The compose stack already solved this: `secrets-init` generates the machine
secrets once into a volume and never overwrites them. This gives the same
guarantee to a deployment that mounts no secrets volume, by extending the
`/run/secrets` hydration already in `wait-for-db.sh` rather than adding a second
mechanism next to it.

Precedence is unchanged, and explicit env still wins:

```
environment  ->  /run/secrets/jwt_secret  ->  db/jwt.secret (generated if absent)
```

`openssl` is not in `node:22-alpine`; `/dev/urandom` + `base64` are. 48 bytes →
64 base64 characters, comfortably past validateEnv's 32-character floor.

**The write is load-bearing.** `JWT_SECRET` is exported only when the file
actually persisted. Exporting an unpersisted secret would mint a new one on every
restart and sign every session out each time — worse than failing validation,
which at least says what is wrong. An unwritable data directory now prints what
to do about it.

The read-back is length-checked and whitespace-trimmed rather than a bare
`[ -s ]`: a short write (ENOSPC, SIGKILL mid-boot) would otherwise leave a
partial file that every later boot accepts, and validateEnv only *warns* below
32 characters, so the install would sign with a truncated key indefinitely.

Creation writes the full content to a private temp file and then **hard-links**
it into place. Two properties are needed at once and each idiom alone gives only
one: the file must never be readable half-written (a plain exclusive-create
redirect makes the inode first and writes after, so a concurrent boot can read
it in between and end up with nothing — measured, 5 of 8 boots), and the create
must not clobber (which is what `mv -f` does, leaving two containers each
signing with their own key). The link is atomic, fails with `EEXIST` when
another container won, and the loser adopts the winner's value by re-reading.
Measured: 12 concurrent boots over 5 rounds converge on exactly one secret,
none empty.

`JWT_SECRET` only, deliberately: `DB_PASSWORD` has to match whatever the Postgres
volume was initialised with, so generating one at boot would break the connection
rather than fix it, and the AIO default engine is SQLite with Redis unused.

## Decisions worth stating

**`scripts/install.sh` is deprecated in place, not deleted.** It refers to two
files that do not exist (`docker-compose.prod.yml`, `scripts/setup-ssl.sh`), and
its `sed` calls now match the *commented* lines in the current `.env.example`, so
they emit `#JWT_SECRET=<random>` and set nothing. Nothing in the repo references
it. It now prints a signpost to `picpeak-setup.sh` and exits non-zero. Deleting it
outright is equally fine — the body below the banner is unreachable — I left it as
a stub so an old bookmark gets a pointer instead of a broken install. Say the word
and I will drop the file.

**`CONTRIBUTING.md` was pointed at what works rather than resurrecting a deleted
file.** It instructed `docker compose -f docker-compose.dev.yml up`; that file is
in `.gitignore` (`:128`) and absent from the tree, so following it verbatim gives
"no such file". It was committed historically and later removed, so I assumed the
removal was deliberate and documented the loop that works today — compose for
Postgres and Redis, `npm run dev` on the host for both apps — with the gitignored
dev compose described as the local-only thing it is. If you would rather ship a
`docker-compose.dev.yml` again, that is a different change and I am happy to do it.

**Two false statements in `docs/single-container.md` are corrected**, both of
which send a NAS user somewhere that does not exist:

- It said the setup token "is also printed to the container log on first start".
  `setupService.js:96` deliberately does **not** do that — logging it
  unconditionally wrote a live first-admin credential into `combined.log` and
  `security.log`, both on the mounted volume and never rotated out. The log line
  names the file instead. The doc now says where the file is, which is the answer
  that fits a GUI deployment anyway: it sits on the volume, so any file manager
  opens it.
- It told NAS users to pull `aio:stable`, in three places including the primary
  `docker run`. That tag does not exist — the image publishes `:main`, `:beta`,
  and version tags. #1098 corrected the README and missed the doc, which is the
  single artifact carrying the whole NAS story. Now `:main`, with a note on why so
  it does not get "corrected" back.

## What changed

**Backend**
- `backend/wait-for-db.sh` — generate and persist `JWT_SECRET` when neither the
  environment nor `/run/secrets` supplies one.

**Compose**
- `docker-compose.production.yml` — defaults on the three bind mounts
  (`${APP_STORAGE:-./storage}` and friends). Without them `docker compose config`
  aborts with `invalid spec: :/app/storage: empty section between colons`, which
  reads as a broken compose file rather than a missing variable. Also corrects a
  comment naming `backend:3001` as the nginx proxy target; it is `3000`.

**Scripts**
- `scripts/picpeak-setup.sh` — reuse existing secrets instead of regenerating.
  Both the Docker and native paths, and both now keep a `.env.backup-<stamp>`
  before the wholesale heredoc rewrite (mirroring `update_docker_installation`) —
  reusing three secrets does not make replacing the rest of a hand-edited file
  safe. It wrote `.env` with an unconditional `cat >`, so re-running the installer
  against a live install rotated `JWT_SECRET` (invalidating every admin session
  and gallery link) and `DB_PASSWORD` (which then no longer matched the
  initialised Postgres volume, so the stack stopped booting). `--update` was
  always the safe path, but re-running an installer is an easy mistake and should
  not cost someone their data access. The new `read_env_value` helper ignores
  commented lines — the exact trap `install.sh` fell into.
- `scripts/install.sh` — deprecation banner, exits 1.

**CI**
- `.github/workflows/docker-build.yml` — a `smoke-aio` step covering the
  no-`JWT_SECRET` boot.

**Docs**
- `docs/single-container.md` — the two corrections above; `JWT_SECRET` moved from
  required to generated in the run commands, the env table and the NAS notes; a
  new **Secrets** section; a new section on using a photo library that is already
  on the disk; and a note on skipping boot-time ownership adoption for large
  libraries.
- `README.md` — the AIO run command drops the `-e` flag, and the setup-token line
  gains the no-shell route.
- `CONTRIBUTING.md` — the dev-setup rewrite above.

### The external photo library was already implemented, just undocumented

Worth calling out because it is #705's actual differentiator and it reads as new:
mounting a library at `/external-media` has worked since the feature landed.
`server.js:535` defaults `EXTERNAL_MEDIA_ROOT` to that path, the route is mounted
unconditionally and gated on `photos.view`, and `externalMediaService` falls back
to it whenever the directory exists. Nothing documented it for the image where
"the photos already live on this NAS" is the entire reason to choose it — the only
"external" in the document was external PostgreSQL.

The `:ro` in the documented mount is accurate rather than cautious: I traced the
write paths, and thumbnails are generated through the storage layer onto the data
volume, so originals are never modified, renamed or moved. The two caveats in the
doc — mount it outside `/data`, and a network share is fine *here specifically* —
both follow from the boot-time ownership adoption, which is what makes the same
share fatal under `/data`.

## Testing

- **`JWT_SECRET` generation, exercised in `node:22-alpine`** (not reasoned about)
  across all four paths: fresh boot generates 64 base64 chars at mode `0600`; a
  second boot reuses the same value; an explicit `JWT_SECRET` overrides both; and
  a compose-shaped run with `/run/secrets/jwt_secret` present takes that value and
  writes no file at all. The unwritable-directory case was re-run as a non-root
  UID after I noticed root bypasses directory permissions and my first result was
  meaningless — it warns and leaves the variable unset, so validation fails with a
  clear message rather than the secret silently rotating.
- **The compose defaults, by running it:** `docker compose -f
  docker-compose.production.yml config` with an empty `.env` fails on `main` and
  succeeds here. (`env_file: .env` still requires the file to exist; `required:
  false` would relax that, but it needs Compose ≥ 2.24 and felt like your call
  rather than mine.)
- **The secret-preserving helper** on all three shapes: existing value reused,
  missing file generates fresh, commented-only line generates fresh rather than
  returning `#…`.
- **A regression test for the guard**, in the `smoke-aio` CI job — and it was
  wrong on the first attempt. The "an explicit `JWT_SECRET` still wins"
  assertion had its `$` expanded by the runner before the container's script
  ran, so it reduced to `[ x = x ]` and passed even against a stub that
  deliberately clobbers the variable. Escaped now, and re-checked against that
  stub so it fails when it should. The existing
  smoke boot passes `JWT_SECRET` explicitly, so it could not exercise the case
  this image is deployed in. The new step asserts, in one container, that it
  becomes healthy with no secret at all, that the file lands at `0600`, that the
  value is byte-identical across a restart, and that an explicit `JWT_SECRET`
  still wins. I built the image locally and ran those assertions against it
  before committing rather than letting CI shake it out.
- Backend unit suites: 92 passing. `sh -n` / `bash -n` clean on all three
  scripts; the workflow YAML parses.
- **Not verified:** I have no NAS to hand, so the Container Manager / Container
  Station flows are reasoned from the image's behaviour, not exercised. Saying so
  rather than implying coverage.

No UI files are touched, so there is no screenshot.

## Against #705

Closes three acceptance criteria — production compose path defaults, install
script deprecated, and (with #1104) the minimal `.env`. It does not close the NAS
bundles; it removes the reasons they could not be written yet.

For the record, one item on that list turns out not to be development at all:
`aio:stable` does not exist because `Dockerfile.aio` has never been promoted to
the `stable` branch, and stable's workflow has no `build-aio` job. The tag will
appear on its own the next time `main` is promoted. #1121 is a stable-lineage
release and does not carry the image.


Luca-Timo:
Ran a pre-review pass over my own diff before asking you to spend time on it. It
found four real defects, one of them in the test I had just added. All fixed in
`b8dfbaac`; rebased onto `3.111.0-beta.0`, 0 behind.

Posting the findings rather than just the fixes, because two of them are the
kind that read as working.

## 🔴 The regression test could not fail

`.github/workflows/docker-build.yml` — the assertion for "an explicit
`JWT_SECRET` still wins" had its `$` inside a single-quoted runner string:

```sh
sh -c './wait-for-db.sh sh -c "[ \"$JWT_SECRET\" = explicit-… ]"'
```

The runner expands `$JWT_SECRET` before the container's script ever runs, so the
comparison reduces to `[ explicit-… = explicit-… ]`. I checked it against a stub
that deliberately clobbers the variable:

```
current form   -> PASSED against the clobbering stub   (asserts nothing)
$ escaped      -> correctly FAILED against the same stub
```

Escaped now. An assertion that cannot fail is worse than no assertion, because
it reads as coverage in the diff.

## 🔴 Existence-only check on the generated secret

The read-back tested `[ -s "$_jwt_file" ]` — non-empty, nothing more. A short
write (ENOSPC, or a SIGKILL mid-boot) leaves a partial file that every later
boot then accepts, and `validateEnv.js:42` only *warns* below 32 characters
rather than refusing. The install would sign tokens with a truncated key
indefinitely and nothing would ever say so.

Length-checked on read now, so a partial file regenerates instead of being
adopted. Verified with a deliberately truncated file.

## 🟡 Non-atomic write, and divergent secrets under concurrency

The write went straight to the destination and exported the value in hand rather
than the value on disk. Two containers sharing one volume and starting together
would each sign with their own generated secret while the file held only one of
them — tokens minted by one rejected by the other, and half the sessions dying
at the next restart.

Temp file plus rename now, and the export reads the file back so they converge
on whichever write won. Same change closes the partial-write window above.

## 🟡 The installer fix covered only half the paths

`setup_docker_installation` reused an existing secret; `setup_native_installation`
still called `generate_jwt_secret` unconditionally and then rewrote `.env`
wholesale. My PR body stated the decision unscoped, so it claimed more than the
code did.

Both paths now reuse. Both also keep a `.env.backup-<stamp>` before the heredoc
rewrite, mirroring what `update_docker_installation` already does — rescuing
three secrets does not make replacing the rest of a hand-edited file safe.
`read_env_value` also tolerates the shapes a hand-edited `.env` actually takes
(leading whitespace, an `export` prefix, spaces around the `=`); reading one of
those as "absent" would regenerate the secrets and then overwrite the evidence,
which is precisely the failure the helper exists to prevent.

## 🟢 Also corrected

- A comment claimed the secret "shares the database's backup and lifecycle".
  It does not: `backupService.js:127` archives `STORAGE_PATH` and the database
  dump, never `DATA_DIR`. An AIO user restoring from the built-in backup would
  have silently lost every outstanding share link. Said plainly now, in the
  comment, the docs, and the boot message.
- `/external-media` resolves its root once and caches it, so adding the mount to
  a running container does nothing until a restart. Documented.
- Two version-pin examples that were already stale (`3.107.4-beta.0`,
  `v3.108.0-beta.0`) replaced with generic wording.
- `set -e` could skip the friendly warning if the generation pipeline failed;
  `docker volume rm` could race `docker rm -f` in the smoke step.

## Verified by running, not reading

- Truncated file regenerates; reuse across boots is byte-identical; explicit env
  wins; `/run/secrets/jwt_secret` still takes precedence and writes no file.
- The tautology reproduced against a clobbering stub, then confirmed fixed.
- `docker compose -f docker-compose.production.yml config` with an empty `.env`:
  fails on `main`, succeeds here.
- Backend unit suites 92 passing; `sh -n` / `bash -n` clean; workflow YAML parses.
- `smoke-aio` is green on this head, so the new no-`JWT_SECRET` assertions run in
  CI rather than only on my machine.

`bypass-size-gate` is red at ~335 lines against its 300-line cap — the
review-size policy, not a code failure.

Luca-Timo:
Second review pass. Three corrections to my own previous comment, two of which
were the fix not working and one of which was a claim that wasn't true. Head is
now `925c5df0`.

**First, a correction to the comment above:** it cites `b8dfbaac`. That SHA was
orphaned by a rebase and is on no branch — sorry, that would have sent you
looking for nothing. The work is in `e2f7608a`, `4197d44f` and `925c5df0`.

## 🔴 The concurrency fix did not work

I claimed re-reading the file after `mv -f` made two containers converge. It
does not: interleave A-mv, A-read, B-mv, B-read and each still signs with its
own key while the file holds only one.

Worse, my first replacement — an exclusive `set -C` create — was a regression.
Creation is atomic but the content write is not, so a loser can open the file
between the two and export an **empty** secret. Measured across 8 concurrent
boots: 5 came up with nothing.

Both properties are needed and each idiom alone gives only one:

- the file must never be readable half-written, or a concurrent boot gets a
  partial secret — which rules out a bare exclusive-create redirect
- the create must not clobber, or two containers each believe they own it —
  which rules out `mv -f`

Now the full content goes to a private temp file and is **hard-linked** into
place: the link is atomic, fails with `EEXIST` when another container already
won, and the content is complete before the name exists. The loser adopts the
winner's value by re-reading rather than trusting what it generated.

Measured after: 12 concurrent boots over 5 rounds converge on exactly one
secret every time, none empty, all matching the file on disk.

## 🔴 Clearing a corrupt file could delete a good one

The cleanup was itself racy. A container that read before the winner linked
would then see the file present, remove it, and install its own. Removal now
happens only on a second pass, after a link attempt and a re-read have already
failed — at which point the file is genuinely corrupt rather than a race in
flight. `rm -f` plus a bare `rmdir`, so a directory that holds anything is
refused rather than deleted.

## 🟡 The installer could corrupt the secret it was rescuing

`read_env_value` trimmed leading whitespace only. A `.env` edited on Windows
yields `abc\\r`; compose treats the CR as a line terminator, so the running
stack uses `abc`, but re-emitting `abc\\r` into a fresh LF file makes it part of
the value. The reused secret then silently differs from the one in use — which
is the exact failure the reuse exists to prevent. Trailing whitespace and CR are
stripped now.

Two more silent failures on the same path: an unreadable `.env` was
indistinguishable from a missing key, so the secrets would regenerate and then
overwrite the file that held them; and the pre-rewrite backup swallowed its own
failure, so the original could be destroyed with no copy and no message. The
read warns now, and a failed backup aborts rather than proceeding into a
destructive rewrite.

## 🟡 A claim of mine that was not true

My previous comment said the docs "now say" the built-in backup does not include
the generated secret. Only the boot message said it. The Secrets section said
"back it up along with the rest of the volume" while the volume table lists
`/data/backup` as "built-in backup output" — so a reader could reasonably infer
coverage, restore from it, and find every share link dead. Now stated explicitly
in the docs.

## 🟢 And the explanation was wrong even though the fix was right

The comment I added for the escaped `$` in the smoke assertion blamed the
runner. It is not the runner — its single quotes do protect the variable. The
**container's outer `sh -c`** substitutes it while parsing the double-quoted
region, before `wait-for-db.sh` runs. Confirmed by printing what actually
arrives at the script:

```
unescaped -> wait-for-db.sh received: [sh -c [ "FROM_CONTAINER_ENV" = wanted ]]
escaped   -> wait-for-db.sh received: [sh -c [ "$JWT_SECRET" = wanted ]]
```

The conclusion and the fix were right, but a permanent code comment that
misdescribes the mechanism sends the next person to the wrong shell. Corrected.

Also: the restart step now re-asserts health rather than falling through to
`docker exec` (a container that never recovered would have reported a confusing
`cat` failure); `/external-media` is resolved lazily on first use, not "at
start-up"; and `docker image ls` lists local images, not registry tags, so that
pointer was useless.

## Verified by running

12-way concurrency over 5 rounds; truncated, whitespace-only and
directory-shaped paths; explicit env still winning; no temp litter; CR,
trailing-whitespace, commented-key and prefix-key cases on the env reader; the
smoke assertion failing correctly against a stub that clobbers the variable and
passing against an honest one; 92 backend unit tests.

One note on my own process, since it cost three iterations: my first three
concurrency measurements said "still divergent" and were all wrong — I was
counting the winner's four-line `[secrets]` notice as distinct secrets. The
mechanism had been correct since the hard-link change. I only found that by
instrumenting rather than iterating on the fix.

the-luap:
Reviewed the full diff, plus an external pass. The `wait-for-db.sh` design is right —
generate-once, persist next to the database, trust the file over the value in hand —
and the CI test that asserts 0600, survival across a restart, and explicit-env
precedence is the right test. Two things block, and they are both cases where the
change does the opposite of what it sets out to do.

## 🔴 Blockers

### 1. Re-running the installer rotates the secrets of the documented install

`scripts/picpeak-setup.sh:637-639` — the reuse only works for an install that has the
secrets **written in `.env`**. The documented one does not: `.env.example:52-62` says
`DB_PASSWORD` and `REDIS_PASSWORD` are *"OPTIONAL. Leave unset and it is auto-generated
on first run"*, and ships them commented out along with `JWT_SECRET`. For that install
the real values live only in the `picpeak-secrets` volume.

So on a re-run `read_env_value` returns empty, and the new code generates fresh values
and writes them into `.env` as explicit ones. What then happens:

- `secrets-init` sees `/run/secrets/db_password` already non-empty and keeps the **old**
  password. Postgres was initialised with it and still uses it (`POSTGRES_PASSWORD_FILE`).
- `wait-for-db.sh:14-16` only falls back to `/run/secrets` when the variable is empty, so
  the backend now uses the **new** explicit `DB_PASSWORD` and can no longer authenticate.
- `JWT_SECRET` goes the same way: a fresh explicit value overrides the volume, signing out
  every admin session and invalidating every gallery link.

That is precisely the outcome the comment above it says this exists to prevent
(*"re-running the installer is an easy mistake to make and it should not cost the user
their data access"*) — it currently only holds for installs that pinned their secrets by
hand.

Blank has to stay blank for an auto-secret install. Distinguishing "key absent" from
"key present but empty" is not enough on its own, since the documented file has them
commented out; the reliable signal is whether the `picpeak-secrets` volume exists.

### 2. `ln` into a directory makes the mistyped-`-v` case unrecoverable

`backend/wait-for-db.sh:136-138` — the retry comment says a directory here *"is normally
one Docker auto-created for a mistyped `-v` target and is empty"*, and falls back to
`rmdir`. But POSIX `ln source target` where target is a **directory** links *into* it, so
the first pass creates `jwt.secret/jwt.secret.$$.tmp` and the directory is no longer empty
— `rmdir` can never succeed afterwards.

Verified against the runtime image, not just theory:

```
$ docker run --rm alpine sh -c 'cd /tmp; mkdir jwt.secret; printf "s\n" > jwt.secret.1.tmp
    ln jwt.secret.1.tmp jwt.secret && ls jwt.secret/
    rm -f jwt.secret 2>/dev/null || rmdir jwt.secret 2>/dev/null || echo CLEANUP-FAILED'
busybox ln SUCCEEDED into dir
-rw-r--r--    2 root root  2 jwt.secret.1.tmp     <- now non-empty
CLEANUP-FAILED
```

`_jwt_read` then returns empty on every boot (`[ -f ]` is false for a directory), the loop
exhausts its two tries, and the container fails validation permanently with no in-container
way out. Rejecting a non-regular file before the `ln` — and only then falling back to
`rmdir` — restores the recovery the comment describes.

## 🟡 Concerns

### 3. The temp file is not private across containers

`wait-for-db.sh:136` — `$$` is namespaced, so it is `1` in every container:

```
$ for i in 1 2; do docker run --rm alpine sh -c 'echo $$'; done
1
1
```

Every container sharing the volume therefore picks the *same* `jwt.secret.1.tmp`, and the
hard link preserves the **mutable inode**, not a snapshot: a second process that truncates
that path after the first has linked it is writing through `jwt.secret` itself. Two
containers can end up exporting different secrets — which behind a load balancer means
sessions that work or fail depending on which container answers.

In fairness to the measurement standard in the PR description: I could not reproduce it.
12 concurrent boots × 5 rounds with the suffix forced identical, then 3 more rounds with
the write deliberately slowed (truncate, 50 ms pause, finish) — 96 boots, all converged on
one secret, none empty. The window is narrow. The mechanism is still real and the fix is
one line: put randomness in the name (`head -c 8 /dev/urandom | base64 | tr -dc a-z0-9`)
rather than `$$`.

### 4. The retry `rm` can delete a secret another container just created

`wait-for-db.sh:110-117` — `[ -e "$_jwt_file" ]` and the `rm -f` are separated by an `echo`.
A container that entered pass two on a stale short read can remove a *good* secret that
another container linked in the meantime, then link its own — leaving the two exporting
different keys. Same narrow window, same silent symptom; also never reproduced here.
Re-reading immediately before the unlink, and skipping it if the file is now valid, closes it.

### 5. `compose config` still fails without `.env`, so the volume defaults do not deliver their stated purpose

`docker-compose.production.yml:113-117` — the new `${APP_STORAGE:-./storage}` defaults are
justified by making `docker compose -f docker-compose.production.yml config` work before a
`.env` exists. It still does not, because `backend` (and `secrets-init`) carry
`env_file: .env`:

```
$ docker compose -f docker-compose.production.yml config
env file /…/.env not found: no such file or directory
(exit 1)
```

The defaults are harmless, but the comment claims a behaviour the file does not have.
`env_file: [{path: .env, required: false}]` is the piece that actually makes it true.

### 6–8. The new CONTRIBUTING dev loop does not run as written

All three are in the same block (`CONTRIBUTING.md:78-88`):

- **Working directory.** The block ends with `cd ../frontend && npm install`, so
  `docker compose up -d postgres redis` runs in `frontend/`, which has no compose file, and
  the following `cd backend` resolves to `frontend/backend`. Needs a `cd ..`.
- **Wrong `.env`.** `cp .env.example .env` therefore lands in `frontend/`. All three of
  `./`, `backend/` and `frontend/` ship a `.env.example`, so it silently copies the wrong
  one and the backend starts with no `.env` at all — `npm run dev` then exits on
  validateEnv, since (unlike the container entrypoint this PR adds) it generates nothing.
- **Ports.** `PORT=3001` in `backend/.env.example` makes the backend URL right, but Vite
  serves on **5173**, not 3000, and `frontend/vite.config.ts` proxies `/api` to
  **7101**, not 3001 — so API calls from the documented dev frontend reach nothing.

Worth knowing before fixing the last one: the `7101` target is not yours. It arrived in
`0da45e69` ("feat: Add CSS template system…") as an unrelated local-environment leak —
`3001` → `7101`. So the doc cannot be made correct without also deciding whether that
proxy target goes back to 3001. Happy for that to be a separate PR, but this one will
document a loop that cannot work until it is.

## 🟢 Nits

- `read_env_value` does not strip surrounding quotes, so `JWT_SECRET="abc"` round-trips as
  `JWT_SECRET="abc"`. Compose strips matching quotes on read, so it is consistent in
  practice — worth a thought only for the native path, which `dotenv` also handles.
- The unreadable-`.env` warning says *"the install will get fresh ones"*, but the `cp`
  immediately below fails and aborts instead. The abort is the better behaviour; the
  message should say so.
- `.env.backup-$(date +%Y%m%d-%H%M%S)` collides if the installer is run twice inside one
  second. Unlikely; `mktemp`-style suffixing removes the question.

## ✅ Verified

Directly inspected and/or executed:

- `ln`-into-directory behaviour, run in `alpine` (the runtime base) — blocker 2.
- `$$ == 1` in separate containers, and 96 concurrent-boot trials against a faithful copy
  of the secret block — concern 3.
- `docker compose config` without `.env` — concern 5.
- `.env.example` secret defaults vs `secrets-init` vs `wait-for-db.sh` precedence, by
  reading all three — blocker 1.
- `vite.config.ts` port/proxy on `main`, and `git log -L` for the provenance of `7101`.
- The re-exec path: the secret block runs before `exec su-exec`, the export survives it, and
  the second pass correctly skips the block because `JWT_SECRET` is then set. Correct.
- File creation ordering vs `chown -R`: the secret is created before the chown and is adopted
  with everything else. Correct, and the CI assertion on 0600 pins it.

Trusted at description level, not re-run: the "5 of 8 boots" and "12 concurrent boots over
5 rounds" measurements in the PR body.

CI is green; `bypass-size-gate` is the >300-line policy gate asking for a maintainer
approval, not a defect.

## Verdict

Close blockers 1 and 2 and I am comfortable merging. Concerns 3 and 4 are narrow races I
could not reproduce but whose mechanism holds and whose fixes are a line each — I would take
them in this PR while the code is open. 5–8 are documentation correctness and can ride along
or follow, as you prefer.

Luca-Timo:
Both blockers closed in `677969fc`. You were right that each one did the opposite
of what it set out to do — and in both cases I had reasoned about the primitive
instead of running it.

I reproduced every finding before fixing it rather than taking the diagnosis on
trust, since that is the standard you applied.

## 🔴 1 — the reuse only helped the install nobody has

Confirmed. `.env.example:52-62` ships all three commented out, so `read_env_value`
returns empty for the documented install and the old code read that as "no
secrets exist". I had pictured a hand-pinned `.env` the whole time I wrote it.

Blank now stays blank whenever the `picpeak-secrets` volume is there to own it:

```sh
if docker volume ls --format '{{.Name}}' | grep -qE '(^|_)picpeak-secrets$'; then
    # leave JWT_SECRET / DB_PASSWORD / REDIS_PASSWORD unset
```

and the keys are emitted **commented**, matching `.env.example`, rather than as
`KEY=` — an empty assignment is still *defined*, and I would rather not depend on
every future reader treating defined-but-empty as unconfigured. Checked that
compose then falls back:

```
#DB_PASSWORD=          ->   DB_PASSWORD: FELL_BACK
```

Only a genuinely fresh install — nothing in `.env`, no volume — gets generated
secrets. An install that *did* pin them by hand still has them reused.

## 🔴 2 — `ln` into a directory

Confirmed against `alpine`, exactly as you had it: the link lands inside, the
directory becomes non-empty, and `rmdir` can never succeed afterwards.

Non-regular paths are now rejected **before** the link rather than cleaned up
after, which is what restores the recovery the comment described:

- empty directory → removed, boot proceeds, secret generated
- directory holding anything → refused and reported, contents untouched
- the `[ -f ]` in `_jwt_read` is why the old code then read empty forever

Verified all three.

## 🟡 3 — `$$` is 1 in every container

Confirmed (`1 1 1` across three runs). Suffix is random now. I could not
reproduce the divergence either — but the inode-sharing mechanism is real and it
was a line, so it is in.

## 🟡 4 — the retry `rm`

Closed the way you described: re-read immediately before the unlink, and stop if
the file has become valid in the meantime.

## 🟡 5 — `compose config`

You are right that the defaults delivered nothing while `env_file` was a hard
requirement. `required: false` on both services, and it now genuinely works:

```
$ docker compose -f docker-compose.production.yml config   # no .env present
PASS
```

## 🟡 6-8 — the CONTRIBUTING loop

All three fixed: the block returns to the repo root, copies `backend/.env.example`
explicitly rather than whichever of the three the current directory holds, and
names 5173.

**On the `7101` leak — I took it here, and I want to flag that as a decision
rather than have you find it.** `frontend/vite.config.ts` now proxies to 3001.
Your provenance is right: it arrived in `0da45e69` unrelated, and
`backend/.env.example` has said `PORT=3001` throughout. I did it in this PR
because the alternative was knowingly documenting a loop that cannot work. If you
would rather review that on its own, say so and I will pull it out — it is a
three-line revert and the CONTRIBUTING block can note the discrepancy instead.

## 🟢 Nits

All three: `read_env_value` strips matching surrounding quotes (harmless for
compose, but the native path is read by dotenv); the unreadable-`.env` warning now
says the install stops, which is what actually happens; backup filenames carry a
pid suffix.

## What I changed in how I check my own work

Three of your eight findings were shell or container primitives behaving
differently from how I had assumed — `ln` into a directory, `$$` under a PID
namespace, and (from my own earlier round) root ignoring file permissions, which
made one of my test results meaningless. That is a pattern rather than three
accidents, so the pre-review checklist I run before filing now has a section for
it, with the rule that a change turning on the exact semantics of a builtin gets
executed in the runtime image rather than reasoned about.

Two of the checks now fire on the pre-fix tree at the same lines you cited
(`picpeak-setup.sh:637`, `wait-for-db.sh:136` and `:138`) and are silent on the
fix. It would not have caught blocker 1's *reasoning* — but it would have made me
look at the line and ask which install I had in mind, which is the part I got
wrong.

## Verified

Executed, not read: the volume-detection branch and compose fallback; empty,
non-empty and regular-file cases at the secret path; 12 concurrent boots × 4
rounds still converging on one secret with none empty; truncated file
self-healing; explicit env still winning; `compose config` with no `.env`; CR,
trailing-whitespace, quoted, commented and prefix-key cases on the env reader;
92 backend unit tests; frontend build.

Not verified: the native installer path end to end — I have no non-Docker install
to run it against, so that change is reasoned from the shared helper rather than
exercised. Saying so rather than implying otherwise.

the-luap:
Re-verified `677969fc` at the lines cited. Both blockers are closed, all six concerns
and all three nits are closed, and I executed the fixes rather than reading them —
same standard as the first round.

One new finding, introduced by the fix for concern 5. It is the only thing outstanding.

## 🔴 Blockers — closed

**1. Secret rotation.** Blank now stays blank when the volume owns it, and the keys are
emitted commented rather than as empty assignments. I ran `env_secret_line` over the
mixed case that matters — one key pinned by hand, two managed by the volume:

```
# JWT_SECRET is auto-generated into the picpeak-secrets volume on first run.
#JWT_SECRET=
DB_PASSWORD=pinned-by-hand
# REDIS_PASSWORD is auto-generated into the picpeak-secrets volume on first run.
#REDIS_PASSWORD=
```

Right on both axes: the hand-pinned value survives, the managed ones stay out of the
way. `command_exists` is defined (`picpeak-setup.sh:281`), so the volume probe cannot
silently fall through to the generate branch — worth checking, because under
`set -euo pipefail` a missing helper there would have re-opened the blocker quietly.

**2. `ln` into a directory.** All four cases, run against the verbatim block in `alpine`:

| fixture | result |
|---|---|
| empty directory (the mistyped `-v`) | removed, boot proceeds, 64-char secret |
| directory with a file in it | refused, **`important.txt` intact**, warning printed |
| truncated regular file | self-heals to 64 chars |
| clean first boot | `mode=600`, 64 chars |

The second row is the one I cared about — refusing without touching contents.

## 🟡 Concerns — closed

**3 & 4.** Random suffix and the re-read before unlink are both in. 12 concurrent boots
× 5 rounds against the new code:

```
round 1: distinct-exported=1 file-len=64 empty-exports=0 leftover-tmp=0
... (5/5 identical)
```

60 boots, one secret, nothing empty, no stranded temp files.

**5. `compose config`.** Genuinely works now — `exit=0` with no `.env` present, where
it was `exit=1` before. See the new finding below.

**6–8. CONTRIBUTING.** `cd ..` restores the working directory, the copy is explicitly
`backend/.env.example`, and the port reads 5173 with the proxy target explained.

**Nits.** All three. `read_env_value` over seven shapes:

```
"quoted"      -> quoted-value      #commented   -> (empty)
'single'      -> single-quoted     MY_JWT_SECRET=… -> not matched
trailing sp   -> plain             export K = v -> spaced
CR-terminated -> has (no ^M)
```

## 🟡 New — `required: false` needs Compose ≥ 2.24, and older Compose rejects the whole file

This is not a regression in the old sense, but it trades a bigger thing for a smaller
one and I do not think it should merge unqualified.

`env_file: [{path, required:false}]` is Compose 2.24+ syntax. On anything older the file
does not merely lose the default — it fails schema validation outright:

```
$ docker compose -f minimal.yml config     # compose 1.29.2
The Compose file './minimal.yml' is invalid because:
services.app.env_file contains {"path": ".env", "required": false},
which is an invalid type, it should be a string
```

That is `config`, but it is the same validation `up` runs, so the stack does not start
at all. And the repo states no minimum Compose version anywhere — I grepped README,
`docs/`, CONTRIBUTING — and `picpeak-setup.sh` has no version gate.

What makes it worth flagging rather than shrugging at: the audience for this PR is
people installing from a NAS GUI, and Synology/QNAP are exactly where an older Docker
Compose is most likely to be sitting. So the change buys `docker compose config`
working before a `.env` exists, and risks `docker compose up` failing on the platforms
the PR is written for. An existing install that pulls this and re-runs compose is down
until its Compose is upgraded.

Either resolution is fine by me:

- **Keep it** and state the requirement — README, `docs/single-container.md`, a comment
  in the compose file, and ideally a version check in `picpeak-setup.sh` that fails with
  "needs Docker Compose 2.24+" rather than letting the schema error surface.
- **Drop it** back to `env_file: .env` and let the docs say "copy `.env.example` to
  `.env` first", which is what the onboarding flow does anyway.

I lean to the first — the defaults are worth having and 2.24 is January 2024 — but it
needs to be said out loud somewhere a NAS user will hit before the error.

## On taking the `7101` fix here

Keep it here. It is three lines, `backend/.env.example` has said `PORT=3001` throughout,
and the alternative is this PR shipping a CONTRIBUTING block that documents a loop which
cannot work. Splitting it would be process for its own sake. Flagging it rather than
letting me find it was the right call.

## Verdict

Close the Compose-version item — a documented minimum plus a check in the installer, or
a revert of that one hunk — and I am comfortable merging.

Everything else is verified: I executed the secret block in the runtime image (4 fixture
cases, 60 concurrent boots), `env_secret_line`, `read_env_value` across 7 shapes, and
`compose config` with and without `.env`. Not re-run: the native installer path, for the
same reason you gave — and you were right to say so rather than imply otherwise.

Luca-Timo:
@the-luap good catch, and I went the other way from your lean — dropped it rather than
documenting the requirement. `bd2bbd94`. Reasoning below so you can overrule it
cheaply if you disagree; restoring it plus a version gate is a small commit.

## Why drop rather than document

Three things pushed me away from keeping it:

**The beneficiaries and the risk-bearers are disjoint.** What `required: false`
buys is `docker compose config` working with *no* `.env` at all. The onboarding
flow copies `.env.example` as its first step, so nothing normal reaches that
state. What it risks is an existing compose install on older Compose failing
schema validation on the next pull — people who gain nothing from this PR going
down until they upgrade.

**The audience it was argued for does not use this file.** The single-container
path is `docker run` with one volume; `docs/single-container.md` never mentions
`docker-compose.production.yml`. So the NAS users this PR is written for get no
benefit here — while the NAS installs that *do* run the compose stack are, as you
say, exactly where an older Compose sits. That inverts the case for keeping it.

**It would introduce a project-wide requirement as a side effect.** I checked:
no minimum Compose version is stated in the README, `docs/`, CONTRIBUTING or
`picpeak-setup.sh`, and there is no version gate anywhere. Adopting one is a
decision about the whole project, and a deployment-docs PR is the wrong place to
smuggle it in. If you want the minimum stated, I would rather do it deliberately
in its own change — with the gate in the installer you described, so the failure
is "needs Docker Compose 2.24+" rather than a schema error.

## What stayed

The volume defaults, which were the substantive part and work on any Compose
version. They fix the real-world case — a `.env` that simply does not set those
variables, which previously aborted with `invalid spec: :/app/storage: empty
section between colons`:

```
empty .env present  -> config exit=0
no .env at all      -> config exit=1   (env_file still requires it)
```

And I corrected the comment, which had justified the defaults as "so `config`
works before a .env exists". Without `required: false` that is no longer true,
and leaving it would have been a comment claiming behaviour the file does not
have — the thing you caught me on last round. It now states what the defaults do,
and records why `required: false` is not there so the next person does not
helpfully add it back.

## On the 7101 call

Noted, thank you — staying in.

## Checklist update

This went in as its own class: newer syntax that older tooling *rejects* rather
than *degrades*, carrying the three questions that decided it — does the repo
state a minimum anywhere, does old tooling degrade or reject, and do the
beneficiaries and the risk-bearers overlap. It fires on
`docker-compose.production.yml:17` at the commit that had `required: false` and
is silent on the revert. Same shape covers a Postgres function past the
documented minimum, ES syntax past the Node floor in `engines`, or a knex API
newer than the pinned major.

Two things fell out of adding it that are worth more than the rule itself:

**Your three rounds here have all been about what a fix introduced, not what it
missed** — round 2 found my concurrency fix didn't converge and my replacement
for it was worse, and round 3 found this. So the checklist now has a re-review
mode that scans only what changed since the head you last verified, rather than
re-reporting the whole branch. The three lines I just wrote are the ones worth
looking at; the fifty I already defended are not.

**And it caught a gap in itself.** The new class had a pattern check but no
reading step attached, so only the syntaxes I had explicitly enumerated would
ever have been flagged — a knex or Postgres version gate would have sailed
through. There is now a self-check that refuses to pass if any class is missing
either half. Same failure mode as the one you found in the compose file, one
level up.

## Verified

`config` with and without `.env` on the reverted file, both compose files and the
workflow parse, 92 backend unit tests, and no `required:`/2.24-only syntax left
anywhere in the tree.

the-luap:
Verified `bd2bbd94`. The open item is closed and I have nothing further — **merge cleared
from my side.**

I took your suggestion and scoped this to what moved since `677969fc`, the head I last
verified: one file, 13 insertions, 16 deletions. Everything else stands from round 2.

## Drop vs. document — your call was better than my lean

I said I leaned to keeping it and documenting the minimum. Your second argument is the
one that changes my mind, and I checked it rather than accepting it:

```
$ git grep -n "docker-compose.production" -- docs/single-container.md
(nothing)
```

The single-container path never touches this file. So the NAS audience this PR is written
for gains nothing from `required: false`, while the NAS installs that *do* run the compose
stack carry all of the risk. Beneficiaries and risk-bearers genuinely disjoint — that
inverts the trade, and "don't adopt a project-wide Compose floor as a side effect of a
deployment-docs PR" is the right instinct on top of it.

## Verified by execution

The A/B that matters, same Compose that rejected it last round:

```
compose 1.29.2, reverted file          -> ACCEPTED (exit 0)
compose 1.29.2, pre-revert 677969fc    -> REJECTED
    services.backend.env_file contains {"path": ".env", "required": false},
    which is an invalid type, it should be a string
```

Your two config claims, exact:

```
empty .env present -> exit=0
no .env at all     -> exit=1
```

And the error the new comment quotes, reproduced by stripping the defaults back out —
because a comment asserting a specific message is worth checking, given that is what the
last two rounds have been about:

```
invalid spec: :/app/storage: empty section between colons
```

No `required:` / `path:` env_file syntax anywhere in the tree (grep sanity-checked against
a known string, so the empty result is a real negative rather than a broken query).

## On the comment you corrected

That was the right call and the more valuable half of this commit. The defaults now claim
only what they do, and the note records why `required: false` is *absent* — so the next
person reads a decision instead of an omission and does not helpfully restore it. Removing
a justification that had stopped being true is exactly the failure mode from round 2, one
file over.

## On the checklist

The re-review mode is the part I would keep. You are right that all three of my rounds
here found something a fix introduced rather than something it missed — and the corollary
is that re-reading a branch I have already defended is mostly a way to spend attention
where the risk is not. Scoping to the delta is what let this round be four checks instead
of forty.

The self-check that refuses a class with a pattern but no reading step is a better catch
than the rule it was added to, for the same reason: it fails closed. That is the shape the
compose comment was missing.

## Verdict

Merge cleared. Both blockers, all six concerns and all three nits verified closed across
three rounds — the secret block executed in the runtime image, 60 concurrent boots, the
env reader across seven shapes, and now the Compose compatibility A/B.

Not re-run this round, unchanged since `677969fc` and verified there: the JWT secret
block, `picpeak-setup.sh`, CONTRIBUTING and the vite proxy. Still not exercised by either
of us: the native installer path, for the reason you gave in round 2.

the-luap (approved):
Approving after three review rounds.

Both blockers, all six concerns and all three nits verified closed — the secret block
executed in the runtime image across four fixture cases, 60 concurrent boots converging
on one secret, the env reader across seven input shapes, and the Compose compatibility
A/B (1.29.2 accepts the reverted file; the pre-revert head is rejected).

The call to drop `required: false` rather than document a Compose floor was the right
one, and better than my stated lean — the single-container path never touches that
compose file, so the audience this PR is written for gained nothing from it while
carrying all of the upgrade risk.

Not exercised by either of us, and stated as such in the thread: the native installer
path.

PR #1126, opened 2026-08-21
fix(faces): dark-mode styling for the People surfaces (#1106)

the-luap:
Closes #1106.

Every frontend surface added by #1074 shipped light-mode-only. The headline defect is that **"Detect people in this gallery" — the switch that starts GDPR Art. 9 processing — has no readable label in the admin dark theme**, and the Manage-people modal renders as a light panel over a dark page.

## The two halves use different mechanisms, on purpose

The rule is already stated in `PeopleStrip.tsx:92` and this PR just applies it consistently:

- **Admin** (`FaceRecognitionCard`, `PeopleManagerModal`) — each neutral gets a `dark:` counterpart, using the pairings already dominant in `components/admin/`.
- **Guest** (`PeopleStrip`, `PeopleSheet`) — gallery **theme tokens** (`var(--color-text)`, `--color-surface`, …). Galleries carry their own dark themes that the admin `dark` class knows nothing about, so `dark:` would be wrong here.

## Beyond the issue's inventory

The issue was filed before #1119 and #1120 merged, so a few things it could not have listed:

1. **`PeopleManagerModal` had 44 hardcoded neutrals, not 24** — the cover picker and the face-in-context viewer arrived after filing. Both are covered.
2. **The magnifier chip on a cover tile** (`:458`) is `bg-white/90`. A mechanical `text-neutral-700 → dark:text-neutral-300` pass would have put light-grey glyphs on a white pill — it needed `dark:bg-neutral-900/90` as well.
3. **`PeopleSheet`'s panel is a hardcoded `bg-white`** — the same shell defect as the admin modal, not mentioned in the issue.
4. **The selected avatar's ring-offset** defaulted to white, ringing the selected face with a bright halo on a dark gallery. Strip and sheet now drive `--tw-ring-offset-color` from `--color-background`.
5. **The rename input** had no background class, so it rendered white-on-white in dark.

## Inverted hover

Both dismiss buttons used `text-neutral-400 hover:text-neutral-600` — which gets *darker* on hover, so on a dark gallery the control disappeared exactly when reached for. They now run off `--color-text` with an opacity change, so hover adds contrast under either theme. That is the issue's own suggested shape.

`text-neutral-400` is deliberately left bare throughout: it reads on both themes, and the repo leaves it unpaired 393 times against 17 pairings. Pairing it to `dark:text-neutral-500` would have *reduced* dark contrast.

## Screenshots

[...]

Issue #1106, opened 2026-08-20
faces: the People-in-this-gallery UI has no dark-mode styling — toggle labels invisible, Manage-people modal renders light


Luca-Timo:
Follow-up to #1074 — presentation only, no logic involved.

## Summary
Every frontend surface added by the face-recognition feature (#1074) is styled for
light mode only. Not one of the four components ships a single `dark:` variant, so
in the admin dark theme the **checkbox labels of all three toggles render invisible**
(dark grey on near-black), the section heading is barely legible, and the two
explanatory callouts stay light-on-white rectangles pasted into a dark page.

The "Manage people" modal is worse: its panel is a hardcoded `bg-white`, so the whole
dialog renders in light mode on top of a dark admin page — confirmed on prod.

The primary control of the feature — "Detect people in this gallery" — has no
readable label in dark mode. An admin is asked to switch on biometric processing
without being able to read what the switch says.

## Severity
**Medium.** No data loss and no security impact, but the feature's main switch is
unlabeled in a supported theme, and the two toggles below it are indistinguishable
from each other. Given that this particular switch starts GDPR Art. 9 special-category
processing and sits directly under a consent notice, "toggle the unlabeled checkbox
and find out" is a bad failure mode. Purely presentational to fix.

## Environment
- Observed on prod (multi-container deploy, `main` channel), admin dark theme.
- Per-event settings → Overview tab → "People in this gallery" card.
- Reproduces regardless of sidecar state; independent of `picpeak-ml` itself.

## Steps to reproduce
1. Enable the `faces` feature flag.
2. Switch the **admin** theme to dark.
3. Open any event → Overview → "People in this gallery".
4. Enable detection so the second and third toggles render.

**Expected:** all labels, hints and callouts legible, as on every other admin card.
**Actual:** the three bold labels ("Detect people in this gallery", "Show the people
bar to guests", "Sort photos into categories automatically") are effectively
invisible; only their smaller grey hint lines can be read. Selecting the text with
the mouse is the only way to see a label — the selection highlight supplies the
contrast the CSS does not.

## Root cause
`text-neutral-*` utilities do not switch with the theme; the repo's convention is to
pair each one with an explicit `dark:` variant. The faces components skip that
entirely:

| File | `dark:` variants | hardcoded `*-neutral-*` |
|---|---|---|
| `frontend/src/components/admin/FaceRecognitionCard.tsx` | **0** | 20 |
[...]

PR #1127, opened 2026-08-21
feat(faces): consolidate look-alike clusters after a scan, and suggest the rest (#1107)

### LINKED ISSUE #1107: faces: consolidate look-alike clusters after a scan, name autocomplete, and open the source photo from a face
Follow-up to #1074. Three requests from running the feature on a real gallery
(114 photos, 14+ clusters). They are all the same surface — the "Manage people"
modal and the naming/merging workflow that follows a scan — so they are bundled
rather than filed separately.

---

## 1. Consolidate look-alike clusters automatically, with a warning

**Observed:** a 114-photo gallery produced 14+ unnamed people, several of which
are visibly the same person split across clusters. Merging them is currently
manual: tap two faces, press Merge, repeat.

**The interesting part is that the code already does this — it just never runs.**
`faceClustering.consolidate()` exists precisely for this case, and its own comment
describes the symptom exactly:

> Greedy assignment can open "Anna in daylight" and "Anna at the party" as two
> clusters when the first few photos of each were dissimilar. Once both have
> absorbed enough faces their centroids converge, and this pass catches that.

It runs at a deliberately stricter threshold than assignment
(`face_match_threshold + 0.08`, capped at 0.95, `faceClustering.js:257`).

But `consolidate()` is called from exactly one place: `recluster()`
(`faceClustering.js:514`). The scan path never calls it — `faceProcessor.js:184`
calls `assignFaces()` and nothing else. So after a normal background scan the
centroids have converged and nobody looks. The consolidation only happens if the
admin happens to press "Re-group people" manually, which is not something the UI
suggests doing, and the button reads like a destructive re-do rather than a
cleanup.

**Requested:**
- Run `consolidate()` when a scan drains, so the first thing the admin sees is
  already consolidated rather than 14 clusters that should be 8.
- For the band *below* the auto-merge threshold — the genuinely uncertain pairs —
  surface them in the modal as **suggestions** rather than merging them: "These
  two look like the same person" with the two faces side by side, and
  accept/dismiss. That is the "with a warning" part of the request: an over-eager
  merge is much harder to unpick than a missed one (the code comment says as
  much), so the uncertain band should never merge silently.
- If an automatic consolidation pass did merge clusters, say so somewhere the
  admin can see it — a count, an undo, or both. Silent merging of biometric
  clusters is the wrong default even when the confidence is high.

## 2. Name autocomplete

**Observed:** naming people means typing each full name into a fresh empty input,
once per cluster. With a wedding-sized guest list and clusters that still need
merging, the same name gets typed repeatedly.

The rename control (`PeopleManagerModal.tsx:338`) is a plain `<input>` with no
`list`/datalist and no suggestion source.

**Requested:** suggest from names already used, so the second and third occurrence
of a name is a keystroke or two rather than the whole thing.

**Scoping question for whoever picks this up:** names from *this event* are
obviously safe. Names from *other events* would be more useful — the same family
recurs across shoots — but that surfaces client names from galleries the current
admin may not have access to, which collides with the event-scoped permissions
work in #743. Suggest starting event-scoped, and treating cross-event suggestions
as a separate decision rather than an implementation detail.

## 3. Open the source photo from a face

**Observed:** there is no way to see the photo a face came from. The thumbnail is
a face crop, so when two clusters look similar the only way to judge whether they
are the same person is to squint at two 40px circles.

**Requested:** open the full photo — in a lightbox — from a person's face
preview.

**The catch:** in the modal, clicking a face is already taken. `FaceThumb` is
wrapped in a button bound to `toggleSelect` (`PeopleManagerModal.tsx:313`), and
the footer tells the admin "Tap two or more faces to merge them into one person."
So this needs a second affordance rather than a straight `onClick` — a hover
"view" control, a modifier-click, or splitting select onto an explicit checkbox
and giving the thumbnail the photo. Worth a moment's design rather than picking
one blind; the merge-selection gesture is the more important of the two and
should not get harder.

Note there is no shared admin lightbox to reuse — `PhotoLightbox` is gallery-side,
and `TransferPhotoPicker` is the only other admin surface with anything similar.

---

Related: #1106 (this whole surface has no dark-mode styling).

the-luap:
Closes #1107.

> **Stacked on #1126.** Targets `fix/1106-faces-dark-mode`, not `main` — the suggestion card is new UI on a surface #1126 is making theme-aware, and writing it light-only would immediately re-open #1106. Retarget once #1126 merges.

**Item 3 of the issue is already done.** "Open the source photo from a face" shipped in #1120 — the magnifier in the row action cluster opens the face outlined in its own photo, with prev/next across that person's appearances. So this covers items 1 and 2.

## 1. Consolidation now runs when a scan drains

The issue's diagnosis was exactly right: `consolidate()` already existed, already described this symptom in its own comment, and simply never ran. Its only caller was `recluster()`.

There is no "scan finished" event to hook — the queue claims photos one at a time, so a backfill is just a lot of independent claims. The available signal is a worker finding nothing left to claim, and that signal is **not sufficient on its own**: with `FACE_PROCESSOR_CONCURRENCY > 1` the other workers may still be busy, and a photo released back to `pending` by a down sidecar is still owed. So an idle worker asks whether the events it touched have *actually* drained (`faceQueue.js`), and an event that is still busy keeps its place for the next tick.

Consolidation failures never propagate into the worker loop — a bookkeeping problem must not stop the queue.

## 2. The uncertain band asks rather than acts

Pairs scoring between the assignment threshold and the stricter auto-merge threshold (`match + 0.08`) become suggestions in the manager: both faces side by side, "Same person" / "Not the same". This is the "with a warning" half of the request — the code already argued that an over-eager merge is much harder to unpick than a missed one, so the band it is *not* confident about never merges by itself.

Dismissals are sticky (new `event_people_merge_dismissals` table), because a pair told "not the same" coming back after every scan is worse than not asking.

What it refuses to ask about, each pinned by a test:

- pairs across different embedding spaces (`model_version`)
- two people the photographer **named differently** — a human assertion this heuristic does not get to overrule, same rule `consolidate()` already applies
- anyone marked "not a real person"
- **the same person twice in one round** — otherwise a genuine three-way fragment offers A-B, A-C and B-C, and accepting A-B leaves two suggestions naming a person the merge just deleted

Accepting picks the **named** side as the merge target so the name survives, falling back to the larger cluster. `doMerge` has to use click order; here the order is ours to choose, so it is chosen correctly.

## 3. Nothing merges silently

A pass that merged anything reports it on the card. It points at **Split** as the way to unpick a wrong merge rather than claiming an undo that does not exist. The count is per-scan: a later pass that merges nothing clears it, rather than leaving a stale number standing.

## Name autocomplete

Needs no endpoint — the people list already open in front of the user *is* the source. One `<datalist>` for the whole modal rather than a copy per row.

Deliberately **event-scoped**, as the issue suggested. Names from other events would be more useful, but that surfaces client names from galleries the current admin may not be allowed to open — the permissions decision in #743, not an implementation detail.

## Also

`photoCount` had no `_one` form, so any single-photo cluster read "1 photos". Already wrong in the people rows; the suggestion card puts it somewhere harder to miss. Added to both locales.

## Verified

**Live, against a real 62-photo gallery with 16 clusters** (screenshots in the first comment):

- Triggered two real re-scans. Both drained, and both wrote `faces_last_consolidated_at` — so the drain hook fires in the real system, not just in tests.
- Both merged **0 pairs, correctly**: this gallery's top pairwise similarity is `0.5177`, well under the `0.68` auto-merge line. These are genuinely different people, and forcing a merge by dropping the threshold would only demonstrate that a bad threshold merges badly. The merge-and-record path is covered by unit test instead, with centroids built to an exact cosine similarity.
- The suggestions endpoint returned a real pair at `0.6020` — inside the band, correctly *not* auto-merged. Dismissed it in the UI; the row persisted and the pair stopped being offered.
- Autocomplete: the rename input carries `list="picpeak-people-names"` and the datalist held exactly the three names in use, deduped and sorted.
- **Honest note on one screenshot:** the consolidation notice was captured with the count set directly in the database, because this gallery legitimately has nothing to auto-merge. The number's provenance is covered by the unit tests above, not by that screenshot.

**Tests** — 19 new, all passing:

- `faceMergeSuggestions.test.js` (13) — band edges, the four refusals, dismissal normalisation and idempotency, ordering, and consolidation reporting including the clear-on-zero case.
- `faceQueueDrainConsolidation.test.js` (6) — the drain condition: waits on `pending`, waits on `processing`, fires once when empty, does not repeat, survives a throwing consolidation, and treats events independently.

Full face suite 101 passing across 14 files; frontend 198 passing; `tsc` clean; ESLint clean on every touched file; build clean.

**Migration 180** verified idempotent and reversible against the real PostgreSQL database — re-ran `up` on an existing schema, then `down`, then `up` again, asserting the table and both columns each time.

One design note worth a second opinion: the suggestion band is `[match_threshold, merge_threshold)`, which at the default `0.60` is a narrow window (`0.60`–`0.68`). Widening the floor would surface more candidates at the cost of more obviously-wrong pairs. I kept it principled — below the match threshold, two centroids are further apart than the distance at which a single face would have joined the cluster at all — but it is a tuning decision, not a fact.

The local pre-push E2E gate was bypassed: the compose `picpeak-backend` cannot start on this host (`/app/storage is not writable by UID 501`), unrelated to and predating this change.

### DIFF
--- a/backend/__tests__/integration/faceMergeSuggestions.test.js
+++ b/backend/__tests__/integration/faceMergeSuggestions.test.js
@@ -95,6 +95,28 @@ async function insertPersonWithFace(eventId, centroid, overrides = {}) {
   return personId;
 }
 
+/** An additional face on an existing person, so a split has something to move. */
+async function addFaceTo(eventId, personId, centroid) {
+  const [p] = await db('photos').insert({
+    event_id: eventId,
+    filename: `${Math.random()}.jpg`,
+    path: '/tmp/x.jpg',
+    type: 'individual',
+  }).returning('id');
+  const photoId = typeof p === 'object' ? p.id : p;
+  const [f] = await db('photo_faces').insert({
+    photo_id: photoId,
+    event_id: eventId,
+    person_id: personId,
+    bbox_x: 0, bbox_y: 0, bbox_w: 200, bbox_h: 200,
+    det_score: 0.99,
+    embedding: clustering.packEmbedding(centroid),
+    model_version: 'test-v1',
+    created_at: new Date().toISOString(),
+  }).returning('id');
+  return typeof f === 'object' ? f.id : f;
+}
+
 const suggest = (eventId) => clustering.suggestMerges(eventId, { thresholds: THRESHOLDS });
 
 describe('face merge suggestions (#1107)', () => {
@@ -321,6 +343,37 @@ describe('face merge suggestions (#1107)', () => {
     });
   });
 
+  describe('manual splits survive the automatic pass', () => {
+    /**
+     * The regression that matters most once consolidation runs on every scan:
+     * a photographer splitting a wrongly-merged cluster produces two people
+     * who are look-alikes BY CONSTRUCTION, so their centroids sit above the
+     * merge threshold and the very next scan would put them straight back.
+     */
+    it('records a split as a separation, so consolidation leaves it alone', async () => {
+      const eventId = await seedEvent('split-protected');
+      const base = new Float32Array(DIM); base[0] = 1;
+
+      // One cluster holding two near-identical faces.
+      const personId = await insertPersonWithFace(eventId, base);
+      const extraFaceId = await addFaceTo(eventId, personId, base);
+
+      const newPersonId = await clustering.splitPerson(eventId, personId, [extraFaceId]);
+      expect(newPersonId).toBeTruthy();
+
+      const rows = await db('event_people_merge_dismissals').where({ event_id: eventId });
+      expect(rows).toHaveLength(1);
+      expect([rows[0].person_a_id, rows[0].person_b_id].sort())
+        .toEqual([personId, newPersonId].sort());
+
+      // Identical centroids — nothing but the recorded separation can stop
+      // this merge.
+      const merged = await clustering.consolidate(eventId, { thresholds: THRESHOLDS });
+      expect(merged).toEqual([]);
+      expect(await db('event_people').where({ event_id: eventId })).toHaveLength(2);
+    });
+  });
+
   describe('consolidation reporting', () => {
     it('records what an automatic pass merged, so it is not silent', async () => {
       const eventId = await seedEvent('report-merged');
@@ -373,6 +426,14 @@ describe('face merge suggestions (#1107)', () => {
       expect(await db('event_people').where({ event_id: eventId }).count({ c: '*' }).first())
         .toEqual(expect.objectContaining({ c: 2 }));
     });
+
+    // NOT covered by a test: reporting what a pass merged before it died
+    // partway. `consolidate` calls `mergePeople` through the module-local
+    // binding, so a spy on the export cannot intercept it, and no realistic
+    // database failure lands on the second merge only. The recording therefore
+    // sits in a `finally` — each mergePeople is its own transaction, so a pass
+    // that throws has still committed what it did, and the alternative is a
+    // real merge going unreported. Verified by reading, not by assertion.
 
     it('clears a previous count when a later pass merges nothing', async () => {
       const eventId = await seedEvent('report-cleared');
--- a/backend/__tests__/integration/faceMergeSuggestions.test.js
+++ b/backend/__tests__/integration/faceMergeSuggestions.test.js
@@ -1,0 +1,288 @@
+/**
+ * Automatic consolidation reporting and the suggestion band (#1107).
+ *
+ * Centroids are built to an EXACT cosine similarity rather than jittered
+ * towards one, because every assertion here is about which side of a threshold
+ * a pair falls on. `pairAtSimilarity` returns two unit vectors whose dot
+ * product is the requested number to floating-point precision, and each pair
+ * is built on its own orthogonal basis so two different pairs are never
+ * accidentally similar to each other.
+ */
+
+const path = require('path');
+const fs = require('fs');
+const os = require('os');
+
+process.env.NODE_ENV = 'test';
+process.env.TEST_DATABASE_PATH = path.join(
+  fs.mkdtempSync(path.join(os.tmpdir(), 'picpeak-facesuggest-')), 'db.sqlite',
+);
+process.env.JWT_SECRET = process.env.JWT_SECRET || 'facesuggest-test-secret';
+
+const { bootCrmDb } = require('./helpers/crmDb');
+
+let db; let cleanup; let clustering;
+
+// Mirrors the service: merge at match + 0.08, so with a 0.60 floor the
+// suggestion band is [0.60, 0.68).
+const THRESHOLDS = {
+  face_match_threshold: 0.6,
+  face_quality_min_score: 0.7,
+  face_quality_min_px: 40,
+};
+
+const DIM = 64;
+
+/** Two unit vectors whose dot product is exactly `target`, on basis (i, i+1). */
+function pairAtSimilarity(target, basis) {
+  const a = new Float32Array(DIM);
+  const b = new Float32Array(DIM);
+  const orth = Math.sqrt(1 - target * target);
+  a[basis] = 1;
+  b[basis] = target;
+  b[basis + 1] = orth;
+  return [a, b];
+}
+
+async function seedEvent(slug) {
+  const [row] = await db('events').insert({
+    slug,
+    event_type: 'wedding',
+    event_name: slug,
+    event_date: '2026-01-01',
+    host_email: 'h@example.com',
+    admin_email: 'a@example.com',
+    password_hash: 'x',
+    share_link: `${slug}-share`,
+    expires_at: new Date().toISOString(),
+  }).returning('id');
+  return typeof row === 'object' ? row.id : row;
+}
+
+async function insertPerson(eventId, centroid, overrides = {}) {
+  const [row] = await db('event_people').insert({
+    event_id: eventId,
+    centroid: clustering.packEmbedding(centroid),
+    face_count_total: 5,
+    model_version: 'test-v1',
+    created_at: new Date().toISOString(),
+    updated_at: new Date().toISOString(),
+    ...overrides,
+  }).returning('id');
+  return typeof row === 'object' ? row.id : row;
+}
+
+/** One person with one real face, so merge/split have something to move. */
+async function insertPersonWithFace(eventId, centroid, overrides = {}) {
+  const personId = await insertPerson(eventId, centroid, overrides);
+  const [p] = await db('photos').insert({
+    event_id: eventId,
+    filename: `${Math.random()}.jpg`,
+    path: '/tmp/x.jpg',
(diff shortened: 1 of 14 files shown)


the-luap:
## External review pass (Codex), 3 rounds — 12 findings fixed

Ran the branch through an external reviewer three times, verifying each finding
against the source before acting. Everything below is committed in `7e142a1e` and
`a0f101ff`.

The theme throughout: **this change makes `consolidate()` run by itself after every
scan**, so anything it could quietly override went from "only if an admin presses
Re-group people" to "automatically, on every gallery". Most of these are that.

### Automatic merging could overrule explicit decisions

1. **It absorbed clusters marked "not a real person."** `mergePeople` ORs
   `is_ignored` onto the survivor, so a false positive crossing the threshold with a
   real person marked *that real person* ignored — dropping them out of the
   guest-facing strip. Now skipped, matching what suggestion generation already did.
   `is_hidden` is deliberately left OR-ing: two clusters of one person staying hidden
   is correct and reversible.
2. **It merged pairs the photographer had dismissed.** Dismissals were consulted only
   when generating suggestions, so a rejected pair that drifted over the threshold —
   or a threshold someone tuned — was merged anyway.
3. **It undid manual Splits.** `splitPerson` recorded nothing, and two clusters pulled
   apart by hand are look-alikes *by construction*, so the next scan put them back. A
   split now records a separation. This was the sharpest regression of the set.
4. **A dismissal could lose the race with the pass it was meant to stop.** The set was
   read once per pass; the pair is now re-read immediately before each merge.
5. **The dismissal read failed open.** A timeout returned "nothing dismissed", quietly
   restoring the merge the dismissal existed to prevent. Only a missing table (i.e. a
   pre-migration install) is allowed to fail open now.

### It could run when it shouldn't, or not run at all

6. **Skipped/failed photos marked the event.** They never went through detection.
   Notably this meant that after an admin switched detection **off** mid-drain, the
   remaining photos drained as `skipped` and consolidation ran anyway. The drain also
   rechecks the per-event toggle now.
7. **A finished gallery waited for the whole install.** The worker only drains when it
   can claim nothing *anywhere*, so with the default single worker one gallery
   finishing during a large backfill waited for every other gallery — potentially
   forever under continuous ingestion. The janitor drains too now.

### It could erase or lose its own report

8. **Concurrency > 1 could overwrite the count with zero.** The last photo is committed
   `done` before auto-categorisation runs, so another worker saw the event as drained,
   consolidated, recorded its count — then the still-running worker re-marked the
   event and the next pass recorded 0. Workers now count themselves in flight per event.
9. **A transient failure lost the pass entirely**, because the event was dropped from
   the touched set before consolidation ran. Dropped only on success now, with a
   backoff so a permanently failing event doesn't warn on every idle tick.
10. **A pass that died partway reported nothing**, despite each `mergePeople` being its
    own committed transaction. Recorded in a `finally`.

### Data lifecycle

11. **The new table escaped the face-data exclusions.** `photo_faces` and
    `event_people` are deliberately excluded from exports and backups; the dismissals
    table was not, so references to person ids the target never receives would attach
    to whatever ids its own re-scan minted. Added to all three lists
    (`picpeakExportService`, `picpeakImportService`, `databaseBackup`).
12. **Purging face data left it behind.** That is the GDPR erasure path, and a stale
    count also had the card reporting merges beside an empty people list. Both cleared,
    inside `purgeEvent` so archival gets it too.

Plus: the dismissal insert swallowed *every* error as "duplicate", so a missing table
or read-only database reported "kept separate" for a decision that was never stored.

### Accepted, not fixed

- **Separations don't survive a full re-scan** — they're keyed on person ids, which
  re-clustering re-mints. Filed as #1132. Within a scan cycle (the normal naming
  workflow) they do hold. I hit this live: a dismissed pair was re-scanned, one side
  was absorbed, and the row was left dangling.
- **The in-flight guard is process-local.** Across pods two workers can still both
  consolidate one event and overwrite the count with zero. Merges stay correct — only
  the notice is affected — and this file already treats multi-pod consolidation as
  "safe rather than coordinated".

### Verification

19 → **26 tests** on this branch; full face suite **120 passing** across 14 files.
Frontend 198 passing, `tsc` clean, ESLint clean on every touched file, build clean.

One gap stated plainly: **the partial-failure reporting (10) has no test.**
`consolidate` calls `mergePeople` through the module-local binding, so a spy on the
export cannot intercept it, and no realistic database failure lands on the second
merge only. It's a `finally`, verified by reading rather than by assertion, and the
test file says so where the test would have been.

Round 3 also flagged the export/backup and purge items as fixed and found no new
issues in those areas; round 2 of #1126 came back clean.

the-luap:
## Open items

Two known limitations, both deliberate. Neither blocks merge, but both should be
read before approving — they bound what this feature actually promises.

### 1. Separations do not survive a full re-scan → #1132

Both kinds of "not the same person" — the explicit dismissal and the implicit one a
**Split** now records — are stored as a pair of `event_people.id` values. Person ids
do not survive a re-scan: `processPhotoFaces` replaces a photo's faces,
`recomputeCentroid` drops clusters that end up empty, and `assignFaces` re-derives
the rest with fresh ids.

Worse than the row merely going stale: `assignFacesLocked` does not consult
separations at all, and a dismissed pair scores at least `face_match_threshold` by
definition — that is the bottom of the suggestion band — so the replacement face can
rejoin the very cluster it was separated from.

Not hypothetical. While testing this branch, a dismissed pair `(157, 158)` was
re-scanned, person 158 was absorbed, and the dismissal row was left dangling.

**What holds today:** within a scan cycle — which is the normal naming workflow —
separations do bind. `consolidate()` reads them on every automatic pass and re-reads
the specific pair immediately before merging it.

**Why not here:** the fix is to enforce separation *during assignment*, keyed to
something that survives re-clustering (face ids rather than person ids). That changes
how assignment behaves — a face would sometimes be pushed away from its nearest
centroid — and wants its own test matrix rather than riding along on a consolidation
change. Scoped in #1132.

### 2. The in-flight guard is process-local

`inFlightByEvent` stops one worker consolidating an event while another is still
inside `processPhotoFaces` — the window where the last photo is already committed
`done` but auto-categorisation is still running. That guard is a `Map` in one
process.

Across pods it does not apply: pod B can see the event as drained, consolidate, and
record its count, after which pod A re-marks the event and the next pass records `0`.

**Blast radius is the notice, not the data.** Merges themselves stay correct —
`mergePeople` is transactional and a pair whose source was already absorbed merges
nothing. What can be wrong is the "merged N look-alike pairs" count on the card,
which can read `0` when a merge did happen.

**Why not here:** closing it means DB-backed coordination (a `pg_advisory_xact_lock`
on the event, as `withEventLock` already does for assignment), which would put a
transaction around the whole pass and change the per-merge transaction boundaries.
`faceClustering.js` already states the position this follows — multi-pod
consolidation is safe rather than coordinated.

Happy to take either one in this PR instead if you would rather not carry them.

PR #1133, opened 2026-08-21
fix(gallery): a missing thumbnail tier must not take the backend down (#1128)

### LINKED ISSUE #1128: [BUG] Backend process exits on first ?w= thumbnail tier load (#1109) — unhandled ReadStream ENOENT after a regenerate race
**Describe the bug**

The first load of a gallery whose responsive thumbnail tiers (#1109) do not exist yet can **terminate the backend process**. Not a 500 on one tile — the whole Node process exits, so every gallery goes blank until the container restarts.

Two separate defects stack here: a race that deletes a tier file another request has already validated, and a lazy `fs.createReadStream` whose `error` event nothing is listening for, which turns what should be a 404 into a process exit.

**To Reproduce**

1. Install with `thumbnail_width` at its default, so `w900` is a genuine tier rather than the canonical width
2. Have a gallery with no `thumb_w900_*` files on disk yet
3. Open it in a browser — the grid issues concurrent `?w=900` requests, one per tile

79 tiles was enough. It is a race, so it does not fire every time.

**Expected behavior**

A thumbnail tier that is missing or mid-generation should produce a 404 or a fallback to the canonical thumbnail for that one request. It should never take the process down.

**Environment**
- OS: Debian (Proxmox LXC)
- Browser: Safari 17 (iOS) and Chrome — irrelevant, the crash is server-side
- PicPeak Version: 3.111.0-beta.0 (backend + frontend + ml), Node v22.23.2
- Deployment Method: Docker Compose (`docker-compose.production.yml`, GHCR images)
- Database: PostgreSQL 15
- Storage: local backend; all photos are `source_origin=external`, originals on an NFS mount

**Logs**

`RestartCount=1`, crash immediately after migrations completed, i.e. on the first real traffic:

```
Migration core/179_person_cover_manual_only.js completed successfully
All migrations completed successfully
node:events:497
      throw er; // Unhandled 'error' event
      ^

Error: ENOENT: no such file or directory, open '/app/storage/thumbnails/thumb_w900_p6677_REDACTED.jpg'
Emitted 'error' event on ReadStream instance at:
    at emitErrorNT (node:internal/streams/destroy:170:8)
    at emitErrorCloseNT (node:internal/streams/destroy:129:3)
    at process.processTicksAndRejections (node:internal/process/task_queues:89:21) {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '/app/storage/thumbnails/thumb_w900_p6677_REDACTED.jpg'
}

Node.js v22.23.2
```

(Basenames redacted throughout — they carry a real person's name. Structure is otherwise verbatim.)

After the automatic restart the tiers finished generating (79/79 present) and the gallery renders fine, so this only bites on first load.

**Additional context — why it is fatal rather than a 404**

Thumbnail route in `backend/src/routes/gallery.js`:

```js
res.setHeader('Content-Length', stat.size);
const stream = await storage.get(thumbnailPath);
stream.pipe(res);
```

On the local backend `storage.get()` returns an `fs.createReadStream`, which is lazy: the ENOENT is emitted on a later tick, after the `await` has resolved and outside the route's `try/catch`. So `errorResponse(res, error, 500, ...)` never runs, nothing is attached to the stream's `error` event, and Node terminates.

The codebase already documents this exact hazard in the bulk-download path in the same file (#895 review):

> Verify the source exists BEFORE appending — but only for local sources: fs.createReadStream is lazy, so its error fires outside this try/catch [...]

The thumbnail route has neither that preflight nor a `stream.on('error')`.

**Additional context — why the file is missing at all**

`ensureThumbnailAtWidth()` treats the tier as a pure cache:

```js
try {
  if (await storage.stat(key)) return key;
} catch (e) {
  // regenerate below
}
```

and on a miss calls `generateThumbnail(..., { regenerate: true, outputBasename, width, height })`. `generateThumbnail` opens by deleting the target:

```js
if (options.regenerate) {
  await storage.delete(thumbnailRelKey).catch(() => {});
}
```

There is no in-flight dedupe, and the grid fires one request per tile concurrently. For a single photo:

- request A misses the stat, enters generation, writes and renames the file into place
- request B stats the same key, hits, returns it, heads for `storage.get()`
- request C also missed the stat earlier, is still inside `generateThumbnail`, and its `regenerate: true` **deletes the file A just published**
- B's lazy stream opens a path that no longer exists → unhandled ENOENT → process exit

`Content-Length` comes from the same earlier `stat`, so it is stale in that window too.

Consistent with this, the crashed process left 74 orphaned intermediates in `storage/thumbnails/`, 17 of them zero-byte:

```
thumb_w900_p6665_REDACTED.jpg.tmp.7.c55173bf
thumb_w900_p6627_REDACTED.jpg.tmp.7.647117df
thumb_w900_p6649_REDACTED.jpg.tmp.7.7c958508
...
```

`7` is the PID of the process that died. Nothing cleans these up on the next start.

**Possible Solution**

1. Attach a `stream.on('error')` in the thumbnail route, and mirror the local-storage preflight already used in the bulk path. This alone downgrades the failure from "backend down" to "one broken image", which seems worth doing independently of the rest.
2. Drop `regenerate: true` from `ensureThumbnailAtWidth`'s call — on a cache miss there is nothing to regenerate, and that unconditional `storage.delete()` is what removes a file another request has already validated.
3. Optionally dedupe in-flight generation per key, so a grid of N tiles does not run N Sharp passes over the same source. On external/NFS sources each of those also re-reads the original.
4. Sweep stale `*.tmp.*` intermediates in the thumbnails directory on startup.

Happy to test a build — I have three more galleries (37, 3230 and 3297 photos) with no tiers generated yet, so I can reproduce at larger scale on request.

the-luap:
Closes #1128.

@BraynArts diagnosed this one down to the interleaving, including the orphaned
`.tmp.7.*` intermediates naming the PID of the process that died. Both defects it
identified are fixed, because either one alone still leaves a route to the other.

## The reader: a lazy stream nobody was listening to

`LocalFsStorage.get()` returns an `fs.createReadStream`. It resolves immediately and
opens the file on a **later tick** — so the ENOENT lands after the `await` returned
and outside the route's `try/catch`. An EventEmitter emitting `error` with no listener
throws, and an uncaught throw from an I/O callback is not something Express can catch:
Node exits. One missing tile took every gallery on the install down until Docker
restarted it.

New `pipeStreamToResponse` attaches the handler that was missing:

- source gone → 404 (410 for a prepared download), with the now-meaningless
  `Content-Length` and `ETag` stripped so the error body is not self-contradictory
- anything else → 500
- **already streaming** → destroy the connection. The status line is spent; a
  truncated image the client would cache is worse than a broken connection
- client hangs up → destroy the source, which on an abandoned grid was one leaked fd
  per tile

Applied to **all nine** streaming responses in `gallery.js`, not just the thumbnail
route. They are the same construct on the same guest-facing surface, and fixing one
crash while leaving eight identical ones next to it is not a fix. (The two
`archive.pipe` sites already had handlers.)

There are ~14 more unguarded `.pipe(res)` sites in admin routes. I left them: they are
authenticated, lower-concurrency, and none has the delete-race below. Worth a separate
sweep.

## The writer: `regenerate: true` on a cache miss

`ensureThumbnailAtWidth` is only reached when the tier is **absent**, so there was
never anything to regenerate — but that flag makes `generateThumbnail` open by
`storage.delete()`-ing the target. With one request per tile in flight: A publishes the
tier, B stats it and heads for the read, and C — still inside generation from its own
earlier miss — unlinks the file B is about to open.

Dropping the flag makes the write a plain put, which stages to a temp file and renames.
A concurrent reader sees the old file or the new one, never a hole.

## Plus: one Sharp pass per tier, not per request

Generation is now deduped per tier key. The reporter's point 3, and it is also what made
the race easy to hit — 79 tiles meant 79 decodes of the same source, each one an extra
full read off the NFS mount for an external photo.

Measured in the test: **5 concurrent generations for 8 requests before, 1 after.**

## Verified

- **6 new tests** on the stream helper, over a real `fs.createReadStream` against a real
  missing path — a hand-rolled emitter that fires synchronously would pass while proving
  nothing, since the lazy-open timing *is* the bug.
- **3 new tests** on concurrent tier generation: the file is never absent once published,
  a burst runs one generation, and a failure is not cached.
- The dedupe test was checked against the unfixed code to confirm it discriminates:
  `Expected: <= 1, Received: 5`.
- One honest note on the helper's first test: an unhandled `error` surfaces as a *suite*
  failure under jest, not through a `process.on('uncaughtException')` listener — such a
  listener never fires and the assertion could never fail. The test file says so where a
  reader would otherwise assume the opposite.
- Full route suite 264 passing across 33 files; tier suite 24 passing; ESLint clean.

## Not addressed

The orphaned `*.tmp.*` intermediates already on disk. Nothing writes new ones once the
process stops dying, but sweeping the existing ones is separate janitorial work — say
the word and I will add it here or file it.

### DIFF
--- a/backend/__tests__/integration/previewTiers.test.js
+++ b/backend/__tests__/integration/previewTiers.test.js
@@ -371,5 +371,88 @@ describe('preview tiers (#1095)', () => {
       await imageProcessor.deleteThumbnailTiers(await db('photos').where({ id: photo.id }).first());
       expect(fs.existsSync(abs)).toBe(false);
     });
+
+    /**
+     * The crash in #1128 needed two things: a tier that disappears, and a
+     * reader that dies on it. The reader is fixed in streamResponse; this is
+     * the half that stops the file disappearing in the first place.
+     */
+    describe('concurrent generation (#1128)', () => {
+      it('never leaves the tier absent once it has been published', async () => {
+        const photo = await seedThumbPhoto();
+        const key = imageProcessor.thumbnailTierKeys(photo).find((k) => k.includes('_w600_'));
+        const abs = path.join(process.env.STORAGE_PATH, key);
+
+        // A grid fires one request per tile at once, and on a cold gallery
+        // every one of them misses the cache. Previously each carried
+        // `regenerate: true`, whose first act is to DELETE the target — so a
+        // later arrival unlinked the file an earlier one had already published
+        // and handed to a reader.
+        const watcher = [];
+        const poll = setInterval(() => watcher.push(fs.existsSync(abs)), 1);
+
+        const results = await Promise.all(
+          Array.from({ length: 12 }, () => imageProcessor.ensureThumbnailAtWidth(photo, 600))
+        );
+        clearInterval(poll);
+
+        expect(results.every((r) => r === key)).toBe(true);
+        expect(fs.existsSync(abs)).toBe(true);
+
+        // Once true, never false again: no window where a validated file is gone.
+        const firstSeen = watcher.indexOf(true);
+        if (firstSeen !== -1) {
+          expect(watcher.slice(firstSeen).every(Boolean)).toBe(true);
+        }
+      });
+
+      it('runs one generation for a burst of requests, not one per request', async () => {
+        const photo = await seedThumbPhoto();
+        const thumbDir = path.join(process.env.STORAGE_PATH, 'thumbnails');
+        await fs.promises.mkdir(thumbDir, { recursive: true });
+
+        // Counted through the staging files LocalFsStorage writes:
+        // `<key>.tmp.<pid>.<hex>`, one per put, each a distinct random suffix.
+        // So distinct temp names == distinct generations, which is the thing
+        // the dedupe is supposed to collapse. (Spying on generateThumbnail
+        // would not work — ensureThumbnailAtWidth calls it through the
+        // module-local binding, so an export spy never sees it.)
+        const seen = new Set();
+        const poll = setInterval(() => {
+          for (const f of fs.readdirSync(thumbDir)) {
+            if (f.includes('_w900_') && f.includes('.tmp.')) seen.add(f);
+          }
+        }, 1);
+
+        const results = await Promise.all(
+          Array.from({ length: 8 }, () => imageProcessor.ensureThumbnailAtWidth(photo, 900))
+        );
+        clearInterval(poll);
+
+        const abs = path.join(
+          process.env.STORAGE_PATH,
+          imageProcessor.thumbnailTierKeys(photo).find((k) => k.includes('_w900_'))
+        );
+        expect(fs.existsSync(abs)).toBe(true);
+        expect(new Set(results).size).toBe(1);
+        // 8 requests, at most one Sharp pass. Before the dedupe this was 8 —
+        // and on an external photo, 8 full reads of the original.
+        expect(seen.size).toBeLessThanOrEqual(1);
+      });
+
+      it('does not cache a failure — a later request retries', async () => {
+        const photo = await seedThumbPhoto();
+        // Source removed underneath: generation fails and must not poison the
+        // key for the lifetime of the process.
+        const src = path.join(process.env.STORAGE_PATH, 'events/active', photo.path);
+        const saved = await fs.promises.readFile(src);
+        await fs.promises.unlink(src);
+
+        expect(await imageProcessor.ensureThumbnailAtWidth(photo, 600)).toBeNull();
+
+        await fs.promises.writeFile(src, saved);
+        expect(await imageProcessor.ensureThumbnailAtWidth(photo, 600)).toContain('_w600_');
+      });
+    });
   });
 });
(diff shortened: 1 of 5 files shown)


the-luap:
## External review (Codex) — converged in 2 rounds

**Round 1** found one real defect, fixed in `6f5383b5`. **Round 2: clean** —
*"No actionable defects were found in the changed code."*

### The error path was sending the file's headers with a JSON body

The handler removed `Content-Length` and `ETag` but left everything else the route
had staged. Express does **not** overwrite a `Content-Type` that is already set, so:

- image routes emitted the 404 as `image/jpeg`
- zip routes emitted it as an `application/zip` attachment — a download that saves to
  disk as a corrupt archive

And the one that actually bites: **`Cache-Control` survived.** The image routes stage
`max-age=1800`, the hero route `3600`. So a 404 caused by the regeneration race — the
transient case this helper exists for — could be cached as a broken tile for **up to an
hour after the tier finished generating**. Precisely inverting the fix.

The error response now clears `Content-Type` and `Content-Disposition` and sets
`no-store`.

Verified against real Express rather than reasoned about: with a staged
`Content-Type: image/jpeg`, `res.status(404).json(...)` really does emit
`content-type: image/jpeg; charset=utf-8` and keep `max-age=1800`.

I had found the Content-Type and Cache-Control halves independently while re-reading
the helper; Codex added the `Content-Disposition` case, which I had missed.

31 tests passing across the two suites, ESLint clean.

PR #1134, opened 2026-08-21
fix(thumbnails): regenerate external photos instead of dropping their tiers (#1129)

### LINKED ISSUE #1129: [BUG] POST /admin/thumbnails/regenerate is a no-op for external/reference photos — and deletes their ?w= tiers before failing
**Version:** 3.111.0-beta.0 (self-hosted, Docker, Postgres). Code path is unchanged on `main` as of today.

## Summary

`POST /api/admin/thumbnails/regenerate` resolves the source image as `storage/events/active/<photo.path>`. For rows with `source_origin` `external` or `reference` that file does not exist — the originals live under the mount referenced by `events.external_path`. Every such photo fails the `fs.access` check and is counted as an error, so the canonical thumbnail is never rebuilt.

Because `deleteThumbnailTiers()` runs **before** that check (deliberately, per the comment, so S3 and external rows aren't skipped), the endpoint still drops the `?w=` tiers from #1095 / #1109. On an install where all photos are external, pressing the button therefore leaves the library in a worse state than before it ran: tiers gone, canonical renditions stale — and the admin UI reports success, since the response is sent before the background loop starts.

This is the same class of bug as #1078, on the thumbnail side.

## Code

`backend/src/routes/adminThumbnails.js` (L157 on `main`):

```js
const originalPath = path.join(storagePath, 'events/active', photo.path);
// ...
try {
  await fs.access(originalPath);
} catch (err) {
  logger.warn(`Original file not found for photo ${photo.id}: ${originalPath}`);
  errorCount++;
  continue;
}
const thumbnailPath = await generateThumbnail(originalPath, { regenerate: true });
```

The neighbouring `/regenerate-previews` (L218 ff. in the same file) already has the fix for this: it selects `source_origin` / `external_relpath` / `filename` and hands them to `ensurePreviewImage()`, which branches on them. Its comment even spells out why — *"without them every external photo looks managed here and generation is skipped"*.

`/regenerate` selects the exact same three columns, but its comment says they are only there for `deleteThumbnailTiers`, and the resolution below ignores them.

## Steps to reproduce

1. Event with `source_mode: reference`, photos imported with `source_origin: external`.
2. Change a thumbnail setting (e.g. `thumbnail_fit` from `cover` to `inside`).
3. Admin → Settings → Thumbnails → regenerate.

Result: one `Original file not found for photo <id>: /app/storage/events/active/<slug>/<file>` per photo. The `thumb_w600_*` / `thumb_w900_*` tiers are gone, the canonical `thumb_ext<id>_*` files are untouched and still carry the old settings.

Observed on 79/79 photos of one event; that install has 6564 external photos in total, so the endpoint is inert across the whole library.

## Suggested fix

`ensureThumbnail(photo)` in `imageProcessor.js` already does everything this route hand-rolls, and does it correctly for both source types: it resolves via `resolvePhotoFilePath()` / `resolvePhotoStorageKey()`, uses the per-photo `ext<id>_<basename>` output name so two events referencing the same NAS basename can't clobber each other, and writes `photos.thumbnail_path` back itself.

The only mismatch is that it short-circuits on `isThumbnailValid()` — which is exactly what `/regenerate-previews` neutralises by nulling the cached path. So the two endpoints can be made symmetrical:

```js
await deleteThumbnailTiers(photo);
const newThumbnailPath = await ensureThumbnail({ ...photo, thumbnail_path: null });
if (newThumbnailPath) successCount++; else errorCount++;
```

That also removes the route's own `fs.access` and `db('photos').update()`, both of which `ensureThumbnail` covers.

Two things worth carrying over from the previews route while touching this:

- **Skip videos.** `/regenerate` has no `media_type` filter, so a video row reaches `generateThumbnail()` and Sharp is handed a container file. `/regenerate-previews` filters them out in SQL; the select would need `media_type` / `mime_type` added either way.
- **Tiers are still only rebuilt lazily.** Even with the canonical fixed, the `?w=` tiers come back on the next gallery open — which is the burst described in #1128. Pre-warming them in the same background loop would close both, but that is a separate call.

## Why this surfaced (possible motivation for prioritising)

Migration 040 seeds `thumbnail_fit: 'cover'`, while the code default is `inside` (`DEFAULT_THUMBNAIL_FIT`). Under `cover` every thumbnail *and* every derived tier is square — verified on disk: 300x300 canonical, 900x900 for the `w900` tier, since `ensureThumbnailAtWidth()` scales the tier height from the configured 300/300 ratio.

The masonry-based layouts compute the slot from `photos.width`/`height` (which are correct in the DB), but the card renders the image as `width: 100%; height: auto`, so the tile ends up with the *file's* aspect ratio, not the photo's. The net effect is that `gallery-premium`, `masonry` and friends render every photo square and centre-cropped, and are visually indistinguishable from a fixed grid, on any install that kept the seeded default.

Getting out of that state means changing `thumbnail_fit` and regenerating — which is what runs into the bug above.

the-luap:
Closes #1129.

@BraynArts found this one and identified the fix; this is essentially their suggestion,
plus the video filter they flagged as worth carrying over.

## What was wrong

`/regenerate` resolved every source as `storage/events/active/<photo.path>` and
`fs.access`'d it. External and reference rows are not there — their originals live under
`events.external_path` — so every one failed the check and was counted as an error.

Inert would already be bad. What makes it **destructive** is that `deleteThumbnailTiers`
runs first, deliberately, so S3 and external rows are not skipped. On a reference install
the button therefore:

1. dropped every `?w=` tier, then
2. rebuilt nothing, and
3. reported success — the response is sent before the background loop starts

Leaving the library worse than before it ran: tiers gone, canonical renditions still
carrying the old settings. On the reporter's install that is all 6564 photos.

## The fix

The neighbouring `/regenerate-previews` already had the answer — hand the row to
`ensurePreviewImage` and let it branch on `source_origin`. This does the same with
`ensureThumbnail`, which:

- resolves both source kinds via `resolvePhotoFilePath` / `resolvePhotoStorageKey`
- uses the per-photo `ext<id>_<basename>` output name, so two events referencing one NAS
  basename cannot clobber each other
- writes `thumbnail_path` back itself

Nulling `thumbnail_path` is what stops it short-circuiting on `isThumbnailValid` — the
same trick the previews route uses, and necessary rather than cosmetic: the old thumbnail
is normally still perfectly readable at exactly the moment someone presses regenerate.

That retires the route's hand-rolled `fs.access` and its own `photos.update`, and with
them the last uses of `generateThumbnail`, `fs`, `path` and `getStoragePath` in the file.

**Videos** are now filtered out in SQL, matching the previews route. They have no
Sharp-derived thumbnail — `videoProcessor` makes a poster frame — so each video row could
only ever produce an error here.

## Verified

5 new integration tests through the real route with supertest, covering the external
rebuild, the `thumbnail_path: null` contract, tier deletion still running first, video
skipping, and per-event scoping.

**4 of the 5 fail against the current code** — checked by stashing the fix — so they
pin the regression rather than just describing it. (The fifth is tier deletion, which was
already correct and must stay that way.)

ESLint clean.

## Note on the reporter's third point

Tiers are still only rebuilt lazily, on the next gallery open. That burst is #1128, fixed
separately — it no longer crashes the backend and now costs one Sharp pass per tier
instead of one per tile. Pre-warming them in this loop would close both properly but is a
distinct change; happy to do it if you want it here.

### DIFF
--- a/backend/__tests__/integration/adminThumbnails.regenerate.test.js
+++ b/backend/__tests__/integration/adminThumbnails.regenerate.test.js
@@ -1,0 +1,173 @@
+/**
+ * POST /admin/thumbnails/regenerate for external/reference photos (#1129).
+ *
+ * The route used to resolve every source as `storage/events/active/<path>` and
+ * `fs.access` it. External and reference rows do not live there — their
+ * originals sit under `events.external_path` — so every one of them failed the
+ * check and was counted as an error.
+ *
+ * That alone would be inert. What made it destructive is that the tier
+ * deletion runs FIRST (deliberately, so S3 and external rows are not skipped):
+ * on a reference install the button dropped every ?w= tier and rebuilt
+ * nothing, while the UI reported success — the response is sent before the
+ * background loop starts.
+ *
+ * The background work is fired with setImmediate, so every assertion here has
+ * to wait for it to drain rather than trusting the response.
+ */
+
+const fs = require('fs');
+const path = require('path');
+const os = require('os');
+const express = require('express');
+const request = require('supertest');
+
+describe('admin thumbnail regeneration (#1129)', () => {
+  let tmpDir; let db; let cleanup; let app; let imageProcessor;
+
+  beforeAll(async () => {
+    tmpDir = await fs.promises.mkdtemp(path.join(os.tmpdir(), 'picpeak-regen-'));
+    process.env.NODE_ENV = 'test';
+    process.env.TEST_DATABASE_PATH = path.join(tmpDir, 'data', 'test.db');
+    process.env.STORAGE_PATH = path.join(tmpDir, 'storage');
+    await fs.promises.mkdir(path.dirname(process.env.TEST_DATABASE_PATH), { recursive: true });
+    await fs.promises.mkdir(process.env.STORAGE_PATH, { recursive: true });
+
+    jest.resetModules();
+
+    jest.doMock('../../src/middleware/auth', () => ({
+      adminAuth: (req, _res, next) => { req.admin = { id: 1, username: 'tester' }; next(); },
+    }));
+    jest.doMock('../../src/middleware/permissions', () => ({
+      requirePermission: () => (_req, _res, next) => next(),
+    }));
+    jest.doMock('../../src/services/imageProcessor', () => ({
+      ensureThumbnail: jest.fn().mockResolvedValue('thumbnails/thumb_ext1_shot.jpg'),
+      ensurePreviewImage: jest.fn().mockResolvedValue('previews/p.jpg'),
+      deleteThumbnailTiers: jest.fn().mockResolvedValue(undefined),
+      deletePreviewTiers: jest.fn().mockResolvedValue(undefined),
+    }));
+
+    // bootCrmDb, not run-migrations: the latter calls process.exit(0) on
+    // success, which ends the jest worker mid-suite.
+    ({ db, cleanup } = await require('./helpers/crmDb').bootCrmDb());
+
+    imageProcessor = require('../../src/services/imageProcessor');
+    app = express();
+    app.use(express.json());
+    app.use('/admin/thumbnails', require('../../src/routes/adminThumbnails'));
+  }, 180000);
+
+  afterAll(async () => {
+    if (cleanup) await cleanup();
+    await fs.promises.rm(tmpDir, { recursive: true, force: true }).catch(() => {});
+  });
+
+  beforeEach(async () => {
+    jest.clearAllMocks();
+    await db('photos').del();
+    await db('events').del();
+  });
+
+  async function seedEvent() {
+    const [row] = await db('events').insert({
+      slug: 'nas-wedding', event_type: 'wedding', event_name: 'nas',
+      event_date: '2026-01-01', host_email: 'h@example.com', admin_email: 'a@example.com',
+      password_hash: 'x', share_link: 'nas-share', expires_at: new Date().toISOString(),
+      source_mode: 'reference', external_path: 'weddings/2026-08',
+    }).returning('id');
+    return typeof row === 'object' ? row.id : row;
+  }
+
+  async function seedPhoto(eventId, overrides = {}) {
+    const [row] = await db('photos').insert({
+      event_id: eventId, filename: 'shot.jpg', path: 'nas-wedding/shot.jpg',
+      type: 'individual', ...overrides,
+    }).returning('id');
+    return typeof row === 'object' ? row.id : row;
+  }
+
+  /** The work runs in setImmediate; give it room to finish. */
+  const drain = () => new Promise((resolve) => setTimeout(resolve, 150));
+
+  it('rebuilds the canonical thumbnail for an external photo instead of erroring', async () => {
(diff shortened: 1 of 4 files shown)


the-luap:
## External review (Codex) — 3 rounds, 3 findings, all fixed

Each round found something the previous one had created or exposed. Worth reading in
order, because round 2 is the one that mattered.

### Round 1 — regeneration stranded a thumbnail per photo on S3 (`ba9b4e0e`)

For a managed photo on S3, `ensureThumbnail` downloads the source to a randomly-named
temp file, and `withProcessableImage` passes no `outputBasename` for non-RAW input — so
the storage key is derived from that random name and **differs on every run**. Nulling
`thumbnail_path`, which is what forces the rebuild, also hides the old key from
everything that would clean it up.

Being precise about blame: `ensureThumbnail` has always done this on S3. What this PR
changed is that the endpoint now *reaches* that code — before, it died at the
`fs.access` check and did nothing at all on S3. The fix makes it work; this stops it
leaking while it does. The old rendition is now removed when the key moved.

### Round 2 — a failed regeneration destroyed the good thumbnail (`07673bac`)

The sharpest finding of the three, and a **P1**.

`generateThumbnail` deleted the target **before sharp had even opened the source**, and
deleted it *again* in its catch block. For external photos the key is stable
(`ext<id>_<basename>`), so a source that could not be read — a NAS mount that blipped, a
corrupt file — left the previous rendition gone and the database still pointing at it.

Across a bulk regenerate that is not one photo, it is **the whole gallery**: press the
button while the mount is flapping and every canonical thumbnail in a reference install
is stripped. That is the exact "worse than before you pressed it" outcome this PR exists
to fix, so shipping it would have been a poor trade.

Both deletes are gone, and neither was doing anything a correct write does not:

- the pre-delete is redundant — `LocalFsStorage.put` stages to a temp file and renames
  over the target, and an S3 put overwrites by key. It only added a window with no
  thumbnail at all.
- the catch-delete claimed to clean up a partial upload, but there cannot be one: `put`
  is the **last** statement in the try, every throw above it happens before anything is
  written, and `put` unlinks its own temp file on failure. The only object it could
  remove is the previous, valid rendition.

This one applies to every caller of `generateThumbnail`, not just this endpoint — which
is the point.

### Round 3 — the round-1 fix could delete the file it just wrote (`559c9001`)

A **P1** against my own fix. The "key moved, delete the old one" branch compared raw
strings, but both backends canonicalise before addressing an object:
`LocalFsStorage._resolve` and `S3StorageBackend._key` each fold `\` to `/` and strip a
leading `./`. A legacy `thumbnail_path` of `thumbnails\thumb_x.jpg` is therefore the
*same object* as the POSIX key generation returns, while comparing unequal — so the
branch would delete the freshly written thumbnail and leave every regenerated row
pointing at nothing.

Now compared through the same canonicalisation. Covered for backslashes, a leading
`./`, and a doubled separator.

## Verification

13 tests across the two suites, ESLint clean.

Stated plainly: **the round-3 fix has not itself been through another external round** —
the loop stops at three. The 12 failures in `backupService.enhanced.test.js` are
pre-existing on `main` (a mock-shape issue), confirmed by running that suite on a clean
checkout.

PR #1135, opened 2026-08-21
fix(gallery): give masonry tiles their real shape back (#1130, #1131)

### LINKED ISSUE #1130: [BUG] gallery-premium discards the masonry tile height, so every tile takes the rendition's aspect ratio — square thumbnails collapse it into a fixed grid
**Version:** 3.111.0-beta.0 (self-hosted, Docker, Postgres). Unchanged on `main`.

## Summary

`GalleryPremiumLayout` lays its tiles out with `MasonryPhotoAlbum`, which computes each tile's `width` **and** `height` from `photos.width` / `photos.height` and hands both to the render prop. The layout's card then throws the height away:

```tsx
// GalleryPremiumLayout.tsx
height: _height,
// ...
// Note: height is passed but not used as we maintain aspect ratio via width
void _height;
// ...
<motion.div className="gallery-premium-photo-card …" style={{ width: '100%', height: 'auto', display: 'block' }}>
  <AuthenticatedImage … style={{ width, height: 'auto' }} className="w-full h-auto object-cover" />
```

With both the card and the image at `height: auto`, the tile's rendered height is decided by the **intrinsic aspect ratio of the file that happens to be served**, not by the photo's real dimensions. `GalleryPremiumLayout.css` already carries the rule that would do the right thing —

```css
.gallery-premium-photo-card img { width: 100%; height: 100%; object-fit: cover; }
```

— but the inline `height: 'auto'` overrides it.

That assumption ("the thumbnail has the photo's aspect ratio") does not hold: `thumbnail_fit` defaults to `cover` on any install created by migration 040, which seeds `'cover'` even though the code constant is `DEFAULT_THUMBNAIL_FIT = 'inside'`. Under `cover` the canonical rendition is `thumbnail_width` x `thumbnail_height` exactly — square on a default install — and `ensureThumbnailAtWidth()` scales the tier height from that same ratio, so every `?w=` tier is square too.

Net effect: **the masonry layout renders as a fixed square grid.** Every photo is centre-cropped to 1:1, portrait and landscape are indistinguishable, and the layout is visually identical to `grid`.

## Measured

Gallery of 79 photos, mixed orientation in the DB (`4240x2832` for 42 of them, `2832x4240` for 28, plus 9 other sizes). Measured in the live DOM at a 3-column desktop width:

```
card 469x469   img 469x469   img.naturalWidth/Height 900x900   (x79, identical)
```

Every one of the 79 tiles is the same square. The `w900` tier on disk is `900x900`; the canonical is `300x300`.

## Scope

`gallery-premium` is the only one of the seven layouts that does this. The other six set an explicit container height and use `h-full`:

| Layout | image sizing |
|---|---|
| grid, masonry, timeline | explicit height / `aspect-square` + `h-full` |
| justified, mosaic | explicit height + `h-full` |
| story | own CSS |
| **gallery-premium** | **`height: auto` on both card and image** |

So the other layouts still show the crop from square renditions, but at least keep the correct tile geometry. `gallery-premium` loses the geometry as well.

## Suggested fix

Use the height the render prop already provides, and let the existing CSS rule apply:

```tsx
<motion.div className="gallery-premium-photo-card …" style={{ width: '100%', height, display: 'block' }}>
  <AuthenticatedImage … className="w-full h-full object-cover" />   // drop the inline height:'auto'
```

`object-fit: cover` is already in the stylesheet, so a square rendition would then be cropped into a correctly shaped tile rather than dictating the tile's shape. That makes the layout correct independently of `thumbnail_fit`, which seems preferable to depending on a setting.

## Related

- Whether `cover` is the right thing to seed in migration 040 is a separate question — but note that changing it means regenerating, and that path is currently broken for external/reference installs (#1129).
- The same gallery on a `masonry` layout is broken differently, by the bundled CSS templates — filed separately.

## Environment

- PicPeak 3.111.0-beta.0, Docker Compose, PostgreSQL
- Chrome 141, desktop, 3-column breakpoint
- `thumbnail_fit: "cover"`, `thumbnail_width/height: 300/300` (untouched defaults from migration 040)

### LINKED ISSUE #1131: [BUG] Bundled CSS templates force `.photo-card img { height: 200px }`, overriding every layout's h-full — the default template breaks masonry/justified/mosaic
**Version:** 3.111.0-beta.0 (self-hosted, Docker, Postgres).

## Summary

Two of the three seeded rows in `css_templates` — including the one flagged `is_default = true` — contain:

```css
.photo-card img {
  width: 100%;
  height: 200px;      /* ← */
  object-fit: cover;
  transition: transform 0.3s ease;
}
```

Six of the seven gallery layouts size their tile by putting an explicit pixel height on the card and letting the image fill it with `h-full` (`height: 100%`). `.photo-card img` has specificity (0,1,1) and beats `.h-full` at (0,1,0), so the template pins every image to 200px inside a card whose height is still computed correctly from `photos.width` / `photos.height`.

The result is a correctly-shaped card with a 200px image glued to its top edge and empty card background below — or, when the computed card is shorter than 200px, an image taller than its own container.

## Measured

`masonry` layout, `masonryMode: columns`, 79 photos of mixed orientation, live DOM at a 3-column desktop width:

```
card 373x326   img 371x200      ← 126px of empty card below the image
card 373x145   img 371x200      ← image is 55px TALLER than its card
```

`img.className` is `w-full h-full object-cover rounded-lg` in every case; computed `height` is `200px` in every case. Resolving the cascade on the image confirms the winner:

```
img, video              { height: auto }     index-*.css
.h-full                 { height: 100% }     index-*.css
.photo-card img         { height: 200px }    <style id="gallery-custom-css">   ← applied
```

## Affected templates

All three seeded at the same timestamp, so these ship with PicPeak rather than being user-authored:

| id | name | `.photo-card img` fixed height |
|----|------|-------------------------------|
| 1 | Elegant Dark | **yes** (`is_default = true`) |
| 2 | Apple Liquid Glass | no |
| 3 | Liquid Glass Dark | **yes** |

Because template 1 is the default, this is the out-of-the-box result for anyone who picks a layout other than `grid`/`timeline` (where a fixed square happens to look intentional).

## Steps to reproduce

1. Event → CSS template "Elegant Dark".
2. Theme → gallery layout `masonry` (or `justified` / `mosaic`).
3. Open the gallery.

Expected: images fill their aspect-ratio-sized cards.
Actual: every image is 200px tall, top-aligned, with the card's background visible below it.

## Suggested fix

`height: 100%` instead of `height: 200px` in templates 1 and 3. The layouts already guarantee a definite height on `.photo-card`, and `object-fit: cover` is on the same rule, so nothing else needs to change. A fixed pixel height in a template can't be right for any of the aspect-ratio layouts — it only ever matched the old fixed-height grid.

Worth considering as well: the templates are applied via a `<style>` block that wins over the bundle by specificity, so any future layout that relies on utility classes is exposed to the same collision. Scoping template rules to something the layouts don't style (or shipping them as CSS variables rather than element rules) would make that class of breakage impossible.

## Secondary, same cause

The two dark templates predate "People in this gallery" (#1074) and don't style it. On a dark background the strip's heading and the per-person photo counts render near-black on near-black and are effectively invisible. Same fix location, if the templates are being touched anyway.

## Environment

- PicPeak 3.111.0-beta.0, Docker Compose, PostgreSQL
- Chrome 141, desktop
- Layouts confirmed affected: `masonry` (measured), plus `justified` / `mosaic` / `grid` / `timeline` by inspection — all use `h-full`. `gallery-premium` is unaffected by this rule because it sets its own inline height, which is its own bug (filed separately).

the-luap:
Closes #1130. Closes #1131.

Two independent causes of the same symptom — an aspect-ratio layout that does not lay
anything out — so they are fixed together. Both reported by @BraynArts with live-DOM
measurements.

## #1130 — gallery-premium threw the tile height away

`MasonryPhotoAlbum` computes width **and** height from `photos.width`/`height` and hands
both to the render prop. The card took the width, `void`ed the height, and set
`height: 'auto'` inline on both itself and the image — so the rendered shape came from the
intrinsic ratio of whatever rendition happened to be served.

That assumption does not hold: `thumbnail_fit` is seeded `'cover'` by migration 040 (the
code constant is `inside`), which makes the canonical rendition exactly
`thumbnail_width × thumbnail_height` — square on a default install — and
`ensureThumbnailAtWidth` scales each `?w=` tier from the same ratio. So every tier is
square too, and the masonry drew identical squares.

The card now uses the height it is given, and the inline `height: auto` comes off the
image so the stylesheet's existing `.gallery-premium-photo-card img { height: 100% }` can
finally apply. The `object-fit: cover` that was already there then crops a square
rendition **into** a correctly-shaped tile, instead of the rendition dictating the tile's
shape — which makes the layout correct regardless of `thumbnail_fit` rather than dependent
on it.

### Before / after

Measured at a 3-column desktop width, on a gallery whose photos carry mixed orientations
in the DB:

| | tile shapes | distinct |
|---|---|---|
| before | `365x365` ×8 | **1** |
| after | `365x546`, `365x365`, `365x244`, … | **3** |

Screenshots in the first comment.

> The gallery I had locally is all square source files, so I set mixed
> `photos.width/height` on it for the demo and restored them afterwards (verified
> byte-identical). That is faithful to what is being tested: the entire point of the fix
> is that tile geometry comes from the **DB dimensions**, not from the file.

## #1131 — the bundled templates pinned every image to a fixed height

`.photo-card img { height: 200px }` has specificity (0,1,1) and beats the `.h-full`
utility (0,1,0) that six of the seven layouts use to fill a card whose height they
computed correctly. Result: a right-shaped card with a 200px image glued to its top edge
and background below — or, where the computed card was shorter than 200px, an image
taller than its own container.

"Elegant Dark" is seeded `is_default = true`, so this was the out-of-the-box result for
anyone choosing any layout other than grid/timeline. "Liquid Glass Dark" carried the same
rule at **240px**, plus **180px** in a mobile media query (the issue says 200px for both;
the values differ, the defect does not). "Apple Liquid Glass" is clean.

Migrations 052/053 are corrected for fresh installs, and **181** repairs the rows already
seeded. Templates are referenced by `events.css_template_id` and read at serve time rather
than copied onto the event, so fixing the row fixes every gallery using it immediately.

The repair rewrites the **whole rule bodies we shipped**, not every `height: <n>px` — so
the 1px gradient divider and the 8px scrollbar in those same stylesheets survive, and a
fixed height a user wrote themselves is left alone.

## Verified

- 6 migration tests: both templates fixed, divider and scrollbar untouched, a
  user-authored template untouched, idempotent, null-safe, and a no-op before the table
  exists.
- End-to-end against the **real** seeds, replayed from git so it exercises what existing
  installs actually have:
  ```
  BEFORE broken: [ 'Elegant Dark', 'Liquid Glass Dark' ]
  181: relaxed the fixed image height in 2 CSS template(s)
  AFTER  broken: []
  divider 1px kept: true | scrollbar 8px kept: true
  ```
- Frontend: `tsc` clean, 198 tests passing, build clean. Migration suite 17 passing.

## Left for later

The two dark templates predate #1074 and do not style the People strip, so its heading and
per-person counts render near-black on near-black — the reporter's secondary note. Fixing
it properly means the templates carrying the gallery theme tokens, which is a design
decision rather than a typo. Worth its own issue; say the word and I will file it.

### DIFF
--- a/backend/__tests__/migrations/181_fix_css_template_photo_height.test.js
+++ b/backend/__tests__/migrations/181_fix_css_template_photo_height.test.js
@@ -154,6 +154,33 @@ describe('migration 181 — CSS template image height (#1131)', () => {
     expect(css).toContain('.hero { height: 400px; }');
   });
 
+  it('does not rewrite other properties that merely end in -height', async () => {
+    // `line-height: 200px` contains `height: 200px` as a substring, so an
+    // unanchored pattern silently rewrites it — in a migration that cannot be
+    // undone.
+    const mine = [
+      '.photo-card img {',
+      '  line-height: 200px;',
+      '  max-height: 300px;',
+      '  min-height: 14px;',
+      '  --tile-height: 220px;',
+      '  height: 200px;',
+      '}',
+    ].join('\n');
+    await knex('css_templates').insert({ name: 'Adjacent Props', css_content: mine });
+
+    await migration.up(knex);
+
+    const css = await contentOf('Adjacent Props');
+    expect(css).toContain('line-height: 200px');
+    expect(css).toContain('max-height: 300px');
+    expect(css).toContain('min-height: 14px');
+    expect(css).toContain('--tile-height: 220px');
+    // Only the real one moved.
+    expect(css).toContain('height: 100%');
+    expect(css).not.toMatch(/(?<![\w-])height:\s*200px/);
+  });
+
   it('leaves non-pixel heights on the image alone', async () => {
     const mine = '.photo-card img { height: 50vh; }\n.photo-card img { height: auto; }';
     await knex('css_templates').insert({ name: 'Relative', css_content: mine });
--- a/backend/__tests__/migrations/181_fix_css_template_photo_height.test.js
+++ b/backend/__tests__/migrations/181_fix_css_template_photo_height.test.js
@@ -116,13 +116,51 @@ describe('migration 181 — CSS template image height (#1131)', () => {
     expect(css).toContain('height: 8px');
   });
 
-  it('does not touch a template the user wrote themselves', async () => {
+  /**
+   * The case that forced the scope wider. `sanitizeCSS` strips control
+   * characters, so any template ever saved through the editor — including a
+   * save that only changed its name — has had every newline REMOVED. An
+   * exact-text migration finds nothing on those installs, is recorded as
+   * applied, and leaves them broken permanently.
+   */
+  it('fixes a template that has been through the editor, newlines and all', async () => {
+    const { sanitizeCSS } = require('../../src/utils/cssSanitizer');
+    const { sanitized } = sanitizeCSS(ELEGANT_DARK);
+    // Precondition: the sanitizer really did flatten it.
+    expect(sanitized).not.toContain('\n');
+    expect(sanitized).toContain('height: 200px');
+    await knex('css_templates').insert({ name: 'Saved Once', css_content: sanitized });
+
+    await migration.up(knex);
+
+    const css = await contentOf('Saved Once');
+    expect(css).not.toContain('200px');
+    expect(css).toContain('height: 100%');
+  });
+
+  it('relaxes a user-authored fixed height too, but only on .photo-card img', async () => {
+    // Deliberately broader than the seeded text — see the migration header. A
+    // pixel height on the image cannot be right under any of the seven
+    // layouts, whoever wrote it; a height anywhere else is none of our
+    // business.
     const mine = '.photo-card img {\n  height: 220px;\n}\n.hero { height: 400px; }';
     await knex('css_templates').insert({ name: 'My Own', css_content: mine });
 
     await migration.up(knex);
 
-    expect(await contentOf('My Own')).toBe(mine);
+    const css = await contentOf('My Own');
+    expect(css).toContain('height: 100%');
+    expect(css).not.toContain('220px');
+    expect(css).toContain('.hero { height: 400px; }');
+  });
+
+  it('leaves non-pixel heights on the image alone', async () => {
+    const mine = '.photo-card img { height: 50vh; }\n.photo-card img { height: auto; }';
+    await knex('css_templates').insert({ name: 'Relative', css_content: mine });
+
+    await migration.up(knex);
+
+    expect(await contentOf('Relative')).toBe(mine);
   });
 
   it('is idempotent and safe on a row with no CSS', async () => {
--- a/backend/__tests__/migrations/181_fix_css_template_photo_height.test.js
+++ b/backend/__tests__/migrations/181_fix_css_template_photo_height.test.js
@@ -181,6 +181,32 @@ describe('migration 181 — CSS template image height (#1131)', () => {
     expect(css).not.toMatch(/(?<![\w-])height:\s*200px/);
   });
 
+  it('handles a grouped selector list', async () => {
+    // Requiring `{` straight after `img` skipped these entirely — and the
+    // migration is still recorded as applied, so the template kept the bug.
+    const mine = '.photo-card img, .thumbnail img {\n  height: 200px;\n}';
+    await knex('css_templates').insert({ name: 'Grouped', css_content: mine });
+
+    await migration.up(knex);
+
+    const css = await contentOf('Grouped');
+    expect(css).toContain('.photo-card img, .thumbnail img {');
+    expect(css).toContain('height: 100%');
+    expect(css).not.toContain('200px');
+  });
+
+  it('skips a nested rule rather than rewriting the wrong declaration', async () => {
+    // Valid nested CSS that passes the validator. A brace-greedy body would
+    // capture the inner block and rewrite the CAPTION's height, which cannot
+    // be undone. Leaving it untouched is the lesser evil.
+    const mine = '.photo-card img {\n  & + .caption { height: 200px; }\n}';
+    await knex('css_templates').insert({ name: 'Nested', css_content: mine });
+
+    await migration.up(knex);
+
+    expect(await contentOf('Nested')).toBe(mine);
+  });
+
   it('leaves non-pixel heights on the image alone', async () => {
     const mine = '.photo-card img { height: 50vh; }\n.photo-card img { height: auto; }';
     await knex('css_templates').insert({ name: 'Relative', css_content: mine });
--- a/backend/__tests__/migrations/181_fix_css_template_photo_height.test.js
+++ b/backend/__tests__/migrations/181_fix_css_template_photo_height.test.js
@@ -1,0 +1,151 @@
+/**
+ * Repairing the bundled templates' fixed image height (#1131).
+ *
+ * The risk in a migration that rewrites user-visible CSS is doing too much,
+ * so most of what is pinned here is what it must NOT touch: the other pixel
+ * heights inside the very same templates (a 1px divider, an 8px scrollbar),
+ * and any rule a user wrote themselves.
+ */
+
+const path = require('path');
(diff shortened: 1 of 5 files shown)


the-luap:
## External review (Codex) — 3 rounds, 4 findings fixed, 1 retracted

All four fixes are in migration 181. The frontend change came through every round
unchallenged except once, below.

### Round 1 [P1] — the migration would have no-opped on most real installs (`70fb90ee`)

`sanitizeCSS` strips all control characters (`cssSanitizer.js:61`), so the moment a
template is saved through the editor — **including a save that only renamed it** — every
newline is removed from the stored CSS. The shipped
`.photo-card img {\n  height: 200px;` becomes `.photo-card img {  height: 200px;`.

Exact-text matching therefore found nothing on those installs, and the migration would
still be recorded as applied: broken permanently, with no second chance. Verified
against the real sanitizer.

Now a whitespace-tolerant rewrite scoped to `.photo-card img` rule bodies.

**This widened the scope, so stating it plainly rather than burying it:** a fixed pixel
height a user wrote themselves is now rewritten too. There is no layout it can be right
for — all seven give `.photo-card` a definite height and expect the image to fill it — so
a pixel height on the image can only detach it from its card. Non-pixel values (%, vh,
auto) and anything outside a `.photo-card img` body are untouched.

### Round 2 [P2] — the pattern rewrote properties that merely end in `-height` (`6c34863c`)

Unanchored, so it matched the tail of others:

```
line-height: 200px    -> line-height: 100%
max-height: 300px     -> max-height: 100%
min-height: 14px      -> min-height: 100%
--tile-height: 220px  -> --tile-height: 100%
```

In a migration whose `down()` is deliberately irreversible, which makes it worse than an
ordinary bug. Fixed with a lookbehind.

### Round 3 [P2 ×2] — grouped selectors and nested rules (`680cda9b`)

- `.photo-card img, .thumbnail img { … }` was **skipped**, because the pattern required
  `{` straight after `img`. Now a selector list.
- `.photo-card img { & + .caption { height: 200px } }` had the inner block captured, so
  the **caption's** height was rewritten instead. Excluding braces from the body means
  such a rule is now skipped entirely — the template keeps its fixed image height, which
  is the lesser evil against corrupting a declaration `down()` cannot restore.

Also resets the `/g` regex's `lastIndex` around the `.test()` call; a stateful regex
tested per rule body otherwise resumes from an arbitrary offset and silently skips
declarations in later rules.

### Retracted [P1] — double-cropping in gallery-premium

Codex argued the ratio-shaped tile crops an already-square rendition a second time, and
proposed feeding these tiles a preview rendition instead. I pushed back: these are grid
tiles, tens to hundreds per page, and the `?w=` thumbnail tiers exist precisely so a
phone does not pull preview-sized images per tile (#1095/#1109) — that swap is a far
larger regression than extra cropping. It retracted.

**But it conceded a real cost, and it is worth knowing before merge:** for a 2:1 photo on
a `thumbnail_fit: cover` install, `main` shows the whole square rendition while this
branch crops its top and bottom — so a subject near either edge can be lost. That is the
trade the issue author explicitly asked for ("correct independently of `thumbnail_fit`"),
and the other six layouts already show that same crop. The real remedy is
`thumbnail_fit: inside` plus a regeneration — which #1134 is what makes possible for
external installs.

## Verification

11 migration tests. Re-verified end to end against the **real** seeds in both shapes,
pristine and sanitizer-flattened:

```
[pristine]  before=["Elegant Dark","Liquid Glass Dark"] after=[] divider1px=true scrollbar8px=true
[sanitized] before=["Elegant Dark","Liquid Glass Dark"] after=[] divider1px=true scrollbar8px=true
```

Frontend `tsc`, tests and build clean.

Stated plainly: **the round-3 fix has not itself been through another external round** —
the loop stops at three.

PR #1136, opened 2026-08-22
docs(faces): link the face-recognition guidance from where people look (#1125)

### LINKED ISSUE #1125: [DOCS] Nothing in the repo links to the face-recognition docs page — README points at the sidecar README, .env.example points at a file that never existed
**What documentation needs improvement?**
- [x] README.md
- [x] Other: `.env.example`, `docs/single-container.md`, `ml/README.md`, admin UI

## Summary

The face-recognition docs page **already exists** and is in the sidebar as
*Features → People in Galleries*:
<https://docs.picpeak.app/features/face-recognition>

Nothing in this repository links to it. Every other feature routes readers to
`docs.picpeak.app`; face recognition is the one feature where the repo either
points somewhere else, or points at a file that has never existed.

The result: the feature with the highest "read before you enable this" burden of
anything PicPeak ships — it starts biometric processing that is GDPR Art. 9
special-category data in the EU — is also the one where a self-hoster following
the repo's own breadcrumbs never reaches the guidance that was written for them.

## The concrete gaps

**1. `.env.example:252` points at a file that does not exist.** In the middle of
the Art. 9 warning:

```
# Face embeddings are biometric data (GDPR Art. 9 special category in the EU).
# The photographer is the controller and needs a lawful basis for the people
# in their photos — see docs/feature-face-recognition.md before enabling.
```

There is no `docs/feature-face-recognition.md` in this repo, and there never has
been (`docs/` contains only `migration-to-org.md`, `single-container.md` and the
screenshots). This is the single pointer to legal guidance an operator gets while
they are literally editing the variables that turn the feature on, and it is a
dead path. Should be <https://docs.picpeak.app/features/face-recognition>.

**2. `README.md:114` links the sidecar's developer README, not the user docs.**

> …smart search, **People in this gallery** face grouping (opt-in per gallery,
> needs the optional [ML sidecar](ml/README.md))…

`ml/README.md` is the right document for *its* audience — HTTP contract,
`X-Face-ML-Token`, the YuNet/FaceNet-512 licence rationale, how to convert
`facenet512.onnx`. It is not what a photographer clicking "face grouping" is
looking for, and it is silent on the parts that matter to them (the feature flag,
the per-gallery toggle, consent, the people bar, auto-categories).

The same two lines link Live Slideshow, storage backends and webhooks to
`docs.picpeak.app`, so this reads as an oversight rather than a choice. The
relative path also breaks wherever the README is rendered outside the GitHub repo
view (GHCR package page, mirrors).

**3. `README.md:136-148` — no row in the documentation table.** The table lists
Live Slideshow, Webhooks, Storage backends, Backup & Restore, CRM & Accounting.
Face recognition is missing, even though its docs page is already in the nav.

**4. `docs/single-container.md:223`** correctly says "No face recognition" and
explains why, but does not link the feature page for readers who want it and now
need to know they must use the multi-container deployment.

**5. `ml/README.md`** has no backlink to the user-facing page either, so the one
face-related document the README does reach is a dead end in the other direction.

**6. No in-app link.** `frontend/src/components/admin/FaceRecognitionCard.tsx`
renders the consent callout and the "Detect people in this gallery" toggle with no
"Learn more" link — grep for `docs.picpeak.app` in that file returns nothing.
The consent callout is exactly where a link belongs.

## Suggested fix

All in this repo — the docs page itself needs no change:

- [ ] `.env.example` — replace `docs/feature-face-recognition.md` with
      `https://docs.picpeak.app/features/face-recognition`
- [ ] `README.md:114` — link the feature name to the docs page; keep
      `ml/README.md` as the secondary "sidecar internals" link if wanted, but use
      the absolute `https://github.com/PicPeak/picpeak/blob/main/ml/README.md`
      form so it survives outside the repo view
- [ ] `README.md` docs table — add a row, e.g.
      `| 🙂 People in galleries (face grouping) | docs.picpeak.app/features/face-recognition |`
- [ ] `docs/single-container.md:223` — link the page from the "No face
      recognition" bullet
- [ ] `ml/README.md` — one line at the top: "Looking for how to use the feature?
      → docs.picpeak.app/features/face-recognition. This document is the sidecar's
      HTTP contract."
- [ ] Optional: a "Learn more" link in the consent callout of
      `FaceRecognitionCard.tsx` (needs an `en`/`de` string pair)

## Acceptance criteria

Someone who has never seen the feature can reach
`docs.picpeak.app/features/face-recognition` from the README, from
`.env.example`, and from the admin card, and no reference in the repo points at a
path that does not resolve.

## Notes

- Related, but separate: #1106 (dark-mode styling), #1107 and #1096 (people
  manager UX). This one is docs/links only, no behaviour change.
- Line numbers are against `main` at the time of writing.

the-luap:
Closes #1125.

All six pointers, verified against the tracked tree rather than my working copy — which
matters, because the two differ here in a way that changes the fix.

## The `.env.example` path: don't "fix" it by creating the file

The issue is right that `docs/feature-face-recognition.md` does not exist, and right
that `docs/` holds only `migration-to-org.md`, `single-container.md` and the
screenshots. Worth adding *why*, so nobody closes this by writing the file:

```
$ git check-ignore -v docs/feature-external-media.md
.gitignore:89:docs/feature-*.md    docs/feature-external-media.md

$ git ls-files docs/ | wc -l
7
```

`docs/feature-*.md` is **gitignored**. My checkout has eight such files sitting there as
local scaffolding — invisible to anyone who clones. So the reference could never have
resolved for a reader, and creating the file would produce a doc git silently drops. The
absolute URL is the only thing that works.

I also checked whether this was one instance of a wider pattern:

```
$ git grep -nE "docs/(feature-|test-|ENVIRONMENT_VARIABLES|DATABASE_SCHEMA|...)" -- . ':(exclude).gitignore'
.env.example:252: ...
```

One hit — the one reported. No cleanup hiding behind it.

## What changed

| # | Where | |
|---|---|---|
| 1 | `.env.example` | dead path → the docs URL, inside the Art. 9 warning |
| 2 | `README.md` | feature name → docs page; sidecar link kept but made **absolute**, so it survives on the GHCR package page and mirrors |
| 3 | `README.md` | new row in the documentation table |
| 4 | `docs/single-container.md` | the "No face recognition" bullet now says where to read what you would be getting |
| 5 | `ml/README.md` | pointer at the top, so the one face doc the README reached is no longer a dead end in the other direction |
| 6 | `FaceRecognitionCard.tsx` | "Read the guidance" in the consent callout, `en` + `de` |

## On item 6

The checklist marks it optional but the acceptance criteria requires reaching the page
"from the admin card", so I took it as in scope.

Placed **inline at the end of the consent sentence**, not as a separate help link — that
callout is the one screen where someone is about to start Art. 9 processing, and the
link belongs next to the obligation rather than filed under help. `target="_blank"` with
`rel="noopener noreferrer"`.

Screenshot in the first comment.

## Verified

- Both URLs added return **200** (`docs.picpeak.app/features/face-recognition`,
  the absolute `ml/README.md`), checked live by extracting every `https://` added in the
  diff and curling it.
- No tracked file still references a gitignored `docs/` path.
- The rendered link carries the right `href`, `target` and `rel`, read back from the live
  DOM rather than the source.
- Frontend: `tsc` clean, ESLint clean, 198 tests passing, build clean.
- Locale files: one key each, no reformatting of the surrounding JSON.

## Note on merge order

Item 6 touches the consent callout in `FaceRecognitionCard.tsx`, which #1126 also edits
(adding `dark:` variants to that same element). Both are small and in different parts of
the block, so whichever lands second should merge cleanly or need a one-line resolution —
flagging it rather than having you discover it.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -249,7 +249,8 @@ LOGS=./logs
 #
 # Face embeddings are biometric data (GDPR Art. 9 special category in the EU).
 # The photographer is the controller and needs a lawful basis for the people
-# in their photos — see docs/feature-face-recognition.md before enabling.
+# in their photos — read https://docs.picpeak.app/features/face-recognition
+# before enabling.
 #
 # NOT AVAILABLE ON THE ALL-IN-ONE IMAGE. The single-container build sets
 # PICPEAK_SINGLE_CONTAINER=true and the backend refuses to enable face
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ Unlike expensive SaaS solutions, PicPeak gives you:
 
 **For photographers** — drag & drop upload, auto-expiring & password-protected galleries, automated emails, an analytics dashboard, custom themes, a public landing page, and a [Live Slideshow](https://docs.picpeak.app/features/live-slideshow) projector view that auto-picks-up new uploads during live events.
 
-**For clients** — clean mobile-optimized galleries, one-click bulk downloads, smart search, **People in this gallery** face grouping (opt-in per gallery, needs the optional [ML sidecar](ml/README.md)), optional guest uploads, and download protection (watermarking + right-click prevention).
+**For clients** — clean mobile-optimized galleries, one-click bulk downloads, smart search, **[People in this gallery](https://docs.picpeak.app/features/face-recognition)** face grouping (opt-in per gallery, needs the optional [ML sidecar](https://github.com/PicPeak/picpeak/blob/main/ml/README.md)), optional guest uploads, and download protection (watermarking + right-click prevention).
 
 **Technical** — Docker-ready, automatic thumbnail generation, external media reference mode, smart archiving of expired galleries, S3-compatible [storage backends](https://docs.picpeak.app/features/storage-backends), [webhooks](https://docs.picpeak.app/features/webhooks), and security-first defaults (JWT, rate limiting, CORS).
 
@@ -139,6 +139,7 @@ Full documentation lives at **[docs.picpeak.app](https://docs.picpeak.app)** —
 | ⚙️ Admin settings reference | [docs.picpeak.app/guides/admin-settings](https://docs.picpeak.app/guides/admin-settings) |
 | 🎯 Creating events | [docs.picpeak.app/guides/creating-events](https://docs.picpeak.app/guides/creating-events) |
 | 📽️ Live Slideshow | [docs.picpeak.app/features/live-slideshow](https://docs.picpeak.app/features/live-slideshow) |
+| 🙂 People in galleries (face grouping) | [docs.picpeak.app/features/face-recognition](https://docs.picpeak.app/features/face-recognition) |
 | 💾 Backup & Restore | [docs.picpeak.app/guides/backup-restore](https://docs.picpeak.app/guides/backup-restore) |
 | 🔌 API reference | [docs.picpeak.app/api](https://docs.picpeak.app/api) |
 | 🪝 Webhooks | [docs.picpeak.app/features/webhooks](https://docs.picpeak.app/features/webhooks) |
--- a/docs/single-container.md
+++ b/docs/single-container.md
@@ -148,4 +148,6 @@ docker inspect --format='{{.State.Health.Status}}' picpeak
   thumbnail and preview generation in one container would slow everything
   down rather than fail cleanly. The toggle in Settings → Features is
   disabled here and says so. Use the multi-container deployment if you want
-  it.
+  it — see
+  [People in galleries](https://docs.picpeak.app/features/face-recognition)
+  for what the feature does and what enabling it commits you to.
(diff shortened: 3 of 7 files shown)


the-luap:
## External review (Codex) — clean on the first round

> *"The changes consistently add valid documentation links and localized UI text
> without introducing functional regressions."*

No findings, so nothing was changed after review — converged at round 1.

Two things I checked myself while it ran, since a docs PR's one behavioural hunk is the
place a problem would hide:

**The new link matches how this codebase already does external links.** Not asserted —
grepped:

- `target="_blank"` + `rel="noopener noreferrer"`: same as `AdminSidebar`,
  `MigrationBanner`, `PicpeakBackupCard`, `UpdateNotification`
- the `ExternalLink` icon: already used in three other admin components
- **no** "opens in a new tab" announcement: zero of the existing external links have one,
  so adding it here would have been the inconsistent choice rather than the careful one

**Re-checked both URLs at review time**, not just when I wrote them — external links are
the part of a docs PR most likely to rot between authoring and merge:

```
https://docs.picpeak.app/features/face-recognition        200
https://github.com/PicPeak/picpeak/blob/main/ml/README.md 200
```

`tsc` clean, ESLint clean, 198 frontend tests passing.

Issue #839, opened 2026-07-18
[FEATURE] Emoji reactions on photos (guest feedback)


the-luap:
## Motivation

SayCheese's "Engagement Mode" offers emoji reactions + upvotes on guest photos. PicPeak's feedback system currently supports likes, 1–5 star ratings, comments and favorites — but only a single "like" as a lightweight reaction. Emoji reactions are a small, high-visibility upgrade for the guest-event use case where star ratings feel too formal.

## Scope

- Per-photo **emoji reactions** with a small fixed set (e.g. ❤️ 😂 😍 👏 🎉 — final set TBD), one reaction per guest per photo (changeable), stored alongside the existing feedback types.
- Reaction counts per emoji shown on the photo (grid overlay and/or lightbox), consistent with how like counts are displayed today.
- Admin controls in the existing feedback settings: `allow_reactions` toggle next to `allow_likes` / `allow_ratings` / `allow_comments`; respects existing `show_to_guests` and guest-identity modes.
- Included in the engagement metrics on the admin feedback views.

## Non-goals

- Free-form/custom emoji picker (full emoji set) — fixed curated set only.
- Reactions on comments.

## Implementation notes

- Extend the feedback tables/service (`feedbackService.js`, `galleryFeedback.js`) with a `reaction` feedback type + emoji value; per-guest uniqueness mirrors the existing like/favorite constraints (guest identity infrastructure already exists).
- Moderation: reactions are emoji-only, so the word filter doesn't apply; still respect feedback moderation settings for visibility if trivial to do.
- Frontend: gallery photo overlay + lightbox reaction bar; i18n strings for the 8 supported locales.

Issue #520, opened 2026-05-18
An option to enable Guest Feedback By default


Rekoo-PS:
> **+ An option to enable Guest Feedback By default 😅🥲**
> smth like this  : 
> admin panel > General >  Event Creation page .
> Guest Feedback Default Settings : (OFF-ON)
> If On > then show its settings to save as default  >
> 
> <img width="1075" height="844" alt="Image" src="https://github.com/user-attachments/assets/332ff078-2ce8-425c-b2a4-a79bd5d56625" />
>  

 _Originally posted by @Rekoo-PS in [#506](https://github.com/the-luap/picpeak/issues/506#issuecomment-4466707808)_

the-luap:
Implemented in #527 alongside the lightbox fix from #518. Mirrors the existing `event_default_require_password` pattern (#317) — same five-file shape end-to-end: admin toggle under Settings → Events, applied as the initial "Guest Feedback enabled" value on the Create Event form when public settings load. Defaults to off so existing installs don't change behaviour until you flip it.

Kept it to just the master `feedback_enabled` toggle, not the sub-toggles (likes/ratings/comments) — those already default to true in the form, so flipping the master on gives sensible behaviour immediately without a second setting to manage.

PR #1137, opened 2026-08-22
feat(gallery): colour labels for client proofing, and one global default per feedback type (#1044)

### LINKED ISSUE #1044: Color labels for client proofing (Lightroom-compatible colors, keyboard shortcuts, filtering)
Feature request from discussion #1027, started by @boergu: https://github.com/PicPeak/picpeak/discussions/1027

Coming from PicDrop, @boergu describes a color-label proofing workflow as the one core feature blocking a full switch to PicPeak. This issue turns that discussion into an actionable feature request, with a few additions for Lightroom compatibility.

## 1. Color labels for client selections

Clients can mark images with color labels while proofing (e.g. 1st choice / 2nd choice / rejected).

**Use Lightroom's color label set: Red, Yellow, Green, Blue, Purple.** Sticking to Lightroom's exact colors keeps the door open to later exporting/displaying the same labels in Lightroom (e.g. via XMP color label metadata), so a client selection in PicPeak can round-trip into the photographer's Lightroom catalog without remapping.

**Clear grid display:** color badges or borders must be clearly visible on thumbnails in the main gallery view, so clients see their overall selection progress at a glance without reopening every image.

## 2. Keyboard shortcuts with a settings toggle

Pressing a number key in the lightbox/fullscreen view should instantly toggle the respective label — mouse-clicking through large selections is tedious.

Add a **toggle in settings** to choose between two keybind modes:

- **Lightroom mode:** matches Lightroom's default shortcuts — `1`–`5` set star ratings, `6`–`9` set color labels (`6` Red, `7` Yellow, `8` Green, `9` Blue). Photographers and clients who know Lightroom get identical muscle memory.
- **Color-only mode** (as proposed in the discussion): `1`, `2`, `3` directly toggle the color labels (e.g. Green = 1st choice, Yellow/Orange = 2nd choice, Red = rejected) — simplest possible scheme for clients who have never used Lightroom.

## 3. Color filtering

A quick filter in both the **public gallery** and the **admin dashboard** to isolate specific labels (e.g. show only Green), making the final export back to Lightroom effortless.

---

Credit to @boergu for the original request and the detailed workflow description — full context in the discussion thread: https://github.com/PicPeak/picpeak/discussions/1027

Luca-Timo:
Closes #1044. Feature request from discussion #1027 by @boergu — colour-label proofing was the one thing keeping him on PicDrop.

Clients mark photos with Lightroom's five colours while proofing, see their whole selection at a glance in the grid, and filter down to it. The photographer filters and exports by colour on the admin side, and the selection lands in their catalogue as `xmp:Label`.

---

## What this does

**1. Colour labels** — `red` / `yellow` / `green` / `blue` / `purple`, one per guest per photo, changeable (same colour again clears it). Lightroom's set deliberately, so a selection round-trips into the catalogue without remapping.

**2. Keyboard shortcuts, with the per-event scheme toggle the issue asked for**

| key | `colors` (default) | `lightroom` |
|-----|--------------------|-------------|
| 1 / 2 / 3 | green (1st choice) / yellow (2nd) / red (rejected) | ★1 / ★2 / ★3 |
| 4 / 5 | — | ★4 / ★5 |
| 6 / 7 / 8 / 9 | — | red / yellow / green / blue |
| 0 | clear the colour | clear the rating |

Same key again clears. Purple has no shortcut in either scheme — Lightroom has none either, so it stays click-only.

**3. Filtering** — colour chips in the public gallery (guest-scoped: *your* greens) and a colour swatch row with per-colour counts in the admin Feedback Filters, which feeds the existing export.

**4. Lightroom round-trip** — the XMP export already emitted `xmp:Label`; it now carries the real colour, plus a `color-<name>` keyword. "Export only the greens" produces sidecars Lightroom reads directly.

**5. The photographer can mark too** — the same 1-5 stars and five colours from the admin viewer, kept **completely separate** from what the client chose: separate table, separate filter row, separate export source. The gallery never shows them.

## Two independent selections, which is what makes the Lightroom import useful

The client's selection and the photographer's triage are different data all the way down:

| | client selections | your marks |
|---|---|---|
| stored in | `photo_feedback` (per guest) | `photo_admin_marks` (per admin) |
| visible in the gallery | yes, per the event's feedback settings | **never** |
| admin grid | solid dot, bottom-left | dashed dot + star count, beside it |
| admin filter | "Color labels" row | "Your marks" row |
| XMP export | `mark_source: client` (default) | `mark_source: mine` |

So you can cull a shoot green/yellow/red yourself, export **your** sidecars, import into Lightroom, and later export the **client's** greens as a second, independent set — without either pass overwriting the other. CSV and JSON carry both at once (`color_label` + `my_rating` / `my_color_label`), so a single export can drive a script that reconciles them.

The separate table isn't incidental. `photo_feedback` is read by a long tail of guest-facing queries — the gallery payload, the per-photo tallies, the denormalized `photos.average_rating` / `*_count` columns, the feedback exports, moderation. Admin rows in there would leak the photographer's private opinion of a shot into the client's proofing view through whichever of those someone forgot to filter. A separate table makes that impossible rather than merely unlikely, and `photoAdminMarks.test.js` asserts it directly: after marking every photo, `photo_feedback` is still empty, the denormalized counters are still zero, and the guest-facing tallies still return `{}`.

**The admin viewer always uses the Lightroom bindings** (1-5 stars, 6-9 colours) whatever scheme the gallery is set to — that scheme is a choice made *for the client*; this surface belongs to the photographer, who needs both halves on the keyboard.

## Also in here: one global standard for enabling feedback types

There was no global control for the individual feedback types — only the per-event panel — and the *creation* defaults were hard-coded in **four** places that had already drifted:

| where | ratings | likes | comments | favorites | reactions |
|---|---|---|---|---|---|
| `adminEvents/crud.js` destructuring | ✓ | ✓ | **✓** | ✓ | ✓ |
| `v1/events.js` hard-coded insert | ✓ | ✓ | **✓** | ✓ | **missing** |
| `getEventFeedbackSettings` fallback | ✓ | ✓ | **✗** | ✓ | ✓ |
| migration column defaults | ✓ | ✓ | **✗** | ✓ | ✓ |

Same install, different answer to "are comments on by default?" depending on which path created the event — and an event created through the v1 API got reactions only by accident of the column default. Adding colours as a fifth toggle to that would have made it worse, so `services/feedbackDefaults.js` is now the single source of truth (one `FEEDBACK_TOGGLES` table, one batched settings read, one merge helper). **Adding the next feedback type is one row in that table.**

The new keys extend the `event_default_*` family that `event_default_require_password` (#317) and `event_default_feedback_enabled` (#520) already established, and keep its semantics.

## Decisions worth disagreeing with

- **Globals are defaults for NEW galleries, never a retroactive switch.** Flipping one cannot change a gallery a client is mid-way through proofing. If you'd rather have a true master switch, that's a different control and wants an explicit "apply to N existing galleries" confirmation.
- **`allow_color_labels` defaults FALSE**, unlike its `allow_*` siblings. A colour bar appearing unannounced in every gallery that already has feedback on is a visible change to live galleries.
- **`xmp:Label` behaviour changes** where colour labels are used: it was inferred from the average star rating, and a real label now wins. The rating-derived mapping stays as the fallback, so exports for events without labels are byte-identical.
- **Colour filters are their own state slice**, not five more members of `FilterType` — that union is consumed by five components with exhaustive `Record<FeedbackFilterType, …>` maps that would have doubled.
- **Colour labels get their own rate limit (2000/h).** A client working through a 500-photo shoot on the keyboard would have hit the likes-sized 200/h cap mid-session. While there, `feedbackRateLimit`'s happy-path and error-path default tables became one constant — the error copy had silently missed every action type added after it was written, including `reaction`.
- **`mark_source` switches stars and colour together.** A mixed mode — the photographer's stars with the client's colour in one sidecar — is a plausible workflow but nobody has asked for it; it's a couple of lines in the projection if you want it.

## Notable implementation points

- Colour labels reuse the emoji-reaction machinery (#839): `submitFeedback` now runs both through one shared single-value path — same value toggles off, a different one switches, and a check-then-insert race collapses on the next interaction instead of leaving a phantom count.
- The viewer's **own** label rides on every gallery photo row and is deliberately **not** gated on `show_feedback_to_guests` — it isn't shared aggregate data, and hiding it would blank the grid badges on refresh for the guest who set them. The per-colour tallies *are* gated, same as the reaction map.
- Admin colour filtering is an `EXISTS` over `photo_feedback`, not a denormalized count — "has any label" and "has a GREEN label" are different questions. Covered by the index migration 180 adds.
- Shortcuts resolve through `utils/feedbackKeybinds`, shared between the gallery lightbox and the admin viewer so the two can't drift.
- `setMark` reads then inserts, and a triage pass fires those back to back — press 4 then 9 and both requests can find no row and both try to insert. The unique index stops the duplicate; the service catches the violation and converges on the row the winner created, re-resolving so the half this call didn't address keeps the winner's value rather than the stale pre-read one. Without it the second keystroke would 500 and be lost.
- The mark route parses `photoId` before querying: Postgres errors on `where id = 'abc'` (22P02) where SQLite simply fails to match, so an unvalidated id answered 500 on one engine and 404 on the other. Modified keys are never bound (Cmd+1 stays a tab switch) and digits typed into a field never relabel the open photo.
- Grid badge is an inset ring plus a corner dot, in every layout (grid, masonry, justified, mosaic, timeline, story, premium), with the colour **name** in the accessible name so colour isn't the only carrier.

## Where the changes live

**Colour labels, backend** — `constants/colorLabels.js` (the set, the XMP map, both keymaps, the dominant-colour tie-break) · `services/feedbackService.js` (shared single-value path, per-colour tallies, stats, both feedback exports) · `utils/feedbackValidation.js` · `routes/galleryFeedback.js` (settings payload, allow-map, per-photo tallies, `my_feedback.color_label`) · `routes/gallery.js` (`filter=color:<name>`, per-viewer label on each photo row) · `routes/adminPhotos.js` (`color_label=` filter, per-photo tallies) · `utils/photoFilterBuilder.js` + `routes/adminPhotoExport.js` (export filter, per-colour summary) · `services/photoExportService.js` + `services/xmpGenerator.js` (CSV/JSON column, `xmp:Label`, keyword) · `middleware/feedbackRateLimit.js`.

**Global defaults** — `services/feedbackDefaults.js` · the two create routes and the settings fallback that now read it · `routes/publicSettings.js` + `services/publicSettings.service.ts` + `pages/admin/CreateEventPage.tsx`. That last one matters: the admin create form POSTs every sub-toggle explicitly, so it seeds its own panel from the globals — server-side inheritance alone would only have covered the v1 API.

**Gallery UI** — `PhotoColorLabels.tsx`, `ColorLabelBadge.tsx`, `ColorLabelFilterChips.tsx` (new) · `PhotoLightbox.tsx`, `PhotoCard.tsx`, `layouts/story/StoryPhotoCard.tsx`, `layouts/GalleryPremiumLayout.tsx` (badge in every layout) · `GalleryView.tsx`, `PhotoFilterBar.tsx`, `GallerySidebar.tsx` (filter slice + chips) · `utils/feedbackKeybinds.ts` (new) · `services/feedback.service.ts`, `types/index.ts`.

**Photographer's own marks** — `migrations/core/181_add_photo_admin_marks.js` · `services/photoAdminMarksService.js` · `PUT /admin/photos/:eventId/photos/:photoId/mark` in `routes/adminPhotos.js` (gated `photos.edit` + `requireEventOwnership`, and re-checks the photo is in that event so a foreign photo id can't ride in through an event you do own) · the `my_color_label` filter on the admin list and in `photoFilterBuilder` · `AdminPhotoViewer.tsx` (the "Your marks" panel and its shortcuts) · `AdminPhotoGrid.tsx` · `PhotoFilterPanel.tsx` · `PhotoExportMenu.tsx`.

**Admin UI** — `FeedbackSettings.tsx` (toggle + scheme radios) · `PhotoFilterPanel.tsx` + `photos.service.ts` + `EventDetailsPage.tsx` (colour filter) · `AdminPhotoGrid.tsx` (grid badge) · `PhotoExportMenu.tsx` (colour-filtered export) · `EventsTab.tsx` + `useSettingsState.ts` (global defaults) · `routes/adminGuests.js` + `AdminGuestDetail.tsx` + `AdminGuestsList.tsx` + `guests.service.ts` (per-guest colour breakdown, alongside the existing likes/favourites columns) · `routes/adminFeedback.js` (colour labels counted in the analytics totals).

**i18n** — `frontend/src/i18n/locales/en.json`, `de.json`, `fr.json`, `es.json`, `nl.json`, `pt.json`, `ru.json`, `sl.json`.

## Migrations

`181_add_photo_admin_marks.js` — new table, guarded on `hasTable`, dropped in `down()`.

`180_add_color_labels.js` — `hasColumn`-guarded, idempotent, symmetric `down()` (the index is dropped before the column so SQLite's table rebuild doesn't trip). Verified by running up → down → up → up against a real SQLite database.

## Testing

```
backend:  npx jest                     2017 passed
          (8 suites fail identically on picpeak/main — verified by stashing;
           adminSettings.logo, adminPhotos.reference, crmMintPaths,
           imageProcessor.storage, storageBackend, webhookDelivery,
           backupService.enhanced, adminAuth)
frontend: npx vitest run               212 passed
          npm run build                clean
          npm run lint                 93 warnings — identical to base
```

New: `feedbackColorLabels.test.js` (15), `feedbackDefaults.test.js` (11), `feedbackKeybinds.test.ts` (14), `photoAdminMarks.test.js` (10 — including the isolation assertion above).

`events.create.test.js` needed updating rather than extending: it pins `app_settings` lookups positionally with `toHaveBeenNthCalledWith`, and the batched defaults read shifts those indices.

**Also exercised end to end against a local SQLite install** (migrated DB, seeded gallery, real photos): label / switch / toggle / invalid-colour rejection over HTTP, both keyboard schemes driven through a real browser and verified in the database, the typing guard, grid badges, gallery and admin filters, global defaults saving and a new event inheriting them, and CSV / JSON / XMP export producing `xmp:Label="Green"` for a colour-filtered selection. The admin marks got the same treatment: both halves written independently over HTTP, invalid values rejected, a photo id from another event refused, the keyboard drive landing rating 4 + blue on the right row, both filters returning disjoint sets, both XMP sources producing their own sidecars, and the gallery showing *nothing* for admin-marked photos — no badge, no tally, no average, absent from the colour filter.

That pass caught one bug this PR would otherwise have shipped: `getPhotoFeedback`'s explicit `.select()` omitted `color_label`, so `my_feedback.color_label` came back null.

## i18n

`en` and `de` written by hand. `fr`, `es`, `nl`, `pt`, `ru`, `sl` are machine-suggested and **want a native-speaker pass** — I only speak German and English.

## Notes

- ~3500 lines, so the size gate will fail. That's the review-size policy, not broken CI. The three commits are separable if you'd rather take them in sequence: schema → global defaults → feature.
- Pre-existing en/de parity drift (2 keys) left alone — not this branch's.

## Screenshots
01gallerygridbadges
**Gallery grid** — colour badges (inset ring + corner dot) and the colour filter chips with per-colour counts.

02lightboxcolorbar
**Lightbox** — colour bar in the toolbar with the active scheme's shortcut printed on each swatch.

03gallerycolorfilter
**Gallery, filtered to green.**

04admineventfeedbacksettings
**Per-event feedback settings** — the Colour Labels toggle and the keyboard-scheme radios, each with a live keymap preview read from the shared scheme.

05adminphotofilter
**Admin Feedback Filters** — colour swatch row with per-colour counts, feeding the existing export.

06settingsglobalfeedbackdefaults
**Settings → Event Creation** — the global per-type defaults and the default lightbox scheme.

07adminvieweryourmarks
**Admin viewer, "Your marks"** — the photographer's own stars and colours, Lightroom bindings, with the note that the client never sees them.

08adminmymarksfilter
**Admin photos** — "Color labels" (the client's) and "Your marks" as separate filter rows, and a tile carrying both verdicts: the client's solid dot, the photographer's dashed dot, and their star count.


the-luap:
Reviewed the whole diff. The separate-table decision is right and the isolation argument
holds — I verified it structurally rather than taking the test's word for it.

One concern, two nits, one coordination item. Nothing blocking.

## 🟡 Concerns

### 1. The `setMark` race fix covers the insert path but not the update path

`photoAdminMarksService.js:40-100` — the read-then-insert race is handled carefully, and
the comment justifies it with "press 4 then 9 and both requests can find no row". But the
same two keystrokes on a photo that **already has a mark** take the `existing` branch,
which is a plain read-modify-write with no protection: both calls read the old row, both
write the full `{rating, color_label}` pair, last write wins.

That is the more common case during a triage pass — by the second keystroke on a photo
there usually *is* a row.

Measured, driving the real service against SQLite:

```
insert path (no row yet)      -> {"rating":4,"color":"blue"}   both survive
update path (row exists, r=3) -> {"rating":3,"color":"blue"}   rating 4 LOST
```

The colour write clobbered the concurrent rating write with its stale pre-read value —
exactly the failure the insert path was written to prevent.

Bounded: one lost keystroke in a narrow window, no corruption, and pressing again fixes
it. But it is silent, and on the keyboard-driven workflow this feature is built around.

The fix does not need the race machinery — it needs the update to stop writing halves it
was not asked about:

```js
const patch = { updated_at: now };
if (rating !== undefined)     patch.rating = next.rating;
if (colorLabel !== undefined) patch.color_label = next.color_label;
await db('photo_admin_marks').where('id', existing.id).update(patch);
```

Then a colour call never touches rating at all and there is nothing to lose. The
delete-when-both-null branch needs the post-update state rather than the pre-read one.

## 🟢 Nits

**2. Migration 180 is not idempotent across a partial failure.** The index creation sits
inside the `hasColorLabel` guard, so a run that died between the two `alterTable` calls
leaves the column without its index, and a re-run skips both. Verified:

```
up / up / down / up cycles     -> correct
column present, index dropped  -> re-run does NOT restore the index
```

Consequence is only that the "show only the greens" filter loses the index the migration
header says it relies on. Guarding the index separately is a couple of lines.

**3. The mark route classifies errors by matching `error.message`**
(`adminPhotos.js:778`, `/Invalid color label|Rating must be/`). It works, but it couples
the route's status code to the service's copy — reword a message in the service and the
400 silently becomes a 500. A thrown type or an `err.code` would not drift.

**4. Migration numbers collide with two of my open PRs.** #1127 adds
`180_face_merge_suggestions.js` and #1135 adds `181_fix_css_template_photo_height.js`;
this adds `180_add_color_labels.js` and `181_add_photo_admin_marks.js`. Not a defect and
not yours to fix — the runner sorts filenames so both would apply — but whoever merges
second should renumber. **Mine are the ones to move; I will rebase after this lands.**

## 🟢 Observation, not a request

The colour filter ORs the guest's own labels with anyone's
(`gallery.js:780`), so `filter=color:green` reveals the *set* of photos others labelled,
without gating on `show_feedback_to_guests`. I checked before raising it: `commented` on
`main` already does exactly this, and the block matches the sibling tokens as the comment
says. So this PR extends an existing pattern consistently rather than introducing
anything — but colour labels are a more revealing signal than "has a comment", and the
public UI narrows client-side, so the aggregate is only reachable by a crafted request.
If that pattern should respect `show_feedback_to_guests`, it wants one change covering
all the tokens, not this one.

## ✅ Verified

Executed or read directly, not taken from the description:

- **The isolation claim, structurally.** `photo_admin_marks` is referenced by exactly
  three backend files — `adminPhotos.js`, `photoAdminMarksService.js`,
  `photoFilterBuilder.js` — and `photoFilterBuilder` is required only by
  `adminPhotoExport.js`. Zero references in `gallery.js`, `galleryFeedback.js`,
  `feedbackService.js` or `publicSettings.js`. `setMark` writes to nothing else, so the
  denormalized counters cannot move. The guest surface cannot read this table because it
  never queries it — which is the property the separate table was chosen for, and it is
  stronger than the test that also asserts it.
- **Route authorisation:** `adminAuth` + `photos.edit` + `requireEventOwnership`, plus the
  photo-in-event re-check, so a foreign photo id cannot ride in through an event you do
  own. The `photoId` parse before the query is right — 22P02 on Postgres vs a silent
  non-match on SQLite is a genuine engine split.
- **Per-viewer label is guest-scoped** (`guest_id`, else `generateGuestIdentifier`), same
  resolution as likes; the aggregate `color_label_count` *is* gated on
  `showFeedbackToGuests`. The asymmetry is argued correctly.
- **Rate limits:** every existing cap unchanged; the error path now shares the table it
  had drifted from, which quietly fixes `reaction` too.
- **XMP:** real label wins, rating-derived mapping preserved as fallback, and `mapLabel`
  still accepts a bare number for callers outside the export service.
- **Migrations:** guarded, symmetric, up/down/up verified against real SQLite.
- **No new dependencies** — no `package.json`, `server.js`, `Dockerfile` or `nginx.conf`
  changes in 4,000 lines. Only `feedbackRateLimit.js` among the usual high-risk files.
- **The three new backend suites pass:** 36 tests, matching the numbers in the description.

`bypass-size-gate` is red because the PR exceeds the self-merge limit and wants a
maintainer review — the policy gate doing its job, not a defect. Note it only triggers on
push events, so it will stay red through merge even after an approval.

## Verdict

Fix the update-path lost update — or tell me you would rather ship it, since it costs one
keystroke and self-corrects — and I am comfortable merging. The nits are yours to take or
leave.

Luca-Timo:
@the-luap Thanks — all three fixed in 7139bcae. The update-path finding was the
right one to lead with; it is the common case and I had only defended the rarer
one.

## 🟡 Concern 1 — lost update on the existing-row path

Fixed as you described: the update writes only the halves the call was asked
about, so an untouched column is never written and there is nothing to lose. No
race machinery needed.

Reproduced your measurement first, so the test pins the real thing — seed
rating 3, fire `rating=4` and `colour=blue` together:

```
before the fix -> {"rating":3,"color_label":"blue"}   the 4 is gone
after the fix  -> {"rating":4,"color_label":"blue"}
```

The regression test fails on the previous service with exactly your
`Expected: 4 / Received: 3`, and passes after. Confirmed by swapping the service
file back and re-running, not by reasoning about it.

One addition to your sketch. You noted the delete-when-both-null branch needs
post-update state. Rather than re-read then delete, the emptiness test is now a
predicate **on** the delete:

```js
const removed = await db('photo_admin_marks')
  .where('id', row.id).whereNull('rating').whereNull('color_label').delete();
if (removed) return null;
```

A re-read followed by a delete has the same shape of window the original bug
had — a concurrent call that just filled a half back in would be dropped by our
stale view. As a predicate the engine decides, so it cannot.

Both paths now go through one `applyToRow`, so the insert convergence and the
plain update cannot drift apart again.

## 🟢 Nit 2 — migration 180 index

Out of the column guard and into its own idempotent statement. I checked
`CREATE INDEX IF NOT EXISTS` on both engines rather than trusting migration
071's comment, which claims SQLite lacks it — SQLite has supported it since
3.3, verified idempotent on the 3.44 this repo ships. `down()` drops it
unconditionally so a partial `up()` still cleans up.

Your scenario, re-run:

```
after up             -> index: true
index dropped by hand-> index: false
after re-run up      -> index: true    (was false)
down/down/up         -> index: true | column: true
```

## 🟢 Nit 3 — error classification

Service tags caller mistakes with `code: 'INVALID_MARK'`; the route keys on the
code. Rewording a message can no longer turn a 400 into a 500. Test asserts the
code rather than the string.

## 🟢 Nit 4 — migration numbers

Understood, and thanks for taking the renumber. Nothing changed here.

## Your observation — colour filter and `show_feedback_to_guests`

Agreed it is not this PR's to change, and your reading matches the intent: I
matched the `commented` sibling deliberately, and the comment says so. But you
are right that a colour label is a more revealing signal than "has a comment",
and one change covering all four tokens is the correct shape. Happy to open a
follow-up issue for it if you want it tracked.

## Verification

Ran, not read: full backend suite 2030 passed (same 8 suites fail on
`picpeak/main`); the three new suites now 13 tests; the concurrent double-write
driven over HTTP against a live install, both halves surviving; the 400s still
classified correctly; clearing both halves still removing the row.

the-luap:
Re-verified `7139bcae`. All three landed, and the two I could drive I drove rather than read.

## Blockers / concerns — closed

| | | |
|---|---|---|
| 🟡 1 | Lost update on the existing-row path | ✅ |
| 🟢 2 | Migration 180 index outside the column guard | ✅ |
| 🟢 3 | Error classification by message text | ✅ |
| 🟢 4 | Migration numbers | ✅ done here, see below |

**Concern 1.** `applyToRow` builds the patch from `rating !== undefined` / `colorLabel !== undefined`, so an unaddressed column is never written. That is the right shape — the fix is the absence of a write, not machinery around one. Reverted `photoAdminMarksService.js` to `7139bcae~1` and re-ran your test: fails with `Expected: 4 / Received: 3`, the exact signature from the review, and passes after.

Your addition improves on my sketch. Making the emptiness test a predicate **on** the delete rather than a re-read-then-delete removes the same class of window the original bug had. Probed it directly — natural concurrent `clear(rating) || set(colour)` lands `{rating: null, color_label: "blue"}`, nothing lost.

**Nit 2.** Ran the migration against SQLite 3.44, including the partial-state case:

```
after up              -> index: true   col: true
index dropped by hand -> index: false  col: true
after re-run up       -> index: true   col: true     <- was false before
down / down / up      -> index: true   col: true
```

Right call checking `CREATE INDEX IF NOT EXISTS` yourself instead of trusting migration 071's comment.

**Nit 3.** `INVALID_MARK` exported from the service, imported at `adminPhotos.js:18`, keyed on at `adminPhotos.js:777`. Rewording a message can no longer flip a 400 into a 500.

## Migration numbers

Taken care of in `8fecdfae` — 180 and 181 went to `180_face_merge_suggestions` and `181_fix_css_template_photo_height` while this was in review, so colour labels are now 182 and 183. Renames only, plus the two prose references to "migration 180" in `feedbackService.js:86` and `photoFilterBuilder.js:85`.

## No regressions

Full backend suite on this branch vs `main`: identical failure sets — same 6 suites, 29 tests, failing on both. Nothing this PR touched.

## One residual, not blocking

A *clear* racing a *set* is still lossy in a way the fix does not cover. Row `{rating: 3, color_label: null}`, call A sends `rating: null`, call B sends `colorLabel: 'blue'`. If A's delete commits inside B's window, B's `update().where('id', row.id)` matches zero rows, and B returns `null` — the colour keystroke is gone and the response says "no mark".

Being precise about how far I took this: I replicated `applyToRow`'s statements against a real database and got `rowsUpdated=0, after=undefined`, so there is no affected-row guard. I did **not** drive it through `setMark` under real parallelism, and natural single-process ordering does not produce it — the probe above lands correctly. It needs true parallelism plus a clear colliding with a set.

One line if you want it: if the update affects 0 rows the row vanished underneath, so fall through to the insert path. Self-corrects on retry either way, so entirely your call — a follow-up is fine.

Yes to the follow-up issue for the colour filter and `show_feedback_to_guests`, covering all four tokens together.

## Verdict

Merge cleared.

the-luap (approved):
Re-verified `7139bcae` and renumbered the migrations to 182/183 in `8fecdfae`. All findings closed; the residual clear-vs-set race is a follow-up, not a blocker.

Issue #640, opened 2026-06-18
[FEATURE] Cherry-pick upstream candidates from active forks


the-luap:
Tracker for features and fixes worth pulling upstream from active forks. Each item links the source fork + commit so credit is preserved; cleanup notes flag what needs adjusting before merge.

## 1. WhatsApp Business API integration

**Source**: [filpgame#5960af1](https://github.com/filpgame/picpeak/commit/5960af1) (PR #1 on their fork) — 9 files, +837 lines, fully tested.

Adds WhatsApp as a notification channel alongside email — fills a gap that DACH / IT photographers keep asking for.

**Files shipped**:
- `backend/src/services/whatsappService.js` + `whatsappProcessor.js` (queue-based, mirrors `emailProcessor` pattern)
- `backend/src/routes/adminWhatsapp.js`
- Frontend Settings tab
- `backend/src/routes/adminEvents.js` hook to enqueue on event published
- Two new tables: `whatsapp_configs` (phone_number_id, waba_id, access_token, template_name, enabled), `whatsapp_queue`
- Three Jest test files

**Cleanup before upstreaming**:
- ⚠️ **Migration collision**: filed as `107_create_whatsapp_tables.js`. Our `107_crm_consolidated.js` already owns that number. Renumber to next available (135+).
- ⚠️ **Hard FK on `event_id`**: their schema uses `references('id').inTable('events')`. Our pattern for cross-cutting tables (audit log, hours entries, inbound docs) is loose-FK so deletes don't RESTRICT. Normalise to match.
- Wire under a `feature_whatsapp` flag in the dependency-rules pattern (consistent with CRM / accounting flags).
- Decide where Resend button lives (resend dialog on event detail page, alongside the email resend).

**Effort estimate**: 1-2 days cleanup for clean upstream merge.

## 2. Per-category download permissions

**Source**: [8digit#928164b](https://github.com/8digit/picpeak/commit/928164b) + [#751ec75](https://github.com/8digit/picpeak/commit/751ec75)

Adds `category_allow_downloads` flag on photo categories so photographers can have different download policies per category — e.g. previews public, originals client-only.

Real use case from a few support threads. Small surface (one column + UI toggle + gallery-side enforcement).

## 3. Backup restore — stream-extract for >2 GiB archives

**Source**: [8digit#69033c6](https://github.com/8digit/picpeak/commit/69033c6)

[...]

Issue #163, opened 2026-02-02
[FEATURE] Category-Specific Hero Images


patchingfailed:
### Is your feature request related to a problem? Please describe.
Categories already act as a lightweight way to create “sub-galleries” via filters, which is a very elegant solution compared to manually creating separate galleries.

However, once categories are used more intentionally, it would be great if the **Hero area could reflect the currently selected category**.

### Describe the solution you'd like
Allow defining an optional **Hero image per category**.

Behavior idea:
- Each category can have its own Hero image
- When a single category is selected, its Hero image is shown
- When multiple categories (or “All”) are selected, fall back to a default / welcome Hero image

This keeps the current filtering approach intact while adding more visual context.

### Describe alternatives you've considered
- Creating separate galleries per category  
This works, but removes the benefit of having one cohesive event gallery with filters.

### Use Case
- **Who would use it?** Admins / photographers
- **When would they use it?** When presenting larger events with clear sections (e.g. Portraits, Ceremony, Party)
- **Why is it important?**  
  - Gives each category a stronger visual identity  
  - Helps clients understand where they are within the gallery  
  - Enhances storytelling without additional navigation complexity  

### Similar Features
Other tools support explicit sub-galleries with their own headers or cover images.  
This would provide a similar benefit while keeping PicPeak’s category-based filtering approach.

### Mockups or Examples
No mockups yet. Conceptually:
- Default Hero = event / welcome Hero
- Category Hero overrides the default when exactly one category is active

### Additional context
I really like the current filter-based “sub-gallery” concept.  
This feature would not replace it, but enhance it visually.

### Implementation Ideas
- Optional Hero image field on category level
- Simple precedence rule:
  - Single category selected → use category Hero
  - Multiple / none selected → use default Hero
- Keep everything optional, with no breaking changes to existing galleries

PR #1161, opened 2026-08-24
feat(gallery): folders that contain photos instead of filtering them (#1160)

### LINKED ISSUE #1160: [FEATURE] Gallery folders: let a category contain photos instead of filtering them
Categories today are a **filter**: every photo shows in the main grid regardless of its category, and picking a category narrows that grid (`GalleryView.tsx:608`). What D#1086 asks for is a **container**: photos put in a bucket should *leave* the main grid and live behind a tile you click into.

Concretely, with 40 finals and 200 selects moved into a "Selects" folder:

```
today (category = filter)          wanted (folder = container)
─────────────────────────          ───────────────────────────
gallery root: 240 photos           gallery root: 40 photos + [ Selects ▸ 200 ]
  └ filter "Selects" → 200           └ click Selects → 200 photos
```

The 200 foldered photos are **not** in the root grid at all. They are only ever shown once you are inside the folder. That is the feature — not nesting depth.

Both behaviours stay available: a category is a filter until it is flagged as a folder, so nothing existing changes.

> My current use case for the gallery is to share selects with a team on a model test shoot, then once finals are chosen, I upload post-processed files (which is where I then archive the selects into a nested folder). In some instances I may re-edit as different aspect ratios or color vs. black & white and that's where I will create additional folders. I would personally choose this over categorizing images to keep the main gallery less cluttered assuming that the users I'm sharing with are likely not using the filters to narrow down the files.

## Scope decision: containment only, no `parent_id`

The reporter says he never goes deeper than one level. Read literally, "gallery root → Selects folder" **is** depth one — that is plain containment, not nesting. No self-referential FK, no tree walk, no depth validation.

`parent_id` is deliberately deferred until someone asks for folders-inside-folders.

## Why not sub-galleries

A `parent_event_id` on `events` was considered and rejected. Every consumer of `events` would need parent-awareness — expiry and archive jobs, storage accounting, backup, email templates, event lists, analytics, CRM/project links, ZIP building — and the guest side is worse: a sub-gallery is either a second password prompt or a shared-token special case. The reporter also wants the folders visible *inside* the gallery, not behind a second link.

## The migration is one column

Three shipped features already built the substrate:

| Need | Already exists |
|---|---|
| Folder cover image | `photo_categories.hero_photo_id` (#163, migration 066) |
| Per-folder download rules | `photo_categories.allow_downloads` (#640, migration 135) |
| Folder ordering | `display_order` + `event_category_order` (#782, migrations 159/160) |
| Move photos in bulk | `BulkCategoryModal.tsx` |
| One photo, one folder | `photos.category_id` is single-valued |

All that's missing is the containment flag.

```js
// migration 185: is_folder on photo_categories
t.boolean('is_folder').notNullable().defaultTo(false);
```

Default `false` means every existing category keeps today's filter behaviour on upgrade. Follow the `hasColumn`-guarded additive pattern of migration 159.

## Backend

1. **`backend/migrations/core/185_add_category_is_folder.js`** — additive, `hasColumn`-guarded, sane `down()`. No backfill needed (the default *is* the preserved behaviour).
2. **`backend/src/routes/gallery.js:965`** — the `select` list is explicit, so `'c.is_folder'` must be added there, and mapped through at `:968-978`. Easy to miss; `c.*` is not in play.
3. **`backend/src/utils/categoryOrder.js`** — no change. Folders and filter-categories share one ordering model, so a folder can sit in the flow of the day alongside a category.
4. **`backend/src/routes/adminCategories.js`** — accept `is_folder` on create (`:43`) and update; `body('is_folder').optional().isBoolean()`.
5. **Guard `events.upload_category_id`** — `gallery.js:2803` defaults guest uploads to it. If it points at a folder, guest uploads land somewhere the guest can't see. Reject at event-settings save, or warn.
6. **`/download-all` keeps whole-event semantics.** Folder photos stay in the event ZIP. Document it rather than change it — the prebuilt ZIP already ignores per-category `allow_downloads`, and that's a separate known gap.

## Gallery UI

Central win: the category filter is applied in exactly one `useMemo` (`GalleryView.tsx:596-711`), so containment is a single-point change.

- **Root view** — exclude photos whose `category_id` belongs to an `is_folder` category. Uncategorised photos always stay at root.
- **Folder view** — show only that folder's photos.
- **Folder tiles** — cover from `hero_photo_id`, name, count. Render in `GalleryView`/`BaseGalleryLayout` **above** the photo grid, so all eight layouts (Grid, Masonry, Justified, Mosaic, Timeline, Carousel, Story, Premium) inherit it without eight implementations.
- **URL state `?folder=<slug>`** — new; category selection is currently pure React state with no URL round-trip. Must not collide with `token` or `admin_preview`.
- **Breadcrumb + back** out of a folder.
- **`GallerySidebar.tsx:242-290`** — folders listed separately from filter categories; they are not the same affordance.
- **Lightbox** navigation stays scoped to the current view's photo set.
- **Counts and the people strip** derive from the visible set, so a folder's photos don't inflate root counts.
- **Folder download** reuses `POST /gallery/:slug/download-selected` (`photo_ids`, `gallery.js:1763`) — no new endpoint.
- **Empty folders** are already invisible to guests: the payload only returns categories that have photos (`gallery.js:952`).

## Admin UI

- `EventCategoryManager.tsx` / `CategoryManager.tsx` — a "Show as folder" toggle.
- Reuse the existing cover-image picker for the folder cover.
- `BulkCategoryModal.tsx` needs no change — moving photos into a folder is moving them into a category.

## Folders are not access control

Worth stating in the UI copy. A folder hides photos from the grid; it does **not** protect them. Per-photo auth (`photoAuth`, `secureImages.js`) is unchanged, so a direct photo URL still resolves. Anyone who reads "archive the selects into a folder" as "make them private" will be wrong, and the toggle's help text should say so.

## Open decision: `is_folder` on global categories

`is_folder` on a **global** category restructures every gallery on the instance at once. Options:

- **(a) Allow it, warn in the UI on toggle.** Simplest; matches how `allow_downloads` already behaves for globals.
- **(b) Event-scoped categories only.** Safest, and matches the reporter's per-shoot use case, but forces a per-event category for a photographer who wants "Selects" everywhere.
- **(c) Per-event override table**, mirroring `event_category_order`. Most flexible, clearly over-engineered for v1.

Leaning **(a)**. Flagging rather than deciding silently.

## Phases

1. Migration + gallery payload + admin CRUD → verify: `is_folder` round-trips through `GET /gallery/:slug/photos`; migration re-runs clean on both SQLite and Postgres.
2. Gallery root containment + folder tiles + `?folder=` drill-in + breadcrumb → verify: folder photos absent from root, present in folder, URL survives reload and back-button.
3. Admin toggle + cover picker + `upload_category_id` guard → verify: toggling a category to folder moves its photos out of the root grid with no re-upload.
4. Folder downloads, counts, i18n (en, de, es, fr, nl, pt, ru, sl), tests, screenshots.

## Tests

- **Backend** — migration idempotency; `is_folder` present in the gallery payload; category CRUD accepts and persists it; `upload_category_id`-points-at-folder guard.
- **Frontend** — `filteredPhotos` excludes folder photos at root and includes them in folder view; `?folder=` round-trip; counts exclude foldered photos.
- **E2E** — enter a folder, download from it, navigate back.

Per CONTRIBUTING, the UI PR needs before/after screenshots of the root grid and the folder view.

## Related

- D#1086 — the request.
- D#780 — category *reordering*, from another Pixieset migrant. Shipped as #782. Same object, adjacent property; the ordering work is what makes folders cheap here.

the-luap:
> **Bottom of a two-PR stack.** #1208 (shared colour tag, #1197) is stacked on this branch and merges after it. Rebased onto current `main` on 27 Aug so the stack sits on current code — that gap included the #1201 capture-date fix.

---

Closes #1160. Implements the folder request from discussion #1086.

## What changes

A category has always been a **filter**: its photos stay in the root grid and picking the category narrows that grid. `photo_categories.is_folder` makes it a **container** instead — its photos leave the root grid entirely and only render once the guest opens the folder.

```
today (category = filter)          with is_folder (folder = container)
─────────────────────────          ───────────────────────────────────
gallery root: 62 photos            gallery root: 42 photos + [ Selects ▸ 20 ]
  └ filter "Selects" → 20            └ click Selects → 20 photos
```

Defaults to `false`, so **every existing gallery keeps filtering exactly as before**. Folders are opt-in per category.

## Why the migration is one column

The neighbouring features already built the substrate, so this needed nothing else:

| Need | Already existed |
|---|---|
| Folder cover image | `hero_photo_id` (#163, migration 066) |
| Per-folder download rules | `allow_downloads` (#640, migration 135) |
| Folder ordering | `display_order` + `event_category_order` (#782, migrations 159/160) |
| Move photos in bulk | `BulkCategoryModal` |
| One photo, one folder | `photos.category_id` is single-valued |

**No `parent_id`.** The request is "root → Selects folder", which is depth one — plain containment. Folders-inside-folders waits until someone asks.

## Moving photos into a folder

Already worked: a folder *is* a category, so the existing bulk "move to category" flow does it. The only gap was that a folder and a filter category looked identical in that dropdown while having very different consequences, so folder options now read `Selects (folder — hidden from the main grid)`.

## Downloads

Both halves of the requirement:

- **Whole gallery** — "download all" still zips every photo *including* foldered ones (verified: 62 files). A folder never silently shrinks the client's one-click download.
- **Per folder** — a `Download folder (20)` button inside a folder zips only that folder, once (verified: 20 files). Reuses `/download-selected`, so no new endpoint and no second zip-building path.

Honours the per-category opt-out (#640) both ways: a folder with `allow_downloads = false` renders no button, and opted-out photos are excluded from the id list rather than 403-ing mid-zip.

## Scope correctness

Containment lands in the one `useMemo` where the category filter was already applied, so everything downstream inherits it:

- **Counts** — root reports 42, not 62.
- **Lightbox** — inside a folder the counter reads `1 / 20`, not `1 / 62`.
- **People strip** — `face_count` comes from `/people` and spans the whole event, which contradicted the grid twice over: inside a folder a face read "12 photos" but filtered to a handful, and at root a person whose photos *all* lived in a folder appeared and filtered to nothing — a dead chip. Recounted from `photo.person_ids` (already what filtering uses); zero-count people dropped. No backend change.
- **Search and feedback chips** — narrow *within* the current scope, never across a folder boundary.

## All eight layouts

Containment comes from `filteredPhotos`, which both layout branches use — so a branch that hides foldered photos without offering the tiles makes them **unreachable**. The folder nav is built once and rendered by both.

The two full-bleed layouts (Premium, Story) are edge-to-edge by design, and a block of cover cards above the hero wrecks the opening they exist for, so they get a compact chip row instead of tiles. It renders only when the gallery actually has folders, leaving existing full-bleed galleries byte-identical.

## Folders are not access control

Stated in the toggle's help text, and pinned by a test. A foldered photo is served by the same per-photo auth as any other — hiding it from the grid does not make its URL unreachable. Anyone reading "archive the selects into a folder" as "make them private" would be wrong, and the UI says so.

## Decision: `is_folder` is event-scoped only

The toggle appears only on event-specific categories; globals (`Ceremony`, `Reception`, …) cannot be made folders. One click on a global would pull those photos out of the root grid of *every* gallery on the instance, including delivered ones. Rationale and the recommended escalation path (a per-event override mirroring `event_category_order`, **not** a global flag) are recorded in https://github.com/PicPeak/picpeak/issues/1160#issuecomment-5400743740.

## Verification

Exercised against a running stack (62-photo gallery, 20 moved into a folder) — not mocks:

- root grid renders 42 cards, folder renders 20, tiles hidden inside a folder
- `?folder=<slug>` round-trips: deep link opens the folder, Back walks out, `token` and `admin_preview` preserved
- move via the real `POST /admin/events/:id/photos/bulk-update`: root 42 → 40, folder 20 → 22
- folder zip 20 files, gallery zip 62 files
- unknown `?folder=` slug falls back to root rather than emptying the gallery

Migration ran clean and idempotent on **both** SQLite and Postgres.

Tests: 20 new frontend unit tests for the containment rule, 5 new backend tests pinning `is_folder` through the guest payload (the category block uses an explicit `select` list, so a column that isn't added there is silently dropped — that contract is now covered). Full suites green: 122 frontend, 272 backend route tests. `tsc` and eslint clean.

Screenshots in a follow-up comment.

### DIFF
--- a/backend/__tests__/routes/galleryFolderCategories.test.js
+++ b/backend/__tests__/routes/galleryFolderCategories.test.js
@@ -122,6 +122,17 @@ describe('folder categories in the gallery payload (#1160)', () => {
     expect(found.is_folder).toBe(false);
   });
 
+  test('a form-encoded is_folder="false" stays a filter (not !!-coerced to true)', async () => {
+    // express-validator's isBoolean() accepts the STRINGS "false"/"0", and
+    // `!!'false'` is true — so `!!` would flip a caller asking for a filter into
+    // a folder, silently pulling their photos out of the root grid.
+    const { parseBooleanInput } = require('../../src/utils/parsers');
+    expect(parseBooleanInput('false', true)).toBe(false);
+    expect(parseBooleanInput('0', true)).toBe(false);
+    expect(parseBooleanInput('true', false)).toBe(true);
+    expect(parseBooleanInput(undefined, false)).toBe(false);
+  });
+
   test('folders are not an access boundary — the photo is still in the payload', async () => {
     // Containment is a rendering rule, not authorisation. If this ever starts
     // failing, folders have silently become a security feature they are not.
--- a/backend/__tests__/routes/galleryFolderCategories.test.js
+++ b/backend/__tests__/routes/galleryFolderCategories.test.js
@@ -1,0 +1,131 @@
+/**
+ * Gallery folders reach the guest payload (#1160).
+ *
+ * `is_folder` is what tells the frontend a category CONTAINS its photos rather
+ * than filtering them. The category block in gallery.js selects an explicit
+ * column list (not `c.*`), so a new column that isn't added there is silently
+ * dropped — every folder would render as a plain filter and the root grid would
+ * still show the foldered photos. These assertions pin that contract.
+ *
+ * Also pins the SQLite side: the engine stores 0/1, so a strict `=== true`
+ * consumer would see every folder as a filter (the #1028 class of bug).
+ */
+
+const path = require('path');
+const fs = require('fs');
+const os = require('os');
+
+process.env.NODE_ENV = 'test';
+process.env.TEST_DATABASE_PATH = path.join(
+  fs.mkdtempSync(path.join(os.tmpdir(), 'picpeak-folders-')), 'db.sqlite',
+);
+process.env.JWT_SECRET = process.env.JWT_SECRET || 'folders-test-secret';
+process.env.STORAGE_PATH = fs.mkdtempSync(path.join(os.tmpdir(), 'picpeak-folders-storage-'));
+
+const request = require('supertest');
+const express = require('express');
+const cookieParser = require('cookie-parser');
+const { bootCrmDb, seedMinimal } = require('../integration/helpers/crmDb');
+
+const SLUG = 'folders-gallery';
+
+describe('folder categories in the gallery payload (#1160)', () => {
+  let db; let cleanup; let app; let eventId; let folderId; let filterId;
+
+  async function getCategories() {
+    const res = await request(app).get(`/api/gallery/${SLUG}/photos`);
+    expect(res.status).toBe(200);
+    return res.body.categories;
+  }
+
+  async function addPhoto(filename, categoryId) {
+    const row = await db('photos').insert({
+      event_id: eventId,
+      filename,
+      path: `${SLUG}/${filename}`,
+      type: 'individual',
+      category_id: categoryId,
+      uploaded_at: new Date().toISOString(),
+    }).returning('id');
+    return row[0]?.id ?? row[0];
+  }
+
+  beforeAll(async () => {
+    ({ db, cleanup } = await bootCrmDb());
+    await seedMinimal(db);
+
+    const ev = await db('events').insert({
+      slug: SLUG,
+      event_type: 'wedding',
+      event_name: 'Folders',
+      event_date: '2026-08-01',
+      host_email: 'h@example.com',
+      admin_email: 'a@example.com',
+      password_hash: 'x',
+      share_link: `/gallery/${SLUG}/s`,
+      share_token: 'folders-share',
+      expires_at: new Date(Date.now() + 7 * 864e5).toISOString(),
+      is_active: 1,
+      is_archived: 0,
+      is_draft: 0,
+      require_password: 0,
+      created_at: new Date().toISOString(),
+    }).returning('id');
+    eventId = ev[0]?.id ?? ev[0];
+
+    const folder = await db('photo_categories').insert({
+      name: 'Selects', slug: 'selects', is_global: 0, event_id: eventId, is_folder: 1,
+    }).returning('id');
+    folderId = folder[0]?.id ?? folder[0];
+
+    const filter = await db('photo_categories').insert({
+      name: 'Ceremony', slug: 'ceremony', is_global: 0, event_id: eventId, is_folder: 0,
+    }).returning('id');
+    filterId = filter[0]?.id ?? filter[0];
+
+    await addPhoto('in-folder.jpg', folderId);
+    await addPhoto('in-filter.jpg', filterId);
+
+    app = express();
+    app.use(express.json());
+    app.use(cookieParser());
+    app.use('/api/gallery', require('../../src/routes/gallery'));
+  }, 120000);
+
+  afterAll(async () => { if (cleanup) await cleanup(); });
+
+  test('the engine under test stores booleans as 0/1', async () => {
+    expect(['sqlite3', 'better-sqlite3']).toContain(db.client.config.client);
+    const row = await db('photo_categories').where('id', folderId).first('is_folder');
+    expect(row.is_folder).toBe(1);
+  });
+
+  test('a folder category is reported with is_folder true', async () => {
+    const folder = (await getCategories()).find((c) => c.id === folderId);
+    expect(folder).toBeDefined();
+    expect(folder.is_folder).toBe(true);
+  });
+
+  test('a plain category keeps filtering — is_folder is false, not undefined', async () => {
+    const filter = (await getCategories()).find((c) => c.id === filterId);
+    expect(filter.is_folder).toBe(false);
+  });
+
+  test('a category predating the column defaults to filtering, never folder', async () => {
+    const legacy = await db('photo_categories').insert({
+      name: 'Legacy', slug: 'legacy', is_global: 0, event_id: eventId,
+    }).returning('id');
+    const legacyId = legacy[0]?.id ?? legacy[0];
+    await addPhoto('legacy.jpg', legacyId);
+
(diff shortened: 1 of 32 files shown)


the-luap:
## Screenshots

Captured against a running stack — real 62-photo gallery with 20 photos moved into a "Selects" folder. Not mockups.

### Guest — gallery root
The folder tile sits above the grid. The grid itself shows **42** photos, not 62: the 20 in the folder are gone from it, which is the whole point of the feature. Cover comes from the category hero (#163), falling back to the folder's first photo.

issue-1160-1-root-with-folder

### Guest — inside the folder
Breadcrumb back to root, and `Download folder (20)` — zips that folder only, once. The people strip is recounted against the 20 photos actually on screen rather than the whole event, so the counts can't contradict the grid.

issue-1160-2-inside-folder

### Guest — Premium (full-bleed) layout
These layouts return early and render edge-to-edge, so they get a compact chip row instead of cover cards — the hero stays the opening it was designed to be. Without this the foldered photos would be hidden here with **no way in**, since containment applies to this branch too.

issue-1160-3-premium-layout-folder

### Guest — Premium, event-wide download
Once folders exist, the full-bleed layouts have no other single way to get the whole gallery: their only gallery-wide download is select-all, which is correctly scoped to what is on screen. The strip carries an event-wide download at the root. It is withheld when any category opts out of downloads, since that route serves the prebuilt zip which has no per-category filter.

issue-1160-6-premium-download-all

### Guest — mobile (390 px)
Tile grid drops to two columns; the folder card and its count stay legible.

issue-1160-4-mobile-root

### Admin — folder toggle
`Small groups` is a filter (closed grey folder), `Selects` is a folder (open, accented). Note the shared/global categories above have **no** toggle at all — folders are event-scoped by design, so one click can't restructure every gallery on the instance.

The tooltip is deliberately blunt about the limit: *"Click to turn it into a folder — hides them from the grid, does NOT restrict access."*

issue-1160-5-admin-folder-toggle

Issue #745, opened 2026-07-04
[FEATURE] Lightroom round-trip: re-import client-approved selections


Luca-Timo:
Get client-approved selections back into Lightroom without manual re-matching. Today: one-way Lightroom TXT export only (#630).

- [ ] Lightroom publish plugin — export selections + re-import approved images (ties into the companion-app plan)
- [ ] Capture One — optional / out of scope for now

Effort: L (needs a plugin).

_Parked idea — no timeline._

daniel-hmc:
@Luca-Timo may I ask what you mean by re-import (approved) images? 
Aren't all images already in Lightroom and you just need to know which images have been approved (filter for them -> put them into a collection or move to another folder)?
If you re-import images, wouldn't that result in duplicate images in Lightroom?
Maybe I misunderstood your idea, though.

Daniel

Luca-Timo:
Hi @daniel-hmc
Ideally I just want to import the selected images into lightroom, maybe let the user select in the plugin if he wants to import all images or just the selected oned

the-luap:
Both halves are merged. Closing.

## What shipped

**Backend** — #1165 (`8db8527f` on `main`, goes out in the next beta):

- `GET /api/v1/events/:id/photos` returns each photo with the marks it carries — client colour labels and star ratings, the photographer's own admin marks, and a merged `color_label` / `rating` for the requested `mark_source`. Filterable by colour, rating and `marked_only`.
- `replaces_photo_id` on upload: the finished edit replaces the proof **in place**. The photo keeps its ID, the client's ratings and colour labels, its comments and its position in the gallery, and the share link stays valid.
- `photos.source_filename` (migration 193) preserves the camera-original name across a replacement, which is what the matching keys on — `original_filename` is overwritten the first time a render goes back up.

**Plugin** — PicPeak/plugin-lightroom#3 (`5f6afdf6` on `main`):

*Library → Plug-in Extras → Import selections from PicPeak…* — pick the event and the RAW folder, choose all photos or only the marked ones, and the plugin matches each pick to its RAW, adds it to the catalog if it isn't there, applies the colour labels and ratings, and collects them. Publishing an edit back replaces its proof.

## @daniel-hmc — your question, answered by the implementation

You asked whether "re-import" would produce duplicates, since the images are already in Lightroom. That was the right thing to push on, and it shaped the design:

[...]

PR #1165, opened 2026-08-24
feat(api): Lightroom round-trip — read proofing marks, put finished edits back (#745)

Luca-Timo:
Backend half of the Lightroom round-trip. Closes nothing on its own — it is the API surface [PicPeak/plugin-lightroom#2](https://github.com/PicPeak/plugin-lightroom/issues/2) needs, and the first concrete step on #745.

## The workflow this enables

1. Unedited camera JPGs (`IMG_1234.JPG`) are uploaded to an event.
2. The client marks colours/stars while proofing; the photographer triages with their own marks.
3. The editor opens Lightroom, picks the event and a RAW folder, and imports **only the marked photos** (or all) with the stars and colour labels already applied.
4. The editor edits, renames to a delivery name, and uploads the renders back — they **replace** the proof in place, keeping its id, feedback, comments and position.

Steps 1–2 already worked. This PR is steps 3 and 4's server side.

### 1. Keep the camera filename — migration 185

`photos.original_filename` is the only carrier of `IMG_1234.JPG` (the stored `filename` is rewritten by `generatePhotoFilename`). But `replacePhoto()` overwrites `original_filename` with whatever name the incoming file has, so **the first re-upload of a renamed render destroys the key the next round-trip needs**. Invisible on the first pass, which is why it needs a column rather than a convention.

`photos.source_filename` is written once at ingest (both upload paths) and never touched by a replace. Backfilled from `original_filename` so existing galleries still match on their first pass.

The backfill sits **outside** the column guard and keys on `whereNull` — a run that dies partway self-heals on re-run instead of leaving half the rows empty forever. Same reasoning as the index in migration 182.

### 2. Read the marks — `GET /api/v1/events/:id/photos`

Guards copied verbatim from the sibling upload route: `apiTokenAuth` + `requireApiScope('read')` + `requirePermission('photos.view')` + `requireEventOwnership`.

Filters: `marked_only`, `mark_source` (client/mine/either), `color_labels`, `my_color_labels`, `min_rating`, `my_min_rating`, `logic`, `page`, `limit`.

The route filters to a page of ids with `PhotoFilterBuilder`, then enriches only those through `photoExportService.getPhotosWithFeedback`. Both halves already existed and neither does both; going id-first also keeps the per-colour tally query bounded by page size.

**No new token scope needed** — `requireApiScope` is hierarchical, so the `admin` token photographers already use for the export plugin authenticates this unchanged.

### 3. `services/markMerge.js` — one place that decides

A photo can carry three opinions at once (guest colours, guest star average, the photographer's row in `photo_admin_marks`). Lightroom has room for one colour and one rating.

- **Colour goes to the photographer** on a tie — one deliberate choice beats an aggregate a tie-break already had to guess at.
- **Rating takes the max** — a rating is a magnitude, and losing the higher one quietly demotes a photo somebody rated highly.
- `roundRating` is now the single implementation; `XmpGenerator.mapRating` delegates to it rather than keeping a second copy of the same five thresholds, so a sidecar and the API can't drift apart on a photo's star count.

### 4. Put the edit back — `replaces_photo_id`

`POST /api/v1/events/:id/photos` takes an optional `replaces_photo_id` and routes to the existing `replacePhoto()`.

The plugin stores the picpeak photo id on the *catalogue* photo, so the id survives the editor renaming the render — which makes the id, not the filename, the reliable key.

**Scoped to the event in the URL.** A token inherits its owner's powers across every event they can see, so an unscoped body-borne id would let one gallery overwrite another's photo. This is the `req.body` id case that ownership middleware does not cover.

Also drops the multer temp file on the two new early returns — this route only unlinks in its `catch`.

### 5. Opt-in `number_token` matching

For renders whose RAW never went through the plugin, `findReplacementCandidate` gains a `number_token` mode matching on the **longest trailing digit run**, exposed as `match_mode` on the admin upload route.

Deliberately the longest run and never a fixed last-N slice: multi-camera shoots disambiguate by prefixing the camera index *into* the number (`cam11234.jpg` / `cam21234.jpg`), and a last-4 slice reads `1234` from both bodies and reintroduces exactly the collision the prefix removes.

**Ambiguity is refused, never guessed** — two photos sharing a number upload as new, with a message pointing at the cam-prefix scheme.

## Decisions worth objecting to

- **Renders replace the proof rather than landing beside it.** Keeps the client's gallery link and all their feedback intact, and needs no new "finals" concept. The cost is that the proof is gone.
- **`mark_source` defaults to `either`,** with the photographer winning colour ties.
- **`number_token` is off by default.** A digit run is a much weaker key than a filename, so it is opt-in per upload.
- **`my_*` filters are per-admin** (migration 183 is unique on `photo_id + admin_id`), so a second admin's triage is deliberately invisible. Correct for the plugin — worth knowing for anything else.

## Verification

Ran against a **live local instance** (SQLite, all 162 migrations applied, 12 photos in a real event), not just unit tests:

| Check | Result |
|---|---|
| `source_filename` stamped at ingest | ✅ `IMG_1001.JPG` on every upload |
| Replace keeps identity | ✅ id 1 kept, event still has 12 photos — no duplicate |
| **`source_filename` survives a replace** | ✅ stayed `IMG_1001.JPG` while `original_filename` became `Smith_Wedding_1001.jpg` |
| Marks survive a replace | ✅ `green` / 5★ intact afterwards |
| `marked_only` + `mark_source=mine` | ✅ exactly the 6 marked photos |
| `mark_source=client` | ✅ picked up a colour set from the gallery lightbox by keyboard |
| `mark_source=either` | ✅ union of both, 7 photos |
| `my_color_labels`, `my_min_rating`, pagination | ✅ all correct |

Plus: `npx jest __tests__/services/lightroomRoundtrip.test.js` — **18 new tests** on a real temp-SQLite DB with every migration applied (so 185 is exercised); `npx jest` across the related suites — **49 passing**, no regressions; `npx eslint` clean on every changed file. The one pre-existing `no-unused-vars` at `adminPhotos.js:40` is untouched and present verbatim on `picpeak/main`.

### Three defects found and fixed during verification

1. `match_mode` reached the photo-cap pre-count but **not** the loop that picks the replacement target — asking for `number_token` would have been silently ignored. (`00ede1e`)
2. `number_token` did `select('*')` over every photo in the event to compare one digit run. Now reads three columns and re-reads only the winner. (`00ede1e`)
3. **The `@openapi` block was invalid YAML.** An inline `{ "green": 2 }` in an unquoted scalar parses as a flow mapping, so swagger-jsdoc threw and dropped the entire route from the generated spec — while the route kept serving normally. Only visible as a boot warning. (`13ac2c3`)

Number 3 is the argument for booting the thing rather than reading the diff; it would have shipped invisibly.

## Known wart, not fixed

Results sort by the stored `photos.filename`, which `replacePhoto()` rewrites — so a replaced photo changes position in the response. Harmless for the plugin (which matches by name, not order), but sorting on `source_filename` would keep camera order stable across replaces. Left alone rather than changing sort semantics late in review; happy to change it if you'd prefer.

## Notes

- **No frontend changes**, so no screenshots — this PR is API surface only.
- **~836 lines** trips `bypass-size-gate`. That is the review-size policy rather than failing CI.
- `nl`/`pt`/`ru` untouched — no user-facing strings in this PR.


the-luap:
Read the whole diff and ran the branch locally. The design is good — `source_filename` as a separate, replace-immune carrier is the right call, the ambiguity handling refuses to guess rather than overwriting the wrong photo, and moving the rating thresholds into `markMerge.roundRating` instead of copying them is exactly right. The 18 tests pass and the full backend suite shows no new failures against `main`.

Four things need fixing before merge, all in the new v1 branch.

## 🔴 Blockers

### 1. Replacing an external photo returns 200 and silently keeps serving the old file
`backend/src/routes/v1/events.js:678-680` — the target lookup is `where({ id, event_id })` with no origin check, so a `reference`/`external` row is accepted. `replacePhoto()` then writes a managed object but never clears `source_origin` or `external_relpath`, and the resolver gives those precedence:

```js
// photoResolver.js:23
const mode = (photo.source_origin || event.source_mode || 'managed');
if (mode === 'reference' || mode === 'external') return null;   // managed key ignored
```

So the edit is uploaded, the row is updated, the API reports success — and every viewer keeps getting the NAS original. The uploaded file is orphaned. Either reject external targets with a 400, or convert the row to managed (`source_origin: 'managed'`, `external_relpath: null`) as part of the replacement.

### 2. Every replacement leaks its temp file
`backend/src/routes/v1/events.js:694-696` clears `tempPath` on the grounds that "replacePhoto moves the temp file itself". It doesn't — `photoReplacementService.js:189` calls `storage.putFromFile()`, which is `copyFile` on local (`LocalFsStorage.js:72`) and an upload on S3. Neither removes the source, and `replacePhoto` never unlinks `newFileTempPath`. With the route's own cleanup disabled, every successful replacement leaves up to 100 MB in `storage/temp` permanently.

Worth noting the admin path has the same hole — `fs.unlink(file.path)` at `adminPhotos.js:373` sits in the new-files branch, so replaced files leak there too. Cleaning it up inside `replacePhoto` would close both at once.

### 3. The download-all ZIP isn't invalidated
`backend/src/routes/v1/events.js:709` returns without touching the cache. The admin path does this deliberately, for replacements specifically:

```js
// adminPhotos.js:471-473
// Invalidate download zip cache after successful upload or replacement
if (uploadedPhotos.length > 0 || replacedPhotos.length > 0) {
  downloadZipService.invalidate(parseInt(eventId));
}
```

Without it, guests keep downloading the pre-edit photo indefinitely — which defeats the point of putting the edit back.

### 4. The round-trip can't see reference or watcher galleries at all
`backend/migrations/core/185_add_photo_source_filename.js:41-44` backfills `whereNotNull('original_filename')`, but `fileWatcher.processNewPhoto` and `adminExternalMedia` never set `original_filename` — the camera name lives in `photos.filename` for those rows. Neither path sets `source_filename` on insert either, and the GET falls back only as far as `original_filename`:

```js
source_filename: photo.source_filename || photo.original_filename || null,
```

So for NAS-mounted and auto-imported galleries every photo reports `source_filename: null` and the plugin has nothing to match on. Those are arguably the galleries most likely to be driven from Lightroom. `COALESCE(original_filename, filename)` in the backfill, plus setting the column on those two ingest paths, would close it.

## 🟡 Concerns

1. `photoReplacementService.js:81-83` — `number_token` mode selects **every** photo row in the event, once per uploaded file, and twice per file when a photo cap is configured (candidate detection runs again in the cap pre-count at `adminPhotos.js:214`). At the 2000-file upload limit against a 5000-photo event that's up to 20M rows across thousands of sequential queries before any image processing starts. Narrowing with a `LIKE '%<token>'` before the JS check, or building the token map once per request, would keep the semantics and drop the cost.

2. `backend/src/routes/v1/events.js:707` — `logActivity(..., null, ...)` records the replacement as a system-level entry. `adminPhotos.js:432` passes `eventId`. This isn't cosmetic: the dashboard feed deliberately excludes NULL-event rows for scoped callers (`adminDashboard.js:177-186`, the GHSA-jhcf scoping), so an API replacement vanishes from the audit trail of the very photographer who owns the event.

## 🟢 Nits

- Migration number `185` collides with `185_add_category_is_folder.js` on the open `feat/1160-gallery-folders` branch. Not fatal — knex keys on filename, so both would run, ordered lexicographically — but whichever merges second should renumber.
- `higherPriorityColor` is exported from `markMerge.js` and never used anywhere else. Either wire it into the `either` merge or drop it.

## ✅ Verified

Directly inspected, not taken from the PR description:

- Migration 185 is guarded, idempotent, backfill is `whereNull`-safe, `down()` drops the index before the column (correct for SQLite's table rebuild).
- Auth chain on both routes: `apiTokenAuth` → `requireApiScope` → `requirePermission` → `requireEventOwnership`. The replacement lookup is scoped to `event_id`, so the cross-gallery overwrite the comment calls out really is closed.
- `PhotoFilterBuilder` — colours are whitelisted against `COLOR_LABELS`, sort goes through a fixed `sortMap`, everything else is a knex binding. No injection surface.
- `mark_source: 'mine'` with no `admin_id` matches nothing rather than widening to every admin's marks.
- `xmpGenerator.mapRating` genuinely delegates now — the thresholds moved rather than being duplicated.
- 18/18 tests in `lightroomRoundtrip.test.js` pass; full backend suite has the same 6 pre-existing failures as `main`, so nothing new broke.

The failing `bypass-size-gate` check is just the >300-line self-merge policy, not a code problem.

## Verdict

Close blockers 1–4 and I'm comfortable merging. The two concerns are worth an answer but I won't hold the PR on them.

Luca-Timo:
All four blockers closed, both concerns addressed, both nits done. Rebased onto `main` (was 30 behind) and force-pushed.

## 🔴 Blockers

**1. External replacement served the old file.** Fixed in `replacePhoto` rather than at the route, so the admin path gets it too: the row is repointed with `source_origin: 'managed'` and `external_relpath: null` as part of the update. I went with converting rather than a 400 — once an edit has been uploaded, serving the untouched NAS original is wrong in every case, and rejecting would have shut the round-trip out of exactly the galleries blocker 4 is about. The file on the share is never touched; this moves the pointer, not the data.

**2. Temp file leak.** You were right that `putFromFile` copies rather than moves — my comment asserting otherwise was the bug. Cleanup now lives in `replacePhoto` after the put, which closes the v1 and admin paths at once as you suggested. The v1 route also unlinks on the **failure** path, which returned before any cleanup ran at all.

**3. ZIP cache.** `downloadZipService.invalidate(event.id)` after a successful replacement, matching `adminPhotos.js:471`.

**4. Reference and watcher galleries invisible.** Backfill is now `COALESCE(original_filename, filename)`, `fileWatcher.processNewPhoto` and `adminExternalMedia` both set `source_filename` on insert, and the GET falls back the last step to `filename`. Agreed these are the galleries most likely to be driven from Lightroom — that one would have shipped as "works for me".

## 🟡 Concerns

**1. `number_token` full scan.** Narrowed with a `LIKE '%<token>%'` on `source_filename`/`original_filename` before the JS check. The token is a regex-extracted digit run so it cannot carry a wildcard. The LIKE deliberately over-matches; the exact trailing-run check still decides, so semantics are unchanged and only the row count drops. I did **not** dedupe the double call from the cap pre-count — it needs the candidate result before any file is processed, and threading a per-request cache through both call sites looked like more risk than the remaining win. Happy to do it if you'd rather.

**2. `logActivity` event scoping.** Now passes `event.id`. Hadn't connected it to the GHSA-jhcf dashboard scoping — that's a real hole, not cosmetic.

## 🟢 Nits

**Migration collision.** Became real: `185_add_category_is_folder.js` is on `main` now and `main` is at 192, so this is renumbered to **193** — with the header comment, the test that requires the path, and the four call-site comments that cite it. `gallery.js:1134` cites *your* 185 and is untouched.

**`higherPriorityColor`.** Dropped. The `either` merge deliberately gives the photographer's colour precedence rather than ranking, so wiring it in would have changed behaviour to suit an export.

## Verification

- **3 new regression tests** — the external→managed repoint, the temp-file cleanup, and the COALESCE backfill reaching a row with no `original_filename`. 21/21 in `lightroomRoundtrip.test.js`, 52/52 across the related suites.
- Rebase had one real conflict, in `adminExternalMedia.js` where #1172's capture-date work landed on the same insert. Resolved by keeping your block and re-applying the single `source_filename` line.
- `eslint` clean on every changed file. Two errors remain in `adminExternalMedia.js:7` and `fileWatcher.js:43`; both are present with my changes stashed, so they are pre-existing on `main` and I've left them.

the-luap:
Re-verified the four commits on `ec3b412c`. All four blockers, both concerns and both nits landed, and I spot-checked each at its final file:line rather than taking the summary. One new problem falls out of the blocker-1 fix — details below.

## Blockers — closed

| # | Where | ✅ |
|---|---|---|
| 1 | `photoReplacementService.js:243-244` — `source_origin: 'managed'`, `external_relpath: null` in the update | ✅ |
| 2 | `photoReplacementService.js:210-211` — unlink after `putFromFile`; failure path at `v1/events.js:699` | ✅ |
| 3 | `v1/events.js:709` — `downloadZipService.invalidate(event.id)` | ✅ |
| 4 | `193:51` `COALESCE(original_filename, filename)`; `fileWatcher.js:142` and `adminExternalMedia.js:244` set the column on insert | ✅ |

On **1**, the thing I actually went looking for: whether the old-asset delete could reach the file on the share. It can't — `resolvePhotoStorageKey` returns `null` for `reference`/`external` (`photoResolver.js:23-26`), so the `oldOriginalKey` delete at `:184-186` is skipped for exactly the rows this converts. "Moves the pointer, not the data" holds, and converting rather than 400-ing was the right call.

## Concerns — closed

1. `photoReplacementService.js:91-98` — LIKE narrowing, and you re-read the full row at `:106-108` after the narrowed select. That re-read is the bug I expected to find there; good. Not deduping the cap pre-count is fine by me — the remaining win is small and threading a cache through both call sites is the riskier change.
2. `v1/events.js:720` — `event.id`. ✅

## Nits — closed

Renumber to 193 is consistent across all ten call sites plus the test's `require` path. `gallery.js:1134` correctly still cites 185, which is the folders migration and not yours. `higherPriorityColor` dropped, and I agree with the reasoning — ranking would have changed the `either` merge to suit an export.

## 🔴 New blocker — the round-tripped photo gets re-imported as a duplicate

Nulling `external_relpath` is what serves the edit correctly, but it also drops the row out of the import's dedupe key.

`adminExternalMedia.js:194-196`:

```js
const exists = await db('photos')
  .where({ event_id: eventId, external_relpath: relFromRoot })
  .first();
if (exists) { skipped++; continue; }
```

That key — and the migration 186 unique index behind it, which is the real guard — is `(event_id, external_relpath)`. After a replacement the row holds `NULL`, and NULLs don't collide in a unique index on either engine, so neither the fast path nor the index catches it.

Concretely, on a reference gallery holding `shoot/IMG_1234.jpg`:

1. Plugin PUTs the finished edit → row converts to managed, relpath nulled.
2. The file on the share is untouched, by design.
3. Someone re-runs the import on that folder — which the comment at `:187` calls "the common case — a re-import of a folder already in the event".
4. Nothing matches `shoot/IMG_1234.jpg` any more, so the camera original is imported **again**, as a new photo.

The client's gallery then carries the edit *and* the un-edited proof. That is the round-trip defeated, on precisely the galleries blocker 4 brought into scope.

**Fix shape, with the trade-off, because neither side is free:** keep `external_relpath` and let `source_origin: 'managed'` carry the meaning. Both resolvers already give `source_origin` precedence (`photoResolver.js:23` and `:52`), so serving stays correct and the dedupe key survives untouched. The cost is that the row stays visible to `externalRelpathFold.js`, which sweeps `whereNotNull('external_relpath')` and can rewrite or staging-prefix it (`:171`, `:276`) — so that sweep would need to skip `source_origin='managed'`. Your call which side to pay on; I don't have a strong preference between that and a dedicated column.

**Adjacent, pre-existing, and not yours to fix** — flagging it because blocker 4 makes it newly reachable: `fileWatcher.js:125-129` dedupes on `filename`/`path`, and *any* replacement changes `filename`. Watcher galleries therefore carry the same duplicate shape independently of this PR, including through the admin replace path. Worth its own issue rather than growing this one.

## ✅ Verified

Directly inspected on `pr-1165-v2`, not read off your summary:

- The four blocker fixes at the lines in the table, plus the external-delete safety above.
- `21/21` in `lightroomRoundtrip.test.js`, run locally.
- The two remaining eslint errors — `adminExternalMedia.js:7`, `fileWatcher.js:43` — are byte-identical on `main`. I ran eslint on both branches; your "pre-existing" call is correct, leave them.
- Migration 193 renumber completeness, by grepping every `193` and `185` reference across `src/` and the test.

Not independently checked: the rebase conflict resolution in `adminExternalMedia.js` against #1172's capture-date work. I read the resulting insert and it looks right, but I did not reconstruct the pre-rebase state to confirm nothing else was dropped.

`bypass-size-gate` is the only red check and is the >300-line self-merge policy, not code.

## Verdict

Close the duplicate-on-re-import blocker and I'm comfortable merging. Everything from the last round is genuinely done — the external-delete path in particular was the one that could have destroyed a customer's NAS original, and it's safe.

Luca-Timo:
New blocker closed, and the pre-existing one you flagged is filed separately.

## 🔴 The re-imported duplicate

Already fixed in `e20ec63d`, pushed before your review landed — we crossed. `external_relpath` is **kept**; only `source_origin` flips to `'managed'`. Same shape you proposed, same reasoning: both resolvers give `source_origin` precedence, so serving is correct and the dedupe key survives.

**I did not take the second half of your suggestion, and I think skipping is the wrong side to pay on.** `externalRelpathFold` rebases every row from base-relative to root-relative. If managed rows are skipped, they keep a base-relative path while everything else moves — and `adminExternalMedia` computes root-relative when it scans. Nothing would match, and the camera original gets imported again. That is the exact bug keeping `external_relpath` exists to prevent, reintroduced through a different door.

So they stay in the rebase. But looking at what the sweep actually does turned up something worse than a rewrite: `b86df3db`.

## 🔴 The fold could delete the delivered edit

The fold does not merely rewrite — it **deletes** collision losers via `externalPhotoDedupe`. The survivor was whichever row was claimed first, which is iteration order:

```js
// externalRelpathFold.js:229-233
const winner = claimed.get(next);
if (winner != null) { losers.set(row.id, winner); collided++; continue; }
```

A replaced photo now sits in that sweep holding `source_origin: 'managed'` and the file the photographer just delivered. On a collision it could be deleted in favour of the untouched camera original next to it on the share.

Managed rows now claim first and therefore survive. The asymmetry is the point: the external row that loses is still on the share and a re-scan re-imports it — the edit is not recoverable.

Needed adding `source_origin` to that select, which was only fetching `id, external_relpath, size_bytes`.

## 🟡 The watcher duplicate

Filed as #1226 rather than grown into this PR, per your note. Recorded there that it predates the round-trip, applies to the admin `replace_by_name` path too, and that `source_filename` is the obvious dedupe key since it survives a replacement by design.

## Verification

- **22/22** in `lightroomRoundtrip.test.js`, including a new case pinning that a managed row wins a fold collision and the external row is the loser.
- eslint clean on both changed files.
- Thank you for checking whether the old-asset delete could reach the share — that was the failure mode I was least sure of, and having it confirmed at `photoResolver.js:23-26` is worth more than my own re-reading of it.

the-luap:
Re-verified the three commits on `b86df3db`. Both items from my round are closed, and the thing you found chasing the first one matters more than what I raised.

## The blocker — closed

`e20ec63d` keeps `external_relpath` and flips only `source_origin`. The dedupe key survives, and both resolvers give `source_origin` precedence, so serving is correct.

**You were right to reject the second half of my suggestion and I was wrong to offer it.** Skipping managed rows in the fold would leave them base-relative while every other row moved to root-relative — and `adminExternalMedia` computes root-relative when it scans, so nothing would match and the camera original would be re-imported anyway. Same bug, different door. Rebasing them and protecting them from deletion is the fix that keeps both properties.

## Your finding — verified, and worse than mine

I followed the chain rather than taking it on description: `externalRelpathFold.js:281` → `deleteDuplicatePhotos` → `externalPhotoDedupe.js:244`, a real `knex('photos').whereIn('id', ids).del()`. With the survivor decided by iteration order, a replaced photo holding the delivered edit could be deleted in favour of the untouched camera original beside it on the share, with its feedback reparented onto that original. Adding `source_origin` to the select at `:172` was genuinely required — it was fetching only `id, external_relpath, size_bytes`.

Managed-claims-first is the right call, and the asymmetry is the correct justification.

One small correction to the framing, which does not change the fix: the edit is not *quite* unrecoverable. `externalPhotoDedupe.js:256` deliberately leaves storage alone, so the file survives orphaned and unreferenced — gone from the gallery, recoverable only by hand. Still categorically worse than the external row, which a re-scan restores.

## 🟡 The one thing I would still change — the new test does not test the code

`lightroomRoundtrip.test.js` never requires `externalRelpathFold`. The new case re-implements the sort and the claim loop inside the test and asserts against its own copy:

```js
const claimOrder = placements.slice().sort((a, b) => { ... });   // a copy
```

So it pins the *idea*, not the shipped implementation. Change the real sort at `externalRelpathFold.js:241` and this test stays green. For the guard against silently deleting a delivered edit, that is the one case that should drive the real code — even if it means exporting the claim ordering as a small named helper so a test can call it directly.

22/22 pass; 1 of the 22 is not guarding what its name claims. Not holding the merge on it, but worth a follow-up.

## ✅ Verified

- Both fixes at the lines above, read on `pr-1165-v3` rather than from the summary.
- 22/22 in `lightroomRoundtrip.test.js`, run locally.
- `86785615` — good catch, and the right shape: `break` on `ENOENT` instead of spawning the missing binary twice more, and a genuinely preview-less RAW still gets the original message. RAW upload being the only consumer of exiftool is exactly why the wording was the whole diagnosis.
- `#1226` filed for the watcher duplicate rather than grown into this PR — right call, it predates the round-trip and applies to the admin replace path too.

`bypass-size-gate` is the only red check and is the >300-line self-merge policy, not code.

## Verdict

Merge cleared. The test above is the only open item and it is a follow-up, not a blocker.

the-luap (approved):
Both items closed and verified at source. Merging; the test-only follow-up is noted above.

PR #1168, opened 2026-08-25
fix(external-media): store external paths from the media root (#1163)

### LINKED ISSUE #1163: Importing a second folder into an event silently breaks every external photo already in it
**Describe the bug**

Importing a second folder into an event that already holds external photos rebases every previously imported row onto the new folder. Their originals then resolve to paths that do not exist. Nothing errors, and the gallery grid still looks intact, so the damage is easy to miss entirely.

`external_relpath` is stored relative to the folder that was imported, while `resolveExternalPath()` (`src/services/externalMediaService.js:97-102`) always rebuilds the path as:

```
EXTERNAL_MEDIA_ROOT + event.external_path + photo.external_relpath
```

and `src/routes/adminExternalMedia.js` overwrites `events.external_path` on every run. So once folder B is imported, every row from folder A is resolved underneath B.

The comment block at `adminExternalMedia.js:97` already describes this and marks it as an unfixed limitation of one base path per event. I am filing it because the failure mode is much quieter than that note suggests.

**To Reproduce**

1. Create an event in reference mode, import `Trip/Subfolder` — photos work, thumbnails and previews generate
2. Later import `Trip` into the same event (the parent, or any sibling)
3. `events.external_path` is now `Trip`, but the rows from step 1 still carry paths relative to `Trip/Subfolder`
4. Open the gallery: the grid renders fine, the lightbox does not

**Expected behavior**

Either the existing rows keep resolving, or the import refuses and explains why. Silently invalidating an entire event is the part that hurts.

 - OS: Debian 13.6 (LXC container)
 - Browser: Chrome 151
 - PicPeak Version: 3.114.0-beta.0
 - Deployment Method: Docker Compose (`docker-compose.production.yml`)
 - Database: PostgreSQL 15.19

**Logs**

There are none, and that is the core complaint — no warning, no failed import, no error in the UI. I found it by resolving a path by hand inside the backend container:

```
# what the app resolves to after the second import
/external_media/<parent>/<sub>/Leknes/_DSC0818.JPG
  -> No such file or directory

# where the file actually is
/external_media/<parent>/Leknes/_DSC0818.JPG
  -> 9043968 bytes
```

In my case **7547 of 8004 rows** pointed into the void. Only the 457 rows from the most recent import still resolved.

**Additional context**

What makes it hard to spot: **thumbnails keep working**. They are written to local storage during the import, while the base path is still correct, so the grid renders normally afterwards. What breaks is everything that needs the original again:

- preview generation fails silently, so photos never get a `preview_path`
- the lightbox falls back to the upscaled 300px thumbnail — it just looks soft and dark, like a performance problem
- downloads and any regeneration cannot find the source

I spent a while chasing this as a CPU/latency issue before realising the paths were dead. A gallery that looks slow and a gallery whose originals are unreachable are indistinguishable from the outside.

**Possible Solution**

In rough order of preference:

1. Store `external_relpath` relative to `EXTERNAL_MEDIA_ROOT` instead of to `event.external_path`, so a row is self-describing and the event's base path stops being load-bearing.
2. Keep the base path, but rewrite existing rows when it changes, in the same transaction as the update.
3. At minimum, refuse the import when the event already holds external rows under a different base that the new path does not contain, with a message saying so.

Even just a warning in the import dialog would be a large improvement over the current silence.

the-luap:
Closes #1163.

> **Stacked on #1167** — the collision handling below reuses its `deleteDuplicatePhotos`. GitHub will retarget this to `main` automatically when #1167 merges.

Importing a second folder into an event silently invalidated every photo already in it. The reporter had **7547 of 8004 rows** resolving to files that do not exist.

## Root cause

`photos.external_relpath` was stored relative to `events.external_path`, and every import overwrites that column (`adminExternalMedia.js`). `resolveExternalPath()` then rebuilt the path as `root + event.external_path + photo.external_relpath`, so the older rows were rebased onto the new folder.

The comment at `adminExternalMedia.js:112-121` already described this and marked it unfixed. What the report adds is *how quiet* it is: thumbnails are written to local storage during the import while the base path is still correct, so the grid keeps rendering. Only what needs the **original** breaks — preview generation, the lightbox, downloads — which presents as a gallery that looks slow rather than one that is broken. The reporter spent a while chasing it as a CPU problem.

## Fix

`external_relpath` becomes relative to `EXTERNAL_MEDIA_ROOT`. A row is then self-describing, and nothing an admin does to the event afterwards can move an already-imported photo. New `resolveExternalPhotoPath(photo)` replaces `resolveExternalPath(event, rel)` for photo resolution; the latter keeps its meaning ("a path under this event's base") for the import walk and faceProcessor's mount-health probe.

Also drops the duplicate-leaf-segment guess in `photoResolver.js`. It papered over this same double-prefixing, and once relpaths are root-relative it actively corrupts a path whose first segment legitimately repeats (base `Trip`, row `Trip/x.jpg`).

## Migration 187 — what it does, and what it refuses to do

Folding each event's base into its rows is trivial for a **healthy** event: `external_path + relpath` is what the app resolves today, so it changes nothing. The work is entirely in the events that have already been rebased, where that same rule would bake the broken path in permanently.

- **Ancestor repair, with provenance.** Where the current resolution is missing, it walks up the base path for an ancestor under which the file *is* there — the shape the bug produces. But existence alone is **not** accepted: a row whose file an admin simply deleted would otherwise adopt any same-named file one directory up (base `Trip/Sub`, relpath `photo.jpg`, an unrelated `Trip/photo.jpg`) and downloads would serve a **different photo**, which is worse than a dead link. A candidate must also match `photos.size_bytes`, recorded by the import from the very file the row describes. Rows carrying no size are never repaired from an ancestor.
- **It never guesses below current behaviour.** A row it cannot place keeps resolving exactly where it resolves today.
- **Skipped entirely when the mount is unreachable or empty.** An unmounted share makes every file look missing, and repairing off that signal would move every original on a healthy install.
- **Health is decided per row, not by sampling.** A rebased event whose first few rows happen to come from the most recent import would read as healthy, and every older row would be baked in wrong.

## Making it safe against interruption and its own intermediate states

- **Two phases.** Probing is read-only and runs first, so a cold NAS taking minutes does not hold a write transaction open. Every rewrite **and the marker** then commit together — split across commits, a process killed mid-fold leaves converted and unconverted rows with no marker, and the next run folds the converted ones again, putting every original one directory deeper with no undo.
- **Staged rewrites.** Every *final* path is distinct, but a final value can equal another row's *current* one (`photo.jpg` repairing to `Trip/photo.jpg` while the row already holding `Trip/photo.jpg` folds deeper), so a one-pass update violates #1167's unique index halfway through. Rows park on a per-row staging value first.
- **The runner cannot misread a failure.** On Postgres that collision surfaces as 23505, which `run-migrations-safe.js:138` treats as "schema already exists" and records as applied *after the rollback* — leaving every path unconverted with nothing to retry. 187 re-throws without the driver's code.
- **Bulk updates target the ids phase 1 captured**, not `where event_id`. An import completing during the probe inserts an already root-relative row, which the event-wide statement would prefix a second time with the stale base.
- **Collision losers are deleted, not skipped.** Skipping leaves a row holding a base-relative path that the root-only resolver reads as `<root>/<relpath>` — permanently wrong, with the marker claiming success. It is a duplicate by construction, so it goes through #1167's `deleteDuplicatePhotos`, which reparents its feedback and marks and reconciles the face clusters.
- **Idempotency is recorded, not inferred.** The tempting check — "does the relpath already start with the base?" — is wrong for any event with a subfolder named after its parent, and being wrong there corrupts a path in an operation with no undo. `down()` is a documented no-op for the same reason.

## Restores

`knex_migrations` is excluded from `.picpeak` archives, so 187 does not re-run after a restore — a pre-#1163 backup would otherwise land base-relative rows on a migrated instance and make every original unreachable with nothing logged. The fold runs there too, which is why the logic lives in `services/externalRelpathFold.js`.

Two ordering bugs came out of review:

- **The fold must precede the face requeue**, in *both* restore paths (`picpeakImportService` and `restoreService`, where the requeue sat in step 6 ahead of the step 7c migrations). The worker is live throughout; queued first, it claims an external row while the path is still event-relative, resolves it wrongly and burns it to `failed` — a state only an explicit Re-scan clears and the later fold does not undo.
- **A failed conversion is no longer reported as a clean restore.** It returns `externalPathsConverted` / `externalPathError`, suppresses the face requeue, and the restore card surfaces it — previously an admin whose external originals were all unreachable got a green success.

## Tests

`__tests__/migrations/187_external_relpath_from_root.test.js` (18) against a real temp directory tree, since the whole mechanism is "is this file actually there": the repair, the wrong-size refusal, the unmounted root, the staging collision, a row inserted mid-probe, the double-fold guards, and the parent-named subfolder. Plus `__tests__/integration/externalImportSecondFolder.test.js` (5) driving the real route and resolver end to end.

Updated `photoResolver`, `ensurePreviewImage.external`, `faceTransientSource`, `regenerateThumbnailsScript` — all carried base-relative fixtures encoding the old contract.

Full backend suite: 6 pre-existing failures, identical to `origin/main`.

No UI surface touched beyond the restore-failure notice.

### DIFF
--- a/backend/__tests__/integration/externalImportConcurrency.test.js
+++ b/backend/__tests__/integration/externalImportConcurrency.test.js
@@ -173,14 +173,16 @@ describe('concurrent external imports (#1162)', () => {
       path: 'x/a.jpg',
       type: 'individual',
       source_origin: 'external',
-      external_relpath: path.join('individual', 'a.jpg'),
+      // Root-relative, as the route now writes it (#1163) — the competing
+      // writer has to target the same value for the race to be real.
+      external_relpath: path.join('nas', 'individual', 'a.jpg'),
     };
 
     const res = await runImport(eventId);
 
     expect(res.status).toBe(200);
     const counts = await relpathCounts(eventId);
-    expect(counts.get(path.join('individual', 'a.jpg'))).toBe(1);
+    expect(counts.get(path.join('nas', 'individual', 'a.jpg'))).toBe(1);
     // Two imported by us, one lost to the other writer and reported honestly.
     expect(res.body.imported).toBe(2);
     expect(res.body.skipped).toBe(1);
@@ -194,7 +196,9 @@ describe('concurrent external imports (#1162)', () => {
       path: 'x/a.jpg',
       type: 'individual',
       source_origin: 'external',
-      external_relpath: path.join('individual', 'a.jpg'),
+      // Root-relative, as the route now writes it (#1163) — the competing
+      // writer has to target the same value for the race to be real.
+      external_relpath: path.join('nas', 'individual', 'a.jpg'),
     };
 
     await runImport(eventId);
--- a/backend/__tests__/integration/externalImportSecondFolder.test.js
+++ b/backend/__tests__/integration/externalImportSecondFolder.test.js
@@ -1,0 +1,175 @@
+/**
+ * Importing a second folder must not move the photos already in the event (#1163).
+ *
+ * events.external_path is overwritten by every import, and external_relpath
+ * used to be stored relative to it — so a second import silently rebased every
+ * existing row onto the new folder. The reporter had 7547 of 8004 originals
+ * pointing at files that do not exist, and nothing said so: thumbnails are
+ * written to local storage during the import while the base path is still
+ * correct, so the grid carries on rendering.
+ *
+ * Driven through the real route and the real resolver, against a real
+ * directory tree — the failure is entirely about whether a file is where the
+ * app looks for it.
+ */
+
+const fs = require('fs');
+const path = require('path');
+const os = require('os');
+const express = require('express');
+const request = require('supertest');
+
+describe('a second external import (#1163)', () => {
+  let tmpDir; let db; let app; let mediaRoot; let resolvePhotoFilePath;
+
+  const touch = async (rel) => {
+    const full = path.join(mediaRoot, rel);
+    await fs.promises.mkdir(path.dirname(full), { recursive: true });
+    await fs.promises.writeFile(full, 'not-a-real-jpeg');
+  };
+
+  beforeAll(async () => {
+    tmpDir = await fs.promises.mkdtemp(path.join(os.tmpdir(), 'picpeak-ext2nd-'));
+    mediaRoot = path.join(tmpDir, 'media');
+    await fs.promises.mkdir(mediaRoot, { recursive: true });
+
+    process.env.NODE_ENV = 'test';
+    process.env.TEST_DATABASE_PATH = path.join(tmpDir, 'data', 'db.sqlite');
+    await fs.promises.mkdir(path.dirname(process.env.TEST_DATABASE_PATH), { recursive: true });
+    process.env.STORAGE_PATH = path.join(tmpDir, 'storage');
+    process.env.EXTERNAL_MEDIA_ROOT = mediaRoot;
+    process.env.JWT_SECRET = process.env.JWT_SECRET || 'ext2nd-secret';
+
+    jest.resetModules();
+
+    jest.doMock('../../src/middleware/auth', () => ({
+      adminAuth: (req, _res, next) => { req.admin = { id: 1, username: 'tester', roleName: 'admin' }; next(); },
+    }));
+    jest.doMock('../../src/middleware/permissions', () => ({
+      requirePermission: () => (_req, _res, next) => next(),
+    }));
+    jest.doMock('../../src/middleware/ownership', () => ({
+      requireEventOwnership: (_req, _res, next) => next(),
+    }));
+    jest.doMock('sharp', () => () => ({ metadata: async () => ({ width: 100, height: 200 }) }));
+    jest.doMock('../../src/services/imageProcessor', () => ({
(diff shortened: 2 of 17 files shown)


the-luap:
## Screenshot

The restore result card when `foldExternalRelpaths` throws — the state this PR adds, and the one an admin sees when every external original in the restored library is unreachable.

restore-external-paths-failed

**How this was produced:** the `/admin/backup/picpeak/import` response was stubbed in the browser to return the exact body the route sends on a failed conversion (`externalPathsConverted: false` plus the error text). No restore was run — triggering this for real means wiping a database and then breaking the fold mid-way, which is not something to do to a live instance for a screenshot. The component, its props and the rendered markup are the real ones.

One thing the shot makes obvious that the diff does not: the surrounding box is still **green** and still says "Backup restored." That is accurate — the rows and files genuinely are in place — but a green success box carrying a red "every referenced original is currently unreachable" is a mixed signal. Worth deciding whether the box should turn amber when `externalPathsConverted === false`; I left it alone rather than widening the diff.

Issue #1015, opened 2026-08-10
[BUG] Image Fit of Diashow not working


mbpictures:
**Describe the bug**
I have portrait only images in my gallery. Therefor I set the "Image Fit" in the global Diashow settings to "Black Bars (No crop)". But this setting doesn't reflect in the diashow.

**To Reproduce**
1. Go to Settings -> Diashow
2. Set image fit to no crop
3. Launch diashow with portrait images on a landscape monitor
4. Images are still cropped

**Expected behavior**
As the setting suggests black bars are added to the left/right and the image doesn't get cropped.

**Screenshots**

**Environment (please complete the following information):**
 - OS: Ubuntu 22.04
 - Browser: Brave v1.93.132
 - PicPeak Version: v3.45.14
 - Deployment Method: Docker Compose (Dokploy)
 - Database: PostgreSQL 15

**Logs**
Please include relevant logs:
```
# Backend logs
docker-compose logs backend | tail -50

# Frontend console errors
[paste any browser console errors]
```

**Additional context**
Cropping the browser window to portrait dimensions doesn't fix the problem (the image remains in 16:9 landscape crop, even though it's portrait, and black bars are added as needed to remain the landscape version of the image).

**Possible Solution**
If you have an idea how to fix the issue, please describe it here.

mbpictures:
I just discovered: it seems like the aspect ratio is fixed to 16:9 with "No crop" being set, as the setting itself changes the behaviour when the browser window get's resized. Example:

This is the image I uploaded (from lorem picsum):

<img width="400" height="600" alt="Image" src="https://github.com/user-attachments/assets/12ee6011-4736-43a6-a3a3-2e9f5867e6f8" />

This is how it get's displayed with "No crop":

<img width="3822" height="1876" alt="Image" src="https://github.com/user-attachments/assets/3625bf47-6479-4c77-8ecf-18e113f4529d" />

My expactation would be, that the image doesn't get cropped at all and instead just has larger black bars.

the-luap:
Confirmed, and your follow-up diagnosis is right — the aspect really is pinned to 16:9. The crop isn't coming from the fit setting, it's already baked into the file being served.

Chain:

- The slideshow picks its image as `preview_url || hero_url || url` (`SlideshowPage.tsx:29`).
- `preview_url` is only emitted when **Lightbox previews** are enabled (`gallery.js:957`), which is off by default — so on a stock install it's `null`.
[...]

PR #1169, opened 2026-08-25
fix(gallery): stop the lightbox loading originals to display a photo (#1166)

### LINKED ISSUE #1166: [BUG] Gallery loads full-resolution originals instead of optimized preview images
## Summary

The gallery frontend appears to load the full-resolution original image for the normal photo viewer, even though the API provides dedicated thumbnail and preview/slideshow URLs.

This causes the browser to download very large JPEG files (e.g. ~16.5 MB for a single 4160×6240 image) when viewing photos in the gallery.

The backend endpoints themselves appear to work correctly. The problem seems to be in the frontend's image URL selection / rendering logic.

---

## Environment

- PicPeak: self-hosted instance
- Browser: Firefox
- Frontend: production build
- Backend/API: same PicPeak instance
- Gallery: anonymized private gallery
- Example photo ID: `47`

---

## Expected behavior

When browsing photos in the gallery, the frontend should use an appropriately sized image variant, such as the preview/slideshow image, rather than downloading the original full-resolution image.

For example:

```text
/photo/47
```

returns the original:

```text
4160 × 6240
~16.5 MB
```

while the API also provides:

```text
/preview/47
```

which returns:

```text
image/jpeg
~345 KB
```

The thumbnail endpoint is even smaller:

```text
/thumbnail/47
~57 KB
```

The browser should therefore not need to download the 16.5 MB original merely to display the photo in the gallery viewer.

---

## Actual behavior

When opening a photo in the gallery, the browser creates an `` element whose `src` is the full-resolution `/photo/:id` endpoint.

Example observed DOM:

```html
<img
  src="https://example.invalid/api/gallery/<gallery>/photo/47"
  width="190"
  height="285"
  class="shrinkToFit"
>
```

The browser reports:

```text
src:
https://example.invalid/api/gallery/<gallery>/photo/47

currentSrc:
https://example.invalid/api/gallery/<gallery>/photo/47

natural dimensions:
4160 × 6240
```

Therefore the browser is actually downloading the original image despite displaying it at a much smaller size.

---

## API response data

The `/photos` API response contains the following fields for the affected photo:

```json
{
  "id": 47,
  "url": "/api/gallery/<gallery>/photo/47",
  "thumbnail_url": "/api/gallery/<gallery>/thumbnail/47",
  "hero_url": "/api/gallery/<gallery>/hero/47",
  "preview_url": null,
  "slideshow_url": "/api/gallery/<gallery>/preview/47"
}
```

This is important because the API already exposes a dedicated `slideshow_url`.

---

## Measured file sizes

For the same photo:

| Endpoint | Content-Type | Size |
|---|---|---:|
| `/photo/47` | `image/jpeg` | ~16.5 MB |
| `/preview/47` | `image/jpeg` | ~345 KB |
| `/thumbnail/47` | `image/jpeg` | ~57 KB |

The endpoints return HTTP 200 and valid JPEG data.

For example, `/preview/47` returned:

```text
STATUS: 200
CONTENT-TYPE: image/jpeg
CONTENT-LENGTH: 344775
BLOB SIZE: 344775
BLOB TYPE: image/jpeg
```

The original `/photo/47` returned:

```text
STATUS: 200
CONTENT-TYPE: image/jpeg
CONTENT-LENGTH: 16534581
BLOB SIZE: 16534581
BLOB TYPE: image/jpeg
```

So this does not appear to be an image-processing failure or a missing preview.

---

## Important frontend finding

The production frontend contains a separate slideshow bundle:

```text
/assets/SlideshowPage-*.js
```

Inspecting the minified bundle revealed this logic:

```javascript
function O(r) {
    return U(r.slideshow_url || r.preview_url || r.url)
}
```

This appears to be correct for the slideshow component because it prefers:

1. `slideshow_url`
2. `preview_url`
3. `url`

The slideshow therefore uses the dedicated preview image correctly.

However, the normal gallery/photo viewer appears to use:

```text
photo.url
```

directly.

As a result, it resolves to:

```text
/photo/47
```

instead of:

```text
/preview/47
```

---

## Additional observation

The browser's DOM confirms that the displayed image is actually the original:

```javascript
[...document.images][0].src
```

returns:

```text
https://example.invalid/api/gallery/<gallery>/photo/47
```

and:

```javascript
[...document.images][0].currentSrc
```

returns the same URL.

The image reports:

```text
width: 4160
height: 6240
```

even though the rendered `` element is only approximately:

```text
190 × 285
```

This strongly suggests that the browser is receiving the original image rather than a resized variant.

---

## Reproduction steps

1. Open a gallery containing high-resolution images.
2. Open the gallery in Firefox.
3. Open a photo in the normal gallery/photo viewer.
4. Open the browser developer console.
5. Run:

```javascript
[...document.images][0].src
```

6. Observe that the image URL ends in:

```text
/photo/<id>
```

7. Fetch the photo metadata:

```javascript
fetch("/api/gallery/<gallery>/photos")
  .then(r => r.json())
  .then(data => {
    const p = data.photos.find(x => x.id === 47);
    console.log(p);
  });
```

8. Observe that the same photo object contains:

```text
thumbnail_url
preview_url
slideshow_url
```

9. Directly request the slideshow/preview URL:

```text
/api/gallery/<gallery>/preview/47
```

10. Observe that it returns a much smaller JPEG.

---

## Expected implementation

The normal gallery viewer should probably use the preview/slideshow URL rather than `url`.

For example, conceptually:

```javascript
const imageUrl =
    photo.slideshow_url ||
    photo.preview_url ||
    photo.url;
```

The exact implementation is of course up to the project.

The important point is that the frontend should not use the original `/photo/:id` endpoint for normal on-screen viewing when a suitable preview variant exists.

---

## Why this matters

This has a significant performance impact.

A single photo can cause approximately:

```text
16.5 MB
```

of image data to be transferred even though the image is displayed at a small size.

For a gallery containing dozens of photos, this can result in hundreds of megabytes of unnecessary network traffic.

It can affect:

- initial gallery loading
- photo navigation
- mobile users
- users with slower connections
- bandwidth consumption
- browser memory usage
- server bandwidth
- perceived gallery performance

The difference for this example is approximately:

```text
Original: ~16.5 MB
Preview:  ~0.345 MB
```

So the preview is roughly two orders of magnitude smaller.

---

## What appears to be working correctly

The following does not appear to be broken:

- `/thumbnail/:id`
- `/preview/:id`
- `/photo/:id`
- `/photos` API response
- generation/storage of the preview image
- API metadata containing `slideshow_url`

The backend successfully provides the different image variants.

The issue appears to be the frontend choosing the wrong variant for the normal gallery viewer.

---

## Possible affected component

The issue may be located in the component responsible for rendering normal gallery photos rather than the dedicated slideshow component.

The slideshow bundle already contains logic equivalent to:

```javascript
photo.slideshow_url || photo.preview_url || photo.url
```

It may therefore be worth checking whether the normal gallery image component uses `photo.url` directly instead of selecting an appropriate preview URL.

---

## Suggested fix

Use the appropriate derived image URL for each viewing context.

For example:

- Thumbnail/grid → `thumbnail_url`
- Normal photo viewer → `slideshow_url` / `preview_url`
- Hero image → `hero_url`
- Full-resolution download → `url` or the appropriate secure download endpoint

The original image should ideally only be fetched when the user explicitly needs the full-resolution image.

---

## Evidence

The following observations were made directly in the browser and against the production frontend/API.

### Photo ID

```text
47
```

### Original

```text
GET /photo/47
HTTP 200
Content-Type: image/jpeg
Content-Length: 16534581
Blob size: 16534581
Dimensions: 4160 × 6240
```

### Preview

```text
GET /preview/47
HTTP 200
Content-Type: image/jpeg
Content-Length: 344775
Blob size: 344775
```

### Thumbnail

```text
GET /thumbnail/47
HTTP 200
Content-Type: image/jpeg
Content-Length: ~57 KB
```

The frontend DOM nevertheless uses:

```text
/photo/47
```

as the image `src`.

---

## Conclusion

This appears to be a frontend image-selection bug rather than a backend image-generation or storage problem.

The API already exposes the appropriate smaller image variants, but the normal gallery viewer appears to bypass them and request the original full-resolution image.

Could you please check the image URL selection logic in the normal gallery/photo viewer and ensure that an appropriate preview/slideshow asset is used for on-screen viewing?

the-luap:
Closes #1166.

The lightbox loads the full-resolution original to display a photo on screen. The reporter measured **16.5 MB where the preview is 345 KB**.

## Root cause

`PhotoLightbox.tsx` read `preview_url`, which the server only emits once an admin has flipped `lightbox_preview_enabled` (`gallery.js:1146`) — **off by default** (migration `104_add_lightbox_preview_tier.js:9`). So a stock install falls straight through to `url`. The reporter's API dump shows exactly that: `preview_url: null`, `slideshow_url` set.

Worse than measured: the lightbox renders its neighbour slides too, so opening one photo pulled **three** originals.

## Fix

Prefer `slideshow_url` before `url`. It is the same `/preview/:id` URL — same watermark query, built from the same string server-side — but emitted unconditionally for images since #1015, because the slideshow never had a fallback worth taking. So this fixes every existing install with no migration and no admin action.

Extracted as `lightboxImageUrl()` in `imageTiers.ts`, which turned out to matter: **two other surfaces were bypassing the lightbox entirely.**

- **Premium galleries** return early from `PhotoGridWithLayouts` and never render `PhotoLightbox` — they build their own slides with `src: photo.url`, so they kept pulling originals. Fixing that also required carrying the photo id on the slide: `handleDownloadFromLightbox` recovered the photo by matching `slide.src === photo.url`, which a derivative src breaks, and Download would have become a silent no-op.
- **The Story layout** rendered the full original as its **grid tile**, at `object-cover` in a small card. That is the one place where the reporter's "hundreds of megabytes for a gallery" was literally true, and neither the issue nor the external review found it. It uses the preview tier now — not the thumbnail, deliberately: `thumbnail_fit` is seeded to `cover` on every install, so thumbnails are square centre-crops, and these cards are not square (400×500 in the carousel, fixed-height in the desktop grid), so a thumbnail would be cropped a second time by the card's own `object-cover` and reframe every photo.

`url` remains the last resort for videos, where both derivative URLs are null.

## Verification

On the local rig with the toggle **off**, so `/photos` returns `preview_url: null` exactly as filed. Opening one photo:

```
before   GET /photo/82, /photo/81, /photo/21                     (3 originals)
after    GET /preview/82?w=1280, /preview/81?w=1280, /preview/21?w=1280
```

Per-image payload on that gallery: **397 KB → 23 KB**. (Its test photos are small; the reporter's 4160×6240 originals give the ~48× he measured.)

## Formats that keep the original

`generatePreviewImage` unconditionally encodes JPEG (`imageProcessor.js:668`), which drops **animation and alpha**. GIF, APNG and PNG therefore bypass the preview tier: PNG because transparency is the norm there and because an APNG is normally reported as `image/png` rather than `image/apng`, so the specific check alone would miss the common upload path. PNG is rare enough in an event gallery that the bandwidth given up is small.

Animated or alpha **WebP** declares `image/webp` exactly like an ordinary still and cannot be told apart from MIME — a guess this layer can never make correctly. **#1171 fixes that at the source** (WebP encoding for alpha/multi-page sources in `generatePreviewImage`) and deletes this bypass entirely, giving PNG galleries the bandwidth win back and fixing the slideshow, admin previews and face avatars at the same time.

## Two corrections to the report

Worth recording so nobody chases the wrong thing:

- **The DOM evidence is not from the gallery.** `class="shrinkToFit"` is Firefox's generated *standalone image document* markup — a raw image opened in a tab, not PicPeak's viewer. The conclusion is right; that evidence does not support it.
- **The default grid is not affected.** `PhotoCard.tsx:367` uses `thumbnailUrlForTile`. Only the lightbox — and, as above, the Story cards.

## One review finding rejected

External review flagged this as bypassing the secure-image route on `enhanced`/`maximum` galleries. It does not. `AuthenticatedImage` collects `requiresToken` and `secureUrlTemplate` into an explicitly-voided `unusedProps` and never substitutes `{{token}}` — so on those protection levels `photo.url` was a literal `.../secure/82/{{token}}`, which returns **HTTP 400**. Verified against a live maximum-protection gallery. The lightbox was falling back to the 300px thumbnail, not to a protected image; the real transition is *broken fallback → access-controlled, watermarked 1920 preview*, and the preview route does apply watermarks (`gallery.js:2641`) behind `verifyGalleryAccess`. The reviewer withdrew the finding on that evidence.

## Settings copy

The toggle no longer decides whether the lightbox uses previews, and it never generated anything on save. It now says what it does — unlock the regenerate button, keep `preview_url` emitted — in en/de/fr/sl, the locales carrying those keys.

## Story galleries load on approach

Every card in a Story gallery mounts at page load — `whileInView` gates the *animation*, not the render — and `AuthenticatedImage` fetches from an effect on mount, so all of them requested at once. Tolerable while they pointed at `photo.url`, because nothing was generated; pointing them at the preview tier would mean a gallery with cold previews Sharp-decodes every original in one burst.

The image now waits until the card is within 200px of the viewport, using framer-motion's `useInView` — the same observer the entrance animation already relies on — with `once` so a card never unloads on scroll-away. Measured on a 62-photo Story gallery: **3 images fetched at load, growing to 15 as you scroll**, where all 62 would have fired before.

Confirming that turned up the last one: `StoryHero` rendered `photo.url` as a full-bleed `object-cover` background — a full original on the critical path for first paint of every Story gallery — when `hero_url` exists for exactly this and is emitted unconditionally (`gallery.js:1139`). That gallery now issues **no `/photo/` request at all**.

## Known limitation

The premium layout memoizes its slide URLs, so rotating the device before opening the lightbox can leave a photo on the tier chosen for the old geometry. The result is one tier of sharpness, and the fix is either recreating the slides array every render (which risks resetting the lightbox mid-view) or a resize subscription the layout does not otherwise need — not worth it.

Screenshots follow in a comment.

## Tests

`__tests__/lightboxImageUrl.test.ts` (10): the reported install shape, the watermark query surviving, videos still reaching `url`, the animated/alpha bypasses, and the #1095 per-device tier sizing surviving the new fallback. 4 of the original 6 fail on the old logic.

Frontend suite: 34 files / 222 tests. `tsc --noEmit` clean.


the-luap:
## Screenshots

Captured against the running stack with `lightbox_preview_enabled` **off**, so `/photos` returns `preview_url: null` — the reporter's exact configuration.

### Lightbox, rendering from the preview tier
Network for this open: `GET /preview/82?w=1280` plus both neighbours. No `/photo/:id` request at all. Before the change the same open issued `GET /photo/82`, `/photo/81`, `/photo/21` — three originals.

issue-1166-lightbox-after

### Story layout — hero and cards
The hero is on `hero_url` (it rendered a full original before), the cards are on the preview tier, and only the ones near the viewport have loaded: **3 images fetched for 62 cards**, rising to 15 as you scroll. Zero `/photo/` requests on the whole page.

issue-1166-story-after

### Settings → Thumbnails
The toggle no longer governs whether the lightbox uses previews, so its copy no longer claims to. It now says what it still does — unlock the regenerate button below it — since saving it never generated anything on its own.

issue-1166-settings-after

Issue #1162, opened 2026-08-24
External import: concurrent runs insert every file twice (check-then-insert, no unique constraint)


BraynArts:
**Describe the bug**

Two overlapping `POST /api/admin/events/:id/import-external` runs against the same event insert every file twice. My event ended up with 8004 photo rows for 6012 distinct `external_relpath` values — 1992 duplicates.

The pairs are unmistakable: identical path, sub-millisecond apart.

```
Letzter abend/_DSC5547.JPG   21:02:05.463941
Letzter abend/_DSC5547.JPG   21:02:05.463190
```

**To Reproduce**

1. Create an event in reference mode and point it at an external folder
2. Start the external import (admin UI, or `POST /api/admin/events/:id/import-external`)
3. While the first run is still walking the tree, start the same import again — a double click, a client retry, or simply pressing it again because a slow run looks stuck
4. Both runs finish, and the event holds two rows per file

A large tree makes this easy to hit, because the first run takes long enough to look hung.

**Expected behavior**

Re-importing the same folder should be idempotent. The second run should either join the first, be rejected, or skip everything already inserted.

**Environment (please complete the following information):**
 - OS: Debian 13.6 (LXC container)
 - Browser: Chrome 151
 - PicPeak Version: 3.114.0-beta.0
 - Deployment Method: Docker Compose (`docker-compose.production.yml`)
 - Database: PostgreSQL 15.19

**Logs**

Nothing is logged — both runs report success. The evidence is in the data:

```sql
SELECT count(*) AS rows,
       count(DISTINCT external_relpath) AS files,
       count(*) - count(DISTINCT external_relpath) AS duplicates
FROM photos WHERE event_id = <id>;

 rows | files | duplicates
------+-------+------------
 8004 |  6012 |       1992
```

**Additional context**

The guard exists but is not atomic. `src/routes/adminExternalMedia.js` (~line 158):

```js
const exists = await db('photos')
  .where({ event_id: eventId, external_relpath: f.rel })
  .first();
if (exists) { skipped++; continue; }
...
const inserted = await db('photos').insert({ ... })
```

Check-then-insert with no transaction, and nothing else closes the window:

1. **No unique constraint.** `photos` carries twelve indexes, none covering `(event_id, external_relpath)` — so the database cannot reject what the check misses.
2. **No in-flight guard.** The route does not track a running import per event, so a second request just starts a parallel walk of the same tree.

Both requests read, both find nothing, both insert.

[...]

PR #1170, opened 2026-08-25
fix(admin): make "Storage used" report storage used (#1164)

### LINKED ISSUE #1164: Dashboard "Storage used" reports catalogued original bytes, not storage used
**Describe the bug**

The admin dashboard tile labelled "Storage used" does not report storage used. It reports the summed byte size of catalogued originals — which, for a reference-mode install, live entirely on external storage and have no relationship to the disk PicPeak runs on.

`src/routes/adminDashboard.js:93`:

```js
const storageUsed = await applyEventScope(db('photos'), req.admin, 'event_id')
  .sum('size_bytes as total')
```

**To Reproduce**

1. Run an install in `source_mode: reference` with originals on external storage
2. Import a few thousand photos
3. Compare the dashboard tile with `df -h` on the PicPeak volume

**Expected behavior**

A tile named "Storage used" should describe storage that is actually used, or be named after what it measures.

 - OS: Debian 13.6 (LXC container)
 - Browser: Chrome 151
 - PicPeak Version: 3.114.0-beta.0
 - Deployment Method: Docker Compose (`docker-compose.production.yml`)
 - Database: PostgreSQL 15.19

**Logs**

Not log-visible. On my install the tile read roughly 80 GB while actual local usage was 21 GB.

**Additional context**

Three things pull the number away from reality, all in the same direction:

- In reference mode the originals are never copied. Those bytes sit on the external share.
- Duplicate rows inflate it, since the same file is counted once per row (see the concurrent-import issue I just filed).
- It ignores what PicPeak genuinely does write locally: thumbnails, previews, hero images, and the per-event download cache.

So the one number an admin would reach for when wondering "am I running out of disk" points away from the answer, and omits precisely the things that fill the disk.

Worth mentioning alongside, since it is the same blind spot: `.download-cache/all.zip` (`downloadZipService.js:44`) has no TTL and no size cap. One of my events held an 11.8 GB zip from a single "download all" click. Nothing would ever have reclaimed it, and the storage tile did not account for it. Deleting it by hand was safe — `getZipInfo()` stats the file, clears `events.download_zip_path` on a miss, and rebuilds on next request — but I only found it because I went looking with `du`.

**Possible Solution**

Either works; the second is more useful:

1. Relabel the tile to something like "Catalogued media", so the number means what it computes.
2. Report real local usage — sum the derived artefacts under the storage root — and show catalogued original size as a separate figure next to it.

A cap or TTL on the download cache would fit naturally with option 2.

the-luap:
Closes #1164.

The dashboard tile labelled "Storage used" did not report storage used. `adminDashboard.js` summed `photos.size_bytes` — the catalogued size of the **originals**, which in reference mode live entirely on external storage and have no relationship to the disk PicPeak runs on. The reporter's tile read ~80 GB against 21 GB of real local usage.

## Bigger than the label

The reporter framed this as naming. It is worse: the same number drove the **soft-limit warning bar** (`AnalyticsPage.tsx`) and, via `/storage/info`, the **recommended soft limit** (`adminSettings.js:1899-1906`). A reference-mode install got a disk-capacity recommendation computed from bytes that are not on the disk. Real `df` stats were already in that handler and simply weren't what the tile showed.

Three things pull the old figure away from reality, all in the same direction — reference-mode originals that were never copied, duplicate rows counting the same file twice (#1162), and everything PicPeak genuinely writes locally being invisible.

## Fix

New `services/localStorageUsage.js` walks the storage root and reports the total plus a breakdown.

Walking rather than summing DB columns is the point: `thumbnail_path` / `preview_path` / `hero_path` record a *key* and never a byte count, and orphans from a deleted event or an interrupted import are real bytes on a real disk. A `du` is the only honest answer, and the only one that notices what PicPeak has forgotten about.

## The exclusion that matters most

**`EXTERNAL_MEDIA_ROOT` is not walked** when it lies inside the storage root. Its compose default is `<storage>/external-media`, where the NAS is bind-mounted — a plain **directory, not a symlink**, so the symlink guard did not cover it. Walking it would put every referenced original back into a figure whose entire purpose is to leave them out, and compare NAS bytes against `statfs()` of the local disk: the exact over-count this replaces, reappearing inside its own fix. A directory that merely shares the name is still counted, because those really are local bytes.

Other details worth review:

- **`.download-cache` gets its own line.** It lives *inside* the event directory, so the naive rule files an 11.8 GB zip as photography — and it is the one bucket that is pure disposable cache, which makes it the one an admin most wants to see separately.
- **Categories use the names the writers actually use** — `heroes` from imageProcessor, `watermarks` from watermarkService, `business-docs`. Getting one wrong is not a crash, it is a silent 30 MB in "other".
- **Symlinks are not followed**, so a link into the media mount cannot put the NAS back in the total.
- **Cached 5 minutes, and concurrent cold-cache callers share one walk.** `/dashboard/stats`, `/storage/info` and the sidebar are routinely requested together, and each was starting its own stat-per-file traversal of the whole library.
- **Not event-scoped**, unlike everything else on `/stats`: disk is not divisible by which admin owns which event. Pinned in `dashboardScope.test.js` so a future reviewer reading "everything here is scoped" doesn't turn it back into a sum of one editor's photos.

## S3 backends keep the catalogued figure

With `STORAGE_BACKEND=s3` the originals, renditions, archives and download caches are objects in the bucket and `STORAGE_PATH` holds only incidental local files — so the walk would report near-zero and drag the recommendation with it. Those installs keep the catalogued approximation they had before this PR, the walk is **skipped before it runs** (that endpoint is polled), and the response says which measurement it is (`storage_measurement: 'disk' | 'catalog'`) so the UI labels it rather than implying a disk measurement that never happened.

## Honesty about what is not known

- A failed measurement reads as **"unavailable"**, not `0 Bytes` and not the catalogued figure — they are different quantities and on a reference install they are supposed to diverge.
- A **partial** walk (an unreadable subtree) makes the total a floor. It is marked `+` in all three surfaces — dashboard tile, analytics card, sidebar, status tab — because a floor silently compared against a soft limit reads as "safely under".
- The analytics bar no longer renders an absent measurement as `0%` with the over-limit state suppressed, which read as plenty of room at exactly the moment nothing was known.

The catalogued figure is kept and shown next to the real one, so the difference stops being invisible.

## On the local rig

```
Storage used   64.37 MB          (was: 15.75 MB)
Catalogued     15.75 MB

originals      17.7 MB      watermarks    27.9 MB
archives        5.1 MB      downloadCache  6.8 MB
thumbnails      5.1 MB      heroes         1.1 MB
previews        3.8 MB      other              0
```

27.9 MB of watermarks and 6.8 MB of download cache — none of which the old figure could see.

Screenshots follow in a comment.

## Not addressed

- `.download-cache/all.zip` (`downloadZipService.js:44`) still has no TTL or size cap. It is only cleared on invalidation or event delete/archive, so a live event holds it indefinitely — the reporter's second observation, and correct. Custom-resolution job archives *do* get swept (`downloadJobCleanupService.js`); the pre-built zip does not. This PR at least makes it visible, which is what makes the case for capping it.
- `settingsService.formatBytes` tops out at TB, so a bogus capacity renders as e.g. "1.4 undefined". Pre-existing and unrelated; noted because it is visible in the same tile.

## Tests

`__tests__/services/localStorageUsage.test.js` (11) against a real temp tree: the breakdown, the download-cache split, orphans, a missing root, the cache and its in-flight coalescing, the symlink escape, and the nested-media-mount exclusion.

`dashboardScope.test.js` updated — its `storageUsed === 1000` assertion encoded the old meaning; it now asserts `catalogedBytes` carries the scoping, that `storageUsed` deliberately does not, and that an S3 backend reports the catalogued figure.

Full backend suite: 6 pre-existing failures, identical to `origin/main`. Frontend: 33 files / 212 tests, `tsc --noEmit` clean.

### DIFF
--- a/backend/__tests__/routes/dashboardScope.test.js
+++ b/backend/__tests__/routes/dashboardScope.test.js
@@ -127,7 +127,23 @@ describe('dashboard scoping (GHSA-c2jj / gqx7 / jhcf)', () => {
     expect(res.status).toBe(200);
     expect(Number(res.body.totalEvents)).toBe(1);
     expect(Number(res.body.totalPhotos)).toBe(1);
-    expect(Number(res.body.storageUsed)).toBe(1000);
+    // The catalogued original bytes — this is what carries the per-event
+    // scoping, and what `storageUsed` reported before #1164.
+    expect(Number(res.body.catalogedBytes)).toBe(1000);
+  });
+
+  it('/stats reports disk usage unscoped, because disk is not per-event', async () => {
+    // storageUsed is a measurement of the storage root (#1164), so it is the
+    // same number for every admin by design. Pinned so a future reviewer
+    // reading "everything on this endpoint is scoped" does not turn it into a
+    // sum of this editor's photos again — which is the bug that was fixed.
+    const res = await request(app)
+      .get('/api/admin/dashboard/stats')
+      .set('Authorization', `Bearer ${editorToken}`);
+
+    expect(res.status).toBe(200);
+    expect(res.body.storageUsed).not.toBe(1000);
+    expect(res.body).toHaveProperty('storageBreakdown');
   });
 
   it('/analytics does not expose a foreign gallery name or slug', async () => {
--- a/backend/__tests__/routes/dashboardScope.test.js
+++ b/backend/__tests__/routes/dashboardScope.test.js
@@ -146,6 +146,27 @@ describe('dashboard scoping (GHSA-c2jj / gqx7 / jhcf)', () => {
     expect(res.body).toHaveProperty('storageBreakdown');
   });
 
+  it('/stats reports the catalogued figure on an S3 backend, not a near-zero disk walk', async () => {
+    // STORAGE_PATH holds only incidental local files when objects live in a
+    // bucket, so walking it would report near-zero and drag the soft-limit
+    // recommendation with it (#1164 review).
+    const prev = process.env.STORAGE_BACKEND;
+    process.env.STORAGE_BACKEND = 's3';
+    try {
+      const res = await request(app)
+        .get('/api/admin/dashboard/stats')
+        .set('Authorization', `Bearer ${editorToken}`);
+
+      expect(res.status).toBe(200);
+      expect(res.body.storageUsed).toBeNull();
+      expect(res.body.storageMeasurement).toBe('catalog');
+      expect(Number(res.body.catalogedBytes)).toBe(1000);
+    } finally {
+      if (prev === undefined) delete process.env.STORAGE_BACKEND;
+      else process.env.STORAGE_BACKEND = prev;
+    }
+  });
+
   it('/analytics does not expose a foreign gallery name or slug', async () => {
     const res = await request(app)
       .get('/api/admin/dashboard/analytics?days=7')
--- a/backend/__tests__/services/localStorageUsage.test.js
+++ b/backend/__tests__/services/localStorageUsage.test.js
@@ -31,12 +31,15 @@ describe('localStorageUsage (#1164)', () => {
   beforeEach(async () => {
     root = await fs.promises.mkdtemp(path.join(os.tmpdir(), 'picpeak-usage-'));
     process.env.STORAGE_PATH = root;
+    delete process.env.EXTERNAL_MEDIA_ROOT;
+    jest.resetModules();
     resetLocalStorageUsageCache();
   });
 
   afterEach(async () => {
     await fs.promises.rm(root, { recursive: true, force: true }).catch(() => {});
     delete process.env.STORAGE_PATH;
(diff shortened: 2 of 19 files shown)


the-luap:
## Screenshots

Captured against the running stack. Real numbers from the local install, not mockups.

### Dashboard — Storage Used
**64.37 MB** actually on disk, with **15.75 MB catalogued** underneath it. 15.75 MB is precisely what this tile used to show — so the old figure was understating real usage by 4x here, and on the reporter's reference-mode install it *over*stated it by 4x in the other direction. Both are the same bug: the number was never measuring the disk.

issue-1164-dashboard-after

### Analytics — Storage Usage
Same split: "Used" is the real measurement and drives the soft-limit bar, "Catalogued media" is the original-bytes figure kept as its own line.

issue-1164-analytics-after

Issue #1096, opened 2026-08-20
[FEATURE] People manager: let me see the full photo (and more sample faces) when naming a person


BraynArts:
**Is your feature request related to a problem? Please describe.**

In the admin People manager ("Personen in dieser Galerie"), each cluster is represented by a single 64 px round face crop taken from one cover photo. That is often not enough to tell *who the person is*, and there is no way to find out from within the dialog.

Real example from a 3,200-photo family gallery that has just been scanned: of the first seven clusters, one shows the back of a head turned away from the camera, one shows a bare shoulder with no face in the circle at all, and one is essentially a blurred landscape. The photographer knows the top cluster is "Robin" only because she recognises the hair — she cannot confirm it, and for the other six she has no way to decide at all.

The information needed to decide is right there in the library — the person appears in 263 photos — but the dialog surfaces exactly one tightly-cropped 64 px window into one of them, and clicking it does nothing.

**Describe the solution you'd like**

When naming a person, let me actually look at them:

1. **Click the avatar to open the source photo full-size** (the existing lightbox would do), ideally with the detected bounding box outlined, so I see the face *in context* — who is standing next to them, what the situation was. That alone resolves most "is this Robin or her sister?" questions.
2. **Show several sample faces per cluster in the naming row**, not just the cover — e.g. 4–6 crops from different photos. A profile shot plus two frontal ones is usually instantly identifiable where the profile alone is not. The data is already available: `GET /admin/events/:eventId/people/:personId/faces` returns every face of the cluster, it is just only used by the Split dialog today.
3. **"Set as cover"** on any of those sample faces, so a cluster that got a bad cover can be fixed instead of staying unrecognisable in the guest-facing `PeopleStrip` too.

**Describe alternatives you've considered**

- *Using the Split dialog to browse the faces* — it does show all faces of a cluster, but at 88 px with the same crop pipeline, and its purpose (and its confirm button) is splitting a cluster apart. Using it as a viewer risks accidentally splitting a person you only wanted to look at.
- *Filtering the gallery by that person and looking there* — works, but means leaving the manager, and you lose your place in a list of dozens of unnamed clusters.
- *Just picking a better cover automatically* (e.g. largest / most frontal face) — a good heuristic and worth having, but it does not replace being able to look; the detector's idea of "best" and a human's do not always agree.

**Use Case**

- **Who:** the photographer / gallery owner, right after enabling face recognition on an existing event.
[...]

PR #1171, opened 2026-08-25
fix(previews): preserve alpha and animation in the preview tier

the-luap:
Follow-up to #1169, and stacked on it — this removes the format bypass that PR had to add.

## The bug

`generatePreviewImage` encoded JPEG unconditionally. JPEG has **no alpha channel and no second frame**, so:

- a transparent PNG came back flattened onto a solid background
- an animated GIF came back as its first frame

...for **every** consumer of the preview tier, not just the lightbox: the slideshow (#1015), the admin preview endpoints, and the face avatars that read it specifically because it is an aspect-preserved whole frame (#1096).

It was invisible by default only because the lightbox served originals — which is what #1166 is about. Fixing that surfaced this, and #1169 had to bypass GIF/APNG/PNG to avoid shipping a visible regression. That bypass was a guess made from MIME alone, and there is one case it could never get right: **a still and an animated WebP declare the same type**.

## Fix

Encode for the source. Sources with `hasAlpha`, or `pages > 1`, are written as **WebP** — which carries both, and is still far smaller than the original. Everything else stays JPEG, so the common path pays nothing.

```
transparent PNG  →  preview_logo.webp     (alpha preserved)
animated GIF     →  preview_wave.webp     (all frames preserved)
ordinary photo   →  preview_shot.jpg      (unchanged)
```

Reading is opened with `animated: isAnimated`, so later frames reach the resize instead of being discarded at decode. `limitInputPixels` still applies and sharp counts an animated input as `width × (height × pages)`, so a pathological GIF is rejected rather than decoded — and the caller falls back to the original, as it already does on any generation failure.

Two things had to move with it:

- **The output extension now matches what was written.** A PNG source previously produced `preview_foo.png` holding JPEG bytes — harmless while the route hard-coded `image/jpeg`, wrong once the encoding varies. Existing keys keep working: they are still JPEG and still served as such.
- **The route derives `Content-Type` from the key.** `X-Content-Type-Options: nosniff` is set on that response, so mislabelling shows a broken image rather than being silently corrected by the browser. The watermark branch re-encodes to JPEG and now says so explicitly.

The frontend `ORIGINAL_ONLY` list is deleted, which gives PNG galleries the bandwidth win back.

## Known limitation

`watermarkService.applyWatermark` composites and re-encodes to JPEG, so a gallery with watermarks on serves a flattened still even for an animated or transparent source. The header is correct for what is actually sent; preserving animation through the compositor is a separate problem and a much larger one.

## Verification

On the local rig, a transparent PNG registered into a real gallery:

```
GET /api/gallery/<slug>/preview/574
Content-Type: image/webp
8290 bytes   format: webp   hasAlpha: true   900x700
preview_path = previews/preview_alpha-test.webp
```

An ordinary photo on the same gallery still serves `image/jpeg` from a `.jpg` key.

## Tests

`__tests__/services/generatePreviewImage.formats.test.js` (6), against real Sharp output — the whole question is what is in the encoded bytes, so nothing is mocked. Covers alpha, animation, the unchanged JPEG path, the extension matching the encoding, the width tiers inheriting it, and an unreadable source still returning null rather than throwing.

The animated fixture is a hand-assembled 2-frame GIF89a rather than Sharp output: writing a multi-page image needs `pageHeight` threaded through the pipeline, and a fixture that silently came out single-page would make the test pass for the wrong reason.

Full backend suite: 6 pre-existing failures, identical to `origin/main`. Frontend: 34 files / 223 tests, `tsc --noEmit` clean.


the-luap:
## Screenshot

The change is in what the preview tier *encodes*, so a normal lightbox shot cannot show it — a flattened image and a transparent one look identical against a dark backdrop. This is the endpoint's own output for the same photo, over a checkerboard.

preview-alpha-comparison

Left is a real legacy artefact, not a mock-up: a preview written by the old generator, which kept the source basename (`.png`) while always encoding JPEG. It is the state migration 188 clears. Right is the same photo re-requested after clearing `preview_path`, regenerated by this PR.

The size is incidental but worth noting — the WebP is *smaller* than the JPEG it replaces (11,782 vs 15,575 bytes) while carrying an alpha channel.

Issue #1090, opened 2026-08-19
Face recognition skips every external/reference photo — the preview limitation it cites was fixed in #1078


BraynArts:
## Describe the bug

On a gallery whose photos are all external media, "People in this gallery" never finds anything: the scan reports `0 / 3230 photos` and finishes without an error. A rescan changes nothing.

The cause is the external/reference guard in `faceProcessor.js`:

```js
// External / reference photos live outside managed storage, and
// resolvePhotoStorageKey returns null for them by design (photoResolver.js).
// ensurePreviewImage therefore cannot build a preview, so scanning them is
// unsupported rather than broken — 'skipped', not 'failed', so an external
// gallery does not report every photo as an error.
if (photo.source_origin === 'external' || photo.source_origin === 'reference') {
  await db('photos').where({ id: photoId }).update({
    face_status: 'skipped', face_started_at: null, face_error: null,
  });
  return { status: 'skipped' };
}
```

Every photo in the event is set to `face_status = 'skipped'` before the sidecar is ever contacted:

```
 face_status | count
-------------+-------
 skipped     |  3230
```

**The stated reason no longer holds.** The comment says `ensurePreviewImage` cannot build a preview for externals — but that was fixed in #1078 (v3.107.1-beta.0). `ensurePreviewImage` in `imageProcessor.js` now has an explicit external branch that reads straight off the mount via `resolvePhotoFilePath` and generates the preview with a per-photo `outputBasename`:

```js
const isExternal = photo.source_origin === 'external' || photo.source_origin === 'reference';

if (isExternal) {
  // Mirrors ensureThumbnail's external branch: the source is a direct fs
  // read off the mount, so no withLocalCopy. ...
  localPath = resolvePhotoFilePath(event, photo);
  ...
  newPreviewPath = await generatePreviewImage(localPath, { regenerate: true, outputBasename });
}
```

So the skip guards against a limitation that was removed one release earlier. The guard is still present on `main`.

This makes the feature a no-op for external-media installs — where an entire library can be external, as it is here (6500+ photos, all `source_origin = external`).

## To Reproduce

1. Set up an event backed by external media (`source_origin = external`), photos on a mounted share.
2. Enable the `faces` feature flag and the per-event `face_recognition_enabled` toggle.
3. Run the `picpeak-ml` sidecar and confirm it is healthy.
4. Start a face scan on the gallery.
5. Progress stays at `0 / N`, no error is shown, and every row ends up `face_status = 'skipped'`.

[...]

PR #1174, opened 2026-08-25
fix(external-media): store external paths from the media root (#1163)

the-luap:
Stable twin of #1168 (issue #1163). **Stacked on #1173**, which supplies `deleteDuplicatePhotos`.

## The bug is on this branch

`externalMediaService.js:99` prefixes every photo with `event.external_path`, and `adminExternalMedia.js:179` overwrites that column on every import. So importing a second folder rebases every row already in the event. Nothing errors and the grid keeps rendering — thumbnails are written to local storage during the import while the base path is still correct — so only the things that need the **original** break: preview generation, the lightbox, downloads. The reporter had 7547 of 8004 rows pointing into the void, and spent a while chasing it as a CPU problem.

## What it carries

- `external_relpath` becomes relative to `EXTERNAL_MEDIA_ROOT`, so a row is self-describing and nothing an admin does to the event can move it. New `resolveExternalPhotoPath(photo)`; `resolveExternalPath` keeps its meaning for the import walk.
- **migration 177** folds each event's base into its rows. Where the current resolution is missing it walks up for an ancestor holding a file of the same name **and** the size the import recorded — existence alone would let a deleted file adopt an unrelated namesake and serve the **wrong original**, which is worse than a dead link. Rows it cannot place keep resolving where they resolve today, and the probe is skipped entirely when the mount is unreachable or empty.
- Probing is read-only and runs first; the rewrites and the marker commit **together**, so an interrupted fold cannot be folded a second time — which would put every original one directory deeper with no undo.
- Rewrites are **staged** through a per-row parking value, because a final path can equal another row's current one. Migration 177 re-throws without the driver's error code, which `run-migrations-safe.js` would otherwise read as "schema already exists" and record as applied after the rollback.
- The fold also runs after a `.picpeak` restore (`knex_migrations` is excluded from the archive), and a failure there is **reported** — surfaced through the API and the restore card — rather than presented as a clean restore.
- Drops the duplicate-leaf-segment guess in `photoResolver`, which papered over this same double-prefixing and corrupts a root-relative path whose first segment legitimately repeats.

**No face-scan requeue reordering.** Face recognition is main-only, so the hazard main had to fix — queueing rows against unconverted paths and burning them to `failed` — does not exist here, in `picpeakImportService` or in `restoreService`. Everything else is present.

Note also that this branch writes `external_path` *after* the import loop, where main writes it before (a #1090 change). That difference is inert once relpaths are root-relative, which is the point of this PR.

## Verification

Run on **this branch**. 23 new tests pass. Four existing suites carried base-relative fixtures encoding the old contract and were updated: `externalImportConcurrency`, `regenerateThumbnailsScript`, `ensurePreviewImage.external`, `photoResolver`.

Full backend suite leaves the same **5 pre-existing failures as `origin/stable`**.

### DIFF
--- a/backend/__tests__/integration/externalImportConcurrency.test.js
+++ b/backend/__tests__/integration/externalImportConcurrency.test.js
@@ -173,14 +173,14 @@ describe('concurrent external imports (#1162)', () => {
       path: 'x/a.jpg',
       type: 'individual',
       source_origin: 'external',
-      external_relpath: path.join('individual', 'a.jpg'),
+      external_relpath: path.join('nas', 'individual', 'a.jpg'),
     };
 
     const res = await runImport(eventId);
 
     expect(res.status).toBe(200);
     const counts = await relpathCounts(eventId);
-    expect(counts.get(path.join('individual', 'a.jpg'))).toBe(1);
+    expect(counts.get(path.join('nas', 'individual', 'a.jpg'))).toBe(1);
     // Two imported by us, one lost to the other writer and reported honestly.
     expect(res.body.imported).toBe(2);
     expect(res.body.skipped).toBe(1);
@@ -194,7 +194,7 @@ describe('concurrent external imports (#1162)', () => {
       path: 'x/a.jpg',
       type: 'individual',
       source_origin: 'external',
-      external_relpath: path.join('individual', 'a.jpg'),
+      external_relpath: path.join('nas', 'individual', 'a.jpg'),
     };
 
     await runImport(eventId);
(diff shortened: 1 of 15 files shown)


the-luap:
## Screenshot

The restore result card when `foldExternalRelpaths` throws — the state this PR adds, and the one an admin sees when every external original in the restored library is unreachable.

restore-external-paths-failed

**How this was produced:** the `/admin/backup/picpeak/import` response was stubbed in the browser to return the exact body the route sends on a failed conversion (`externalPathsConverted: false` plus the error text). No restore was run — triggering this for real means wiping a database and then breaking the fold mid-way, which is not something to do to a live instance for a screenshot. The component, its props and the rendered markup are the real ones.

One thing the shot makes obvious that the diff does not: the surrounding box is still **green** and still says "Backup restored." That is accurate — the rows and files genuinely are in place — but a green success box carrying a red "every referenced original is currently unreachable" is a mixed signal. Worth deciding whether the box should turn amber when `externalPathsConverted === false`; I left it alone rather than widening the diff.

PR #1175, opened 2026-08-25
fix(gallery): stop the lightbox loading originals to display a photo (#1166)

the-luap:
Stable twin of #1169 (issue #1166).

## The bug is on this branch, on four surfaces

`PhotoLightbox.tsx:846` and `:872` read `preview_url`, which the server only emits once an admin has flipped `lightbox_preview_enabled` — off by default. So a stock install falls through to `url`, the untouched original. A reporter measured 16.5 MB for a photo whose preview is 345 KB, and the lightbox renders its neighbours too, so opening one photo pulled **three** originals.

`slideshow_url` **is** emitted on this branch (#1015 is already here), so the fix works: prefer it before `url`. Same `/preview/:id` URL, watermark query included.

Three other surfaces bypass `PhotoLightbox` entirely and had the same bug:

- **premium** builds its own slides with `src: photo.url` (`:259`). Fixing that also required carrying the photo id on the slide — the download handler recovered the photo by matching `slide.src` against `photo.url`, which a derivative src breaks, and Download would have become a silent no-op.
- **the Story card** rendered the full original as its **grid tile** at `object-cover` in a small card (`:63`) — the one place where "hundreds of megabytes for a gallery" is literally true. It uses the preview tier now, not the thumbnail: `thumbnail_fit` is seeded to `cover`, so a thumbnail is a square centre-crop and these cards are not square, which would reframe every photo.
- **the Story hero** rendered one as a full-bleed background (`:44`) when `hero_url` exists for exactly that.

Story cards also **load on approach** now. Every card mounts at page load — `whileInView` gates the animation, not the render — and `AuthenticatedImage` fetches from an effect on mount, so all of them requested at once. framer-motion 12.33 is on this branch, so `useInView` is available.

**No responsive `?w=` tiers.** #1095 is main-only, so `lightboxImageUrl` here selects a URL and nothing more — no `previewUrlForViewport`, no width tables, and no viewport-tier tests.

It still lives in `components/gallery/imageTiers.ts`, the same path main uses, so that backporting #1095 later merges into this file rather than landing beside it. The file header says so.

## Formats

GIF, APNG and PNG keep the original, because `generatePreviewImage` encodes JPEG — no second frame, no alpha channel. **The backend fix that removes this list is the next PR in this stack.** Merging this one alone without it would put flattened GIFs and black-backgrounded PNGs in front of the widest install base.

## Settings copy

The toggle no longer decides whether the lightbox uses previews, and saving it never generated anything. Reworded to describe what it does — unlock the regenerate button — in en/de/fr/sl, the locales carrying those keys.

## Verification

Run on **this branch**. 8 new tests; frontend suite 21 files / 113 tests; `tsc --noEmit` clean.


the-luap:
## Screenshots

Captured against the running stack with `lightbox_preview_enabled` **off**, so `/photos` returns `preview_url: null` — the reporter's exact configuration.

### Lightbox, rendering from the preview tier
Network for this open: `GET /preview/82?w=1280` plus both neighbours. No `/photo/:id` request at all. Before the change the same open issued `GET /photo/82`, `/photo/81`, `/photo/21` — three originals.

issue-1166-lightbox-after

### Story layout — hero and cards
The hero is on `hero_url` (it rendered a full original before), the cards are on the preview tier, and only the ones near the viewport have loaded: **3 images fetched for 62 cards**, rising to 15 as you scroll. Zero `/photo/` requests on the whole page.

issue-1166-story-after

### Settings → Thumbnails
The toggle no longer governs whether the lightbox uses previews, so its copy no longer claims to. It now says what it still does — unlock the regenerate button below it — since saving it never generated anything on its own.

issue-1166-settings-after

PR #1176, opened 2026-08-25
fix(previews): preserve alpha and animation in the preview tier

the-luap:
Stable twin of #1171. **Stacked on the #1166 twin**, whose format bypass this removes.

## The bug is on this branch

`generatePreviewImage` (`imageProcessor.js:511`) encodes JPEG unconditionally (`:550`). JPEG has **no alpha channel and no second frame**, so a transparent PNG comes back flattened onto a solid background and an animated GIF comes back as its first frame — for every consumer of the preview tier, not just the lightbox.

It is invisible by default today only because the lightbox serves originals. The #1166 twin makes previews the default display path, which is what surfaces it.

## Fix

Encode for the source: `hasAlpha` or `pages > 1` → **WebP**, which carries both and is still far smaller than the original. Everything else stays JPEG.

- The output extension matches what was written. A PNG source previously produced `preview_foo.png` holding JPEG bytes — harmless while the route hard-coded `image/jpeg`, wrong once the encoding varies. Existing keys keep working: still JPEG, still served as such.
- The route derives `Content-Type` from the key. `X-Content-Type-Options: nosniff` is set on that response, so mislabelling shows a broken image rather than being quietly corrected. The watermark branch re-encodes to JPEG and now says so.

The frontend `ORIGINAL_ONLY` list is deleted, which gives PNG galleries the bandwidth win back and removes the guess that layer could never make correctly — a still and an animated WebP declare the same MIME type.

**No width-tier case.** The responsive `?w=` renditions (#1095) are main-only, so this branch has a single canonical preview per photo.

## Known limitation

`applyWatermark` composites and re-encodes to JPEG, so a watermarked gallery serves a flattened still even for an animated or transparent source. The header is correct for what is actually sent; preserving animation through the compositor is a separate and much larger problem.

## Verification

Run on **this branch**. 5 new backend tests against real Sharp output — nothing mocked, since the whole question is what is in the encoded bytes. The animated fixture is a hand-assembled 2-frame GIF89a: Sharp's multi-page writer needs `pageHeight` threaded through the pipeline, and a fixture that silently came out single-page would make the test pass for the wrong reason.

Full backend suite leaves the same **5 pre-existing failures as `origin/stable`**. Frontend 21 files / 114 tests, `tsc` clean.


the-luap:
## Screenshot

The change is in what the preview tier *encodes*, so a normal lightbox shot cannot show it — a flattened image and a transparent one look identical against a dark backdrop. This is the endpoint's own output for the same photo, over a checkerboard.

preview-alpha-comparison

Left is a real legacy artefact, not a mock-up: a preview written by the old generator, which kept the source basename (`.png`) while always encoding JPEG. It is the state migration 188 clears. Right is the same photo re-requested after clearing `preview_path`, regenerated by this PR.

The size is incidental but worth noting — the WebP is *smaller* than the JPEG it replaces (11,782 vs 15,575 bytes) while carrying an alpha channel.

PR #1177, opened 2026-08-25
fix(admin): make "Storage used" report storage used (#1164)

the-luap:
Stable twin of #1170 (issue #1164).

## The bug is on this branch verbatim

`adminDashboard.js:92` and `adminSettings.js:1177` both sum `photos.size_bytes` — the catalogued size of the **originals**, which in reference mode live on external storage and have no relationship to the disk PicPeak runs on. The reporter's tile read ~80 GB against 21 GB of real usage.

Worse than the label: the same number drives the **soft-limit warning bar** and, via `/storage/info`, the **recommended soft limit**. A reference-mode install gets a disk-capacity recommendation computed from bytes that are not on the disk.

## What it carries

New `localStorageUsage` service walks the storage root and reports the total plus a breakdown. Walking rather than summing DB columns is the point: `thumbnail_path` / `preview_path` / `hero_path` record a *key* and never a byte count, and orphans from a deleted event or an interrupted import are real bytes.

- **The external media root is excluded** when it sits inside the storage root. Its compose default is `<storage>/external-media`, where the NAS is bind-mounted — a plain **directory, not a symlink**, so a symlink guard alone does not cover it. Walking it would put every referenced original back into a figure whose entire purpose is to leave them out.
- `.download-cache` gets its own line: it lives *inside* the event directory, so the naive rule files a multi-GB zip as photography — and it is the one bucket that is pure disposable cache.
- Symlinks are not followed; concurrent cold-cache callers share one walk (the dashboard, `/storage/info` and the sidebar are routinely requested together); cached 5 minutes.
- **S3 installs keep the catalogued figure**, and the walk is skipped *before* it runs — the objects are in the bucket and `STORAGE_PATH` holds only incidental local files, so walking it would report near-zero and drag the recommendation down. The response says which measurement it is so the UI can label it.
- An absent measurement reads as **"unavailable"**, and a partial one is marked `+` across the dashboard tile, analytics card, sidebar and status tab — a floor silently compared against a soft limit reads as "safely under".
- Not event-scoped, unlike everything else on `/stats`: disk is not divisible by which admin owns which event. Pinned in `dashboardScope.test.js`.

None functionally — the service is new, so it ports cleanly, and every surface it touches exists here. The only difference is stable's `photos` table has fewer feedback counters, which this PR does not touch.

## Verification

Run on **this branch**. 11 new service tests; `dashboardScope.test.js` updated for the changed contract (its `storageUsed === 1000` assertion encoded the old meaning) plus two new cases for the unscoped and S3 behaviours.

Full backend suite leaves the same **5 pre-existing failures as `origin/stable`**. Frontend 20 files / 104 tests, `tsc` clean.

## Not addressed

`.download-cache/all.zip` still has no TTL or size cap — the reporter's second observation, and correct. This PR at least makes it visible in the breakdown, which is what makes the case for capping it.

### DIFF
--- a/backend/__tests__/routes/dashboardScope.test.js
+++ b/backend/__tests__/routes/dashboardScope.test.js
@@ -127,7 +127,44 @@ describe('dashboard scoping (GHSA-c2jj / gqx7 / jhcf)', () => {
     expect(res.status).toBe(200);
     expect(Number(res.body.totalEvents)).toBe(1);
     expect(Number(res.body.totalPhotos)).toBe(1);
-    expect(Number(res.body.storageUsed)).toBe(1000);
+    // The catalogued original bytes — this is what carries the per-event
+    // scoping, and what `storageUsed` reported before #1164.
+    expect(Number(res.body.catalogedBytes)).toBe(1000);
+  });
+
+  it('/stats reports disk usage unscoped, because disk is not per-event', async () => {
+    // storageUsed is a measurement of the storage root (#1164), so it is the
+    // same number for every admin by design. Pinned so a future reviewer
+    // reading "everything on this endpoint is scoped" does not turn it into a
+    // sum of this editor's photos again — which is the bug that was fixed.
+    const res = await request(app)
+      .get('/api/admin/dashboard/stats')
+      .set('Authorization', `Bearer ${editorToken}`);
+
+    expect(res.status).toBe(200);
+    expect(res.body.storageUsed).not.toBe(1000);
+    expect(res.body).toHaveProperty('storageBreakdown');
+  });
+
+  it('/stats reports the catalogued figure on an S3 backend, not a near-zero disk walk', async () => {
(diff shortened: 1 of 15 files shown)


the-luap:
## Screenshots

Captured against the running stack. Real numbers from the local install, not mockups.

### Dashboard — Storage Used
**64.37 MB** actually on disk, with **15.75 MB catalogued** underneath it. 15.75 MB is precisely what this tile used to show — so the old figure was understating real usage by 4x here, and on the reporter's reference-mode install it *over*stated it by 4x in the other direction. Both are the same bug: the number was never measuring the disk.

issue-1164-dashboard-after

### Analytics — Storage Usage
Same split: "Used" is the real measurement and drives the soft-limit bar, "Catalogued media" is the original-bytes figure kept as its own line.

issue-1164-analytics-after

PR #1179, opened 2026-08-25
fix(external-media): record capture dates on import, and backfill existing libraries (#1172)

### LINKED ISSUE #1172: [BUG] External media import never sets captured_at — "Date Taken" sort silently falls back to import order
## Describe the bug

The external media import (`source_origin='external'`) never reads EXIF, so `photos.captured_at` stays `NULL` for every photo it creates. The gallery sorts "Date Taken" with `COALESCE(photos.captured_at, photos.uploaded_at)`, so on a reference-mode library that sort silently degrades into "sort by import batch" — no error, no warning, and no hint in the UI that the sort key is missing.

In my instance that means the 5555 photos of a 12-day trip were ordered by two import timestamps instead of by capture time. The folders happened to arrive in two batches, so the first two days of the trip ended up at positions 4204–5296 of 5555 while the last days sat at the front.

## Where it comes from

1. `src/routes/adminExternalMedia.js` (~line 176) inserts `event_id, filename, path, thumbnail_path, type, size_bytes, width, height, source_origin, external_relpath`. `captured_at` is not among them. The file is already opened a few lines above for `sharp(f.full).metadata()` to get the dimensions, so this is exactly the point where the capture date could be read too.
2. `processPhoto()` in `src/services/photoProcessor.js:485` is the only caller of `extractCaptureDate()`, and it never runs for external rows.
3. `src/routes/gallery.js:680`: `orderByRaw('COALESCE(photos.captured_at, photos.uploaded_at) ' + sortOrder)`.

Managed uploads are unaffected. In the same instance, a manually uploaded event has `captured_at` on 79/79 photos, while three external events had it on 0 of 8864.

## Secondary issue: no tiebreaker

That `COALESCE(...)` is the only `ORDER BY` term. A bulk import writes hundreds of rows inside the same second, and with `captured_at` NULL an entire batch collapses onto nearly identical values, so the returned order is whatever Postgres happens to produce. Adding `photos.id` as a secondary key would at least make the result stable and roughly insertion-ordered.

## To Reproduce

1. Create an event in external/reference mode pointing at a folder of photos that carry `DateTimeOriginal`.
2. Run the external media import.
3. `SELECT count(*) FROM photos WHERE event_id = <id> AND captured_at IS NULL;` → equals the total number of photos.
4. Open the gallery and pick "Date Taken (Oldest First)" → photos come back in import order, not capture order.

## Expected behavior

External imports populate `captured_at` from EXIF just like managed uploads do, and existing libraries can be backfilled.

## Suggested fix

- Read the capture date in the external import loop, next to the existing `sharp().metadata()` call, and include it in the insert. `extractCaptureDate()` already exists and is exported from `imageProcessor.js`.
- Offer a backfill for libraries that were imported before the fix, along the lines of the existing `POST /api/admin/thumbnails/regenerate-previews`. Note that `POST /api/admin/thumbnails/regenerate` is not a usable model here: it resolves originals under `storage/events/active/<photo.path>`, which never exists for external rows (see #1129).
- Add `photos.id` as a secondary sort key in `gallery.js:680`.

## Environment

- PicPeak 3.114.0-beta.0 (backend + frontend), PostgreSQL, Docker Compose (production compose file)
- Reference/external mode, originals on an NFS-mounted NAS
- ~8.8k photos across three external events, plus one managed event as a control

## Additional context

As a workaround I read `DateTimeOriginal` out of the files myself, wrote it into `photos.captured_at` and set `events.default_photo_sort = 'capture_date_asc'`. The gallery order then matched the trip exactly, including a spot where the camera's file counter rolled over from `_DSC9999` to `_DSC0001` mid-morning — that one is only resolvable from EXIF, never from the filename.

One detail worth deciding on while fixing this: `extractCaptureDate()` picks only `DateTimeOriginal` / `CreateDate` / `DateTimeDigitized` / `ModifyDate` and not `OffsetTimeOriginal`, so the EXIF wall-clock time ends up stored as if it were UTC. I verified this on a managed upload: EXIF `09:45:03` with offset `-03:00` is stored as `09:45:03+00`. That is self-consistent within a library and fine for ordering, but it does mean `captured_at` is not a true instant, which matters if it is ever displayed in the viewer's local timezone.

---

_The analysis, code references and text of this report were produced by Claude (Anthropic), working on my instance at my request. I reviewed it before filing._

the-luap:
Closes #1172. **Stacked on #1168** — it changes the same INSERT statement, so it cannot land independently without conflicting.

## Root cause

The external import never read EXIF, so `photos.captured_at` stayed **NULL for every row it created**. The gallery sorts "Date Taken" with `COALESCE(captured_at, uploaded_at)`, and on a bulk import `uploaded_at` is the import timestamp — so the sort silently degraded into *"order by import batch"*, with no error and nothing in the UI to say the sort key was missing.

The reporter's 12-day trip came back with its first two days at **positions 4204–5296 of 5555**, because those folders happened to be imported second.

All three of his pointers check out: `captured_at` appears nowhere in `adminExternalMedia.js`; `extractCaptureDate()` exists at `imageProcessor.js:1059` and is exported, but its only callers are `photoProcessor` and `photoReplacementService`, neither of which runs for external rows.

## Fix

**The import reads the date** next to the `sharp().metadata()` call that already opens the file, so this costs one more read of the same source rather than a second pass over the mount. Best-effort like the dimensions beside it: a source with no EXIF imports with `captured_at` NULL, exactly as before.

**`POST /api/admin/photos/repair-capture-dates`** backfills existing libraries, modelled on the dimension repair sitting next to it in the same file — background pass, in-flight guard, status endpoint, and `resolvePhotoFilePath`, which is what reaches an external row at all. The reporter is right that `/regenerate` is not a usable model: it resolves under `storage/events/active/<photo.path>`, which never exists for external rows (#1129).

Not a migration, deliberately: the originals sit on a mount that may be down at upgrade time, reading 8000+ of them would block the boot, and a run that found nothing has to be repeatable once the mount is back.

- **"no EXIF date" is counted separately from "could not read the file."** An operator needs to tell *these files carry no date* from *the mount is broken* before deciding whether re-running will help.
- The update is guarded `whereNull`, so an import or a replacement finishing mid-run is not overwritten by a slower pass.

**Every sort branch gets `photos.id` as a tiebreaker** — not just `capture_date`. The reporter flagged this as secondary; it is broader than he noticed. A bulk import writes hundreds of rows inside one second, so `uploaded_at` collapses *and* the COALESCE fallback collapses onto it, and `filename` ties too. The visible symptom is a grid that reshuffles between page loads. `id` is insertion order, which also makes the fallback ordering meaningful rather than arbitrary.

A card in Settings → Status drives the backfill, mirroring the dimensions one.

## Not addressed — and worth its own issue

`extractCaptureDate` reads no `OffsetTimeOriginal`, so a naive EXIF timestamp is resolved by exifr **against the host timezone**. I verified this while building the fixtures: EXIF `09:45:03` came back as `07:45:03Z` on a CEST machine. So `captured_at` is not a true instant, and the same file imported on two machines yields two values.

That predates this change and applies to managed uploads equally — the reporter's own footnote describes a related symptom. Fixing it changes the meaning of every existing `captured_at`, so it does not belong here. The tests deliberately assert **ordering**, never an absolute instant, so they do not encode the bug or break on a UTC runner.

## Tests

`externalImportCaptureDate.test.js` (3) — real JPEGs carrying real EXIF through the real route, including the reported two-batch ordering case. Note the fixtures write `DateTimeOriginal` into **IFD2**, not IFD0: exifr does not see it anywhere else.

`captureDateBackfill.test.js` (5) — the external row the thumbnail regenerator cannot reach, "no EXIF" vs "unreadable" counted apart, nothing-to-do, and the `whereNull` guard against a concurrent write.

Full backend suite: 6 pre-existing failures, identical to `origin/main`. Frontend 33 files / 212 tests, `tsc` clean.

### DIFF
--- a/backend/__tests__/integration/captureDateBackfill.test.js
+++ b/backend/__tests__/integration/captureDateBackfill.test.js
@@ -1,0 +1,150 @@
+/**
+ * Backfilling captured_at on a library imported before #1172.
+ *
+ * The point of the endpoint, rather than a migration: it resolves originals
+ * through resolvePhotoFilePath, which is the only path that reaches an
+ * external row. The thumbnail regenerator resolves under
+ * storage/events/active/<photo.path>, which never exists for those (#1129) —
+ * so it cannot be the model.
+ */
+
+const fs = require('fs');
+const path = require('path');
+const os = require('os');
+const express = require('express');
+const request = require('supertest');
+const sharp = require('sharp');
+
+describe('capture date backfill (#1172)', () => {
+  let tmpDir; let db; let app; let mediaRoot;
+
+  const writeJpegWithExif = async (abs, iso) => {
+    await fs.promises.mkdir(path.dirname(abs), { recursive: true });
+    const d = new Date(iso);
+    const pad = (n) => String(n).padStart(2, '0');
+    const exifDate = `${d.getUTCFullYear()}:${pad(d.getUTCMonth() + 1)}:${pad(d.getUTCDate())} `
+      + `${pad(d.getUTCHours())}:${pad(d.getUTCMinutes())}:${pad(d.getUTCSeconds())}`;
+    await sharp({ create: { width: 60, height: 40, channels: 3, background: { r: 9, g: 9, b: 9 } } })
+      .withExif({ IFD2: { DateTimeOriginal: exifDate } }).jpeg().toFile(abs);
+  };
+
+  const settle = async () => { for (let i = 0; i < 60; i++) { await new Promise((r) => setTimeout(r, 50)); const s = await status(); if (!s.body.isRunning) return s; } throw new Error('backfill did not settle'); };
+  const status = () => request(app).get('/api/admin/photos/repair-capture-dates/status');
+
+  beforeAll(async () => {
+    tmpDir = await fs.promises.mkdtemp(path.join(os.tmpdir(), 'picpeak-capfill-'));
+    mediaRoot = path.join(tmpDir, 'media');
+    await fs.promises.mkdir(mediaRoot, { recursive: true });
+
+    process.env.NODE_ENV = 'test';
+    process.env.TEST_DATABASE_PATH = path.join(tmpDir, 'data', 'db.sqlite');
+    await fs.promises.mkdir(path.dirname(process.env.TEST_DATABASE_PATH), { recursive: true });
+    process.env.STORAGE_PATH = path.join(tmpDir, 'storage');
+    process.env.EXTERNAL_MEDIA_ROOT = mediaRoot;
+    process.env.JWT_SECRET = process.env.JWT_SECRET || 'capfill-secret';
+
+    jest.resetModules();
+    jest.doMock('../../src/middleware/auth', () => ({
+      adminAuth: (req, _res, next) => { req.admin = { id: 1, username: 'tester', roleName: 'admin' }; next(); },
+    }));
+    jest.doMock('../../src/middleware/permissions', () => ({
+      requirePermission: () => (_req, _res, next) => next(),
+    }));
+    jest.doMock('../../src/utils/logger', () => ({
+      debug: jest.fn(), info: jest.fn(), warn: jest.fn(), error: jest.fn(),
+    }));
+
+    ({ db } = await require('./helpers/crmDb').bootCrmDb());
+
+    app = express();
+    app.use(express.json());
+    app.use('/api/admin/photos', require('../../src/routes/adminPhotoDimensions'));
+  }, 180000);
+
+  afterAll(async () => {
+    if (db) await db.destroy?.();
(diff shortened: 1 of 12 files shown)


the-luap:
## Screenshot

Settings → Status, the new **Capture Dates** card. The amber box is the backlog this PR exists to clear: photos imported from external media before capture dates were read, whose galleries sort by import order instead of when the photos were taken.

issue-1172-capture-dates-card


The result line under the boxes is worded as "no date found" / "unreachable" rather than "without EXIF" / "unreadable": `extractCaptureDate` returns null both for a file that genuinely carries no date and for one it cannot parse, so those two land in the same bucket. The last number is the one that means the storage could not be reached.

Issue #1178, opened 2026-08-25
Guest-set colour labels don't appear in gallery grid for other guests (only in lightbox)


boergu:
First of all, huge thanks for adding colour labels for client proofing — I'm the one who requested this feature, and seeing it actually shipped just days later is amazing. Exactly what I needed for client selection workflows. Really appreciate the work you're putting into this project!

Now, a bug I ran into with it:

Version: v3.115.0-beta.0
"Show Feedback to Guests" is enabled (under Guest Feedback Settings).

When one guest sets a colour label (e.g. red) on a photo:
- The admin sees it correctly in the admin UI.
- Other guests see it correctly when opening the photo in fullscreen/lightbox view.
- Other guests do NOT see it in the gallery grid/thumbnail view.

My own colour labels show correctly in the grid for me, and grid filtering by my own colours works fine. It's specifically other guests' labels that are missing from the grid view, even though they're visible in lightbox.

---

Additional feature idea, related to the above:

My use case is built around simple, single-verdict feedback, not per-guest group tallying. Right now it looks like multiple guests can each leave their own separate colour rating on the same photo, so ratings pile up per guest.

I checked the existing Identity Mode settings ("Simple feedback" vs. "Per-guest selections") — neither covers this. Both are still identity-scoped: "Simple feedback" shares state only across guests on the same physical device (anonymous device fingerprint), and "Per-guest selections" tracks named individuals. In both cases, feedback is still stored per-identity, just with a different definition of "identity".

What I'm asking for is a third model with no identity dimension at all for the shared tag: one single, identity-less colour tag per photo, stored server-side, that any guest on any device can overwrite. Not "everyone shares a device's state" — "everyone, on any device, shares the photo's state." So if guest A marks a photo green and guest B later marks the same photo orange, the shared tag simply becomes orange — B's action overrides A's, rather than both marks being kept separately.

Would be great to have this as an alternative feedback mode (maybe a third option under Identity Mode), since some use cases — like mine — want one collaboratively-agreed tag per photo rather than per-person tallies.

the-luap:
Implemented and merged in #1180.

The grid now shows the colours other guests gave a photo, as small dots beside your own mark, so a client's picks are visible while scanning rather than only after opening the lightbox. Gated on the gallery's "show feedback to guests" setting, and your own identity is excluded from the "others" list so your mark is never double-counted.

Two things worth recording from the review:

[...]

PR #1180, opened 2026-08-25
fix(gallery): show other guests' colour labels in the grid (#1178)

the-luap:
Closes the bug half of #1178. The feature request in the same issue is **not** implemented — see the bottom.

## Root cause

A colour set by one guest was visible to others in the lightbox and invisible on the tile. The two surfaces read different endpoints:

| | Source | Shows |
|---|---|---|
| Lightbox | `/photos/:id/feedback` → `color_labels` (`getPhotoColorLabelCounts`, everyone) | all guests', as per-colour tallies |
| Grid | `/photos` → `my_color_label` | **the viewer's own, only** |

There was no field in the grid payload carrying other guests' colours at all, so `PhotoCard` could not have rendered them. The feature was simply never extended to the grid.

## Fix

`/photos` now also returns **`other_color_labels`** — the DISTINCT colours other viewers put on each photo — gated on `show_feedback_to_guests` like every other aggregate. `my_color_label` stays ungated, because a viewer's own selection is not shared data; that existing distinction is unchanged.

Rendered as small dots beside the badge:

```
┌──────────────┐
│  [photo]     │
│         ●RED │  ← your label: unchanged badge + inset ring
│  ●●          │  ← others: green, blue
└──────────────┘
```

Design choices worth a look:

- **Distinct colours, not counts.** A tile has room for a couple of marks; *who* marked it and *how many* is a question the lightbox already answers properly.
- **Capped at three dots**, so a heavily-marked photo cannot flood the tile.
- **The viewer's own colour is excluded from the dots**, so the badge and the dots never say the same thing twice.
- **Opposite corners**, so the two do not read as one group. The inset ring stays the viewer's own signal — that is what the badge was built for ("the client watches their selection progress across the grid").
- An unknown colour is ignored rather than rendered as a blank dot, so a colour added server-side later cannot produce an empty circle on an older build.

`PhotoCard` was not the only render site. `GalleryPremiumLayout` and `StoryPhotoCard` keep their own copies of the badge, and both needed the same prop — otherwise this would have fixed the default grid and left the two full-bleed layouts showing nothing, which is the same shape of gap as the original bug. Found by driving a real gallery and then grepping for the remaining call sites.

## Verified on a running instance

Two guest identities labelled one photo red and green; the viewer is a third with no label of their own.

```
sharing ON  → other_color_labels: ['green', 'red']    my_color_label: null
sharing OFF → other_color_labels: []
```

Screenshot of the tile follows in a comment.

## Not implemented: the identity-less shared tag

The issue also asks for a third Identity Mode — one colour tag per photo, no identity dimension, any guest on any device overwrites it.

The reporter's reading of the existing modes is correct, and I confirmed it: `simple` scopes by device fingerprint, `guest` scopes by `guest_id` (migration 078). Both store one row per identity, so neither produces a single shared verdict.

That is a genuine gap, but it is a third identity model touching the feedback schema, the per-guest caps (#655), moderation, and every admin aggregate that currently counts per person. It needs its own design and its own PR rather than riding along on a display fix.

## Tests

`ColorLabelBadge.test.tsx` (6): the reported case (others' label with none of your own), own-plus-others without repeating your colour, the cap, the unknown-colour guard, and nothing rendered when there is nothing to show.

Backend suite: 6 pre-existing failures, identical to `origin/main`. Frontend 34 files / 218 tests, `tsc` clean.

### DIFF
--- a/backend/src/routes/gallery.js
+++ b/backend/src/routes/gallery.js
@@ -918,6 +918,46 @@ router.get('/:slug/photos', verifyGalleryAccess, resolveGuest, async (req, res) 
       });
     }
 
+    // OTHER viewers' colour labels, per photo (#1178).
+    //
+    // The lightbox has always shown these — /photos/:id/feedback returns
+    // per-colour tallies across everyone — but the grid had no field carrying
+    // them, so a label set by one guest was visible in fullscreen and invisible
+    // on the tile. With sharing on, that is just a hole.
+    //
+    // DISTINCT colours, not counts: a tile has room for a couple of dots, and
+    // "who else marked this, and how" is a lightbox question. The viewer's own
+    // colour is excluded here so the badge and the dots never say the same
+    // thing twice — the frontend renders `my_color_label` as the badge and
+    // these beside it.
+    //
+    // Gated on showFeedbackToGuests, like every other aggregate: this is other
+    // people's feedback, unlike my_color_label above.
+    const otherColorLabelsByPhoto = {};
+    if (photos.length > 0 && showFeedbackToGuests) {
+      const othersQuery = db('photo_feedback')
+        .where({ event_id: req.event.id, feedback_type: 'color_label', is_hidden: false })
+        .whereIn('photo_id', photos.map(p => p.id))
+        .whereNotNull('color_label');
+      if (req.guest?.id) {
+        othersQuery.where(function () {
+          this.whereNot('guest_id', req.guest.id).orWhereNull('guest_id');
+        });
+      } else {
+        const mine = generateGuestIdentifier(req);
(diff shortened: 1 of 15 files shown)


the-luap:
## Screenshot

Two guest identities labelled this photo red and green; the viewer is a third with no label of their own. Before this change the tile showed nothing at all — the colours were only visible after opening the photo in the lightbox.

issue-1178-grid-other-labels


The dots sit next to the viewer's own colour dot, in the corner the colour-label feature already owns. Review flagged an earlier bottom-left placement: that corner carries Timeline's timestamp chip on every tile and the media-type badges elsewhere, both of which paint over it.

PR #1183, opened 2026-08-25
fix(external-media): record captured_at on import and add a backfill (stable)

the-luap:
Stable twin of #1179. Stacked on #1174, mirroring the main-side stack (#1179 sits on #1168's branch).

## Why stable needs this

The bug in #1172 reproduces here exactly as filed:

| | stable |
|---|---|
| `photos.captured_at` column | present (migration `071_add_captured_at.js`) |
| `adminExternalMedia.js` import route | present |
| gallery sorts on `captured_at` | yes — `COALESCE(photos.captured_at, photos.uploaded_at)` |
| import writes `captured_at` | **no — 0 references** |

So a stable install offering "Date Taken" silently sorts external galleries by import order, which is what the reporter hit.

## What is ported

- **`adminExternalMedia.js`** — reads the capture date at import time, off the file already opened a few lines above for the dimensions, so it costs one extra read of the same source rather than a second pass over the mount. Best-effort: a source with no EXIF, or one exifr cannot parse, imports with `captured_at` NULL and falls back to `uploaded_at` as before.
- **Backfill endpoint** (`POST /repair-capture-dates` + status) — without it, existing installs would get correct dates only on *new* imports and no way to fix the galleries already there, which is the actual complaint. Carries every fix that came out of review on #1179: managed originals resolve through `resolvePhotoStorageKey` + `withLocalCopy` so S3 installs work, archived events are excluded (archiving deletes their originals, so they would fail every run and keep the button lit), and the run flag is claimed before the candidate query so two concurrent POSTs cannot both start a pass.
- **`gallery.js`** — `photos.id` as a tiebreaker on all three sorts. A bulk import writes hundreds of rows inside the same second, so `uploaded_at` ties are the normal case, and with `captured_at` NULL the COALESCE collapses onto it too. The visible symptom is a grid that reshuffles between page loads.
- **Settings → Status card** + en/de/fr/sl strings.

No migration: `captured_at` already exists here.

## One deliberate difference from main

The backfill is gated on `settings.edit` / `settings.view`; main uses `system.manage` / `system.view`, which do not exist on this branch.

That is the same gate under its older name. `system.manage` is one of the permissions `settings.edit` was split into, and main's migration 175 projects every `settings.edit` holder forward onto it (`175:161-170`). On this branch `settings.edit` is the restrictive one — `admin` carries only `settings.view` (`056_add_role_permissions_table.js:63`), and `editor` has neither (`056:69-78`). So both branches let exactly the same people through.

Not gated on `photos.edit` — `editor` holds that, and this walks every event in the install.

## Verification

- `npx jest __tests__/integration/captureDateBackfill.test.js __tests__/integration/externalImportCaptureDate.test.js` — **9 passed**, both suites, run on this branch.
- `npx tsc --noEmit` — clean.
- `npx eslint` on all three changed backend routes — clean, except one pre-existing `getExternalMediaRoot is assigned a value but never used` in `adminExternalMedia.js:7` that is present on the unmodified base branch too (verified by stashing). Not touched here.
- Directly inspected on this branch rather than assumed from main: the `captured_at` column, `extractCaptureDate`'s export, `withLocalCopy` and `resolvePhotoStorageKey`, `events.is_archived`, `photos.media_type`, and the role→permission seeding.

Diff is at full parity with #1179 — same ten files, same shape.

### DIFF
--- a/backend/__tests__/integration/captureDateBackfill.test.js
+++ b/backend/__tests__/integration/captureDateBackfill.test.js
@@ -133,6 +133,43 @@ describe('capture date backfill (#1172)', () => {
     expect((await status()).body.withoutCaptureDate).toBe(0);
   });
 
+  it('skips a watcher-imported video, which carries media_type "image"', async () => {
+    // fileWatcher.processNewPhoto sets type='video' and a video/* mime but
+    // never media_type (fileWatcher.js:128-130), so the row keeps the 'image'
+    // default from migration 048. Filtering on media_type alone queued it every
+    // run: extractCaptureDate returns null for a video, captured_at stays null,
+    // and the backlog never cleared.
+    const { eventId } = await seed({ relpath: 'clip.jpg', exifIso: '2026-06-01T09:45:03Z', writeFile: false });
+    await db('photos').del();
+    await db('photos').insert({
+      event_id: eventId, filename: 'clip.mp4', path: 'capfill/clip.mp4',
+      type: 'video', media_type: 'image', mime_type: 'video/mp4',
+      source_origin: 'external', external_relpath: 'trip/clip.mp4',
+      uploaded_at: new Date().toISOString(), captured_at: null,
+    });
+
+    const res = await request(app).post('/api/admin/photos/repair-capture-dates');
+    expect(res.body.count).toBe(0);
+
+    const s = await status();
+    // And it is not counted as a permanent backlog either.
+    expect(s.body.total).toBe(0);
+    expect(s.body.withoutCaptureDate).toBe(0);
+  });
+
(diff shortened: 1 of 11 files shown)


the-luap:
## Screenshot

Settings → Status, the new **Capture Dates** card — identical to the main twin (#1179), whose `StatusTab.tsx` change ported here unchanged.

issue-1172-capture-dates-card

Captured against the main twin's branch. The frontend diff between the two is byte-identical, so this is what the card looks like on stable as well; the only branch difference in this PR is the backend permission gate (`settings.edit` instead of `system.manage`), which is not visible here.

PR #1194, opened 2026-08-26
fix(images): respect EXIF orientation in thumbnails, heroes and previews

### LINKED ISSUE #1185: Thumbnails/hero/previews ignore EXIF orientation - .rotate() missing (download path already handles it correctly)
**Describe the bug**
`generateThumbnail`, `generateHeroImage` and `generatePreviewImage` in `backend/src/services/imageProcessor.js` never call sharp's `.rotate()` before resizing, so any photo whose EXIF `Orientation` tag is not 1 (i.e. rotated in-camera — very common for portrait shots on bodies that don't physically rotate the sensor) gets thumbnailed/previewed/hero'd from the raw, unrotated pixels. The same pipeline then calls `.withMetadata(false)`, which strips the orientation tag from the output too — so the browser has no hint left to auto-correct it either. Net result: the image renders sideways or upside-down, with nothing downstream able to fix it.

The download path in the same file already does this correctly (`resizeToBox`):
```js
const image = animated
  ? sharp(inputBuffer, { ...opts, animated: true })
  : probe.rotate();
```
None of the three gallery-facing generators have an equivalent call — `generateThumbnail` (~line 257), `generateHeroImage` (~line 511), `generatePreviewImage` (~line 651) all go straight from `sharp(imagePath, {...})` to `.resize()`.

This is not specific to external/reference imports — `photoProcessor.js` (managed admin uploads) calls the exact same shared `generateThumbnail`. It only reads as an "external media" bug because our reference libraries happen to contain a lot of rotated originals; a library without any camera-rotated shots wouldn't show it.

**To Reproduce**
1. Import or upload a photo with EXIF `Orientation` != 1 (any portrait shot taken by physically rotating a camera body that writes orientation tags instead of rotating the sensor data — e.g. Sony ILCE-7M3, Sony ZV-E1).
2. Open the gallery (or Admin).
3. Thumbnail, preview and hero image all show the photo sideways/upside-down. The same photo fetched through the download route is correctly oriented.

**Expected behavior**
Thumbnails/previews/hero images should respect EXIF orientation the same way the download route already does.

**Evidence**
Sampled `sharp().metadata().orientation` directly against files backing two live events (random sample, not cherry-picked):

| Event | Camera | Sampled | orientation != 1 |
|---|---|---|---|
| "Norwegen 2025" | Sony ILCE-7M3 | 24 | 13 (54%) |
| "Rio 2026" | Sony ZV-E1 | 8 | 3 (38%) |

Every sampled row also has `photos.width`/`photos.height` in the DB equal to the raw (pre-rotation) sensor dimensions — so any masonry/justified layout that sizes grid tiles from those DB columns compounds the same bug: the tile aspect ratio is wrong for exactly the same photos, on top of the image itself being unrotated.

**Suggested fix**
Mirror the existing `resizeToBox` handling in the same file: call `.rotate()` on the sharp pipeline before `.resize()` in `generateThumbnail`, `generateHeroImage` and `generatePreviewImage` for still images (skip for animated sources, same reasoning `resizeToBox` already documents — `.rotate()` flattens multi-frame images).

**Environment**
- OS: Debian (LXC), Docker Compose host
- PicPeak Version: 3.115.0-beta.0
- Deployment Method: Docker Compose (production compose, GHCR images)
- Database: PostgreSQL 15

the-luap:
Closes #1185.

## The bug

`generateThumbnail`, `generateHeroImage` and `generatePreviewImage` went straight from `sharp(imagePath, {...})` to `.resize()` with no `.rotate()`. sharp decodes pixels **as stored**, not as displayed, so any photo whose EXIF `Orientation` is not 1 was resized from the raw frame and came out sideways. All three then call `.withMetadata(false)`, which strips the tag from the output — so the browser had no hint left to correct it either.

The reporter's diagnosis was exact, including why the same photo looks fine on download: `resizeToBox` already does `probe.rotate()` for stills. That asymmetry is the whole bug.

Not external-media specific — `photoProcessor` calls the same shared `generateThumbnail` for managed admin uploads.

## The fix

`.rotate()` before `.resize()` in all three, guarded to stills for the reason `resizeToBox` already documents: `.rotate()` flattens a multi-frame source to its first frame.

**And the half that compounds it.** `photos.width`/`height` were taken straight from `metadata.width`/`height`, which for orientation 5-8 are swapped relative to what a viewer sees — so a portrait photo was stored as landscape and masonry/justified sized its tile with the wrong aspect ratio *on top of* the image being unrotated. A shared `orientedDimensions()` helper now does that conversion at all four capture sites:

| Site | Path |
|---|---|
| `photoProcessor.js` ×2 | managed upload + background processing |
| `adminExternalMedia.js` | external/reference import |
| `adminPhotoDimensions.js` | the dimension repair job |

So the stored numbers describe the rotated result the generators now produce.

## Scope limit worth knowing

Existing rows keep their pre-rotation dimensions until the photo is reprocessed — the dimension repair only visits rows where width/height is NULL, and it cannot tell a correct landscape from a mis-stored one without re-reading every original. The **images** correct themselves on the next thumbnail/preview regeneration. If you want existing dimensions corrected in bulk, that wants its own pass and I'd rather it were a deliberate one.

## Verification

- `__tests__/services/imageProcessorOrientation.test.js` — **7 tests**. Fixture is 400×200 stored with `orientation: 6`, i.e. displaying as 200×400, the exact shape the reporter's Sony bodies produce. Asserts the fixture really is tagged, that `orientedDimensions` swaps for 5-8 and leaves 1 alone and survives unusable metadata, that thumbnail and preview both come out portrait, and that the output carries no surviving orientation tag (so nothing double-rotates).
- **Mutation-checked**: reverting the three `.rotate()` calls and the swap fails 4 of the 7. They do not pass by accident.
- Full backend suite: same 6 pre-existing failures as `main`, diffed. One test needed updating — `photoProcessor.processPhoto.test.js` mocks `imageProcessor`, so its mock now carries `orientedDimensions`; mirrored inline rather than `requireActual`'d, which would drag the real module's database dependency into the mock factory.
- `eslint` clean on every changed file. (`adminExternalMedia.js:7` reports a pre-existing unused `getExternalMediaRoot` — present on unmodified `main`, untouched here.)


the-luap:
## Scope narrowed after review

Round 2 added a `recompute` mode plus a face requeue, to correct libraries that predate this fix. Round 3 found five problems with it and it has been **reverted** — split out to #1198 with all five recorded.

The one that settled it was an own-goal: requeueing face scanning makes `processPhotoFaces` call `ensurePreviewImage`, which returns the **cached** pre-fix preview whenever it is still a valid image. The rescan therefore read unrotated pixels and scaled those boxes by the newly-corrected dimensions — worse than leaving the data alone. The rest need work that does not belong in a bugfix: S3 and RAW originals are unreadable by that job, archived events have no originals at all, orientation 2/3/4 changes pixels without changing dimensions so a delta check never fires, and the two writes are not atomic.

**What this PR now is**, unchanged since round 1 and clean in every round:

- `.rotate()` before `.resize()` in `generateThumbnail`, `generateHeroImage` and `generatePreviewImage`
- `orientedDimensions()` at all **eight** ingest sites (round 2 found four I had missed — filesystem watcher, S3 auto-importer, v1 upload API, replace-by-name; I had grepped `metadata.width` and the v1 route aliases it to `meta`)
- the animated guard removed from thumbnail and hero, which never open sources with `animated: true` and so already flatten — guarding there protected an animation that was being discarded anyway while leaving the output unrotated. Kept in the preview generator, which genuinely does preserve animation.

Existing photos still need #1198; new and reprocessed ones are correct from here.

PR #1196, opened 2026-08-26
fix(gallery): show colour labels in the Carousel layout

### LINKED ISSUE #1189: Carousel layout never shows colour labels, in any gallery
## Problem

The Carousel layout never shows colour labels. Not the viewer's own mark from #1044, and not other viewers' marks from #1178 — a photo the client flagged green looks identical to one nobody touched, and the only way to see the label is to open the lightbox.

Every other layout shows them:

| Layout | Colour labels |
|---|---|
| Grid, Justified, Masonry, Mosaic, Timeline, Story | ✅ via `PhotoCard` |
| Premium | ✅ renders `ColorLabelBadge` directly |
| **Carousel** | ❌ **none** |

`CarouselGalleryLayout.tsx` renders its own markup — an `AuthenticatedImage` for the main frame (`:86`) and another for each thumbnail in the strip (`:253`) — rather than going through `PhotoCard`, so it inherits nothing. `grep -c ColorLabelBadge` on that file is 0, on `main` and on every branch.

Carousel is a first-class option in the layout dispatch (`PhotoGridWithLayouts.tsx:291`), so this is a gallery a photographer can pick and then silently lose the proofing signal in.

## Not a regression

This predates #1178. The colour-label feature shipped in #1044 through `PhotoCard`, and Carousel has never used `PhotoCard`, so it has been missing since the feature landed. #1178 (other viewers' labels) inherits the same gap for the same reason — it was raised in review there and split out rather than widening that PR.

## Fix shape

Render `ColorLabelBadge` in both places the carousel paints a photo:

- the **thumbnail strip** (`:242-260`) — this is the one that matters, since it is the only place the carousel shows more than one photo at a time and therefore the only place a label can be scanned;
- the **main frame** (`:86`), for the photo currently in view.

The badge takes `colorLabel` and `otherColorLabels` and positions itself absolutely, so it needs a positioned ancestor and nothing else. Worth checking the thumbnail size first: at strip scale the 20px own-dot plus up to three 10px dots may need a smaller variant rather than the grid-tile treatment.

## Note on placement

`ColorLabelBadge` hard-codes `absolute top-2 left-2`. In the tile layouts that corner is shared with `FeedbackCountIndicators` and the selection checkbox, which was reviewed at length in #1178 and settled as "no corner is collision-free, so share the one the feature already owns". The carousel's own top-left corner carries its overlay controls (`:114`), so whoever picks this up should check that the badge does not land under them — and if it does, that is the point at which making the position overrideable per layout starts to pay for itself.

the-luap:
Closes #1189.

## The gap

`CarouselGalleryLayout` renders its own `AuthenticatedImage` for the main frame and for each strip thumbnail rather than going through `PhotoCard`, so it inherited none of the colour-label treatment — not other viewers' marks from #1178, and not the viewer's own from #1044.

Not a regression from #1178: the feature shipped through `PhotoCard` and Carousel has never used it, so a photo the client flagged green has looked identical to an untouched one since #1044 landed.

## What changed

Rendered in **both** places the carousel paints a photo. The **thumbnail strip** is the one that matters — it is the only place this layout shows more than one photo at once, so it is the only place a label can actually be scanned.

Two additions to `ColorLabelBadge`, both defaulting to current behaviour so every existing layout renders identically (its 6 existing tests pass untouched):

| Prop | Why |
|---|---|
| `size="sm"` | The strip's tiles are 80px. The grid-sized 20px own-dot plus three 10px dots covers most of the image. |
| `position` | This layout has different corners free. |

**Placement had to be worked out per surface**, because the carousel's corners are already busy:

- **main frame → `bottom-4 left-4`**, the only free corner: top-left carries the counter and category chips, top-right the play/fullscreen buttons, and both sides the prev/next controls.
- **strip → `top-1 left-1`**, where nothing competes.

This is precisely the per-layout position override that #1178's review concluded would start to pay for itself the first time a layout genuinely needed a different corner — so it is introduced here rather than retrofitted across the tile layouts, which still share the corner the feature already owns.

## Verification

- `CarouselColorLabels.test.tsx` — **6 tests**: defaults untouched (the guard that keeps every other layout unchanged), the small variant shrinking both dot kinds, the position override applying and the default *not* applying, no markup at all for an unmarked photo (the strip maps over every photo, so an empty positioned span would be wrong), other viewers' marks showing when the viewer has none of their own — the #1178 case that makes this gap visible — and the inset ring surviving the small variant.
- Existing `ColorLabelBadge.test.tsx`: **6/6 pass unmodified**.
- Full gallery + settings suites: 82 tests pass.
- `tsc -b` reports no errors in either changed file; `eslint` clean.

## Screenshot

Not attached yet. Producing an honest one needs a gallery switched to the carousel layout with colour-label feedback seeded against it, which is a rig setup rather than a page load — happy to capture it before this merges if you'd like it on the record.

### DIFF
--- a/frontend/src/components/gallery/ColorLabelBadge.tsx
+++ b/frontend/src/components/gallery/ColorLabelBadge.tsx
@@ -13,6 +13,19 @@ interface ColorLabelBadgeProps {
   otherColorLabels?: string[];
   /** Extra classes for positioning inside the tile. */
   className?: string;
+  /**
+   * Dot scale. 'sm' is for surfaces smaller than a grid tile — the carousel's
+   * 80px thumbnail strip (#1189), where the default 20px dot plus three 10px
+   * ones covers most of the image.
+   */
+  size?: 'md' | 'sm';
+  /**
+   * Where the dot row sits inside its positioned ancestor. Overridable because
+   * the tile layouts and the carousel have different corners free: the
+   * carousel's own top-left carries its counter and category chips, so the
+   * default would land underneath them (#1189).
+   */
+  position?: string;
 }
 
 /**
@@ -27,6 +40,8 @@ export const ColorLabelBadge: React.FC<ColorLabelBadgeProps> = ({
   colorLabel,
   otherColorLabels = [],
   className = '',
+  size = 'md',
+  position = 'top-2 left-2',
 }) => {
   const { t } = useTranslation();
 
@@ -46,6 +61,9 @@ export const ColorLabelBadge: React.FC<ColorLabelBadgeProps> = ({
     colors: others.map((c) => t(`feedback.colorLabels.${c}`, c)).join(', '),
   });
 
+  const mineDotClass = size === 'sm' ? 'w-3.5 h-3.5 border' : 'w-5 h-5 border-2';
+  const otherDotClass = size === 'sm' ? 'w-2 h-2' : 'w-2.5 h-2.5';
+
   return (
     <>
       {mine && swatch && (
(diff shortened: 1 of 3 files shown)


the-luap:
## Screenshot

Carousel layout with colour labels, both surfaces at once.

issue-1189-carousel-color-labels

**Main frame** — the viewer's own blue mark as the inset ring plus the dot row bottom-left, with the two other guests' green and red beside it. Bottom-left because it is the only corner this layout leaves free: top-left carries the counter and category chips, top-right the play and fullscreen buttons, and both sides the prev/next controls.

**Thumbnail strip** — the same photo's three dots at top-left in the small variant, and two more photos further along carrying a single mark each. This is the surface that matters: the strip is the only place the carousel shows more than one photo at a time, so it is the only place a label can actually be scanned.

Before this, neither surface showed anything — Carousel paints its own markup instead of going through `PhotoCard`, so it has shown no colour labels at all since the feature landed in #1044.

PR #1204, opened 2026-08-26
fix(images): fence the capture-date backfill on the file it read (#1201)

### LINKED ISSUE #1201: Capture-date backfill can write a stale date onto a photo replaced mid-run
## The bug

The capture-date backfill can write one photo's EXIF date onto a different photo's file.

`adminPhotoDimensions.js:433-436` (main; `:425-428` on stable) commits its result keyed only on the row id:

```js
const updated = await db('photos')
  .where({ id: photo.id })
  .whereNull('captured_at')
  .update({ captured_at: captured.toISOString() });
```

The job snapshots every candidate up front, then walks them one at a time reading originals off S3 or a NAS mount — a pass that can run for many minutes. `photoReplacementService.replacePhoto()` swaps a **new file under an existing row id** and rewrites `path`/`filename`; it is reachable from the `replace_by_name` upload path (`adminPhotos.js:279`).

So: the backfill reads photo 500's original and extracts `2026-06-03`. Before it writes, an editor re-uploads over that same row. The replacement has no EXIF date, so `captured_at` is still NULL and the `whereNull` guard passes. The old file's capture date lands on the new photo.

Silent — nothing errors, the run reports it as a success, and the gallery just sorts that photo to the wrong place.

## Scope

- **main** — merged in #1179
- **stable** — merged in #1183

Narrow trigger (a replace-by-name upload has to land inside the backfill's window on the same photo) but the failure is wrong data with no signal.

## Fix

The pattern already exists on main. #1199 added exactly this fence to the orientation backfill for the same reason:

```js
const fenceOf = (photo) => ({ id: photo.id, path: photo.path, filename: photo.filename });
```

Matching path and filename as well as id means a replaced row simply matches zero rows and is skipped, rather than being given the previous file's metadata. The capture-date job already selects `photos.path` and `photos.filename`, so this needs no query change — and `successCount` should come from the affected-row count so a fenced-out row is not reported as updated.

Same change on both branches.

## How this was missed

Raised during review of #1179 and dismissed by me on the grounds that "there is no photo-replacement endpoint". That was wrong: there is no dedicated endpoint, but `replace_by_name` calls `replacePhoto` on an existing row, which is the same thing for this purpose. Recording it here rather than leaving the incorrect rejection standing in that thread.

the-luap:
Closes #1201.

## The bug

`adminPhotoDimensions.js:433` committed the capture-date backfill keyed only on the row id:

```js
const updated = await db('photos')
  .where({ id: photo.id })
  .whereNull('captured_at')
  .update({ captured_at: captured.toISOString() });
```

The job snapshots every candidate up front, then walks them one at a time reading originals off S3 or a NAS mount — a pass that can run for many minutes. `photoReplacementService.replacePhoto()` swaps a **new file under an existing row id** and rewrites `path`/`filename`; it is reachable from the `replace_by_name` upload path (`adminPhotos.js:279`).

So: the backfill reads photo 500's original and extracts `2026-06-03`. Before it writes, an editor re-uploads over that same row. The replacement has no EXIF date of its own, so `captured_at` is still NULL and the `whereNull` guard passes — the old file's capture date lands on the new photo. Silent: nothing errors, the run reports it as a success, and the gallery sorts that photo to the wrong place.

## The fix

Fence the write on the identity that was actually read, not just the id — the same fence #1199 put on the orientation backfill for the same reason:

```js
.where({ id: photo.id, path: photo.path, filename: photo.filename })
.whereNull('captured_at')
```

A replaced row now matches zero rows and is skipped instead of being given the previous file's metadata. The candidate query already selects `photos.path` and `photos.filename`, so no query change. `successCount` already comes from the affected-row count (`if (updated) successCount++`), so a fenced-out row is not reported as updated.

Knex renders a null value in the object form as `is null`, so a row with a NULL `path` still matches itself.

## Test

`captureDateBackfill.test.js` — new case: a replacement landing mid-run (path/filename rewritten, `captured_at` still NULL) leaves `captured_at` NULL and is not counted as a success. Verified it fails on the unfenced code and passes with the fence; the other 8 cases in the file are unchanged and green.

No UI surface touched, so no screenshot.

## Stable

Stable carries the same shape (`:425-428`, merged in #1183) and gets the twin separately.

---

## Follow-ups from external review

An external review pass (Codex, 4 rounds) cleared the fence itself and surfaced one adjacent gap, fixed in the commits above.

**Accounting.** `replacePhoto` is not the only writer of `photos.path`/`filename` — `eventRenameService.js:201-207` rewrites both on an event rename, which is not a content change. Either of those, or another writer filling `captured_at` first, makes the update affect zero rows. Those candidates fell out of the run's arithmetic entirely: `success + noExif + failed` no longer added up to the count the operator was shown when they started the job. They are now counted (`skipped`) and surfaced on the card, shown only when non-zero — the same shape the orientation job uses for `staleTiers`.

Deliberately **not** done: retrying the row. A retry means a second S3/NAS read, it can lose the same race again, and for the already-dated case there is nothing to retry. The fenced row keeps `captured_at` NULL, so the status endpoint keeps reporting it as backlog and the next run picks it up.

**Checked and dismissed:** whether a NULL `path` breaks the fence. Knex renders a null value in the object form as `is null` on both the pg and sqlite3 clients (verified against the generated SQL on both), so such a row still matches itself.

Locale coverage for the new string matches the existing `staleTiers` key — en, de, fr, sl — with the `defaultValue` carrying the rest. Screenshots in a comment below.

### DIFF
--- a/backend/__tests__/integration/captureDateBackfill.test.js
+++ b/backend/__tests__/integration/captureDateBackfill.test.js
@@ -198,6 +198,8 @@ describe('capture date backfill (#1172)', () => {
 
     expect(new Date((await db('photos').where({ id: photoId }).first()).captured_at).toISOString()).toBe(claimed);
     expect(done.body.lastResult.success).toBe(0);
+    // Read but not written, so it is accounted for rather than dropped.
+    expect(done.body.lastResult.skipped).toBe(1);
   });
 
   it('does not date a row whose file was replaced while it was reading (#1201)', async () => {
@@ -218,5 +220,8 @@ describe('capture date backfill (#1172)', () => {
 
     expect((await db('photos').where({ id: photoId }).first()).captured_at).toBeNull();
     expect(done.body.lastResult.success).toBe(0);
+    // Not an error and not "no EXIF" — the date was found, another writer just
+    // got there first. It stays in the backlog for the next run.
+    expect(done.body.lastResult).toMatchObject({ noExif: 0, failed: 0, skipped: 1 });
   });
 });
--- a/backend/__tests__/integration/captureDateBackfill.test.js
+++ b/backend/__tests__/integration/captureDateBackfill.test.js
@@ -199,4 +199,24 @@ describe('capture date backfill (#1172)', () => {
     expect(new Date((await db('photos').where({ id: photoId }).first()).captured_at).toISOString()).toBe(claimed);
     expect(done.body.lastResult.success).toBe(0);
   });
+
+  it('does not date a row whose file was replaced while it was reading (#1201)', async () => {
+    // replacePhoto swaps a NEW file under an existing row and rewrites
+    // path/filename (reachable from replace_by_name). The replacement carries
+    // no date of its own, so captured_at is still NULL and the whereNull guard
+    // alone would let the previous file's EXIF date land on it. The write is
+    // fenced on the identity that was read, so the row is skipped instead —
+    // and not counted as updated either.
+    const { photoId } = await seed({ relpath: 'orig.jpg', exifIso: '2026-06-03T09:00:00Z' });
+
+    const res = await request(app).post('/api/admin/photos/repair-capture-dates');
+    expect(res.body.count).toBe(1);
+    // Simulate the replacement landing before the loop writes.
+    await db('photos').where({ id: photoId })
+      .update({ path: 'capfill/replaced.jpg', filename: 'replaced.jpg' });
+    const done = await settle();
+
+    expect((await db('photos').where({ id: photoId }).first()).captured_at).toBeNull();
+    expect(done.body.lastResult.success).toBe(0);
+  });
 });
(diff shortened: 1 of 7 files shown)


the-luap:
## Screenshot

The new line on the Capture Dates card, shown only on a run that skipped something.

issue-1201-skipped-en

**Light, EN** — `Last run: 0 updated, 0 with no date found, 0 unreachable` on its own used to be the whole report for a run that read a photo and deliberately declined to write it. The amber line underneath is what closes that hole. Same shape the orientation job already uses for `staleTiers` a few lines below: rendered only when the count is non-zero, so an ordinary run looks exactly as it did before.

issue-1201-skipped-dark

**Dark, EN** — `text-amber-600 dark:text-amber-400`, the same pair the Missing Capture Date tile above it uses.

issue-1201-skipped-de

**Light, DE** — locale coverage matches the `staleTiers` key: en, de, fr, sl, with the `defaultValue` carrying the rest.

Rendered from this branch against the real payload shape (`lastResult: { success: 0, noExif: 0, failed: 0, skipped: 1 }`), not from a live install — the local rig is on migration 183 and predates this card entirely.

**Wording note:** it says "were not updated", not "will be retried". Two ways to land in that counter — the row was dated by another writer, or its file changed under us — and only the second is still backlog. The Missing Capture Date figure above is the one that says whether work is left.

PR #1205, opened 2026-08-26
fix(images): fence the capture-date backfill on the file it read (#1201)

the-luap:
Stable twin of #1204. Same change, same test — #1201 says "same change on both branches".

## The bug

`adminPhotoDimensions.js:425` committed the capture-date backfill keyed only on the row id:

```js
const updated = await db('photos')
  .where({ id: photo.id })
  .whereNull('captured_at')
  .update({ captured_at: captured.toISOString() });
```

The job snapshots every candidate up front, then walks them one at a time reading originals off S3 or a NAS mount — a pass that can run for many minutes. `photoReplacementService.replacePhoto()` swaps a **new file under an existing row id** and rewrites `path`/`filename`; it is reachable from the `replace_by_name` upload path (`adminPhotos.js`).

A replacement landing inside that window carries no EXIF date of its own, so `captured_at` is still NULL and the `whereNull` guard passes — the old file's capture date lands on the new photo. Silent: nothing errors, the run reports it as a success, and the gallery sorts that photo to the wrong place.

## The fix

Fence the write on the identity that was actually read, not just the id:

```js
.where({ id: photo.id, path: photo.path, filename: photo.filename })
.whereNull('captured_at')
```

A replaced row now matches zero rows and is skipped. The candidate query already selects `photos.path` and `photos.filename`, so no query change. `successCount` already comes from the affected-row count (`if (updated) successCount++`), so a fenced-out row is not reported as updated.

Knex renders a null value in the object form as `is null`, so a row with a NULL `path` still matches itself.

Note: the orientation backfill this fence pattern comes from (#1199) is main-only, so this diff is the capture-date write alone — no other surface on stable carries the shape.

## Test

`captureDateBackfill.test.js` — new case: a replacement landing mid-run (path/filename rewritten, `captured_at` still NULL) leaves `captured_at` NULL and is not counted as a success. Verified on this branch that it fails on the unfenced code and passes with the fence; the other 8 cases are unchanged and green.

No UI surface touched, so no screenshot.

---

## Follow-ups from external review

Mirrors the follow-up commits on #1204, verified independently on this branch (backend suite 9/9, tsc, eslint, and the card re-rendered from this branch's code).

**Accounting.** `replacePhoto` is not the only writer of `photos.path`/`filename` — `eventRenameService.js` rewrites both on an event rename, which is not a content change. Either of those, or another writer filling `captured_at` first, makes the update affect zero rows, and those candidates fell out of the run's arithmetic: `success + noExif + failed` no longer added up to the count the operator was shown when they started the job. Now counted (`skipped`) and surfaced on the card, shown only when non-zero.

The wording states what is known — changed by something else, not updated — rather than promising a retry: for the already-dated case there is nothing to retry, and the Missing Capture Date figure above is the one that says whether work is left.

**Checked and dismissed:** whether a NULL `path` breaks the fence. Knex renders a null value in the object form as `is null` on both the pg and sqlite3 clients, so such a row still matches itself.

Screenshot in a comment below.

### DIFF
--- a/backend/__tests__/integration/captureDateBackfill.test.js
+++ b/backend/__tests__/integration/captureDateBackfill.test.js
@@ -198,6 +198,8 @@ describe('capture date backfill (#1172)', () => {
 
     expect(new Date((await db('photos').where({ id: photoId }).first()).captured_at).toISOString()).toBe(claimed);
     expect(done.body.lastResult.success).toBe(0);
+    // Read but not written, so it is accounted for rather than dropped.
+    expect(done.body.lastResult.skipped).toBe(1);
   });
 
   it('does not date a row whose file was replaced while it was reading (#1201)', async () => {
@@ -218,5 +220,8 @@ describe('capture date backfill (#1172)', () => {
 
     expect((await db('photos').where({ id: photoId }).first()).captured_at).toBeNull();
     expect(done.body.lastResult.success).toBe(0);
+    // Not an error and not "no EXIF" — the date was found, another writer just
+    // got there first. It stays in the backlog for the next run.
+    expect(done.body.lastResult).toMatchObject({ noExif: 0, failed: 0, skipped: 1 });
   });
 });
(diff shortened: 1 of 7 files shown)


the-luap:
## Screenshot

Same card, same line, rendered from the stable branch.

issue-1201-skipped-stable-en

**Light, EN** — identical to #1204's. The only branch difference is the card's permission gate (`canEditSettings` here, `canManageSystem` on main after #1182); the result line and its locale keys are the same.

Rendered from this branch against the real payload shape (`lastResult: { success: 0, noExif: 0, failed: 0, skipped: 1 }`), not from a live install — the local rig is on migration 183 and predates this card.

PR #1208, opened 2026-08-27
feat(feedback): a third identity mode with one shared colour tag per photo (#1197)

### LINKED ISSUE #1197: Feedback: a third identity mode with one shared, overwritable colour tag per photo
Split out of #1178, where @boergu raised it alongside the grid bug. The bug is fixed and shipped; this is the other half of that report and it deserves its own thread.

## The request, in the reporter's words

> What I'm asking for is a third model with no identity dimension at all for the shared tag: one single, identity-less colour tag per photo, stored server-side, that any guest on any device can overwrite. Not "everyone shares a device's state" — "everyone, on any device, shares the photo's state." So if guest A marks a photo green and guest B later marks the same photo orange, the shared tag simply becomes orange — B's action overrides A's, rather than both marks being kept separately.

The use case is a single collaboratively-agreed verdict per photo, rather than per-person tallies.

## Why the existing modes don't cover it

`identity_mode` accepts exactly two values (`feedbackValidation.js:213-214`):

| Mode | Identity | Shared across |
|---|---|---|
| `simple` | anonymous device fingerprint | guests on the same physical device |
| `guest` | named individual | that individual, on any device |

@boergu checked both before filing, and his reading is correct: they differ in what "identity" *means*, not in whether feedback is identity-scoped. Every colour label is a `photo_feedback` row carrying a `guest_identifier`/`guest_id`, so two guests marking the same photo always produce two rows. There is no configuration that collapses them.

## What a third mode would touch

Sketching this out because the storage decision is the whole design, and it is not obvious:

1. **Where the shared tag lives.** Either a `photo_feedback` row under a reserved identity (keeps the existing filter/aggregate machinery, at the cost of a magic identifier), or a column on `photos` (honest about being one value per photo, but a second storage location for the same concept, and every read path has to know which mode it is in).
2. **The aggregates stop meaning what they mean.** `photos.color_label_count` and `dominant_color_label` exist to tally *across* guests. With one tag there is nothing to tally and no tie to break — the API would need to return a single value rather than a per-colour map, and every consumer of it (grid, lightbox, admin filters, the Lightroom round-trip in #745) has to handle both shapes.
3. **Filtering collapses.** `color_labels` vs `my_color_labels` are the same question in this mode. The guest filter chips would need to become one set.
4. **Last-write-wins needs a rule for concurrency.** Two guests tapping different colours within the same second must land on a defined answer, not whichever transaction commits second by luck.
5. **Switching an event's mode is the awkward part.** An event that already has per-guest labels has N marks per photo and no single correct collapse — first? most recent? most common? Or is the mode locked once feedback exists? This wants deciding before implementation, not during.
6. **Attribution disappears by design.** The admin currently sees who marked what. In this mode that is the point — but it should be a deliberate, documented consequence rather than a surprise.

## Scope note

Worth treating as a genuine feature rather than a settings toggle: it changes what a colour label *is*, and #745 (Lightroom round-trip) now reads these values too, so the API shape question in (2) reaches further than the gallery.

Filed on @boergu's behalf — the original request and the reasoning above are his.

the-luap:
> **Stacked on #1161.** This PR targets `feat/1160-gallery-folders`, so the diff above is this change alone. Merge #1161 first; GitHub retargets this to `main` automatically when it does.

---

Closes #1197.

Third identity model, split out of #1178: one identity-less colour tag per photo that any guest on any device can overwrite. Not "everyone shares a device's state" but "everyone shares the photo's state" — a single agreed verdict instead of per-person tallies.

## The three decisions the issue asked to settle first

**(1) Where the tag lives** — an ordinary `photo_feedback` row under a reserved `guest_identifier`, not a column on `photos`. This is the choice that makes the rest of the feature small: the per-colour tally simply has exactly one entry, so `dominant_color_label`, `color_label_count`, the admin colour filter and the XMP/CSV export #745 reads all keep their current shapes. **No API shape change, and no consumer learns a second shape** — which also disposes of decision (2).

The identifier cannot be claimed by a guest: real ones are sha256 hashes or server-minted UUIDs read back from the guests table, never client-supplied. The per-guest write path rejects it outright anyway (`feedbackService.js`), and there's a test for that.

**(4) Concurrency** — the write runs in a transaction that locks the photo row. Without it, two guests tapping different colours in the same instant both read "no tag", both insert, and the photo carries two shared tags — the exact per-guest tally this mode exists to remove. The test for this **fails with 3 rows** against the un-transactioned version; SQLite ignores `forUpdate` but serialises writers anyway, on Postgres it does real work.

**(5) Switching modes** — non-destructive, as chosen. Existing per-guest labels are left alone and not read while shared is on; the tag starts empty rather than collapsing marks nobody agreed on. Switching back restores every original mark exactly. An event can hold both sets, only one live.

## Decision (3) needed nothing — the premise was off

The issue expected `color_labels` and `my_color_labels` to collapse into one. They don't, on either side:

- **Guest gallery** — the filters and chip counts already read `my_color_label`, which now carries the shared tag. Zero frontend change.
- **Admin** — `my_color_labels` filters `photo_admin_marks`, the admin's own private marks (#1183), not guest identity. It was never the same question as `color_labels`.

## Scope

`identity_mode='shared'` changes **the colour tag only**. Likes, ratings, comments, favourites and reactions stay per-visitor exactly as in `simple` — that's what was asked for, and widening it would change what every other control means. There's a test asserting two devices can like independently in shared mode.

## Behaviour details

- **Re-tapping the current colour clears it**, from any guest — the same toggle every other colour path uses, and the only way to remove a tag without inventing a second control. Any guest clearing any tag is the mode, not a hole in it.
- **The tag stays visible with `show_feedback_to_guests` off**, arriving through the per-viewer channel because it's the photo's own state rather than someone else's opinion. The per-colour tallies stay hidden. The colour filters answer from it for the same reason — otherwise a gallery with sharing off would show colours on tiles that no filter could find.
- **The shared tag is not re-rendered as another viewer's dot** (#1178's grid dots are skipped in this mode), or every tile would show the same colour twice.
- **Attribution is gone by design** (decision 6) and the settings panel says so *before* the operator picks the mode, along with the promise that existing labels survive.

## Tests

**15 new backend integration tests** (`sharedColorTag.test.js`) covering overwrite, cross-device visibility, no attribution stored, clear-by-any-guest, the concurrency race, per-photo isolation, tally shape, no double-render, the sharing-off surfaces, the non-destructive mode switch both ways, likes staying per-guest, and the reserved-identity guard.

**6 new frontend tests** for the settings panel.

Full suites: backend **2344 passed**, frontend **235 passed + 6 new**. The backend's 22 failures are pre-existing — the same suite list fails on a clean `origin/main` (settings logo, adminPhotos reference, CRM mint paths, webhook delivery, backup service, admin auth), and none touch feedback or colour labels. Confirmed by running the full suite on a stashed tree and diffing the failing sets across three runs.

Screenshots in a comment below.


the-luap:
## Screenshots

The only new UI surface: the third identity mode in the event's feedback settings.

issue-1197-settings-shared-en

**Light, EN** — the option, and the consequences stated before the operator commits to them. Attribution disappearing is the point of the mode rather than a bug, but it is not something to discover afterwards from the admin view; the same line promises that existing per-visitor labels survive the switch, which is the non-destructive behaviour decision (5) settled on.

issue-1197-settings-shared-dark

**Dark, EN** — `text-amber-600 dark:text-amber-400`, the pair used elsewhere for consequential-but-not-error notes.

issue-1197-settings-shared-de

**Light, DE** — the three new strings are in all eight locales that carry the sibling `identityMode*` keys (en, de, fr, sl, es, nl, pt, ru).

Rendered from this branch against the real settings payload, not from a live install — the local rig predates this code.

**Nothing changes in the gallery.** That is the design working: the badge, the lightbox swatch and the keyboard shortcuts all read "the colour on this photo, from my point of view", which is exactly what the shared tag is — so they were already correct without being touched. What changes is that a second device now sees the same colour, and that is asserted in `sharedColorTag.test.js` rather than in a screenshot, where two browser profiles side by side would prove less.

PR #1216, opened 2026-08-27
fix(guests): surface duplicate guest registrations, and stop making so many (#1210)

### LINKED ISSUE #1210: Guest registration creates a new identity every time, even with the same email — favourites/likes fragment across duplicate guests
## Guest registration creates a new identity every time, even when the same email is given — favourites/likes fragment across duplicate guests

**Summary**

On a gallery with `identity_mode: 'guest'`, opening the gallery from a new browser/device, or simply after the guest session token expires, always creates a brand-new row in `gallery_guests` — even when the visitor enters the exact same name and email as a previous visit. There's no lookup against an existing (non-deleted) guest for that event before inserting. Over the lifetime of a single gallery this produces many duplicate "guests" for what is really one person, and their likes/favourites/comments end up scattered across those duplicate identities instead of being recognised as one person's selections. For a photographer using the gallery's hearted photos to drive an invoicing/editing workflow, this means the admin has to manually detect and merge duplicates before the "final selection" can be trusted.

**Where this lives in the code** (paths relative to `picpeak-backend`):

- `routes/galleryGuests.js`, `POST /:slug/guest` — always does a plain `db('gallery_guests').insert(...)`, with no prior `where({ event_id, email })` lookup. This is the path a returning visitor lands on whenever their token is missing/expired.
- `middleware/guestAuth.js`, `signGuestToken()` — default `expiresIn = '24h'`, and every call site (registration, invite redemption, recovery verify) uses this default with no override. So even the *same* browser loses its identity after 24h of inactivity.
- Admin-issued invites (`guest_invites` table) are single-use — once `redeemed_at` is set, a client re-opening the same emailed gallery link a second time can't silently re-authenticate with it, and falls through to the plain registration form above.
- A proper fix already exists for part of this: `POST /:slug/guest/recover` + `POST /:slug/guest/verify` emails a 6-digit code and correctly reuses the existing guest row by email (`resolveGuest`/`verify` route). The frontend does expose this as an "I've been here before" link right on the registration form (`public.entry.alreadyHere` i18n key) — but it's opt-in. Nothing nudges a returning guest toward it instead of just re-typing their name, so in practice it goes largely unused and duplicates keep accumulating even for guests who *did* give an email the first time.
- `event_feedback_settings.require_name_email` defaults to `false` on every newly created event (`routes/adminEvents/crud.js`, `routes/v1/events.js`, `services/feedbackService.js` all hardcode `false`). Since recovery is keyed by email, any guest who wasn't asked for one is permanently ineligible for it — worsening the fragmentation for exactly the guests (family/friends given the same shared gallery link) most likely to revisit without a personal invite.

**Reproduction**

1. Create an event with `identity_mode: 'guest'`, `require_name_email: true`.
2. Register as a guest with name "Test" and `email: test@example.com`, like a couple of photos.
3. Wait past the guest token's 24h expiry (or clear the browser's storage/cookies to simulate a new device), then reopen the same gallery URL.
4. Enter the same name and email again via the plain registration form (not "I've been here before").
5. Observe: a second row is created in `gallery_guests` with the same email, and the earlier likes stay attributed to the first (now orphaned) guest row — the gallery's aggregate hearted-photos view is now split across two guests for one real person.

**Suggested directions** (any of these would meaningfully help; not mutually exclusive):

- On `POST /:slug/guest`, when an email is supplied, look up an existing non-deleted guest for `(event_id, lower(email))` before inserting. If found, treat it as a recovery — either reuse the row directly, or (to preserve the existing verification-by-email guarantee) auto-route into the `/recover` + `/verify` flow instead of silently reusing it.
- Make the guest JWT expiry longer by default (or configurable per event) — 24h is short relative to how long a client typically takes to finish reviewing/selecting a gallery.
- Surface the "I've been here before" option more assertively — e.g. detect that the typed email already exists for the event and switch the form into recovery mode inline, rather than requiring the guest to notice and click a separate link first.

**Environment**

Self-hosted, Docker Compose deployment. Happy to provide exact version/commit if useful — checked against the `picpeak-backend`/`picpeak-frontend` images running on our own install (not the latest `main`, so line numbers above may have shifted slightly upstream).

the-luap:
Closes #1210.

> **Stacked on #1214** (`fix/1211-uncategorized-filter`). The diff above is this change alone. Merge #1214 first; GitHub retargets this to `main` automatically when it does.

## The problem

Guest registration always inserts. A client whose token expired — or who opens the gallery on a second device — becomes a new `gallery_guests` row, and their likes and favourites split across the copies. The photographer's "final selection" is then only trustworthy if somebody notices two Tinas with half the picks each.

## What I did — and what I deliberately didn't

The report's first suggestion is to reuse the row when a typed email matches. **I did not do that**, and would push back on it: anyone who knows an address could type it in and inherit that person's identity and selections. Answering differently for a known email also leaks which addresses are in the gallery — precisely what `/guest/recover` goes out of its way to avoid by always returning 200.

The registration path is untouched. Instead, two halves:

### 1. Say which rows are the same person

Merging already worked — endpoint *and* merge mode in the UI. What was missing was knowing **which** rows to merge; a split selection is invisible until someone spots it.

The guests list now marks each row with the others sharing its email and returns a count. The admin list shows a banner and per-row badges, and hands the group to the merge mode that already exists.

Matching rules, each of which is a decision rather than a default:

- **Case-folded and trimmed** — the same person types `Tina@` one day and `tina@` the next, and a trailing space is invisible in the list. Both would otherwise read as different people.
- **Email only.** Two guests called Anna are not evidence of anything.
- **No email, no grouping.** `require_name_email` is off by default, so a shared gallery link produces plenty of email-less guests; grouping those would merge strangers.
- **Deleted rows ignored** — the survivor is not "a duplicate of" something the admin cannot see or merge.

It **preselects, it does not merge**. Which row survives decides the name and verification state the merged guest keeps, and that is the admin's call.

### 2. Create fewer of them

The guest token was 24h, and every call site took that default — so even the same browser lost its identity after a day of inactivity, which is the churn that produces the duplicates in the first place.

Now **30 days**, `GUEST_TOKEN_TTL` to override, documented in `.env.example`. A guest token is scoped to one event, carries no admin capability, and the gallery is already behind whatever password or share link protects it; 30 days is the shape of a real proofing cycle. An operator with stricter requirements can set `24h` and get the old behaviour back.

## Not done, and why

Preventing duplicates *at the entry path* needs the verification round-trip — email a code on every registration, not just known addresses, or the response itself leaks. That is a real option but it makes every guest check their inbox to get in and hard-couples gallery entry to working SMTP. Worth deciding separately rather than smuggling into a bug fix.

## Tests

13 across both ends. 8 of the 9 backend ones fail without the change (the ninth pins that the signed token still verifies, which must hold either way).

- `guestDuplicateDetection.test.js` — pairs point at each other, group/row counts, case and whitespace folding, no grouping without an email, shared names ignored, deleted rows ignored.
- `guestTokenTtl.test.js` — defaults to 30 days not 24 hours, honours an override, still signs a token the middleware accepts. Pinned because a silent revert to 24h would restore the churn without breaking anything visible.
- `AdminGuestsList.duplicates.test.tsx` — the banner counts, the badges match what it claims, both stay away when nobody is duplicated, and Review opens merge mode **without merging**.

Backend 2342 pass (failures are the known pre-existing baseline, unchanged). Frontend 243 pass. `tsc` and eslint clean.

**The frontend tests earned their place immediately** — they caught a real bug while being written: my new `useMemo` sat below the loading early-return, so the hook count changed between renders and the component threw "Rendered more hooks than during the previous render" as soon as data arrived. Hoisted above the guard.

## CI note

Only the branch-agnostic checks run here — `tests.yml` and friends filter on `pull_request: branches: [main, beta, stable]`, so no stacked PR triggers them. The suites were run against this branch locally as above; I can dispatch the workflow at the branch too if you want a CI record before merge.


the-luap:
## Screenshots

issue-1210-guests-duplicates-en

**The banner and the badges.** Tina registered twice — once as "Tina Ferrarelli" with a verified email, once as plain "Tina" five days later, after her token expired. The numbers are the point: 18 likes and 6 favourites on one row, 12 and 4 on the other. Neither is her selection. Before this, the only way to notice was to read the list and spot the same address twice.

The badge sits on the email because the email is the evidence — the names differ, and two guests called Anna would prove nothing.

issue-1210-guests-merge-preselected

**Review hands the group to the merge that already existed.** Both rows ticked, "2 selected", and nothing merged. Which row survives decides the name and the verification tick the merged guest keeps, so that stays the admin's choice — the banner does the finding, not the deciding.

issue-1210-guests-duplicates-dark

**Dark mode**, same amber pair used elsewhere for consequential-but-not-error notices.

Rendered from this branch against the real API payload shape, mounting the component directly rather than the whole admin shell — the local rig's database predates this code.

PR #1217, opened 2026-08-27
fix(gallery): make the returning-guest recovery findable (#1210)

the-luap:
Part of #1210. The cheap half of what the reporter asked for third: *"surface the 'I've been here before' option more assertively."*

> **Stacked on #1216**, which is stacked on #1214. Merge order: #1214 → #1216 → this.

## Why this and not entry-path verification

The reporter's other suggestion — dedupe at registration by emailing a verification code — I'd hold off on, and #1216 explains why in more depth. Short version: it hard-couples gallery entry to working SMTP (misconfigure your mail server and nobody can get into a gallery), it puts friction on every guest to help the minority of galleries that collect an email at all, and #1216's 30-day token already removes the dominant cause of the duplicates.

This does the part that costs nothing and risks nothing.

## The change

`GuestNamePromptModal` already offered recovery — as a `text-sm` link under the submit button, which reads as fine print. Guests skipped it and registered again, which is how duplicates accumulated *even for guests who had given an email the first time and were eligible for recovery*.

Now it gets its own block under a divider, worded around what the guest loses by missing it:

> Been here before? Your earlier picks are still saved.
> **Get them back**

"I've been here before" reads as a greeting. This reads as a reason to stop.

**Still a choice, not a check.** It does not look up whether the typed address is already registered — that would answer "is this person in this gallery" to anyone who asked, which is exactly why `/guest/recover` always returns 200 and cannot be used that way.

## The retired key

`gallery.guestPrompt.alreadyHere` is removed rather than reworded. A key called `alreadyHere` holding "Get them back" would mislead the next translator. Both new strings (`returningHint`, `recoverPicks`) are in all seven locales that carried the old one; `es` has no `guestPrompt` block at all and is untouched, as before.

## Tests

Three: the hint renders, the affordance routes into recovery rather than registering, and an ordinary first-time registration still just registers — the nudge is an offer beside the form, not a step in front of it.

Frontend suite 246 pass, 39 files. `tsc` and eslint clean.

## Screenshot — owed, not delivered

This one **does** add a visible element, so the repo rule applies and I have not met it. I tried to render the modal in the local rig and stopped after several attempts: reaching it means getting a stubbed gallery past the password gate with enough of the app's endpoint surface mocked to avoid the error boundary, and the rig's own database predates this code.

Happy to bring the rig up properly if you want the screenshot before merge — say the word. Flagging it rather than quietly skipping it, since unlike #1213 (where the change had no visual element) this one genuinely warrants one.

**The same gap applies to #1216**, which adds the duplicate banner and row badges to the admin guests list. I should have called that out there and didn't.


the-luap:
## Screenshots

issue-1210-recovery-nudge-en

**English.** The divider, the line, and the affordance. Previously this was `I've been here before` as a small link directly under the buttons, with no divider and nothing explaining why it mattered — which is why returning guests skipped it and registered again.

It is worded around what the guest loses, not around who they are: "Been here before? Your earlier picks are still saved." A greeting is easy to scroll past; a saved-work notice is not.

issue-1210-recovery-nudge-de

**German**, which is why this got screenshotted rather than trusted to the tests. The dialog addresses the guest as *Sie* throughout — "Willkommen — wie heißen Sie?", "Ihre Auswahl wird unter diesem Namen gespeichert" — and my first draft of the new line came back in *du*. Mixed formality in one modal, invisible in a diff and obvious on screen. Fixed in the last commit.

Both strings are in all seven locales that carried the retired `alreadyHere` key.

Rendered from this branch by mounting the modal with its real provider and stylesheet; the theme variables are the ones the running app sets, read from it rather than invented.

PR #1240, opened 2026-08-29
fix(archives): take the restored category from the manifest

munin92:
---

Description

The archive writer already persists category_name per photo in photos_manifest.json — that is why the manifest exists, and the comment above it says so: "(and category linkage) can't be derived from the extracted files alone". The restore route then read only original_filename out of it and kept deriving the category from the ZIP's first path segment.

Archives store photos exactly as they sit on disk, so an event whose photos live in the gallery root produces a flat zip. path.dirname() is '.' for every entry, no category is resolved, and every restored photo lands with category_id = null — silently, behind a 200.

Seen on a real restore: 596 photos back, 0 with a category, while the nine category rows sat untouched in the tableoriginal_filename out of it and kept deriving the category from the ZIP's first path segment.

Archives store photos exactly as they sit on disk, so an event whose photos live in the gallery root produces a flafor every entry, no category is resolved, and every restored photo lands with category_id = null — silently, behinda 200.

Seen on a real restore: 596 photos back, 0 with a category, while the nine category rows sat untouched in the table

Now the manifest is the source of truth and the first path segment is the fallback, so foldered archives and legacy archives without a manifest behave exactly as before. The find-or-create is pulled into resolveCategoryId so both paths share it and each name is resolved once per restore.

Related to #640 (the manifest itself landed in e4e79a0b3).

Type of change

How Has This Been Tested?

backend/__tests__/integration/adminArchives.restoreCategories.test.js builds real ZIPs with archiver and drives POST /admin/archives/:id/restore against a SQLite database:

1. flat archive with manifest → categories come from category_name
2. flat archive with an existing category row → reused, not duplicated
3. foldered archive without manifest → still resolved from the first path segment

Removing the route change makes cases 1 and 2 fail while 3 keeps passing, so the fallback is demonstrably untouched.

- [x] Unit tests pass (npm test)

Test Configuration:
- PicPeak Version: 3.46.7
- Node.js Version: 24.20.0
- Database: SQLite (tests), PostgreSQL 18.4 (the install this was found on)
- Browser: n/a — backend only

Checklist:

- [ ] I have made corresponding changes to the documentation — none needed
- [x] I have added tests that prove my fix is effective
- [ ] I have updated the CHANGELOG.md file — see note below

Additional Notes

npm test on this tree has 7 pre-existing failing suites (webhookDelivery, storageBackend, adminAuth, backupService. adminPhotos.reference, imageProcessor.storage). I ran the full suite with and without this change: the failing setis byte-identical, so none of them are mine.

src/routes/adminArchives.js:9 has a pre-existing lint error ('archiver' is assigned a value but never used) — left alone so this PR does one thing.

CHANGELOG.md not touched, since the repo generates it via release-please. Happy to add an entry if you'd rather have it in the PR.

### DIFF
--- a/backend/__tests__/integration/adminArchives.restoreCategories.test.js
+++ b/backend/__tests__/integration/adminArchives.restoreCategories.test.js
@@ -1,0 +1,177 @@
+/**
+ * Restoring an archive must put the photos back into their categories.
+ *
+ * The archive writer already persists `category_name` per photo in
+ * `photos_manifest.json` — that is why the manifest exists, and the comment
+ * above it says so: "(and category linkage) can't be derived from the
+ * extracted files alone". The restore route then read only
+ * `original_filename` from it and kept deriving the category from the ZIP's
+ * first path segment.
+ *
+ * Archives store photos exactly as they sit on disk, so an event whose photos
+ * live in the gallery root produces a FLAT zip. `path.dirname()` is '.' for
+ * every entry, no category is resolved, and every restored photo lands with
+ * `category_id = null` — silently, with a 200 response.
+ *
+ * These pin the manifest as the source of truth, with the directory as the
+ * fallback that keeps foldered and legacy archives working.
+ */
+
+const fs = require('fs');
+const path = require('path');
+const os = require('os');
+const express = require('express');
+const request = require('supertest');
+
+describe('archive restore restores categories (flat archives included)', () => {
+  let tmpDir; let db; let cleanup; let app; let storagePath;
+
+  beforeAll(async () => {
+    tmpDir = await fs.promises.mkdtemp(path.join(os.tmpdir(), 'picpeak-restore-cat-'));
+    storagePath = path.join(tmpDir, 'storage');
+    process.env.NODE_ENV = 'test';
+    process.env.TEST_DATABASE_PATH = path.join(tmpDir, 'data', 'test.db');
+    process.env.STORAGE_PATH = storagePath;
+    await fs.promises.mkdir(path.dirname(process.env.TEST_DATABASE_PATH), { recursive: true });
+    await fs.promises.mkdir(path.join(storagePath, 'archives'), { recursive: true });
+
+    jest.resetModules();
+    jest.doMock('../../src/middleware/auth', () => ({
+      adminAuth: (req, _res, next) => { req.admin = { id: 1, username: 'tester' }; next(); },
+    }));
+    jest.doMock('../../src/middleware/permissions', () => ({
+      requirePermission: () => (_req, _res, next) => next(),
+    }));
+    jest.doMock('../../src/middleware/ownership', () => ({
+      requireEventOwnership: (_req, _res, next) => next(),
+    }));
+
+    ({ db, cleanup } = await require('./helpers/crmDb').bootCrmDb());
+    // bootCrmDb points STORAGE_PATH at its own tmp dir; follow it rather than
+    // fighting it, so the archives the tests write are where the route looks.
+    storagePath = process.env.STORAGE_PATH;
+    await fs.promises.mkdir(path.join(storagePath, 'archives'), { recursive: true });
+
+    app = express();
+    app.use(express.json());
+    app.use('/admin/archives', require('../../src/routes/adminArchives'));
+  }, 180000);
+
+  afterAll(async () => {
+    if (cleanup) await cleanup();
+    await fs.promises.rm(tmpDir, { recursive: true, force: true }).catch(() => {});
+  });
+
+  beforeEach(async () => {
+    await db('photos').del();
+    await db('photo_categories').del();
+    await db('events').del();
+  });
+
+  /** A one-pixel JPEG is enough; the route only stats the extracted file. */
+  const PIXEL = Buffer.from(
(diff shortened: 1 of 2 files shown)


the-luap:
Good catch, and a well-built PR — real ZIPs through the real route against a real database, which is the right way to test this and not the easy way. Every claim in the description that I checked held up.

## ✅ Verified against source, not taken from the description

- **The manifest really does carry it.** `archiveService.js:45` selects `photo_categories.name as category_name` and `:50` serialises `manifestRows` straight into the JSON. The write side always knew; the read side never used it. The comment at `adminArchives.js:202` even says so.
- **`archive.id` is the event id** — it is an `events` row — so scoping the find-or-create to it is right.
- **Your fail-without-fix claim is exact.** I ran it: removing the manifest read fails cases 1 and 2 and leaves case 3 passing. Thank you for stating it precisely enough to check.
- **The lint error at `:9` is genuinely pre-existing** — `const archiver = require('archiver')` is byte-identical on `main`. Correct call to leave it.
- The `resolveCategoryId` extraction is behaviour-preserving and still resolves one query per unique name.

## 🟡 One thing to fix before merge

The directory fallback reads the **storage subfolder**, not a category. Archive entry names come from the storage key minus `events/active/{slug}`, and that layout is `individual/{filename}` / `collages/{filename}` — categories have never been directories there.

I built a realistically-shaped archive and restored it rather than reasoning about it:

```
CATEGORIES CREATED: ["individual","collages"]
e.jpg category: individual
f.jpg category: collages
```

And the version that matters for this PR — a manifest that explicitly records a photo as uncategorized:

```
MANIFEST SAID NULL -> got category: individual
```

So the manifest is authoritative for *"category X"* but not for *"no category"*: an uncategorized photo comes back filed under a category named `individual` instead of coming back uncategorized. That is the one remaining case sitting directly on this PR's own premise.

To be clear, **this is pre-existing and not something you introduced** — and your change already makes it far rarer, because any archive with a manifest now takes the manifest's answer. It is just that the last gap lands on the goal.

The fix is small: treat a manifest entry as authoritative including its null, and confine the fallback to photos with no manifest entry at all.

```js
let categoryId = null;
if (manifestEntry) {
  categoryId = await resolveCategoryId(manifestEntry.category_name);
} else if (dirPath && dirPath !== '.') {
  categoryId = await resolveCategoryId(dirPath.split(path.sep)[0]);
}
```

## 🟢 Nits, none blocking

- The third test uses `Drohne/d.jpg`, a shape a real PicPeak archive never produces. It passes, but it blesses a fallback that does not match reality — pointing it at `individual/` would document what actually happens on a legacy archive.
- No unique constraint on `photo_categories(event_id, name)`, so two concurrent restores can still duplicate a category. Pre-existing.
- `resolveCategoryId` runs before the `existingPhoto` check, so re-restoring an archive whose photos already exist can create category rows nothing references. Pre-existing order, marginally more reachable now that flat archives resolve categories at all.

## Next step — your call

Happy either way:

- **You take it** — the snippet above plus flipping the third test to `individual/` is the whole change, and I will re-review as soon as it is pushed.
- **I take it** — say the word and I will push the change to this branch (or land it as a follow-up commit on merge), with you keeping authorship of the PR.

No preference from my side; whichever gets it in sooner. **Ready to merge as soon as that one is in** — everything else here is done.

Thanks for the report quality too: the reproduction, the pre-existing-failures note, and the "I ran the suite with and without" check are exactly what makes a fix like this quick to trust.

munin92:
@the-luap 
Thanks for the Feedback. 

You can take it and push the change. Its fine for me

the-luap:
Pushed as `b4fb8ef0` — thanks for handing it over. Your commit is untouched, so the PR is still yours.

## What the commit does

Confines the directory fallback to photos with **no manifest entry at all**, so a manifest entry is authoritative including when its `category_name` is null:

```js
let categoryId = null;
if (manifestEntry) {
  categoryId = await resolveCategoryId(manifestEntry.category_name);
} else if (dirPath && dirPath !== '.') {
  categoryId = await resolveCategoryId(dirPath.split(path.sep)[0]);
}
```

## Two test changes

**The legacy case now uses `individual/`** rather than `Drohne/`. That is the shape a real archive actually has — entry names are the storage key minus `events/active/{slug}` — so the test now documents what the fallback genuinely does on a legacy archive (invents a category called `individual`) instead of a category-shaped folder no archive produces. It is not a pretty outcome, but it beats losing every category on an archive that has no manifest, and pinning the real behaviour is more useful than pinning a hypothetical.

**A new case** pins that a manifest recording a photo as uncategorized leaves it uncategorized *and* creates no category row as a side effect.

Verified the same way you verified yours: reverting just the route change fails the new case while the legacy fallback keeps passing.

```
✕ honours a manifest that says UNCATEGORIZED, instead of inventing one from the directory
✓ still falls back to the directory for legacy archives with no manifest
```

4 tests green, CI green, and the `archiver` lint error at `:9` is still the pre-existing one — left alone as you had it.

## Untouched, as flagged

The two remaining nits are pre-existing and out of scope for this PR: no unique constraint on `photo_categories(event_id, name)` so concurrent restores can still duplicate, and `resolveCategoryId` running before the `existingPhoto` check so a re-restore can create category rows nothing references.

Merging now. Thanks for the fix and for the report quality — the checkable fail-without-fix claim made this quick from the first read.

the-luap (approved):
Manifest-authoritative fix pushed and verified. Merging.

PR #1302, opened 2026-09-05
feat(newsletters): warn about deliverability before a large send

the-luap:
The composer already paces sends — the rate is clamped to what the queue can actually drain and `scheduled_at` is staggered per recipient — and the help text said as much. But pacing answers the wrong risk.

**Spam filtering reacts to a domain's volume and reputation, not to the interval between messages.** A perfectly throttled send of several hundred near-identical mails, from a domain that normally emits only gallery notifications, is exactly the shape that gets junked or blocked. Nothing in the UI told the operator that, and the one hint that existed (`newsletters.rateHelp`) talked only about provider rate limits — which is the problem we had already solved.

## What this adds

A warning above the queue button once the **server-resolved** recipient count reaches 50:

- what actually goes wrong — reputation, and that providers may throttle, junk or block the whole batch
- that a bad run damages delivery of gallery email too, not just the campaign
- the SPF/DKIM/DMARC prerequisite
- the suggestion to split a first campaign across smaller sends
- the real duration at the chosen rate, **and that other email queues behind it** — the email queue is global (`processEmailQueue` takes 10 rows a minute across all types), so a long campaign delays gallery invitations and password resets. That consequence was previously invisible.

The existing rate hint now mentions looking like spam, not only being rate-limited.

## Why 50

Deliberately low. It is not a provider limit — the queue's own pacing handles rate. The operators who most need this warning are the ones sending their first campaign from a domain with no bulk-sending history, and that is exactly when a few dozen messages is enough to matter. The constant carries this reasoning in a comment.

## Why a warning rather than a hard cap

Chunked sending is already implemented and enforced server-side; the rate is clamped to 1–10/min regardless of what the client sends. What was missing was informed consent, not another limit. Blocking a legitimate 300-person send would be wrong — the operator may well have a warmed-up domain.

## Verification

- `npx tsc -b` — 0 errors
- `npx eslint` on both changed files — 0 errors
- `npm run test` — 76 files, 551 tests passing, including two new cases: the warning appears at 120 recipients with the duration and queue-contention text, and is absent at 12
- EN and DE strings added; both locale files re-validated as JSON
- E2E smoke suite green on push (13 passed)

## Screenshots

Both at the same framing, so the difference is just the warning. Real data from the local rig — 63 seeded recipients vs 12, at the default 10/min.

1302-large-send-warning**63 recipients — warning shown**, directly above "Queue campaign". Note it reports the true duration (~7 minutes at 10/min) and that other email queues behind it.

1302-small-send-no-warning**12 recipients — no warning.** It is conditional, not permanent chrome.

The send-rate help text visible in both shots is the reworded one.

## Note on the local rig

The rig had no `newsletters` row in `feature_flags` at all, and `clients` was off, so the feature was unreachable there. I enabled both, seeded 63 marked customers and one draft campaign to reach the threshold, took the shots, then deleted all of it and put both flags back exactly as they were (`clients=false`, `newsletters` row removed). Verified against counts recorded before seeding: 1 customer account, 0 campaigns, 0 campaign recipients.

### DIFF
--- a/frontend/src/i18n/locales/de.json
+++ b/frontend/src/i18n/locales/de.json
@@ -6978,7 +6978,7 @@
       "title": "Großer Versand — prüfen Sie zuerst Ihre Versandreputation",
       "body": "{{count}} Personen auf einmal von einer Domain anzuschreiben, die sonst nur Transaktionsmails versendet, lässt Spamfilter aufmerksam werden. Anbieter können den gesamten Versand drosseln, als Spam einstufen oder blockieren — und ein schlechter Lauf beeinträchtigt auch die Zustellung Ihrer Galerie-E-Mails.",
       "advice": "Stellen Sie sicher, dass SPF, DKIM und DMARC für Ihre Versanddomain eingerichtet sind, senden Sie sich zuerst einen Test, und teilen Sie eine erste Kampagne nach Möglichkeit in mehrere kleinere Sendungen auf.",
-      "duration": "Bei {{rate}}/Minute dauert dies etwa {{minutes}} Minuten; andere E-Mails — Galerie-Einladungen, Passwort-Zurücksetzungen — warten so lange in der Warteschlange."
+      "duration": "Bei {{rate}}/Minute dauert dies etwa {{minutes}} Minuten. Die Versand-Warteschlange wird geteilt: Während der Kampagne können andere E-Mails — Galerie-Einladungen, Passwort-Zurücksetzungen — dahinter verzögert werden."
     }
   }
 }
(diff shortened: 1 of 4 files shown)


the-luap:
Screenshots are in the working tree (git-excluded, so not in the diff):

- `pr-screenshots/1302-large-send-warning.png` — 63 recipients, warning shown above "Queue campaign"
- `pr-screenshots/1302-small-send-no-warning.png` — 12 recipients, no warning

Same scroll position and window size in both, so they line up as a before/after. Drop them into the PR body where the two placeholders are.

Issue #1287, opened 2026-09-03
[BUG] Large gallery on mobile (iOS Safari): tiles fail to load, blank page after refresh


leonlivevocalist-svg:
> **This is a symptom report, not a diagnosis.** I could not isolate the cause and I would rather say so than send you after a theory I have already disproved three times. What follows is what reproduces, what I measured, and — probably more useful — what I have ruled out with evidence.

**Describe the bug**

On a large gallery (546 photos), a subset of grid tiles never render, and they do not recover. Loading advances in bursts while scrolling and then stops; the unrendered tiles stay blank indefinitely, through further scrolling and while stationary.

There is no error anywhere: nothing in the browser console, no failed network request, nothing in the backend logs.

For the client this reads as "the gallery is broken". The report that started this was: *"every time I scroll down it doesn't show all the photos as in they haven't loaded, so I refresh the page and it just goes blank."*

**To Reproduce**

1. Create an event with several hundred photos (this one has 546)
2. Open the gallery as a guest and scroll steadily to the bottom
3. Some tiles load; a set of them never do
4. Stop scrolling and wait — the outstanding tiles do not load
5. Scroll a little in either direction — they still do not load

**Expected behavior**

Every tile eventually renders, and any that has not loaded is retried when it is scrolled back into view.

**Screenshots**

Not applicable — the failure is blank tiles and stalled loading rather than anything rendered incorrectly.

**Environment (please complete the following information):**
 - OS: Ubuntu 24.04 (server); clients on Windows 11 and iOS 18.1.1
 - Browser: Chrome 152 (reproduced); originally reported on iOS 18.1.1 Safari
 - PicPeak Version: v3.122.7-beta.0
 - Deployment Method: Docker Compose, behind Caddy as a plain reverse proxy
 - Database: PostgreSQL

**Logs**

None to give, which is part of the problem. `docker compose logs backend` shows nothing during the failure, the browser console is clean, and no request appears as failed.

**What I ruled out (with evidence)**

- **Not the server.** Thumbnail requests average ~190 ms each sequentially (12 samples, 174–219 ms), and twelve issued in parallel complete in 209 ms of wall clock — so it parallelises properly and is nowhere near saturated.
- **Not image size.** The grid uses correctly-sized thumbnails: 400×600 natural, displayed at 286×358, ~60 KB per file. It is not pulling full-size images.
- **Not a blob leak.** Images are delivered as `blob:` URLs; those blobs are properly revoked after load.
[...]

PR #1303, opened 2026-09-05
fix: remove the image-fragmentation surface

### LINKED ISSUE #1300: [BUG] fragmentation_level is configurable, stored, and consumed by nothing
**Describe the bug**

`fragmentation_level` is configurable in Settings → Image security ("1-10, higher = more protection"), stored per event, and served to the gallery client — and no code anywhere multiplies by it. The number is decorative on both sides of the wire.

**The chain, end to end**

Configuration and storage work:

- `ImageSecurityTab.tsx:167-177` renders the 1-10 control with the help text above
- migration `038_add_enhanced_image_protection.js:15` creates `events.fragmentation_level` (default 3) and seeds `default_fragmentation_level` at :81
- `gallery.js:1230` serves `fragmentation_level: req.event.fragmentation_level || 3` in the gallery's `protectionSettings`

Consumption does not:

- **Server** — `secureImageService.fragmentImageBuffer` hardcodes the grid:
  ```js
  // Create 3x3 grid of fragments
  const cols = 3;
  const rows = 3;
  ```
  `secureImageService.js:293-295`. The event's level never reaches it. It is called from `processProtectedImage` only when `protectionLevel === 'maximum'`.
- **Client** — `fragmentation_level` appears exactly once, as a type declaration at `types/index.ts:293`. Nothing reads it. The only component that could reassemble fragments is `ProtectedImage`, which hardcodes `gridSize = 4` (`ProtectedImage.tsx:30`), is exported from `components/common/index.ts:27`, and is rendered nowhere.

So an operator setting 1 or 10 gets byte-identical output, and the two hardcoded grids (3 server, 4 client) do not even agree with each other.

**Why it matters**

This is a security control that reports protection it does not provide. A dead *feature* is a backlog item; a dead *security* control is worse than an absent one, because an operator who raises it believes they have hardened the gallery and stops looking.

**Related, worth deciding together**

The routes that would consume this are themselves unreachable from the app: `secureToken.service.ts` is imported by nothing, so `/api/secure-images/...` is never called by the client. If that path is ever wired up, note that `processProtectedImage` returns `{ type: 'fragmented', fragments: [...] }` as **JSON** rather than image bytes (`protectedImages.js:155-166`) — a caller expecting an image gets a JSON body. Nothing hits that today.

Same class as #1297 (a prop surface accepted and discarded), one layer further down: there the props were dead, here the whole feature is.

**What to do**

Three honest options, in the order I would take them:

1. **Stop advertising it** (immediate) — disable or clearly mark the Settings control so the "higher = more protection" promise is not made. No behaviour change; removes the misleading claim.
2. **Remove the surface** — drop the control, the `gallery.js` field, and the unreachable fragmenter. Correct if fragmentation is not on the roadmap.
3. **Finish it** — make the level drive the grid *and* build the client-side reassembly, with server and client agreeing on the grid. This is a feature with real design questions (what does level 7 mean? what is the performance cost of a 10x10 reassembly on mobile, given #1287?), not a cleanup.

Option 1 is being done now; 2 vs 3 is a product decision.

**Version**

Present on `main` as of 3.123.0-beta.0. Surfaced while reviewing the #1296 fix, which made these settings apply at creation time — `fragmentation_level` is the one of the four that still lands nowhere.

the-luap:
Closes #1300.

`fragmentation_level` was configurable in Settings → Image security ("1-10, higher = more protection"), stored per event, served to the gallery client — and consumed by nothing.

## Correcting the issue on one point

#1300 called this dead on both sides. That was too strong, and it matters for the decision. Both halves exist and are individually coherent:

- **Server**: `secureImages.js` has a real fragment API — `?fragment=N` serves an individual fragment, omitting it returns metadata (`handleFragmentedImage`)
- **Client**: `ProtectedImage.tsx` has a real canvas reassembly path, `scrambleFragments` and all

They were never connected — `ProtectedImage` is rendered nowhere and `secureToken.service.ts` is imported by nothing — and **they disagree**: the server cut a fixed 3×3 grid, the client reassembled a 4×4 one. Wired together as they stood, they would have produced scrambled images rather than protection.

## Why remove rather than finish

Finishing it buys nothing that is worth the cost:

1. **It is not protection.** The client fetches the whole image, then redraws it in pieces on a canvas. The full original has already crossed the wire before any "protection" is applied. That is obfuscation of the rendering, not a control on the asset.
2. **The cost lands on mobile.** Per-fragment canvas work on a several-hundred-photo gallery is precisely the memory profile under investigation in #1287.
3. **The number was never the hard part.** Threading `fragmentation_level` through is small; building and reconciling the client is the actual work, and it would be new mobile canvas code shipped against an open memory bug.

A dead feature is a backlog item. A dead *security* control is worse than an absent one — an operator who raises it believes the gallery is hardened and stops looking.

## What goes

**Backend** — `secureImageService.fragmentImageBuffer` and the branch that called it; the `?fragment=N` path and `handleFragmentedImage` in `secureImages.js`; the fragmented-JSON response in `protectedImages.js`; `fragmentation_level` from the gallery `protectionSettings` payload; `default_fragmentation_level` from the image-security settings allow-list; the `PUT /:id` validator.

**Frontend** — the operator control and its strings in all eight locales; `ProtectedImage`'s fragment renderer and its props; `FragmentConfig` and the fragment fields in `types/protection.ts`; `fragmentation_level` from the gallery type; `fragmentGrid` from `useImageProtection`.

Two obsolete `ProtectedImage` tests go with it — they asserted `drawImage` was called, which stayed true after fragmentation was gone, so they were pinning nothing and had started leaking removed props onto the DOM.

## What deliberately stays

**No migration.** `events.fragmentation_level` and the `default_fragmentation_level` row remain. Dropping a column is irreversible and the stored values are harmless once nothing reads them; if they should go, that is a deliberate data decision deserving its own migration.

**`fragmentGrid` on `AuthenticatedImage` and the gallery layouts.** #1299 already removes it as part of the inert prop surface. Touching it here would only collide.

## Relationship to the other open PRs

- **#1301 is superseded** — it disabled this control; this removes it. Closing it in favour of this PR.
- **#1298 adds `fragmentation_level` to the creation defaults.** They do not conflict textually, but whichever merges second leaves the other's fragmentation handling stranded. If #1298 goes in first, I will strip fragmentation from its resolver as a follow-up; if this goes first, the same edit applies before #1298 merges. Flagging rather than pre-emptively coupling the two branches.

## Screenshot

1300-fragmentation-removed**Settings → Image security after** — the control is gone; Default Image Quality keeps its half-width column.

## Verification

- `npx tsc -b` — 0 errors
- `npx eslint` on every changed file — 0 errors (2 pre-existing unused-catch warnings in `useImageProtection.ts`, untouched)
- Frontend: 76 files, 547 tests passing
- Backend: 279 suites, 2987 tests passing, 0 failures
- All 8 locale files re-validated as JSON
- Confirmed in the running admin UI that the control no longer renders and the tab layout holds
- Grepped the tree: no `fragmentation_level` / `fragmentImage` / `handleFragmentedImage` / `scrambleFragments` outside migrations, and the only remaining `fragmentGrid` references are the ones #1299 owns


the-luap:
Screenshot for this PR (git-excluded, so not in the diff):

- `pr-screenshots/1300-fragmentation-removed.png` — Settings → Image security with the control removed

For a before/after, `pr-screenshots/1300-fragmentation-before.png` from the now-closed #1301 shows the control as it is on main.

PR #1401, opened 2026-09-11
fix(upload): let the csrf gate pass application/octet-stream chunks

### LINKED ISSUE #1377: [BUG] chunked upload always returns 415, the endpoint has never accepted a chunk
**Describe the bug**

The chunked upload endpoints cannot accept a chunk. The CSRF gate mounted on `/api` rejects any mutating request whose body is not `application/json` or `multipart/form-data`, and the only client sends `application/octet-stream`, so every chunk POST is answered 415 before it reaches the route and before authentication.

- gate: `backend/src/middleware/csrf.js` (was inline in `backend/server.js` before 3.13x)
- route reads the raw request stream: `backend/src/routes/adminPhotos.js`, `for await (const chunk of req)`
- client: `frontend/src/services/photos.service.ts`, `'Content-Type': 'application/octet-stream'`

**To Reproduce**

Against the backend directly, so no proxy is involved. No credentials needed, the gate runs ahead of `adminAuth`:

```
$ curl -s -X POST -H 'Content-Type: application/octet-stream' --data-binary 'chunkbytes' \
    http://127.0.0.1:3000/api/admin/photos/1/chunked-upload/abc/chunk/0
{"error":"Unsupported Content-Type. Use application/json or multipart/form-data."}   # 415

$ curl -s -X POST -H 'Content-Type: application/json' -d '{}' \
    http://127.0.0.1:3000/api/admin/photos/1/chunked-upload/abc/chunk/0
{"error":"No token provided"}                                                        # 401
```

The 415 is the API's own error and it lands before auth, so no valid token changes the outcome.

**Expected behavior**

The chunk is accepted and written.

**Environment**

- OS and version: Ubuntu 24.04
- Browser and version: n/a, reproduced with a plain HTTP client
- PicPeak version and Docker image tag: `ghcr.io/picpeak/picpeak/backend:stable` at v3.45.15, and confirmed by reading the code on `main` (3.131.3-beta.0)
- Deployment method: Docker Compose
- Database and version: PostgreSQL 15

**Logs**

None. The request never reaches the route, so nothing is logged. On a deployment with 48 days of logs there is not one chunked upload request, and `storage/chunks/` was never created.

**Additional context**

No client side workaround exists:

- `multipart/form-data` passes the gate but the route reads the raw stream, so the MIME boundary lands in the chunk. Sending a 1024 byte chunk this way delivers 1247 bytes starting with `--`.
- `application/json` is consumed by `express.json` before the route.
- `Transfer-Encoding: chunked` with no `Content-Length` used to slip past the length check. On `main` the gate counts `transfer-encoding` as a body, so that is closed too.

`uploadLargeFile` in `frontend/src/services/photos.service.ts` has no callers, so the web UI has never exercised this path. That fits an API that shipped broken rather than a regression.

**Possible Solution**

Either mount the chunked upload routes ahead of the gate, the way `/api/analytics/tracker` already is in `server.js`, or change the chunk route to accept `multipart/form-data` and read the part with multer. The first keeps the existing client working.

AtelyPham:
## Description

The chunked upload endpoint has never accepted a chunk. The client posts each chunk as `application/octet-stream` and the route reads the raw request stream, but the CSRF gate on `/api` answers any non-JSON, non-multipart body with 415 before the request reaches the route or auth.

The gate's origin check is the CSRF defence. Its Content-Type list only has to keep out what a cross-site page can send without a preflight, and octet-stream is not one of those: an HTML form cannot produce it, and `fetch()` or `sendBeacon` with it is not CORS-safelisted, so the browser preflights and the origin check rejects it. This change adds `application/octet-stream` to the list. `express.json` leaves an octet-stream body unread, so JSON routes see `{}` as they do for a bodiless request.

The issue suggested mounting the chunk routes ahead of the gate. That also skips the origin check, the request logger, the maintenance gate and the session timeout for those routes, so I widened the gate instead. Client untouched.

Fixes PicPeak/picpeak#1377

## Type of change

## How Has This Been Tested?

`backend/__tests__/middleware/csrfChunkedUpload.test.js`: same-origin octet-stream reaches a raw-stream route byte for byte (415 on `main`), cross-site octet-stream is still 403, `text/plain` is still 415, a JSON route sees an empty body on an octet-stream post. `galleryAccessPolicy` and `securityHardeningBatch2` still pass.

Manual, against this branch on SQLite: the issue's curl now gets 401 from auth instead of 415. Logged in, `init` then two octet-stream chunks then `complete` for a 73 byte PNG. Both chunks 200, `complete` created the photo, stored file is byte-identical to the source.

**Test Configuration**:
* PicPeak Version: main (3.131.3-beta.0)
* Node.js Version: 24
* Database: SQLite
* Browser: n/a, curl

## Checklist:

- [ ] I have made corresponding changes to the documentation

## Additional Notes:

Now that chunks arrive, `adminPhotos.js` buffers each whole chunk in memory before `uploadChunk` applies the size cap. Behind `adminAuth` and `photos.upload`, pre-existing, not changed here.

### DIFF
--- a/backend/__tests__/middleware/csrfChunkedUpload.test.js
+++ b/backend/__tests__/middleware/csrfChunkedUpload.test.js
@@ -1,0 +1,75 @@
+/**
+ * The chunked upload client posts each chunk as application/octet-stream and
+ * the route reads the raw request stream. The CSRF gate answered every such
+ * request 415 before it reached the route, so the endpoint never accepted a
+ * chunk (PicPeak/picpeak#1377).
+ *
+ * The gate's origin check is the CSRF defence. The Content-Type list only
+ * has to keep out what a cross-site page can send without a preflight, and
+ * application/octet-stream is not on that list: an HTML form cannot produce
+ * it and fetch() with it is not CORS-safelisted.
+ */
+const express = require('express');
+const request = require('supertest');
+
+jest.mock('../../src/utils/logger', () => ({ error: jest.fn(), warn: jest.fn(), info: jest.fn(), debug: jest.fn() }));
+
+function buildApp() {
+  const app = express();
+  // Same order as server.js: scoped JSON parser, then the gate on /api.
+  app.use(['/api/admin', '/api/v1'], express.json({ limit: '50mb' }));
+  app.use(express.json({ limit: '2mb' }));
+  app.use('/api', require('../../src/middleware/csrf'));
+  // Mirrors the chunk route in adminPhotos.js: consume the raw stream.
+  app.post('/api/admin/photos/:eventId/chunked-upload/:uploadId/chunk/:chunkIndex', async (req, res) => {
+    const chunks = [];
+    for await (const chunk of req) chunks.push(chunk);
+    res.json({ received: Buffer.concat(chunks).toString('base64') });
+  });
+  app.post('/api/admin/other', (req, res) => res.json({ body: req.body }));
+  return app;
+}
+
+const CHUNK_PATH = '/api/admin/photos/1/chunked-upload/abc/chunk/0';
+
+describe('CSRF gate and application/octet-stream', () => {
+  it('lets a same-origin octet-stream chunk reach the route byte for byte', async () => {
+    // Not valid UTF-8, so a text decode anywhere on the path would show up.
+    const payload = Buffer.concat([Buffer.from('chunkbytes'), Buffer.from([0xff, 0x00, 0xfe])]);
+    const res = await request(buildApp())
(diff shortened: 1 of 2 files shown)


the-luap (approved):
Reviewed and verified earlier: the octet-stream exemption is scoped to the CSRF gate's Content-Type allowlist and leaves the primary origin check untouched, and declining the issue's suggested route-remount shortcut was the right call.

Merging now that the streaming and size-cap hardening it makes reachable has landed ahead of it, so the chunk route is bounded before this opens it to traffic.

PR #1417, opened 2026-09-11
fix(video): regenerate missing thumbnails for already-uploaded videos

### LINKED ISSUE #1414: [BUG] - Video thumbnail missing for old uploaded videos
**Describe the bug**
The thumbnail of a video is missing and it is not showing in the gallery

**To Reproduce**
1. upload a video
2. open any gallery that includes that video (preview, customer view, guest view) and you'll see the video present but not the thumbnail 

**Expected behavior**
I'd expect to have the thumbnails on the video as for the phots.

**Screenshots**

Image

 - OS and version:
 - Browser and version:
 - PicPeak version and Docker image tag (if applicable):
Frontend: [v3.131.3-beta.0](https://github.com/PicPeak/picpeak/releases/tag/v3.131.3-beta.0)
Backend: [v3.131.3-beta.0](https://github.com/PicPeak/picpeak/releases/tag/v3.131.3-beta.0)
PICPEAK_CHANNEL=main
 - Deployment method: [Docker Composel]
 - Database and version: [PostgreSQ]

**Logs**
Please include relevant logs:

Image

Image

# Backend logs (Docker Compose)
docker compose logs --tail=50 backend
nothing here

# Frontend console errors
[paste any browser console errors]
not applicable

**Additional context**

**Possible Solution**
If you have an idea how to fix the issue, please describe it here.

shreyanshj10:
## Description

Videos uploaded before the upload pipeline learned to fall back to a placeholder thumbnail still show up in galleries without one — a broken tile — while a video uploaded today gets its poster frame fine. Two things kept those older rows from ever recovering:

1. `ensureThumbnail()` is the path that repairs a row whose thumbnail is missing or unreadable (the gallery thumbnail route, the admin "regenerate thumbnails" button and the OG image builder all go through it), but it had no video branch: it resolved the source and handed it to Sharp, which throws on an mp4. So regeneration could never produce anything for a video. I gave it a video branch that runs the same `processUploadedVideo()` a fresh upload uses, so it gets the real poster frame, and the existing SVG placeholder fallback when ffmpeg cannot read the file. External/reference videos keep the per-photo `ext<id>_` key prefix the image branch already uses, and the key otherwise matches the upload pipeline's so a regeneration overwrites the old rendition instead of orphaning it.

2. The gallery payload only emitted `thumbnail_url` when `thumbnail_path` was already set, so for these rows the frontend never even called the route that would have repaired them — it fell back to `thumbnail_url || url` and rendered the original video file into an ``. That is the broken tile in the report, and it also means opening the gallery downloads the whole video. Videos now always get the thumbnail URL; still images keep the old behaviour, since for them the original is a usable fallback.

No migration or backfill needed — the first view of an affected gallery regenerates and persists the thumbnail.

Fixes #1414

## Type of change

## How Has This Been Tested?

- `backend/src/services/__tests__/imageProcessor.videoThumbnail.test.js` (new): covers a managed video with no thumbnail, a legacy row identified only by `mime_type`, an external video's key, the failure path returning `null` rather than throwing, and a still image staying on the image path. Three of the five fail on the unpatched code.
- `backend/__tests__/routes/galleryProtectedVideoUrls.test.js`: two assertions on the gallery payload — a video with no `thumbnail_path` is still offered the thumbnail route, a still image is not.

**Test Configuration**:
* Database: SQLite / PostgreSQL
* Browser: n/a (backend only)

## Checklist:

- [x] I have added tests that prove my fix is effective

## Additional Notes:

Backend only, so no screenshot. `CHANGELOG.md` is left to release-please as CONTRIBUTING describes.

### DIFF
--- a/backend/__tests__/routes/galleryProtectedVideoUrls.test.js
+++ b/backend/__tests__/routes/galleryProtectedVideoUrls.test.js
@@ -160,4 +160,19 @@ describe('videos stay playable under enhanced/maximum protection (#1370)', () =>
       expect((await photoPayload(imageId)).url).toBe(`/api/gallery/${SLUG}/photo/${imageId}`);
     });
   });
+
+  // Both rows here were seeded with no thumbnail_path, which is the state
+  // every video uploaded before the pipeline gained its placeholder fallback
+  // is still in.
+  describe('a video with no stored thumbnail (#1414)', () => {
+    test('is still offered the thumbnail route, which regenerates it lazily', async () => {
+      const photo = await photoPayload(videoId);
+      expect(photo.thumbnail_url).toBe(`/api/gallery/${SLUG}/thumbnail/${videoId}`);
+    });
+
+    test('while a still image keeps falling back to its original', async () => {
+      const photo = await photoPayload(imageId);
+      expect(photo.thumbnail_url).toBeNull();
+    });
+  });
 });
--- a/backend/src/services/__tests__/imageProcessor.videoThumbnail.test.js
+++ b/backend/src/services/__tests__/imageProcessor.videoThumbnail.test.js
@@ -1,0 +1,131 @@
+/**
+ * Lazy thumbnail regeneration for videos (#1414).
+ *
+ * ensureThumbnail is the one path that repairs a photo row whose thumbnail is
+ * missing or unreadable — the gallery thumbnail route, the admin "regenerate
+ * thumbnails" button and the OG-image builder all go through it. It had no
+ * video branch, so it resolved the source and handed it to Sharp, which throws
+ * on an mp4. Videos that reached that state — uploaded before the upload
+ * pipeline learned to fall back to a placeholder, or with their rendition
+ * since lost — could never get a thumbnail back, however many times they were
+ * viewed or regenerated.
+ *
+ * They now go through the same processUploadedVideo a fresh upload uses, so
+ * regeneration yields the poster frame (or, if ffmpeg cannot read the file,
+ * the same SVG placeholder the upload path falls back to).
+ */
+
+jest.mock('../../utils/logger');
+jest.mock('../../database/db', () => {
+  const photosUpdate = jest.fn().mockResolvedValue(1);
+  const db = jest.fn((table) => {
+    if (table === 'events') {
+      return { where: () => ({ first: () => Promise.resolve({ id: 7, slug: 'summer-wedding' }) }) };
+    }
+    if (table === 'photos') {
+      return { where: () => ({ update: photosUpdate }) };
+    }
+    throw new Error(`unexpected table: ${table}`);
+  });
+  db.photosUpdate = photosUpdate;
+  return { db };
+});
+jest.mock('../storage', () => ({
+  getStorage: () => ({
+    kind: () => 'local',
+    resolveLocalPath: (key) => `/storage/${key}`
+  })
+}));
+jest.mock('../videoProcessor', () => ({
+  processUploadedVideo: jest.fn()
+}));
+jest.mock('../photoResolver', () => ({
+  resolvePhotoStorageKey: jest.fn(),
+  resolvePhotoFilePath: jest.fn()
+}));
+
+const { db } = require('../../database/db');
+const { processUploadedVideo } = require('../videoProcessor');
+const { resolvePhotoStorageKey, resolvePhotoFilePath } = require('../photoResolver');
+const { ensureThumbnail } = require('../imageProcessor');
+
+const managedVideo = {
+  id: 42,
+  event_id: 7,
+  filename: 'clip.mp4',
+  path: 'summer-wedding/individual/clip.mp4',
+  media_type: 'video',
+  mime_type: 'video/mp4',
+  thumbnail_path: null
+};
+
+describe('ensureThumbnail rebuilds a video thumbnail from the video (#1414)', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+    resolvePhotoStorageKey.mockReturnValue('events/active/summer-wedding/individual/clip.mp4');
+    resolvePhotoFilePath.mockReturnValue('/mnt/nas/2026/clip.mp4');
+    processUploadedVideo.mockImplementation(async (_videoPath, thumbnailKey) => ({
+      success: true,
+      metadata: { duration: 12 },
+      thumbnailKey
+    }));
+  });
+
+  it('generates a poster frame for a managed video that has no thumbnail yet', async () => {
+    const result = await ensureThumbnail(managedVideo);
+
+    expect(processUploadedVideo).toHaveBeenCalledWith(
+      '/storage/events/active/summer-wedding/individual/clip.mp4',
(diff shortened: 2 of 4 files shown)


shreyanshj10:
Thanks for running both halves by hand — the placeholder-key check saved me a round.

**1. Sizing.** Worth bounding, agreed, so I've taken your first option. On a non-local backend `regenerateVideoThumbnail` now does one `storage.stat()` — a HEAD, no body — before `withLocalCopy`, and a source over the limit goes straight to the same SVG placeholder the upload path falls back to, under the same key, with nothing downloaded. Local and external sources skip the check, since nothing is copied for them anyway.

The limit is `VIDEO_THUMBNAIL_MAX_SOURCE_BYTES`, read at call time like `RESTORE_MAX_DECOMPRESSED_BYTES` in `restoreService`, documented in `.env.example`. Default is 512 MB: that covers an ordinary phone or camera clip end to end, so the common case still gets a real poster frame, while the multi-GB files behind your concern never touch the request path.

**2. Admin regenerate.** You're right — the route filters videos out before `ensureThumbnail`, so the button never reaches this fix. I have the filter removal and a test for it ready, but I'm keeping it out of this PR so the diff you've already verified stays additive and review-sized. I'll open it as a follow-up the moment this lands, unless you'd rather take it yourself.

On your one condition there: the loop already wraps each photo in its own try/catch, so an unreadable video is logged, counted as an error and skipped — one bad file can't take the batch down.

**3. Backport.** Happy either way — take it yourself, or I'll open it once this lands.

the-luap:
Re-verified `b69d8b69`. The bound is the shape I asked for and it lands where it should. Nothing blocking.

## Concern 1 (sizing) — closed

| Check | Where | |
|---|---|---|
| Bound runs before any body is fetched | `imageProcessor.js:675-693` sits ahead of `withLocalCopy` at `:695` | ✅ |
| It really is a HEAD | `S3StorageBackend.js:96-109` — `HeadObjectCommand`, returns `{ size, mtime }`, `null` on 404 | ✅ |
| Missing object / HEAD error does not download | `null` falls through to `withLocalCopy`, which fails into the existing `catch` → `null`; any other HEAD error throws into the same `catch` | ✅ |
| Placeholder lands under the same key | `:687` strips `thumb_` and `generateVideoPlaceholder` re-derives `thumbnails/thumb_<name>.jpg`; holds for dotted basenames (`clip.final.mp4` → `thumb_clip.final.jpg` both ways) | ✅ |
| Explicit dimensions skip the settings query | `:688-691` — matters, that lookup is the one that deadlocks inside a SQLite transaction | ✅ |
| Local and external exempt | `:664-669` returns before the check; `kind() === 'local'` skips it | ✅ |
| Env read at call time, bad values fall back | `:677-680` — non-numeric, `0` and negatives all resolve to the 512 MB default | ✅ |

Tests: 18/18 across the two suites, lint clean. Revert check — putting `imageProcessor.js` back to `7c24fa5f` fails exactly one test, the over-bound one. The under-bound test passes on both, which is correct: the old code downloaded too. Note the over-bound test runs the real `generateVideoPlaceholder` through real Sharp and only mocks storage, so it proves a JPEG buffer is produced, not just that a branch was taken.

CI: I approved the fork runs for the new commit. `backend`, `frontend`, `e2e-smoke`, `fresh-install`, `smoke-aio`, `upgrade-from-bootstrap`, both `nginx` jobs and all six image builds pass.

## Concern 2 (admin regenerate) — agreed, follow-up

Keeping it out of this PR is the right call. Your try/catch claim holds: `adminThumbnails.js:183-240` wraps each photo, counts the failure and moves on.

One thing the follow-up has to handle, and it is a consequence of this commit, so flagging it now rather than in that review:

`adminThumbnails.js:209` calls `ensureThumbnail({ ...photo, thumbnail_path: null }, { force: true })` for **every** row, not just broken ones. Once the video filter is gone, that goes through the new bound. On an S3 install, a video over 512 MB that got a real poster frame at upload time (the upload path has the file locally, so no bound applies there) would have that poster **replaced by the placeholder** the first time an admin presses "Regenerate thumbnails". That is a regression the button does not have today, because today it skips videos entirely.

The bound is justified by the guest request path. An admin-initiated batch is not that path. So in the follow-up, either let the admin route opt out of the bound (an option threaded through `ensureThumbnail` → `regenerateVideoThumbnail`), or skip videos that already have a valid thumbnail. I'd take the first.

## 🟢 Nits — none need a change here

- The bound is per object, not per gallery. There is no cross-video cap, so a cold gallery of twenty 400 MB clips can still run twenty concurrent downloads into `os.tmpdir()`. One-time, and the same shape already exists for images at a smaller scale. A process-wide limit of 1–2 concurrent video regenerations would close it; fine as part of the follow-up or not at all.
- An over-limit row persists the placeholder via the `thumbnail_path` write, so raising `VIDEO_THUMBNAIL_MAX_SOURCE_BYTES` later does not heal rows already placeholdered. Once the admin button reaches videos, that becomes the way to heal them, which is one more reason for the opt-out above.
- The variable is in `.env.example` but not on the docs site's environment variables page. I'll add it there.

## Verified vs. trusted

Directly inspected: everything in the table, the admin loop, the S3 `stat` contract, both test suites run locally plus the revert check. Not exercised: a real S3 or MinIO backend with an object over the limit — the HEAD-then-placeholder path is covered by the mocked-storage test and by reading `S3StorageBackend.stat`, not by an end-to-end run.

## Verdict

From my side, merge cleared. I'll take the stable backport once this lands; the admin-button follow-up is yours, with the opt-out above as the one requirement.

the-luap:
Correction to my last comment. Two things in it no longer hold, and I'm taking them back.

**1. "The admin-button follow-up is yours" — I took it myself: #1557.** Apologies for the switch after telling you otherwise, you said you had it ready. The reason is the next point: the shape I asked you for was wrong, and I did not want you to build against it.

**2. The opt-out I called "the one requirement" is not the right design on its own.** I recommended letting the admin route skip the size bound and otherwise treating videos like stills. `videoProcessor.js:66-70` says why that is wrong: `generateVideoThumbnail` renders the poster at a fixed `300x300`. It does not depend on the thumbnail settings the button exists to apply. The loop force-rebuilds every row (`adminThumbnails.js:209`, `thumbnail_path: null` + `force: true`), so for videos that means every press re-reads every video, a full object download each on S3, to write back identical bytes.

What #1557 does instead, for a video row:

- repair, not rebuild: `thumbnail_path` is passed through and `force` is off, so `ensureThumbnail`'s own validity check skips a healthy poster and only a missing or unreadable one is regenerated
- the bound is still lifted for that repair (`boundVideoSource: false`), since it is an admin-initiated background job and bounded a large S3 video could only ever get the placeholder
- no tier deletion, videos never take the tier path
- stills are untouched; `/regenerate-previews` keeps its filter

Your try/catch point carries over as-is: a video that returns `null` or throws is counted and the batch moves on, and there is a test for both.

If your ready-made version handles something #1557 misses, a review there is very welcome. Everything else in my last comment stands, including the verdict on the bound itself. The stable backport of this PR is #1556.

PR #1425, opened 2026-09-11
fix(archives): stop restore from dropping videos and rewriting the row

AtelyPham:
## Problem

Restoring an archived event brings back every file but only some of the photos.

The gallery, the admin grid and every download are built from database rows. Restore decides which extracted files get a row by matching the filename against five web image extensions (`backend/src/routes/adminArchives.js`). A video or a DNG lands on disk with no row, so as far as the app can tell it is gone. Both are accepted uploads with their own entries in `ALLOWED_MEDIA_TYPES`.

The rows it does write are wrong in three columns:

- `type` is 'individual' or 'collage', and the download zip groups its folders by it. Restore writes the file extension, so every restored photo lands under "Collages".
- `media_type` is omitted and defaults to 'image', so a restored clip stops being a clip.
- `uploaded_at` gets the restore time, so the gallery order changes.

## Fix

The archive already carries `photos_manifest.json` with each photo's `type` and `uploaded_at`, so restore rebuilds the row from that. Only 'individual' and 'collage' are trusted for `type`: an event restored by the current code and archived again carries the extension in its manifest, and the directory fallback repairs those. `uploaded_at` goes through `toIso`, since on SQLite the manifest can hold epoch milliseconds. For archives written before the manifest existed, the extension set is the fallback. That set now comes from `ALLOWED_MEDIA_TYPES` and `ALLOWED_VIDEO_TYPES` in `fileSecurityUtils` instead of a second hardcoded list, so a format the uploader accepts is restorable the same day.

`media_type` and `mime_type` join the manifest for archives written from here on. Any video signal (manifest `media_type`, a video/* `mime_type`, a video extension) makes the restored row a video. That matters for watcher-imported clips: `fileWatcher` never sets `media_type`, so those rows sit at the 'image' default with only their mime saying otherwise.

## Tests

`backend/__tests__/integration/adminArchives.restoreRowFidelity.test.js` drives real zips through the real route against SQLite. Eleven tests, ten fail on `main` today. The ninth guards that widening the gate does not turn `photos_manifest.json` or the feedback exports into photo rows.

Backend suite and `eslint src/` clean, the two MinIO integration suites aside (they need the MinIO service CI provisions).

Backend only, no UI change.


the-luap (approved):
Re-verified 60ae0c90 at file:line; both concerns and the nit closed. 11 pass on the branch, 10 fail against main's sources. Merge cleared.

PR #1426, opened 2026-09-12
fix(gallery): keep the video controls reachable while paused, and go fullscreen on iOS

natistir:
## Description

Three related problems in the guest `VideoPlayer`, all reproduced on a self-hosted stable 3.46.13 install and fixed together because they share the same surface:

1. **The paused-state play overlay blocked the control bar.** The centre play overlay is `absolute inset-0` and renders after the control bar, so while the video was paused it sat on top of the fullscreen, mute and progress controls and swallowed their clicks. A click on the poster went to the overlay's wrapper div instead of the `<video>`. The wrapper is now `pointer-events-none` and only the button is `pointer-events-auto`.
2. **The auto-hide timer read stale state.** `handleMouseMove` captured `isPlaying` at the time of the last mouse move, so pausing within three seconds of a move hid the controls anyway — leaving no visible play or fullscreen button, with every click going to the `<video>` and toggling playback. And resting the pointer on the bar to click pause let the bar fade out from under the cursor. Pausing now always reveals the controls, the timer reads the live state through refs, hover on the bar defers the hide, and a faded bar is `pointer-events-none` so it can't intercept clicks.
3. **Fullscreen was a no-op on iPhone Safari.** `toggleFullscreen` only ever called `video.requestFullscreen()`, which iOS Safari does not implement. Fullscreen is now requested on the container (so the custom controls come along), falling back to `webkitRequestFullscreen`, then to the video element's `webkitEnterFullscreen` (native iOS player). `fullscreenchange` / `webkitendfullscreen` listeners keep `isFullscreen` in sync when the user exits with Esc or the native UI.

No changes to the lightbox, to protection levels, or to how the video URL is resolved.

## Type of change

## How Has This Been Tested?

- [x] Unit tests pass (`npm test`) — `VideoPlayer.pausedControls.test.tsx` adds 6 tests (overlay click-through, container fullscreen, iOS fallback, Esc sync, pause-reveals-controls with a hide already scheduled, hover defers hide); the existing `VideoPlayer.loadError.test.tsx` still passes (11/11 for the component). `eslint` and `tsc --noEmit` clean for the touched files.
- [x] Tested on production-like environment — built into the frontend image of a self-hosted stable 3.46.13 install; verified play/pause/mute/fullscreen while paused and while hovering the bar on desktop Chrome, and playback + fullscreen on iOS Safari.

**Test Configuration**:
* PicPeak Version: stable 3.46.13 (patch applies cleanly to `main`, this PR is against `main`)
* Node.js Version: 22
* Database: PostgreSQL
* Browser: Chrome (desktop), Safari (iOS)

## Checklist:

- [ ] I have made corresponding changes to the documentation — n/a
- [x] Any dependent changes have been merged and published — none

## Additional Notes:

On `stable` there is a fourth, separate cause for "dead controls" on desktop: the lightbox's translucent bottom toolbar overlays the media area, and a landscape video's control bar lands underneath it. That is already fixed on `main` by #892 and is not part of this PR; I've opened a separate issue asking for its backport to `stable`.

### DIFF
--- a/frontend/src/components/gallery/VideoPlayer.tsx
+++ b/frontend/src/components/gallery/VideoPlayer.tsx
@@ -27,6 +27,7 @@ export const VideoPlayer: React.FC<VideoPlayerProps> = ({
 }) => {
   const { t } = useTranslation();
   const videoRef = useRef<HTMLVideoElement>(null);
+  const containerRef = useRef<HTMLDivElement>(null);
   const [isPlaying, setIsPlaying] = useState(false);
   const [loadError, setLoadError] = useState<string | null>(null);
   const [isMuted, setIsMuted] = useState(muted);
@@ -36,6 +37,10 @@ export const VideoPlayer: React.FC<VideoPlayerProps> = ({
   const [duration, setDuration] = useState(0);
   const [showControls, setShowControls] = useState(true);
   const controlsTimeoutRef = useRef<NodeJS.Timeout>();
+  // The hide timer reads these instead of state so it sees the value at the
+  // moment it fires, not the one captured when the mouse last moved.
+  const isPlayingRef = useRef(false);
+  const hoveringControlsRef = useRef(false);
 
   useEffect(() => {
     const video = videoRef.current;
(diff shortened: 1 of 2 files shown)


the-luap (approved):
Verified. All 6 of the original tests fail against `main`, and both tests in the follow-up fail against your first commit — so each fix is pinned by something that actually catches its absence. 13/13 across both video suites, lint and tsc clean.

You took the prefixed-event concern and the scoping nit together, which closes the last asymmetry: three enter paths, and now a sync listener for each of them.

The `pointer-events-none` wrapper with `pointer-events-auto` on the button is the right shape, and making the faded bar non-interactive is a quiet bonus — at opacity-0 it was still swallowing clicks near the bottom of a playing video.

PR #1427, opened 2026-09-12
fix(frontend): allow blob: in the CSP media-src directive

natistir:
## Description

`AdminAuthenticatedVideo` fetches the file with the auth header and plays it through `URL.createObjectURL(...)`, i.e. a `blob:` URL. The frontend nginx CSP already allows `blob:` in `img-src` for the same pattern (`AdminAuthenticatedImage`), but `media-src` was `'self'` only, so the browser refused the source. The symptom is a grey player with no error in the UI — only a CSP violation in the console — which is easy to mistake for a codec or MIME problem.

This adds `blob:` to `media-src` in the three CSP strings in `frontend/nginx.conf`. Same exposure as the existing `img-src blob:` — only blobs created by the page itself.

## Type of change

## How Has This Been Tested?

- [x] Tested on production-like environment — running on a self-hosted stable install since 3.45.0 (July) via a patched `default.conf`, now built into the frontend image on 3.46.13. Admin lightbox video playback works with the header; the CSP violation is gone. `nginx -t` passes.

**Test Configuration**:
* PicPeak Version: stable 3.45.0 → 3.46.13
* Node.js Version: 22
* Database: PostgreSQL
* Browser: Chrome, Safari

## Checklist:

- [ ] I have commented my code — config-only change
- [ ] I have made corresponding changes to the documentation — n/a
- [ ] I have added tests — nginx config, not covered by the test suites
- [x] Any dependent changes have been merged and published — none

### DIFF
--- a/backend/server.js
+++ b/backend/server.js
@@ -147,7 +147,7 @@ const cspDirectives = {
   connectSrc: ["'self'", 'https://www.google.com', 'https://www.gstatic.com'], // API connections
   fontSrc: ["'self'", "https:", "data:"], // Web fonts
   objectSrc: ["'none'"], // Disable plugins
-  mediaSrc: ["'self'"], // Audio/video
+  mediaSrc: ["'self'", "blob:"], // Audio/video, incl. page-created blobs (matches imgSrc)
   frameSrc: ["'self'", 'https://www.google.com'],
 };
 // Only upgrade insecure requests when HSTS explicitly enabled (HTTPS deployment)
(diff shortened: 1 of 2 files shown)


the-luap (approved):
Verified end to end. The blob premise is real (`AdminAuthenticatedVideo.tsx:31`), all three nginx policies are updated, and a directive-by-directive diff of the policy string shows `media-src` as the only change — `script-src` is untouched, which is the one that would matter. Same exposure as the `img-src blob:` that has been there all along: blobs the page created itself.

The follow-up commit closes the all-in-one gap, so the two policies now agree instead of diverging by deployment shape.

Thanks for the months of production use behind this — that is worth more than the diff size suggests.

Issue #1430, opened 2026-09-12
[FEATURE] Make video a first-class media type: counts, filters, wording, ingest and playback


natistir:
**Is your feature request related to a problem? Please describe.**

Video has been a supported media type in the schema since migration 048 (`photos.media_type`, duration, codec, dimensions), but the rest of the product never caught up, and it shows in the issue tracker: #203, #895, #908, #1370, #1372, #1377, #1403, #1408, #1414, #1426, #1427, #1428 are all video problems that arrived one at a time as bugs. Each fix is right on its own, but the pattern is that video is bolted onto a photo product rather than being a first-class media type.

Concretely, from running PicPeak for a studio that delivers photo *and* video to the same clients:

- **Language.** `en.json` has ~310 strings that say "photo" and ~16 that say "video". A client event with 40 videos shows "Manage Photos", "Upload Photos", "Total Photos: 40". Every other locale inherits the same wording.
- **Counts.** Event cards, dashboard and per-event statistics report one number. There is no way to see "137 photos · 6 videos", let alone total runtime.
- **Filtering.** Neither the admin grid nor the guest gallery can show only videos (or only photos). For a videographer's deliverable that is the primary view, not an edge case.
- **Ingest.** The external-media picker only lists `.jpg/.jpeg/.png/.webp` (`externalMediaService.js:77`), so videos dropped into the media root are invisible to it. In the admin uploader a single file larger than `general_max_upload_batch_size_mb` is still one POST — the batch logic splits *sets* of files, not a file — so a 400 MB clip fails behind Cloudflare Tunnel or any proxy with a request-size cap, even though the chunked API exists (#1377 made it reachable; nothing in the UI calls `uploadLargeFile`).
- **Processing feedback.** ffmpeg failures are logged server-side and otherwise silent; the admin sees a tile with no thumbnail (#1414) and no way to retry.
- **Playback.** The guest player is a bare `<video src>` of the original. HEVC/H.265 phone footage plays only in Safari — `VideoPlayer` already tells the guest to download instead (#1370) — and there is no rendition or transcode step even though ffmpeg is in the image. `downloads.js` already answers `Range` requests, but the gallery video route does not go through it, so scrubbing a large file over a tunnel is poor.

**Describe the solution you'd like**

Treat "video is a media type" as one initiative, delivered as small independent PRs, rather than continuing to fix video one bug at a time. Proposed shape:

[...]

Issue #509, opened 2026-05-16
[BUG] Broken upload batch size for reverse proxy compatibility


Heisendres0:
**Describe the bug**
It appears that the issue described in #208 , which was fixed until recently, is happening again. Maybe something broke?

**To Reproduce**
Follow the exact same steps as #208

Rekoo-PS:
I was using a local server with tunnels.
I let GPT make an uploader using the api tokens
I just drag and drop the whole folder, and its working great as it's uploading 4 photos at a time simultaneously .

But I moved to Hostinger VPS " without Cloudflare tunnels" recently, so I can't confirm if it's working on big folders xD

the-luap:
You were right — fixed in #513. Real regression of #208, not a fresh bug. Traced the timeline:

- 2026-03-05 — PR #214 (commit \`02a46e0\`) shipped \`general_max_upload_batch_size_mb\` (default 95MB for Cloudflare Tunnel headroom), wired into PhotoUpload + the General settings UI.
- 2026-03-11 — "Merge main into beta for release/beta-to-main" (\`28793bb\`) resolved its conflict by keeping main's older tree, silently deleting migration 072, the UI input, the i18n strings, the \`useSettingsState\` field, and replacing the setting read with the hardcoded \`MAX_BYTES_PER_CHUNK = 500 * 1024 * 1024\`. \`git blame\` on that line points back at the original revert (\`cc4503a\`) because the merge resolved by accepting that side. Galleries fronted by Cloudflare have been quietly broken on batch uploads >100MB since then.

Fix is a mechanical re-application of PR #214 to current beta. Existing installs are safe in both directions — the migrations-table row was never cleaned up (no down ran), so the seeded \`app_settings\` row is still there for installs that did go through the original 072. The new migration's \`if (!exists)\` guard makes it idempotent. Added a comment in 072 explaining the regression history so the next merge conflict has a fighting chance.

the-luap:
Fixed in `98f3c3d` (fix(upload): restore configurable batch-size for reverse proxies) and released in v3.44.0 stable (via PR #568).

Please reopen if you still see the #208 regression on v3.44.0+.

PR #1431, opened 2026-09-12
feat(admin): upload single files above the batch cap through the chunked API

natistir:
## Description

Item 5 of #1430, and the first PR of that set.

The batch logic (#509) splits a *selection* into POSTs under `general_max_upload_batch_size_mb`, but a lone file above the cap cannot be split that way and still went out as one multipart request — exactly what a proxy with a request-size limit rejects (Cloudflare Tunnel's ~100 MB body cap being the common case). The chunked API has existed all along and became reachable in #1377; nothing in the admin UI called `photosService.uploadLargeFile`.

Now:

- **`PhotoUpload.tsx`** — files above the batch cap go through `uploadLargeFile` (10 MB parts); everything under it keeps the multipart batch path untouched. Progress is reported over both (each large file is one "unit" alongside the batches), and a failed large file goes into the same failure report as multipart failures without blocking the rest of the selection. Replace-by-name is not supported by the chunked complete step, so with it on a large file is skipped and listed in the report rather than uploaded as a second copy.
- **`photos.service.ts`** — the chunk and complete requests no longer inherit the default axios timeout (a large clip is many sequential parts, and complete runs ffmpeg); `shouldUseChunkedUpload` takes the threshold so the UI can pass the configured batch cap.

No backend changes. The file's MIME is passed through as before; `main` derives it from the extension server-side and ignores the client value.

**Two settings interact here.** The routing threshold is `general_max_upload_batch_size_mb` (default 95 MB, "what fits in one request"). Whether a single file can be stored at all is `general_max_file_size_mb` / `general_max_video_size_mb`, which default to 50 MB. On a default install a file large enough to take the chunked path is therefore still rejected at init with an accurate "maximum size is N MB per file" message — an improvement over the proxy-level network error, but the chunked path only does useful work once the per-file cap is raised above the batch cap.

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

- [x] Unit tests pass (`npm test`) — `PhotoUpload.largeFileChunked.test.tsx` (5 tests: above-cap file goes chunked and is not POSTed; under-cap file stays multipart; mixed selection uses both; a chunked failure lands in the report and the rest still uploads; replace-by-name skips a large file into the report). 72/72 across the upload, service and fileTypes suites; existing `PhotoUpload.*` tests unchanged and passing. `eslint` and `tsc --noEmit` clean; `vite build` succeeds.
- [x] Tested on production-like environment — the same routing has been running on a self-hosted stable install behind Cloudflare Tunnel since 3.45.0 (re-ported here onto `main`'s restructured `PhotoUpload`): multi-hundred-MB `.mp4`/`.mov` uploads that previously failed with a network error complete through the chunked path; small batches behave as before.

**Test Configuration**:
* PicPeak Version: stable 3.45.0 → 3.46.13 (this PR against `main`)
* Node.js Version: 22
* Database: PostgreSQL
* Browser: Chrome (desktop), Safari (macOS)

## Checklist:

- [ ] I have made corresponding changes to the documentation — no new setting; the settings interaction is described above
- [x] Any dependent changes have been merged and published — #1377

## Additional Notes:

- The threshold is the existing `general_max_upload_batch_size_mb` rather than a new setting, since it already encodes "what my proxy will accept in one request".
- The `uploadLargeFile` chunk loop and `complete` step are unchanged; this PR only routes to them. Per-chunk retry is a deliberate follow-up.

### DIFF
--- a/frontend/src/components/admin/PhotoUpload.tsx
+++ b/frontend/src/components/admin/PhotoUpload.tsx
@@ -10,6 +10,7 @@ import { settingsService } from '../../services/settings.service';
 import { useTranslation } from 'react-i18next';
 import { extensionsToMimeTypes, extensionsToAcceptString, extensionsToLabel } from '../../utils/fileTypes';
 import { useUploadProgress } from '../../hooks/useUploadProgress';
+import { photosService } from '../../services/photos.service';
 
 interface PhotoUploadProps {
   eventId: number;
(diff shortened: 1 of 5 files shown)


the-luap (approved):
Verified the amend directly rather than on the summary: `git diff d0066a78 a051539a` is empty, so it really was message-only — trailer gone from the commit, link gone from the description, and CI re-ran 10/10 on the new head.

Content was already settled: the MIME revert is complete (`frontend/src/utils/` diffs empty against `main`), large files with replace-by-name land in the failure report instead of becoming a second copy, and the new test fails against `6eb0fb47` so it pins the behaviour rather than describing it.

This also gives `uploadLargeFile` its first caller. The chunked route has been dead code since it shipped, so the hardening that went in this week — the CSRF gate letting octet-stream through, and the streaming cap that stops a chunk body being buffered before anything checks its size — starts carrying real traffic with this. Good order to have done it in.

Thanks for switching the trailer off in your tooling, and for taking the nit as well as the concern.

PR #1457, opened 2026-09-14
fix(crm): keep the editor's line-item order when a quote or invoice is saved (#1452)

### LINKED ISSUE #1452: [BUG] Reordering line items on a saved quote is lost after saving
**Describe the bug**

On a saved quote, moving line items with the up/down arrows in the editor looks correct until you save. After saving and reloading, the line items are back in their original order. The PDF and the customer's quote page also show the original order.

**To Reproduce**

1. Open an existing quote with at least two line items (CRM → Quotes → open a quote → Edit).
2. Move a line item with the ↑ / ↓ arrows in the line-items table.
3. Save the quote.
4. Reopen the quote, or open its PDF.
5. The line items are in their original order.

**Expected behavior**

The order shown in the editor at save time is the order that gets stored, and it's used everywhere after that: editor, PDF, customer view, and conversion to contract or invoice. When a parent line moves, its sub-items move with it.

**Root cause (from reading the code on upstream `main`)**

The editor keeps two different things in `position`:

- In `frontend/src/components/admin/LineItemsTable.tsx`, `position` is a **stable row identifier** that links sub-items to their parent. The file header says so explicitly: *"once assigned at row creation we never renumber it"*. `move()` (~line 165) only reorders the array; it doesn't touch `position`.
- `QuoteEditorPage.tsx` (~line 148) sends `position: li.position` for every row, which is still the original value.
- `mapPayloadToService` in `backend/src/routes/adminQuotes.js:240` keeps the client's `position` and only falls back to the array index when it's missing (`li.position == null ? idx + 1 : li.position`).
- `updateQuote` deletes and re-inserts all line items with those same positions (`quoteService.js` ~line 780).
- `getQuoteById` reads them back with `.orderBy('li.position', 'asc')` (`quoteService.js` ~line 495).

So the order the user chose (the array order) is thrown away on save, and the original creation order comes back.

**Likely also affected (not verified in the UI)**

- **Invoices / bills:** `BillEditorPage.tsx` uses the same `LineItemsTable`, also sends `position: li.position`, and `backend/src/routes/adminInvoices.js:328` has the identical mapping.
- **New quotes reordered before their first save:** same code path. Rows get positions 1…n on creation, so reordering before the first save should be lost too.

**Possible Solution**

Treat the payload's **array order as the display order** and renumber on the server, rather than trusting the client's `position` for ordering:

- In both `adminQuotes.js` and `adminInvoices.js` (ideally one shared helper, not two copies), assign `position = index + 1` in array order. Build an `oldPosition → newPosition` map and use it to rewrite each sub-item's `parent_position`, so parent/child links survive the renumbering.
- The frontend can keep `position` as its stable row ID; nothing in `LineItemsTable` has to change.
- Renumbering on the server also protects any other API client that sends items in the intended order.

**Acceptance criteria**

- [ ] Reordering top-level line items on a saved quote persists after save and reload.
- [ ] Moving a parent line moves its sub-items with it, and they stay attached after saving.
- [ ] Reordering sub-items within the same parent persists.
- [ ] The quote PDF, the customer quote page and quote → contract / invoice conversion all use the saved order.
- [ ] The same behaviour is verified for invoices in `BillEditorPage`.
- [ ] Backend tests cover: reordered top-level items, reordered sub-items, and a moved parent with sub-items (parent links still correct after renumbering).

**Additional context**

Code references are against upstream `main` (3043a4f3).

Voyagerroc-Lab:
## Description

A reorder made with the up/down arrows in the line-items table was thrown away on save. The editor keeps `position` as a **stable row id** (`LineItemsTable`'s header: *"once assigned at row creation we never renumber it"*, and `move()` only reorders the array), so both editors posted each row's original `position` back, the service stored it verbatim, and the detail endpoints read the items back `ORDER BY position`. The result: the order was right in the editor and reverted on reload, in the PDF and on the customer page.

This makes the payload's array order the display order — the order the user arranged — and stores that:

- New shared helper `backend/src/utils/lineItemPositions.js` → `renumberLineItemPositions()`: renumbers `position` to `1..n` in array order, and rewrites every `parent_position` through the same old → new map so a moved parent keeps its sub-items attached.
- Wired into `mapPayloadToService` in **both** `adminQuotes.js` and `adminInvoices.js` (one helper, not two copies).
- A `parent_position` that names a position absent from the payload is left untouched, so `validateLineItemHierarchy` still reports it as a missing parent (400) instead of the row being silently re-parented onto whatever item ends up on that number.
- The frontend is untouched: `position` stays a stable row id there. Renumbering server-side also protects any other API client that sends items in the intended order.

Covers all four surfaces the issue listed, plus quote → contract/invoice conversion, because they all read the stored order through `getQuoteById` / `getInvoiceById`.

Fixes #1452

## Type of change

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## How Has This Been Tested?

Two new suites, plus the surrounding CRM suites:

```
backend/__tests__/utils/lineItemPositions.test.js        10 passed   (pure helper)
backend/__tests__/integration/lineItemOrder.test.js       5 passed   (real HTTP route → service → SQLite)
__tests__/services/quoteService*+invoiceService*,
__tests__/routes/publicQuotes.test.js,
__tests__/integration/discountLineItems.test.js          100 passed / 8 suites
npm run lint (backend)                                    clean
```

The integration suite drives the real routes: PUT a reordered quote → GET it back and assert the order; a moved parent with sub-items (the sub-item's `parentLineItemId` follows the renumbered parent); reordered sub-items inside one parent; a new quote reordered before its first save; and the same reorder for a scheduled invoice.

**Non-vacuous:** with only the two route files reverted to `main` (helper and tests kept), all 5 integration tests fail; with the fix, all 5 pass.

- [x] Unit tests pass (`npm test` on the suites above)
- [ ] Tested on Docker deployment

**Test Configuration**:
* PicPeak Version: main @ 2fe79e4
* Node.js Version: v22.22.0
* Database: SQLite (temp DB from `__tests__/integration/helpers/crmDb.js`)
* Browser: n/a (backend only)

## Checklist:

- [x] I have made corresponding changes to the documentation

Backend-only change — no user-facing surface is modified, so there is nothing to screenshot. The commit message is a Conventional Commit (`fix(crm): … (#1452)`) so release-please generates the changelog entry, which is why `CHANGELOG.md` is not edited by hand here.

## Additional Notes:

- Scope kept to ordering: no route validators, service signatures or payload keys changed.
- Invoices can only be edited while `scheduled`, so the invoice test uses that path; the mapping it exercises is the same one `BillEditorPage` posts to.
- The `position` field in the API response keeps its meaning (1-based order within the document), so existing consumers are unaffected.


the-luap (commented):
Thanks. The diagnosis is right: `LineItemsTable` keeps `position` as a stable row id, and the service stores it verbatim and reads it back `ORDER BY position`. Renumbering from array order in `mapPayloadToService` is the right place for the fix. The integration tests reproduce the bug: with only the two route files reverted to `main`, all 5 fail.

The problem is the remap in the helper. For two payload shapes the service used to reject with a 400, it now saves and attaches a sub-item to the wrong parent.

## 🔴 Blockers

### 1. A missing `parent_position` is re-parented when its number is between 1 and n
`backend/src/utils/lineItemPositions.js:39-41`

The docblock (lines 17-21) says a `parent_position` that isn't in the payload is left for `validateLineItemHierarchy` to reject. It is left unchanged, but after renumbering that number can belong to a different row, so validation finds a parent and accepts it:

```js
renumberLineItemPositions([
  { position: 10, d: 'A' },
  { position: 20, d: 'B' },
  { position: 30, d: 'C', parent_position: 2 },
]);
// → A:1, B:2, C:3 with parent_position 2  → C is saved under B
```

On `main` this is `400 LINE_ITEM_PARENT_NOT_FOUND` (`quoteService.js:268`). The unit test at `lineItemPositions.test.js:92-100` uses `42`, which is higher than the item count, so it never hits this case.

### 2. Duplicate positions are accepted and can attach a sub-item to the wrong parent
`backend/src/utils/lineItemPositions.js:31-33`

The old→new map is last-one-wins:

```js
renumberLineItemPositions([
  { position: 1, d: 'P1' },
  { position: 2, d: 'S', parent_position: 1 },
  { position: 1, d: 'P1dup' },
]);
// → P1:1, S:2 with parent_position 3, P1dup:3  → S is saved under P1dup
```

On `main` this is `400 LINE_ITEM_POSITION_DUPLICATE` (`quoteService.js:256`). After renumbering, positions are always unique, so that check can no longer fire. Nothing validates `lineItems.*.position` at the route. The route default `li.position == null ? idx + 1 : li.position` can also create a duplicate when a payload mixes rows with and without a position.

The editor doesn't send either payload, but any client calling these admin endpoints can.

**Fix for both:** renumber only when the payload is unambiguous. If positions aren't unique, or a non-null `parent_position` isn't one of the payload's positions, return the items unchanged so `validateLineItemHierarchy` still rejects them:

```js
const oldPositions = items.map((item) => ensureInt(item.position));
const unambiguous = new Set(oldPositions).size === oldPositions.length
  && items.every((item) => item.parent_position == null || item.parent_position === ''
    || oldPositions.includes(ensureInt(item.parent_position)));
if (!unambiguous) return items;
```

Please add a unit test for each case, plus one route test showing a 400 is still returned.

## 🟡 Concerns

1. **`lineItemPositions.js:32,38`: the map keys use `Number()`, but the service compares with `ensureInt` (`quoteService.js:217`).** For `[{position: 2.7}, {position: 3, parent_position: 2}]`, `main` matches the sub-item to the parent via `parseInt`. The helper leaves `2` unmatched, and after renumbering `2` is the sub-item's own position, so the save now fails with `LINE_ITEM_SELF_PARENT`. Use `ensureInt` for both keys and lookups so the helper and the validator agree.
2. **Missing test for the most common way this bug shows up.** `addSubItem` gives a new row `max + 1` (`LineItemsTable.tsx:73`). So a sub-item added later to the first parent posts as `[P1(1), S(3, parent 1), P2(2)]`. On `main` it reloads as P1, P2, S. The PDF groups rows by array order (`pdfService.js:750-763`), so S prints under P2. The PR fixes this (it now saves as P1:1, S:2→1, P2:3), but no test covers it.
3. **Test coverage gaps:**
   - Invoice create is only exercised with an in-order payload, so it doesn't show that the create path renumbers.
   - Nothing tests that a sub-item under a sub-item still gets `LINE_ITEM_NESTING_TOO_DEEP` after renumbering.
   - Test 4 (`lineItemOrder.test.js:172`, "a new quote can be reordered before its first save is lost") has a garbled name and repeats test 1. Either give it a distinct scenario or drop it.

## 🟢 Nits

- **Comment length:** the helper has a 25-line docblock for a 15-line function, and the same explanation is repeated in both routes and both test headers. One short comment at the call site plus a shorter docblock would do.
- **`#1452` in code comments** (`lineItemPositions.js:2`, `adminQuotes.js:242`, `adminInvoices.js:330`, test headers): we don't reference issue numbers in source comments; please drop them.
- **Invoice test name:** "a reordered scheduled invoice keeps the new order after a reload" creates a plain invoice.
- **PR title:** it ends in `(#1452)`, an issue number. On squash-merge GitHub appends the PR number, so the changelog entry would carry both. Please drop it from the title; the body already links the issue.

## ✅ Verified

Directly inspected or run:

- **Tests:** the new suites pass on the PR branch (15/15). With only `adminQuotes.js` and `adminInvoices.js` reverted to `main`, all 5 integration tests fail.
- **Edge cases:** Blockers 1 and 2, Concern 1 and the later-sub-item case in Concern 2 were run against the helper from the PR branch, with the outputs shown above.
- **Coverage of the change:** `mapPayloadToService` feeds quote create, update and preview (`adminQuotes.js:371,414,568`) and invoice create, update and preview (`adminInvoices.js:428,605,952`). The invoice update line-item rewrite (`adminInvoices.js:745-751`) uses that same payload.
- **Validation:** `validateLineItemHierarchy` (`quoteService.js:246`) is the only guard for duplicate, self-parent and missing-parent errors. With unique positions the remap is one-to-one, so the self-parent and nesting checks behave as before.
- **Editors after save:** both navigate away after a successful save (`QuoteEditorPage.tsx:431-438`, `BillEditorPage.tsx:479`), so no second save is posted from stale row ids. The next edit reloads positions from the GET.

Checked by reading only:

- **Paths without the renumbering:** quote→invoice clone, installment plan, storno, hour and expense billing, monthly draft append and `duplicateQuote`. All build positions from rows already read in stored order, so none needs it.
- **Readers:** the PDF, the public quote page and the conversions all read line items `ORDER BY position` and rebuild parent links from `parent_line_item_id`.

## Verdict

Close Blockers 1 and 2 (skip renumbering on ambiguous input, with tests) and switch the keys to `ensureInt`, and I'm comfortable merging.

Issue #1265, opened 2026-09-01
[BUG] Guest identity token is stored in sessionStorage, so returning guests are re-prompted and duplicate gallery_guests rows are created


leonlivevocalist-svg:
**Describe the bug**

In `identity_mode='guest'`, the guest identity JWT is stored in `sessionStorage` (keys `guest_token_<slug>` / `guest_identity_<slug>`). `sessionStorage` is scoped to the browser tab, so a guest who closes the tab and returns via the same emailed link has lost their identity and is shown the registration prompt again. The `?invite=` token in that link is single-use and already redeemed, so it cannot re-establish them.

If they answer the prompt by typing their name rather than using "I've been here before", `POST /gallery/:slug/guest` inserts a brand new `gallery_guests` row — there is no lookup of an existing `(event_id, email)` guest — and their earlier likes now belong to an identity they can no longer act as. They cannot remove them.

This is the same end-user symptom as #1210. The duplicate flagging in #1216 and the recovery-link discoverability work in #1217 both help *after* the split has happened, but the split itself is still triggered by something entirely routine: closing a tab.

A note on `GUEST_TOKEN_TTL`, since it looks like it should cover this and doesn't: the `30d` default in `middleware/guestAuth.js` governs how long the JWT stays *valid*, not how long the browser *keeps* it. With the token in `sessionStorage` the practical lifetime is "until this tab closes", so the 30-day validity is rarely reached.

**To Reproduce**

1. Create an event with `identity_mode='guest'` and `allow_likes = true`
2. Open the gallery as a guest, register (or redeem an invite), and like a few photos
3. Close the tab, then open the same gallery link again in a new tab and enter the password
4. The "who are you?" prompt appears. Enter the same name and email as before
5. A second `gallery_guests` row is created for the same person. The likes from step 2 are still counted on the photos, but cannot be removed from this new session

**Expected behavior**

A returning guest on the same device and browser is still recognised as themselves, and can remove likes they made on a previous visit.

**Screenshots**

Not applicable — this is visible in the database state rather than on screen.

**Environment (please complete the following information):**
 - OS: Windows 11 (client), Ubuntu 24.04 (server)
 - Browser: Chrome 152 — not browser-specific, `sessionStorage` semantics are the same across browsers
 - PicPeak Version: v3.122.2-beta.0 (also present in v3.119.0-beta.0)
 - Deployment Method: Docker Compose
 - Database: PostgreSQL

**Logs**

[...]

PR #1517, opened 2026-09-16
fix(guests): preserve feedback ownership when merging identities

the-luap:
## Description

Merging guest identities moved `photo_feedback.guest_id` but left `guest_identifier` pointing to the discarded identity. The per-photo feedback endpoint then failed to recognise transferred selections as the surviving guest's own, especially when feedback sharing was disabled. Overlapping selections could also remain duplicated and require multiple clicks to clear.

This change:

- Reads per-photo ownership through the verified `guest_id`, including rows affected by earlier merges, and rewrites both identifiers during new merges.
- Preserves the union of photo selections, collapses visible non-comment duplicates per photo/type, and keeps the latest rating/reaction/colour label. Original choice timestamps survive subsequent merges; comments and hidden moderation records remain intact.
- Clears duplicate likes/favourites from earlier merges in one click.
- Commits feedback, counters, invitations and guest updates together. PostgreSQL locks the participating guests in ID order; invalid, deleted and cross-event identities are rejected.

Addresses the separate merge defect reported in [the follow-up to #1265](https://github.com/PicPeak/picpeak/issues/1265#issuecomment-5693856634). The reported frontend bundle mismatch is still being diagnosed. Guest-token storage and registration are unchanged.

## Type of change

- [x] Bug fix

## Validation

- 13 new integration tests pass on SQLite and PostgreSQL 15 (isolated Docker database), using the real routes, authentication and migrated schemas.
- 80 existing feedback and guest integration/route/service tests pass: duplicate detection, visibility, ownership, hidden feedback, ratings, reactions, colour labels and per-guest limits.
- Full backend CI passes: **3,932 tests across 380 suites**. All PR checks completed successfully, including AMD64/ARM64 Docker builds and the all-in-one startup smoke test.
- Failure-injection tests verify rollback after feedback recalculation fails and after later invite/guest updates.
- ESLint passes for all changed JavaScript files; `git diff --check` passes.

No schema migration is required. Previously merged rows are recognised through `guest_id`; existing duplicate selections can be cleared with one interaction.


the-luap (commented):
Review of `a4ec21f7`. Backend only: 4 files, no migration, no UI surface, so no screenshot is needed.

## 🔴 Blockers

None.

## 🟡 Concerns

None that block. One question on the ordering guarantee is listed under nits.

## 🟢 Nits

- **Late write during a merge.** `adminGuests.js:646-650`: the PostgreSQL `forUpdate` locks the `gallery_guests` rows, but a feedback submission from a source guest doesn't take that lock. If that insert commits after the merge's `scope().update(...)` statement and before the transaction commits, it stays attached to a guest that is then soft-deleted. The window is a few milliseconds and the next merge picks the row up, so this is not worth more code; noting it so the "lock in a stable order" comment isn't read as covering submissions.
- **Loose ID coercion.** `mergeIds.map(Number)` coerces `[5]` or `"5"` to `5`, so the positive-integer check accepts them. Harmless, since the event and `is_deleted` check still applies.
- **Pre-registration feedback.** `galleryFeedback.js:105`: once a guest token is present, the per-photo "own feedback" query matches `guest_id` only. A row the device created before registering (`guest_id` NULL) is no longer shown as the guest's own there. That's consistent with `/my-feedback` (`:480`) and the submit-time duplicate check, which already prefer `guest_id`, so this aligns the three rather than regressing one.

## ✅ Verified

**Read directly:**
- **One transaction.** The merge runs inside a single `db.transaction`: identity check, feedback merge, email canonicalisation, invite carry-over, soft delete, and `logActivity(..., trx)`. `mergeGuestFeedback` uses the passed executor throughout, and `updatePhotoFeedbackStats(photoId, trx)` already accepted one. No global `db` or `hasColumnCached` call sits inside the transaction, so there's no SQLite single-connection deadlock.
- **Guard is stricter and safe on both engines.** The guest check now also requires `is_deleted = false` (via `formatBoolean`), so merging into or from a deleted identity is a 400 instead of silently succeeding.
- **Collapse rule matches the submit path.** Dedup keeps comments and hidden rows and collapses one visible row per `photo:type`. That is the same single-row-per-guest rule `submitFeedback` enforces for like, favorite, rating, reaction and colour label (`feedbackService.js:479-603`).
- **Shared colour labels are untouched.** They're stored with `guest_id: null` and the reserved identifier (`:300-309`), so the `whereIn('guest_id', ...)` scope never touches them.
- **Timestamps.** `timestamp()` handles epoch ms, SQLite `YYYY-MM-DD HH:MM:SS` (as UTC), ISO strings and PG `Date`. The merge no longer rewrites `updated_at`, so a later merge still compares real choice times.
- **Toggle-off clears duplicates.** Toggling a like or favorite off now deletes the guest's whole visible set, the same shape the rating and single-value paths already use.

**Ran:**
- **SQLite:** 17 feedback and guest suites, 130/130 tests, including the new `guestMergeFeedback.test.js` (13).
- **Regression check:** with `backend/src` switched to `origin/main`, all 13 new tests fail, so the suite does catch the bug.
- **PostgreSQL 15** (throwaway container, a fresh database per suite):
  - `guestMergeFeedback` 13/13
  - `hiddenOwnFeedback` 8/8
  - `galleryFilterFeedbackVisibility` 12/12
  - `guestDuplicateDetection` 10/10
  - `feedbackOwnership` 4/4
  - `galleryFeedbackVisibility` 5/6. The one failure is `average_rating` returned as `"4.00"`, and it fails identically on `origin/main`.

**Not checked:** concurrent merges under real load. The lock ordering was only reviewed by reading the code.

## Verdict

Merge cleared. This needs a stable twin for 3.132.x.

PR #1525, opened 2026-09-16
fix(guests): preserve feedback ownership when merging identities (stable)

the-luap:
## Description

Stable twin of #1517, which fixes guest-identity merges. It is the same commit, `git cherry-pick -x 65047a74`, and it applied cleanly. Every file it touches is identical on `main` and `stable`: `adminGuests.js`, `galleryFeedback.js`, `feedbackService.js`, `db.js`, `dbCompat.js` and the `crmDb` test helper. No migration.

Merging guest identities moved `photo_feedback.guest_id` but left `guest_identifier` on the discarded identity. The per-photo feedback endpoint then didn't recognise transferred selections as the surviving guest's own, and overlapping selections stayed duplicated. The merge now:
- rewrites both identifiers;
- keeps the union of selections, with one visible row per photo and type (the latest value);
- preserves comments and hidden moderation rows;
- commits feedback, counters, invites and the guest updates in one transaction.

This relates to the merge defect reported in the follow-up comment on issue 1265. The full description and the review are on #1517.

## Review follow-up (second commit)

The security review of this PR found no blockers. Its concerns and nits are fixed on main in #1527, and that commit is cherry-picked here (`git cherry-pick -x 7c5815fb`) so this twin stays in parity with main. The backend files it touches are byte-identical to #1527. It adds:
- strict guest-ID parsing, a duplicate-ID rejection and a 100-guest merge bound;
- an active-survivor check in the service and numeric guest-ID comparison;
- savepoint-guarded audit inserts inside transactions, so a failing audit insert no longer diverges between PostgreSQL and SQLite;
- a `FOR SHARE` re-check so a like racing a merge is not attached to the deleted source guest;
- 15 new tests for rejected merges, permission/ownership, the failing audit insert and the concurrent like.

Re-verified on this branch: 39 feedback/guest/contract/quote/invoice/expense suites pass on SQLite (368 tests). On PostgreSQL 15, `guestMergeFeedback` (28), `hiddenOwnFeedback`, `guestDuplicateDetection`, `feedbackOwnership`, `contractService` and `invoiceService.smoke` pass.

The separate `/my-feedback` column exposure has its own pair: #1526 / #1528.

## Type of change

## How Has This Been Tested?

Re-run on this branch itself:

- **SQLite:** 17 feedback and guest suites, 130/130 tests, including the new `guestMergeFeedback.test.js` (13).
- **PostgreSQL 15** (fresh database per suite): `guestMergeFeedback` 13/13, `hiddenOwnFeedback` 8/8, `galleryFilterFeedbackVisibility` 12/12, `guestDuplicateDetection` 10/10, `feedbackOwnership` 4/4.
- **Revert check:** with `backend/src` switched back to `origin/stable`, all 13 new tests fail.
- **ESLint:** clean on the changed files.

No UI change.


the-luap:
Security review of the guest-merge change. Same patch as the merged main PR (`git patch-id` matches `65047a74`), so everything below applies to `main` too.

## 🔴 Blockers

None.

## 🟡 Concerns

1. **`/my-feedback` hands one merged identity's personal columns to the other** (pre-existing, not introduced here). `galleryFeedback.js:489` selects `photo_feedback.*` for `guest_id = req.guest.id`, which includes `guest_email`, `guest_name`, `ip_address` and `user_agent`. The merge rewrites `guest_id`/`guest_identifier` but not those columns, so after merging B into A, A's guest token receives B's email, IP and user agent. The old merge moved `guest_id` as well, so this predates the PR. Even without a merge, the guest's own IP/UA never needs to reach the browser. Fix: select the columns the gallery actually consumes.
2. **Rejection paths are untested.** `guestMergeFeedback.test.js:223` covers a deleted target and a self-merge only. Missing: guest from another event (`adminGuests.js:651`), deleted source guest, non-integer / ≤0 IDs (`adminGuests.js:634`), caller without `events.edit` / non-owner. The PostgreSQL `forUpdate()` branch (`adminGuests.js:648`) is never exercised by the SQLite CI job.
3. **Audit-log failure diverges by engine.** `logActivity(..., trx)` (`adminGuests.js:697`) swallows its own insert error (`db.js:676`). On PostgreSQL the failed statement aborts the transaction and the merge returns 500; on SQLite the merge commits without an audit row. The rollback test at line 216 mocks `logActivity` to reject, which bypasses that internal catch, so neither behaviour is proven.
4. **A like racing a merge can land on the soft-deleted source guest** (PostgreSQL). `submitFeedback` from a source guest that already passed `resolveGuest` waits on the `FOR UPDATE` lock via the FK and inserts after commit with the deleted `guest_id`. Integrity, not a leak.

## 🟢 Nits

- `adminGuests.js:629`: `.map(Number)` accepts `true`, `"0x10"`, `[[3]]`. Fails closed, but not strict.
- Duplicate IDs in `mergeIds` get the misleading "must belong to the same event" error.
- `feedbackService.js:1349` loads the survivor without `is_deleted: false`; only safe because the route checks first.
- No upper bound on `mergeIds`; `scope()` passes the full list to `whereIn` unchunked.
- `sources.includes(row.guest_id)` relies on `guest_id` being `integer`; a `bigint` column would silently zero the `merged` count.

## ✅ Verified

Directly inspected on this branch:
- Route chain `adminAuth` → `requirePermission('events.edit')` → `requireEventOwnership` (`adminGuests.js:623-625`).
- Same-event / active check runs inside the transaction, after the lock. Fixed error strings, nothing internal leaked.
- Merge scope is `event_id: survivor.event_id` + `guest_id IN (keep, ...sources)`; the duplicate delete only uses IDs from that result set. Shared colour labels carry `guest_id: null` (`feedbackService.js:305-306`) and are untouched.
- Hidden rows and comments are skipped by dedupe; `is_hidden` / `is_approved` are never written.
- Every write goes through `trx` (stats recalculation, invites, soft-delete, email canonicalisation, `logActivity`). No global `db`, no `hasColumnCached`.
- Ownership read (`galleryFeedback.js:103`) uses `req.guest?.id` from the verified token; NULL-`guest_id` rows fall back to `guest_identifier` as before.
- Like/favourite toggle-off (`feedbackService.js:591`) is scoped to visible rows of this guest/photo/event/type.

Trusted at subagent-audit level, not re-read: `requireEventOwnership` has no `admin`-role bypass; `resolveGuest` matches both `guestId` and `eventId` on a non-deleted row. PostgreSQL 15 results trusted at PR-description level.

## Verdict

No blockers. Fixing the concerns and nits in a follow-up pair (main + stable) rather than growing this cherry-pick; linked once open.

the-luap:
Follow-up to the review above: every concern and nit is fixed.

- **Concern 1** (`/my-feedback` personal columns) is a separate concern, so it has its own pair: #1526 (main) / #1528 (stable).
- **Everything else** is in #1527 on main. The same commit is cherry-picked onto this branch as `a00ba01f`, so this twin stays in parity with main.

## Concerns — closed

1. **`/my-feedback` hands one merged identity's personal columns to the other** ✅ #1526 / #1528. The route now selects named columns only; the new `myFeedbackColumns.test.js` fails with the change reverted.
2. **Rejection paths untested** ✅ `guestMergeFeedback.test.js` gains 13 tests:
   - cross-event and deleted source
   - zero, negative, fractional, non-numeric, boolean, hexadecimal, nested-array, duplicate and oversized IDs
   - non-numeric keep ID
   - viewer without `events.edit` → 403, and an editor who doesn't own the event → 403

   The suite also runs against PostgreSQL 15, which exercises the `forUpdate()` branch.
3. **Audit failure diverges by engine** ✅ `logActivity` inserts under a savepoint when given a transaction (`db.js`). New test: renaming `activity_logs` mid-test used to return 500 on PostgreSQL and 200 on SQLite; now the merge commits on both. This also covers the contract, quote, invoice and expense callers that pass a transaction.
4. **Like racing a merge** ✅ The feedback insert re-checks the guest under `FOR SHARE` in its own transaction (`feedbackService.js`) and returns `guest_missing`. The route maps that to the same `401 GUEST_IDENTITY_REQUIRED` that `resolveGuest` gives the next request. New test: the like is submitted while another transaction holds the guest and soft-deletes it. Before the fix it failed on both engines.

## Nits — closed

- ✅ **Strict IDs:** only numbers or decimal strings are accepted. `[[id]]` used to perform a merge.
- ✅ **Duplicate IDs:** they get their own 400.
- ✅ **Merge size:** bounded to 100 guests.
- ✅ **Survivor check:** `mergeGuestFeedback` loads the survivor with `is_deleted: false`.
- ✅ **ID comparison:** the merged-row count compares IDs as numbers.

## Verification

- **Before the fix:** SQLite failed 2 of the new tests and PostgreSQL 15 failed 4.
- **After the fix:** all 28 pass on both, on main and on this branch.
- **SQLite, this branch:** 39 feedback, guest, contract, quote, invoice, expense and activity suites pass (368 tests).
- **PostgreSQL 15, fresh database per suite:** `guestMergeFeedback`, `hiddenOwnFeedback`, `guestDuplicateDetection`, `feedbackOwnership`, `contractService` and `invoiceService.smoke` pass.

## Verdict

Merge cleared, together with #1527 on main.

the-luap:
Added `test(activity): read audit metadata on PostgreSQL in logActivityTrx` (`c31dc3df` on main, cherry-picked as `619c68eb` on the stable twin #1525). `activity_logs.metadata` is `text` on SQLite and `jsonb` on PostgreSQL, where the driver already returns an object, so the suite's `JSON.parse(row.metadata)` failed all four tests on PostgreSQL, with and without this PR. The assertions now read either shape. Verified 4/4 on SQLite and on PostgreSQL 15, on both branches; CI green.

PR #1527, opened 2026-09-16
fix(guests): harden guest merge validation, auditing and concurrent likes

the-luap:
## Description

Follow-up to #1517 from its security review (posted on the stable twin, #1525). No blockers were found there. This closes the concerns and nits.

**Merge route (`adminGuests.js`)**
- Guest IDs are accepted only as numbers or decimal strings. `Number()` also turned `true`, `'0x10'` and `[[3]]` into valid-looking IDs; `[[id]]` actually performed a merge.
- Duplicate `mergeIds` get their own 400 instead of the misleading "must belong to the same event".
- A merge is bounded to 100 guests, which keeps the `IN` lists small. The admin picks duplicates by hand, so this does not limit real use.

**Service (`feedbackService.js`)**
- `mergeGuestFeedback` requires an active survivor itself instead of relying on the only current caller.
- The merged-row count compares IDs numerically, so it no longer depends on the driver's integer type.
- **Concurrent like during a merge.** A like from a source guest that had already passed `resolveGuest` waited on the merge's row lock through the foreign key. It was then inserted onto the soft-deleted guest once the merge committed. The insert now re-checks the guest under `FOR SHARE` in its own transaction and returns the same `401 GUEST_IDENTITY_REQUIRED` the next request from that token would get.

**Audit log (`db.js`)**
- When `logActivity` runs with a transaction, the insert now runs under a savepoint. Before, a failed audit insert aborted the whole PostgreSQL transaction, so the merge returned 500, while SQLite committed the merge without the audit row. Now both engines commit the merge and log the error.
- This also applies to the other callers that pass a transaction: contracts, quotes, invoices and expenses.

## Type of change

## How Has This Been Tested?

**New tests** in `guestMergeFeedback.test.js`, 28 in total:
- rejected merges: cross-event, deleted source, zero, negative, fractional, non-numeric, boolean, hexadecimal, nested-array, duplicate and oversized IDs, and a non-numeric keep ID
- a viewer without `events.edit`, and an editor who doesn't own the event
- a failing audit insert
- a like submitted while a merge is deleting its guest

**Before and after the fix:**
- Before, SQLite failed the nested-array and concurrent-like tests.
- Before, PostgreSQL 15 also failed the hexadecimal and audit-insert tests. That confirms both engine-specific concerns.
- After the fix, all 28 pass on both engines.

**Regression runs:**
- SQLite: all 39 feedback, guest, contract, quote, invoice, expense and activity suites pass (368 tests).
- PostgreSQL 15 (fresh database per suite): `hiddenOwnFeedback`, `galleryFilterFeedbackVisibility`, `guestDuplicateDetection`, `feedbackOwnership`, `contractService`, `invoiceService.smoke`, `monthlyDraftInvoiceCreate` and `expenseService.markup` pass.
- `logActivityTrx` fails on PostgreSQL both with and without this change: it `JSON.parse`s a `jsonb` column, so that suite only works on SQLite.
- Full backend suite on SQLite: 3,891 passed. The only failures were the OIDC/MFA suites, because the local `node_modules` lacks `openid-client`; they fail the same way on an unmodified checkout. CI covers them.

ESLint is clean on the changed files. No UI change.


the-luap:
Added `test(activity): read audit metadata on PostgreSQL in logActivityTrx` (`c31dc3df` on main, cherry-picked as `619c68eb` on the stable twin #1525). `activity_logs.metadata` is `text` on SQLite and `jsonb` on PostgreSQL, where the driver already returns an object, so the suite's `JSON.parse(row.metadata)` failed all four tests on PostgreSQL, with and without this PR. The assertions now read either shape. Verified 4/4 on SQLite and on PostgreSQL 15, on both branches; CI green.

PR #1464, opened 2026-09-14
feat(crm): quote templates, service catalogue, contract designer, signatures v2 and the customer portal

Luca-Timo:
This PR bundles the CRM work agreed for this step: #1451 (phases 1 and 2), #1445, #1446 and #1444.

- **Quote catalogue (#1451 phase 1).**
  - Services carry a unit, a price mode, a pinned rate, a category and a VAT code.
  - There are packages, and text blocks with `{{placeholders}}`.
  - Hour and day rates come from the customer or the business default, and a line can follow the quote's hours or days.
  - Promotions are ticked in the editor, e.g. "Vereinsrabatt −300 CHF" or "−10 %".
  - Quote templates publish immutable versions.
- **Add-ons the customer books (#1451 phase 2).** Optional lines are booked or left out on the public quote page, with live totals, and the choice can be changed for a while after accepting. The portal accepts a quote with its add-ons as shown.
- **PDFs and contract templates (#1445).**
  - Quotes, invoices and contracts share one line table, and PDFs use one theme with Jost italic.
  - Every generated PDF is recorded, and a sent document opens as the file that was sent.
  - Contract templates have versions, attachments and signature slots.
- **Signatures v2 (#1446).**
  - Up to five customer signers plus the issuer, signing in parallel or in order.
  - Each signer gets their own link and email code.
  - A chained signing log that verifies, and a sealed PDF.
- **Customer portal (#1444).** Documents can be shared with the customer, and a logged-in customer answers quotes and signs contracts without a code.
- **Swiss / Liechtenstein invoices.** A service date, a due date, VAT labels, and a VAT row when the business isn't VAT-registered.

**Review status.** Five review rounds, all findings closed. *Review history* at the end lists each one with its commit and the test that pins it. Since round 5 there has been a final sweep, and quotes and contracts now use one "confirm it's you" step. *After round 5* covers both. Screenshots are below.

**No longer stacked — merged with `picpeak/main`.**
- **#1460** is merged and on main. Its request ids, idempotency key and single-request create are what the contract work here builds on.
- **#1457** was closed unmerged, so its line-item order fix for #1452 is carried here. Its review findings are applied in 36a16cc4.
- The merge also brings in **#1462**, **#1465** and **#1469**, which change the same public quote and contract surface as this branch. *Merged with main: how the overlaps were resolved* below says what was kept from each side and why.

## Screenshots

Dark mode, local instance; admin in English, customer pages in German. Taken during the final sweep, some before its last commits; the notes say where that shows.

[...]

PR #1535, opened 2026-09-17
fix(invoices): refuse a payment-check link whose expiry can't be read

Luca-Timo:
`invoice_payment_check_tokens.expires_at` is `NOT NULL` (migration 107), and both reads of it were:

```js
if (row.expires_at && new Date(row.expires_at).getTime() < Date.now()) {
```

For a value that doesn't parse — and for a missing one — that comparison is false, so the payment-check link kept working past its expiry instead of being refused.

**On the production engines this is hardening, not a live bypass.** PostgreSQL hands the column back as a `Date`, and production SQLite stores a bare `Date` as epoch ms; `new Date(...)` reads both. The shape that isn't readable is the one a service's `new Date()` produces under Jest (node-sqlite3 stores a foreign-realm Date as `"[object Object]"`), which is exactly why the guard had no test until now.

## What changed

`backend/src/services/invoice/payments.js`

- **The write** (`queuePaymentCheckEmail`) stores the expiry as an ISO string, so nothing depends on how the driver serialises a `Date`.
- **Both reads** (`getPaymentCheckByToken`, `recordPaymentCheckAction`) go through one `assertNotExpired`, which reads the column with `utils/queueTimestamps.toMillis` — a `Date`, epoch ms, ISO, or SQLite's zone-less `CURRENT_TIMESTAMP` shape — and treats a value it cannot read as **expired**. The `row.expires_at &&` short-circuit is gone with it: the column is NOT NULL, so a missing value is as unreadable as a garbled one.
- **The 24h throttle one column over** (added in round 1) had the same shape: `queuePaymentCheckEmail` wrote `invoices.last_payment_check_at` as a bare `Date` and read it back with `new Date(...).getTime()`. Under the shape SQLite keeps a foreign-realm `Date` as, that read is `NaN`, `now - NaN < 24h` is false, and the throttle stops holding — so every scheduler tick mails the admin about the same invoice again. The stamp is written as ISO and read through `toMillis`, and here `null` deliberately means **not throttled**: a stamp nobody can read must not stop payment checks for that invoice for good, and erring this way costs one extra email rather than a live link.

## Test

`backend/__tests__/integration/paymentCheckLinkExpiry.test.js`

- a live link opens with the expiry stored as ISO, and as whichever shape the engine keeps a `Date` as — epoch ms on SQLite, a `Date` on PostgreSQL, where it is a perfectly good future expiry and the link *should* open;
- an expiry in the past is refused by both entry points;
- four unreadable values are refused by both entry points — `"[object Object]"`, `"not a date"`, `""`, and a bare `Date`, which is what this suite's driver makes of one;
- the expiry `queuePaymentCheckEmail` itself writes reads back as a time, and the link it wrote opens;
- the throttle holds on the stamp the service writes — two calls in a row, the second `throttled_24h` — and a stamp nothing can read leaves the invoice mailable rather than throttled for good.

**The suite runs on both engines.** A `timestamptz` column refuses epoch ms and refuses an unparseable value at insert time, which is itself the argument that only SQLite can hold those shapes, so those cases are gated on the engine (`db.client.config.client === 'pg'`). ISO, the expired case, the token the service writes and the throttle run on both.

Each half is pinned by reverting it:

- without the read fix the unreadable expiry case fails on SQLite (`Expected: "TOKEN_EXPIRED", Received: null`);
- without the ISO write the throttle case fails — the second call sends a second email.

## Verification

- **SQLite** — `npx jest invoice payment accountingHistory workflow`: 22 suites, 219 passed, 1 skipped.
- **PostgreSQL 15** — each suite on a fresh `public` schema (they all run `migrations/core` themselves, so one shared database collides): `paymentCheckLinkExpiry` 5/5, `publicPaymentCheck` 6/6, `accountingHistory*` 21+9+11+19+8+2, `invoiceDunning` 6/6, `monthlyDraftInvoiceCreate` 7/7, `incomingInvoiceRebill` 8/8, `contractInvoiceAtomic` 2/2, `workflowRoutes` 11/11. `workflowEngine` (20 failed) and `workflowApprovalRace` (3 failed) fail on PostgreSQL — **identically at the base commit** (`f02bc9a1`), on fixtures that insert `workflow_nodes` without a `workflow_id`; nothing in this diff touches them.
- `npx eslint` clean on both changed files.
- Branched from `main` (`f02bc9a1`), not from #1464.

The `stable` twin is PicPeak/picpeak#1536 — `payments.js` is identical on both branches, so the diff is the same line for line. They should go out together.

Found while sweeping the same fail-open shape out of the quote and contract paths in #1464 (round 3), where it had hidden a blocker: a test that stored a bare `Date` passed on SQLite and failed on PostgreSQL.

## Review round 1

- **Concern 1 — the suite couldn't run on PostgreSQL.** Fixed and verified on PostgreSQL 15: the engine-specific shapes are gated, the bare `Date` is now checked as a *live* link there, and the unreadable case returns early with the reason written out (`timestamptz` cannot hold any of those values). 5/5 on both engines.
- **Concern 2 — the 24h throttle had the same shape.** Fixed in this PR rather than a follow-up, with `null` → *not throttled* as you'd keep it, and a regression test for both halves.
- **Nits.** `assertNotExpired` moved below `PAYMENT_CHECK_TOKEN_TTL_MS`, so the GHSA comment sits on the constant again; `created_at` is just `nowIso`. The 24h literal became `PAYMENT_CHECK_THROTTLE_MS` next to the TTL.


the-luap (commented):
Reviewed `530c1950` (and `444e6474` on `stable`, whose diff is identical line for line). I read the diff and removed the read fix and the write fix separately to see which test fails. I ran the new suite on SQLite and on PostgreSQL 15, plus the invoice, payment-check, workflow and accounting history suites on both branches. The change is right. One test-side problem and one adjacent spot of the same kind are left.

## 🟡 Concerns

### 1. The new suite can't run on PostgreSQL

Two of the four tests error on PostgreSQL before they reach the service, because the fixtures insert values a `timestamptz` column refuses:

```text
● a live link opens, in every shape the engines store
  date/time field value out of range: "1789896558551"
● an expiry nothing can read is refused, not treated as no expiry at all
  invalid input syntax for type timestamp with time zone: "[object Object]"
```

The bare `new Date(…)` in the unreadable list is also only unreadable under Jest + SQLite. On PostgreSQL it is a valid future expiry and the link *should* open. Gate the epoch-ms and unreadable shapes on the engine (`db.client.config.client !== 'pg'`, since `DATABASE_CLIENT` selects the engine for these suites) and keep ISO, expired and the service-written token on both. That PostgreSQL rejects those shapes is itself the argument that only SQLite can hold them.

### 2. The 24-hour throttle next to it has the same shape

`queuePaymentCheckEmail` still writes `last_payment_check_at: now` as a bare `Date` (`payments.js:287`) and reads it with `new Date(invoice.last_payment_check_at).getTime()` (`:264`). An unreadable value makes `now - NaN < 24h` false, so the throttle doesn't hold and each tick sends another email. It's the same class as the expiry, but it fails open into repeat emails rather than a link, and on the production engines the value reads back fine. Writing ISO at `:287` removes the unreadable source. Reading through `toMillis` makes the choice explicit. I'd keep `null` → *not throttled*, since the other choice would stop payment checks for that invoice for good. Fine in this PR or a follow-up, your call.

## 🟢 Nits

- The new `assertNotExpired` JSDoc went in between the "Payment-check token lifetime (GHSA-wg94-f86h-vq68 hardening)" comment and `PAYMENT_CHECK_TOKEN_TTL_MS` (`payments.js:18-43`), so that comment now sits above the function. Move the function below the constant.
- `created_at: now instanceof Date ? now.toISOString() : now` (`:284`): `now` is always `new Date()` (`:262`), so `now.toISOString()` is enough.

## ✅ Verified

- **Both halves are pinned.** Reverting `assertNotExpired` to the old `row.expires_at && new Date(…) < now` fails "an expiry nothing can read is refused". Writing `expires_at: expiresAt` bare again fails "the token the service writes is readable back".
- **No change for tokens already stored:** a PostgreSQL `Date` and a production-SQLite epoch number give the same milliseconds through `toMillis` (`queueTimestamps.js:28-29`). A stored value that can't be read now returns 410 as intended.
- **Every other writer goes through `queuePaymentCheckEmail`:** the workflow action (`workflows/actions.js:71`), the scheduler (`invoice/scheduler.js:150`) and the admin route (`adminInvoices.js:875`). There are no other `invoice_payment_check_tokens` writes in `backend/src`. The `used_at` claim stays conditional on `NULL`, so a replay is still refused.
- **Test runs:**
  - The new suite is 4/4 on SQLite on both branches, and 2/4 on PostgreSQL (concern 1).
  - The invoice, payment-check, workflow and accounting history suites are 17 suites, 189 tests (1 skipped) on each branch, SQLite.
  - Lint clean.
- An external Codex pass found nothing in the diff itself. Its follow-up on code the diff doesn't touch surfaced concerns 1 and 2, which I re-checked at file:line above.

## Verdict

Make the suite runnable on PostgreSQL (concern 1) and I'm comfortable merging both twins; the throttle can ride along or follow. Same review applies to #1536.

Luca-Timo:
@the-luap Both concerns are in, and the suite now runs on PostgreSQL 15 — `496ccf98` here, `39255dbc` on `stable` (same patch, verified with a `diff` of the two).

## Concern 1 — the suite on PostgreSQL

Gated on the engine, as you suggested, but with one addition: the bare `Date` isn't just dropped on PostgreSQL, it moves to the *other* list. Your point that it's a valid future expiry there is exactly what should be asserted, so the live-link test now takes ISO on both engines plus whichever shape that engine keeps a `Date` as — epoch ms on SQLite, a `Date` on PostgreSQL, where the link must open. The unreadable test returns early on PostgreSQL with the reason written out, since `timestamptz` rejects every one of those four values at insert time.

5/5 on both engines now (PostgreSQL 15.19 in a container, `DATABASE_CLIENT=pg`).

## Concern 2 — the throttle

Taken in this PR rather than as a follow-up, and with your call on the semantics: ISO at the write, `toMillis` at the read, `null` → **not throttled**. Stopping payment checks for an invoice for good is the worse failure, and one extra email is the price of erring the other way. The 24h literal is now `PAYMENT_CHECK_THROTTLE_MS` beside the TTL.

One honest note on what the new test pins. Reverting the **write** fails it — the second call sends a second email, which is the live defect on SQLite. Reverting the **read** alone does *not* fail it, and can't: with an ISO write, the only unreadable values are ones an old writer left behind, and on those the old `NaN` comparison and the new `null` both land on "not throttled" by design. What the `toMillis` read actually buys beyond making that choice explicit is the zone-less `CURRENT_TIMESTAMP` shape — `new Date('2026-09-17 10:06:09')` reads as *local* time, which on a non-UTC deployment shifts the 24h window by the offset, while `toMillis` stamps the zone the value carries. That difference is invisible in a UTC test run, so I've left it unpinned rather than pretend otherwise; the child-process-with-`TZ` trick from the `queueTimestamps` suite is the way to cover it if you want it.

## Nits

- `assertNotExpired` moved below `PAYMENT_CHECK_TOKEN_TTL_MS`, so the GHSA comment sits on the constant it describes again.
- `created_at` is a plain `nowIso` (one `const` now, used for the token row and both invoice columns).

## Runs

- **SQLite** — `npx jest invoice payment accountingHistory workflow`: 22 suites, 219 passed, 1 skipped. Same on both branches.
- **PostgreSQL 15** — per suite on a fresh `public` schema, since `bootCrmDb` runs `migrations/core` itself and one shared database collides between suites: `paymentCheckLinkExpiry` 5/5, `publicPaymentCheck` 6/6, the six `accountingHistory*` suites 70/70, `invoiceDunning` 6/6, `monthlyDraftInvoiceCreate` 7/7, `incomingInvoiceRebill` 8/8, `contractInvoiceAtomic` 2/2, `workflowRoutes` 11/11.
- Two failures I'm not fixing here: `workflowEngine` (20) and `workflowApprovalRace` (3) fail on PostgreSQL on fixtures that insert `workflow_nodes` with no `workflow_id`. I ran them at the base commit `f02bc9a1` in a separate worktree — **identical failures**, so they're pre-existing and outside this diff.
- `npx eslint` clean on both changed files.

Same changes on #1536; the PR bodies of both are updated.

the-luap (commented):
Re-verified `496ccf98`. Both concerns are closed and the honesty note about what the test pins is exactly right.

## Concerns — closed

| # | Where | Check |
|---|---|---|
| 1 | `paymentCheckLinkExpiry.test.js` gated on `isPg()` | **5/5 on SQLite and 5/5 on PostgreSQL 15**, no shim. Moving the bare `Date` into the live-link list on PostgreSQL is the right call — there it is a valid future expiry, and asserting the link opens is worth more than skipping it. ✅ |
| 2 | `payments.js:275` read through `toMillis`, `:300` ISO write, `PAYMENT_CHECK_THROTTLE_MS` | Reverting the write fails "the 24h throttle holds on the stamp the service writes". ✅ |

`null` → not throttled is the right way round, for the reason you give: the alternative strands an invoice's payment checks for good, and the cost here is one extra email per tick.

**Your note on what the test pins holds.** I reverted the read on its own (`new Date(...).getTime()` with no null check) and the suite still passes 5/5, as you said it would. With an ISO write the only unreadable values are ones an older writer left behind, and on those the old comparison and the new one both land on "not throttled". So the read change buys the zone-less `CURRENT_TIMESTAMP` shape and an explicit choice, not a behaviour a test can catch in a UTC run. Saying so beats a test that looks like it covers something it doesn't. The `TZ` child-process variant is worth having, but in the `queueTimestamps` suite where that shape belongs, not here.

## 🟢 Nit

`last_payment_check_at` had to become ISO, but `updated_at` went with it in the same update (`:301`). Every other writer of `invoices.updated_at` still writes a bare `Date`, so on SQLite that column now holds epoch numbers from every other path and ISO strings from this one. Nothing compares it, and the only reader that orders by it is the table-stats panel in `adminSystem.js:392` ("last update" per table, where SQLite sorts all numbers before all text). Harmless, but it widens the shape drift past the column the concern was about — I'd leave `updated_at: now`.

## ✅ Verified

- New suite: 5/5 on SQLite and PostgreSQL 15, on this branch and on #1536.
- `npx jest invoice payment accountingHistory workflow`: **22 suites, 219 passed, 1 skipped** — matches your numbers.
- Reverting the ISO write fails the throttle test; reverting the read alone does not (above).
- `assertNotExpired` now sits below `PAYMENT_CHECK_TOKEN_TTL_MS`, so the GHSA comment is back on its constant, and `created_at` is a plain `nowIso`. Lint clean on both changed files.
- The `workflowEngine` / `workflowApprovalRace` PostgreSQL failures: taken at your description-level word that they reproduce at `f02bc9a1`. I didn't re-run them, since neither suite touches this diff.

## Verdict

Merge cleared. #1536 is identical line for line and gets the same verdict.

PR #1536, opened 2026-09-17
fix(invoices): refuse a payment-check link whose expiry can't be read (stable)

Luca-Timo:
`invoice_payment_check_tokens.expires_at` is `NOT NULL` (migration 107), and both reads of it were:

```js
if (row.expires_at && new Date(row.expires_at).getTime() < Date.now()) {
```

For a value that doesn't parse — and for a missing one — that comparison is false, so the payment-check link kept working past its expiry instead of being refused.

**On the production engines this is hardening, not a live bypass.** PostgreSQL hands the column back as a `Date`, and production SQLite stores a bare `Date` as epoch ms; `new Date(...)` reads both. The shape that isn't readable is the one a service's `new Date()` produces under Jest (node-sqlite3 stores a foreign-realm Date as `"[object Object]"`), which is exactly why the guard had no test until now.

## What changed

`backend/src/services/invoice/payments.js`

- **The write** (`queuePaymentCheckEmail`) stores the expiry as an ISO string, so nothing depends on how the driver serialises a `Date`.
- **Both reads** (`getPaymentCheckByToken`, `recordPaymentCheckAction`) go through one `assertNotExpired`, which reads the column with `utils/queueTimestamps.toMillis` — a `Date`, epoch ms, ISO, or SQLite's zone-less `CURRENT_TIMESTAMP` shape — and treats a value it cannot read as **expired**. The `row.expires_at &&` short-circuit is gone with it: the column is NOT NULL, so a missing value is as unreadable as a garbled one.
- **The 24h throttle one column over** (added in round 1) had the same shape: `queuePaymentCheckEmail` wrote `invoices.last_payment_check_at` as a bare `Date` and read it back with `new Date(...).getTime()`. Under the shape SQLite keeps a foreign-realm `Date` as, that read is `NaN`, `now - NaN < 24h` is false, and the throttle stops holding — so every scheduler tick mails the admin about the same invoice again. The stamp is written as ISO and read through `toMillis`, and here `null` deliberately means **not throttled**: a stamp nobody can read must not stop payment checks for that invoice for good, and erring this way costs one extra email rather than a live link.

## Test

`backend/__tests__/integration/paymentCheckLinkExpiry.test.js`

- a live link opens with the expiry stored as ISO, and as whichever shape the engine keeps a `Date` as — epoch ms on SQLite, a `Date` on PostgreSQL, where it is a perfectly good future expiry and the link *should* open;
- an expiry in the past is refused by both entry points;
- four unreadable values are refused by both entry points — `"[object Object]"`, `"not a date"`, `""`, and a bare `Date`, which is what this suite's driver makes of one;
- the expiry `queuePaymentCheckEmail` itself writes reads back as a time, and the link it wrote opens;
- the throttle holds on the stamp the service writes — two calls in a row, the second `throttled_24h` — and a stamp nothing can read leaves the invoice mailable rather than throttled for good.

**The suite runs on both engines.** A `timestamptz` column refuses epoch ms and refuses an unparseable value at insert time, which is itself the argument that only SQLite can hold those shapes, so those cases are gated on the engine (`db.client.config.client === 'pg'`). ISO, the expired case, the token the service writes and the throttle run on both.

Each half is pinned by reverting it:

- without the read fix the unreadable expiry case fails on SQLite (`Expected: "TOKEN_EXPIRED", Received: null`);
- without the ISO write the throttle case fails — the second call sends a second email.

## Verification

- **SQLite** — `npx jest invoice payment accountingHistory workflow`: 22 suites, 219 passed, 1 skipped.
- **PostgreSQL 15** — each suite on a fresh `public` schema (they all run `migrations/core` themselves, so one shared database collides): `paymentCheckLinkExpiry` 5/5, `publicPaymentCheck` 6/6, `accountingHistory*` 21+9+11+19+8+2, `invoiceDunning` 6/6, `monthlyDraftInvoiceCreate` 7/7, `incomingInvoiceRebill` 8/8, `contractInvoiceAtomic` 2/2, `workflowRoutes` 11/11. `workflowEngine` (20 failed) and `workflowApprovalRace` (3 failed) fail on PostgreSQL — **identically at the base commit** (`f02bc9a1`), on fixtures that insert `workflow_nodes` without a `workflow_id`; nothing in this diff touches them.
- `npx eslint` clean on both changed files.
- Cherry-picked onto `stable` (`7c2620fb`); the diff is identical to PicPeak/picpeak#1535 line for line (verified with `diff` of the two patches).

**The `stable` twin of PicPeak/picpeak#1535**, cherry-picked onto `stable` (`7c2620fb`) — `payments.js` is identical on both branches, so the diff is the same. They should go out together.

Found while sweeping the same fail-open shape out of the quote and contract paths in #1464 (round 3), where it had hidden a blocker: a test that stored a bare `Date` passed on SQLite and failed on PostgreSQL.

## Review round 1

- **Concern 1 — the suite couldn't run on PostgreSQL.** Fixed and verified on PostgreSQL 15: the engine-specific shapes are gated, the bare `Date` is now checked as a *live* link there, and the unreadable case returns early with the reason written out (`timestamptz` cannot hold any of those values). 5/5 on both engines.
- **Concern 2 — the 24h throttle had the same shape.** Fixed in this PR rather than a follow-up, with `null` → *not throttled* as you'd keep it, and a regression test for both halves.
- **Nits.** `assertNotExpired` moved below `PAYMENT_CHECK_TOKEN_TTL_MS`, so the GHSA comment sits on the constant again; `created_at` is just `nowIso`. The 24h literal became `PAYMENT_CHECK_THROTTLE_MS` next to the TTL.


the-luap (commented):
Stable twin of #1535. The diff is identical line for line to `530c1950` on `main`, and I ran the same checks on this branch: the new suite is 4/4 on SQLite and 2/4 on PostgreSQL for the same fixture reason, and the invoice, payment-check, workflow and accounting history suites are 17 suites, 189 tests green. The review on #1535 applies unchanged here. Please carry whatever lands there (the PostgreSQL gating, and the throttle if you take it) into this branch too, so the twins stay in sync.

Luca-Timo:
@the-luap The `stable` twin carries both fixes from your #1535 review — `39255dbc`, the same patch as `496ccf98` there (verified with a `diff` of the two, identical line for line).

- **Concern 1** — the engine-specific expiry shapes are gated on `db.client.config.client === 'pg'`, and the bare `Date` moves to the live-link list on PostgreSQL rather than being dropped, since there it *is* a valid future expiry. 5/5 on SQLite and on PostgreSQL 15.19.
- **Concern 2** — the 24h throttle writes `last_payment_check_at` as ISO and reads it through `toMillis`, with `null` → not throttled, plus a regression test.
- **Nits** — `assertNotExpired` moved below `PAYMENT_CHECK_TOKEN_TTL_MS`; `created_at` is a plain `nowIso`.

Runs on this branch: `npx jest invoice payment accountingHistory workflow` → 22 suites, 219 passed, 1 skipped on SQLite; `paymentCheckLinkExpiry` 5/5 on PostgreSQL 15; eslint clean. The reasoning behind the throttle's `null`, and what the new test does and does not pin, is written out in the [#1535 reply](https://github.com/PicPeak/picpeak/pull/1535#issuecomment-5717088218) — it applies unchanged here.

the-luap (commented):
Re-verified `39255dbc` as the twin of `496ccf98`: I diffed both branches against their bases and the changes are identical line for line.

Ran on this branch: `paymentCheckLinkExpiry` **5/5 on SQLite and 5/5 on PostgreSQL 15**, and `npx jest invoice payment accountingHistory workflow` at **22 suites, 219 passed, 1 skipped**. Lint clean.

The review is on [#1535](https://github.com/PicPeak/picpeak/pull/1535#pullrequestreview-5235034433) — both concerns closed, one nit about `updated_at` going ISO along with `last_payment_check_at`. It applies here unchanged.

Merge cleared, together with its main twin.

PR #1539, opened 2026-09-18
feat(crm): customer groups with colour chips, filtering and archiving

the-luap:
Customer groups for the CRM, as described in issue 1443. An admin manages a group catalogue, puts customers in as many groups as they like, and reads the customer overview by them.

## What changes

**Migration 226.** `customer_groups` (name, description, colour, sort order, archived) and `customer_group_members`, the many-to-many join. A group is referenced by id everywhere, so a rename or a recolour shows up wherever the chip is drawn without rewriting a customer record. Reading the catalogue takes `customers.view`, because that is what it takes to see the customers it organises; changing it takes the new `customers.groups.manage`, granted to super_admin and admin, so a role that may edit a customer record doesn't silently gain the catalogue everyone's overview is filtered by.

**Service.** Create, rename, recolour, reorder, archive, restore, and replacing a customer's groups in one call. Names are unique regardless of case: uniqueness hangs on a `name_key` column folded in JS with a unique index on it, because SQL `LOWER()` is ASCII-only on SQLite and "Ärzte" / "ärzte" have to be one group on both engines. Colours must be hex. Every change goes to `activity_logs`, with the admin who made it as the actor.

**Two rules the rest of the code leans on:**
- an **archived** group stays on the customers that carry it and is refused for a new assignment (`GROUP_ARCHIVED`). That is how a group in use is retired;
- **deleting** a group with members is refused (`GROUP_IN_USE`). The check and the delete are one statement, and the group side of the join is `RESTRICT`, so an assignment that lands in between is refused by PostgreSQL rather than cascaded away. A delete can never reach a customer record;
- **erasing** a customer clears their memberships, so an anonymised row doesn't keep counting as a member.

**API.** The customer list and detail responses carry their groups. The list takes `?groupIds=1,2` (repeatable or comma-separated) and a customer matches **any** selected group, which is what "show me these groups" means in an overview. It is a subquery rather than a join, so the existing event count stays the number of events rather than the number of (event × group) pairs, and it composes with the existing search.

**Admin UI.** A Groups tab on the customers page for the catalogue, chips on each row, a multi-select filter with a clear action, and a Groups card on the customer detail page.

**Two decisions worth your eye:**
- the colour is a **dot beside the name**, never the background of the text and never the only carrier of meaning. That is what makes a chip readable in both themes whatever colour an admin picks, and it is why the same markup needs no light/dark variant of its own;
- reordering is **up/down buttons**, not drag handles: it works with a keyboard and on a phone.

## Screenshots

Light and dark, 1440px, plus 390px for the phone. Taken against the local rig with five seeded customers and five groups, one of them archived.

**1. Overview** — chips per row, the filter pills with their member counts
Customer overview with group chips

**2. Filtered by one group** — the tab count follows, and a Clear action appears
Overview filtered to the Corporate group

**3. The catalogue tab** — member counts, reorder, archive, delete. Delete is disabled while customers are still in a group; the archived group says so and is deletable because it is empty
Group catalogue

**4. Customer detail** — the groups this customer is in, with the picker behind "Change groups"
Groups card on the customer detail page

**5. Dark mode**
Overview in dark mode

**6. Phone width** — name, groups and state in one column, and no horizontal scroll left. Retaken after the review fixes: the state under the name is the same block as the status column, so a passive customer's badge shows on a phone too
Overview at 390px

### Mobile layout, fixed in `0ab93e07`

The first round of screenshots showed three things wrong at 390px, two of them older than this branch:

- the **tabs row spilled out of the card**. It wraps now rather than running off the side;
- **"Create passive customer" was wider than the screen** and cut off. The header stacks below `sm` and both buttons take the full row;
- **the table needed sideways scrolling to read anything**. The columns that only make sense side by side — company, groups, events, last login, status and the row action — step aside below `sm`, and the name cell carries the groups and the state underneath the name. A phone row is now name, groups, state and email, with the horizontal scroll measured at **0px, down from 288px**.

Deactivate is on the customer detail page too, which is where the row links, so hiding that button on a phone takes nothing away. The frontend test pins which columns hide at which breakpoint and that the name cell carries the phone copy of the groups.

## Tests

- **Backend, 14 integration tests** through the real router, so the permission and validation layers are part of what is pinned: the catalogue, case-insensitive names, hex colours, assignment to none/one/several groups, the overview and detail payloads, a rename that leaves the customer record alone, archived behaviour both ways, the filter with and without the search, the refused delete (and that no customer goes with it), `customers.view` reading but not writing, erasure clearing the memberships, and the log with its actor. Green on **SQLite and PostgreSQL 15**.
- **Frontend, 19 tests** (16 for the page, 3 for the detail card): the overview opening on everyone, the filter through the server, one group then two then cleared, the empty-filter message, archived groups offered on a row but not as a filter, the catalogue tab's loading, empty and error states plus the disabled delete, and the responsive layout described above.
- **Migration 226 is idempotent and round-trips** (`up` twice, `down` twice, `up` again) on both engines, permission grant included.
- `npm run build`, ESLint and the whole admin frontend suite (34 files, 197 tests) are clean.

## Upgrade

Additive. Existing customers keep working with no group: the column shows an em dash, and an unfiltered overview lists everyone exactly as before. Nothing is enabled or hidden by a feature flag — the tab appears with the CRM surface the admin already has.

## Notes

- `customer_groups` is admin organisation rather than billing data, so it is deliberately **not** in the accounting change history (migration 219). Group changes are recorded in `activity_logs`.
- The issue asks for tests covering light and dark mode. There is no precedent for asserting a theme in this repo's test suite, and the chips carry no theme-specific markup by design (see the decision above), so the themes are covered by the screenshots rather than by a test.
- Pushed with `PUSH_SKIP_E2E=1`: the pre-push E2E script isn't executable in this worktree.


the-luap:
Self-review at `0ab93e07` before this goes in. One blocker, seven concerns; fixes follow on this branch.

## 🔴 Blockers

### 1. Group activity-log rows carry no actor
`customerGroupsService.js:119,152,174,192,272` — all five call `logActivity(type, meta, null, adminId)` with a bare number. `logActivity` (`db.js:648`) reads `actor?.id`, `actor?.name` and `actor?.type || 'system'`, so every row is stored as `actor_type='system'`, `actor_id=null`, `actor_name=null`. The body says every change goes to `activity_logs`; none of them says who made it, and rows already written can't be repaired. Fix: pass `{ type: 'admin', id, name }` the way `adminActor(req.admin)` does for the document routes. The log test (`customerGroups.test.js:268-285`) plucks `activity_type` only and isn't scoped to its own group, so it can't see this.

## 🟡 Concerns

1. `customerGroupsService.js:164-173` — the delete guard counts members, then deletes, with no transaction, while migration 226:76 has `onDelete('CASCADE')` on `group_id`. An assignment landing in between is removed silently on PostgreSQL, which is exactly what the body says can't happen. Make the delete conditional in one statement.
2. `customerGroupsService.js:64-70` — `LOWER(name)` in SQL against `name.toLowerCase()` in JS. SQLite's `LOWER()` is ASCII-only (`select lower('Ärzte')` → `Ärzte`), so with `Ärzte` in the catalogue, `ärzte` is accepted. A same-case concurrent create also surfaces `customer_groups_name_unique` as a raw 500. A JS-normalised key column with a unique index covers both.
3. `CustomerManagementPage.tsx:171,249` — the Groups tab renders `CustomerGroupsPanel` for everyone. With `customers.view` only, every control is there and every one returns 403. The detail card in this same PR is gated (`CustomerDetailPage.tsx:465`).
4. `CustomerManagementPage.tsx:75,88` + `CustomerGroupChips.tsx:96` — `groupFilter` is never pruned, and the filter returns `null` once no live group is left, Clear button included. Archive the last live group while it is selected and the list stays filtered with nothing to reset it.
5. `CustomerManagementPage.tsx:299-311,319` — the phone copy of the status shows Active / Deactivated only. "Invitation pending" and "Passive — admin only" (issue 1261) live in the `hidden sm:table-cell` and are gone on a phone.
6. Erased customers keep their memberships (migration 226:18-20 says so on purpose). The anonymised row still counts in `memberCount` and blocks deleting the group, and a group name can be personal. Erasure should clear them.
7. `en.json:5941` — `memberCount` has no `_one` form; `customerGroups.test.tsx:207` asserts `'1 customers'` and pins it.

## 🟢 Nits

- `setCustomerGroups` reads outside its transaction — two concurrent replaces → 500, no corruption.
- The viewer-role test asserts 403 for POST and assign only.
- The frontend test mocks `hasPermission: () => true`, which is how concern 3 went unnoticed.

## ✅ Verified

Directly inspected: migration 226 (guarded, additive, 225 grant pattern, sane `down()`; 226 is free, 227 is taken by the permission-boundary PR); route order and guards in `adminCustomers.js:161-230`; description cap and customer-exists check in the service. Ran locally on SQLite: backend 11/11, frontend 10/10.

Trusted at audit level, not re-read line by line: PG string-count coercion, no schema introspection inside a trx, ISO timestamps, the archived-group re-send logic, en/de key parity, cache invalidation. Trusted at description level: the PostgreSQL 15 run and the migration round-trip.

## Verdict

Close the blocker and concerns 1–7 and this is clear to merge.

the-luap:
All of the above is closed in `34aa6c78`.

## Blocker — closed

| | Where | |
|---|---|---|
| 1. Log rows without an actor | `customerGroupsService.js` `adminActor()`; the routes hand over `req.admin` instead of `req.admin.id` | ✅ |

The log test now reads only its own group's rows and asserts `actor_type: 'admin'`, `actor_id` and a name on each. Swapped the old bare-id shape back in to check: it fails.

## Concerns — closed

| | Where | |
|---|---|---|
| 1. Delete guard vs. CASCADE | `remove()` is one conditional `DELETE … WHERE NOT EXISTS`; migration 226 `group_id` is `RESTRICT`; `23503` maps to `GROUP_IN_USE` | ✅ |
| 2. Umlauts / raw 500 on a lost race | `name_key` (folded in JS) with `customer_groups_name_key_unique`; unique violations map to `GROUP_NAME_TAKEN` on create and rename | ✅ |
| 3. Ungated Groups tab | `CustomerGroupsPanel` takes `canManage`; without it the catalogue is read-only | ✅ |
| 4. Stuck filter | `activeGroupFilter` in `CustomerManagementPage.tsx` — the selection intersected with what the filter still offers, used for the query, the pills and the empty message | ✅ |
| 5. Phone row missing badges | `renderStatus()` is the one block behind both the status column and the phone copy | ✅ |
| 6. Erasure | `eraseCustomer` deletes the customer's `customer_group_members` rows inside its transaction | ✅ |
| 7. Plural | `memberCount_one` / `_other` in en and de, `defaultValue_one/_other` at the call site | ✅ |

On concern 1: a single statement is enough on SQLite, not on PostgreSQL — a concurrent insert holds only a key-share lock on the group row, so the delete proceeds on its earlier snapshot. That window is what `RESTRICT` is for. Checked against a throwaway `postgres:15`: migration 226 `up/up/down/down/up` clean, conditional delete with a member → 0 rows, plain delete → `23503` with the membership intact, duplicate `name_key` → `23505`.

Migration 226 changed in place (new column, `RESTRICT`). It hasn't shipped, so that is safe for every install — but a dev database that already ran the earlier 226 has it in the ledger and needs `down` + `up` once.

## Tests

- Backend: 12 in `customerGroups.test.js` (umlaut + rename-into-taken + the index itself, membership intact after a refused delete, FK is `RESTRICT`, erasure, actor). With the suites that exercise erasure: 76 passed.
- Frontend: 13 in `customerGroups.test.tsx`; the four new or changed ones fail against the previous components. `customerInvitationVisibility.test.tsx` now expects the badge twice per row — status column plus phone copy — and checks the tooltip on both. Admin suite: 48 files / 294 tests.
- `npm run build`, ESLint and `tsc` clean on the touched files. `usageCoverageInventory` passes.

Not re-run by me: the full backend suite on PostgreSQL — CI covers that.

Screenshot 6 (390px) predates the badge under the name on a phone row and wants retaking.

the-luap:
Nits closed in `9bb238e5`.

| | Where | |
|---|---|---|
| `setCustomerGroups` read outside its transaction | `customerGroupsService.js` — the diff is worked out from rows read inside the trx; a lost unique-index race is `409 GROUP_ASSIGNMENT_CONFLICT` | ✅ |
| Unbounded id lists | `adminCustomers.js` — `groupIds` cut at 100 on the filter, refused above 100 on assignment, `orderedIds` refused above 500; `groupsForCustomers` looks up in slices of 500 | ✅ |
| Viewer test covered POST and assign only | now PUT, reorder and DELETE too, and the group is unchanged afterwards | ✅ |
| "no longer exists" comment on a test that sent `abc` | both cases are in: a non-id is dropped, a stale id filters to an empty list | ✅ |
| Vacuous `deleteGroup` assertion | the disabled button is clicked; an empty group deleting after its confirmation is its own test | ✅ |
| `CustomerGroupsCard` untested | `CustomerGroupsCard.test.tsx`: read-only without the permission, archived-but-carried stays listed and checked, archived-not-carried isn't offered, one replace call with the whole set, cancel restores | ✅ |
| Palette buttons announced hex values | named colours, `customers.groups.palette.*` in en and de; the key is a literal union so `i18n:extract` can follow it | ✅ |
| Archive clickable while its request runs | disabled on `updateGroup.isPending` | ✅ |
| Spinner flash on every filter toggle | `placeholderData: keepPreviousData` | ✅ |
| Unused `onRemove` / `removeFrom` | removed, key dropped from en and de | ✅ |

The assignment race, for real: 25 rounds of six concurrent `setCustomerGroups` for one customer through the service against a throwaway `postgres:15` — 28 succeeded, 122 got `409 GROUP_ASSIGNMENT_CONFLICT`, none surfaced a raw error, and every round ended with exactly the three membership rows. On SQLite the single connection serialises the saves, so the suite pins the mapping by failing the transaction the way the index does, plus two saves at once ending in one row.

Two I left, on purpose:

- **es / fr / nl / pt / ru / sl.** Those files are 4,300 keys behind en each and carry 0–109 of the 310 `customers.*` keys; the admin CRM has only ever shipped in en and de and falls back to English elsewhere. Adding 40 machine-translated keys there would be the first of their kind, unreviewed.
- **English backend error messages.** That is the convention across the services (`'Customer not found'`, `'Customer group not found'`); the stable part is the `code`.

Backend: 14 in the groups suite, 82 with the adjacent suites. Frontend: admin suite 49 files / 299 tests. Build, ESLint and `tsc` clean on the touched files.

the-luap (commented):
## 🔴 Blockers

None.

## 🟡 Concerns

1. `is_archived` queries use raw `false` instead of `formatBoolean()` (`customerGroupsService.js`). Tests pass on SQLite today; still a footgun vs repo convention.
2. Archive-add guard `added.includes(g.id)` is not `Number()`-normalized while `currentIds` is — integer PKs make this fine today; defensive `Number(g.id)` would match the rest of the service.

## 🟢 Nits

- `reorder` `known.includes(id)` same class as above.
- Route `description` has no `.isLength({ max: 500 })`; service enforces.
- Concurrent `create` can race on `max(sort_order)+1` — duplicate sort orders only.

## ✅ Verified

- Migration 226: `hasTable` guards; unique `name_key` + membership pair; group FK `RESTRICT`, customer FK `CASCADE`; `down()` ordered; permission seeded for `super_admin`/`admin` only.
- Authz: writes need `customers.groups.manage`; GET uses `customers.view`; viewer role integration test pins 403; `PUT /:id` cannot mass-assign `groupIds`.
- Color XSS: `normalizeColor` = `/^#[0-9a-f]{6}$/i`; React text + `style.backgroundColor` only.
- Name race: assert + unique + `isUniqueViolation` → 409.
- Filter: `parseGroupIds` → positive ints, cap 100; bound `whereIn`.

## Verdict

Merge-cleared. Close the `formatBoolean` / `Number(g.id)` concerns if you want belt-and-suspenders.

the-luap:
Review items closed in `319534c6`.

## Concerns — closed

| | Where | |
|---|---|---|
| 1. Raw `false` on `is_archived` | `customerGroupsService.js:105` (list), `:122` (create), `:166` (update, which also wrote a raw boolean) now go through `formatBoolean()` | ✅ |
| 2. Archive-add guard not normalised | `customerGroupsService.js:294`, `added.includes(Number(g.id))` | ✅ |

## Nits

| | Where | |
|---|---|---|
| `reorder` id comparison | `customerGroupsService.js:220-221`, both sides `Number()` | ✅ |
| Route description cap | `adminCustomers.js:180` (POST) and `:205` (PUT), `.trim().isLength({ max: 500 })`. Trimmed first because the service trims before its own check; without it a padded description the service accepts would be refused at the route | ✅ |
| Concurrent `create` racing on `max(sort_order)+1` | left as is: the only outcome is two groups with the same position, which the next reorder fixes | — |

New case in `customerGroups.test.js`: 501 characters on create and update → 400 from the route (not the service's `GROUP_DESCRIPTION_TOO_LONG`), 500 with surrounding whitespace accepted. Fails with the route change reverted. `customerGroups` 15/15, the 12 customer/CRM suites 189/189.

PR #1540, opened 2026-09-19
fix: tighten admin permission boundaries (backup manifest, settings writers, SSO email linking, Super Admin accounts, workflow emails)

Luca-Timo:
Tightens five places where a delegated admin role could reach something the permission model reserves for a Super Admin (or for a narrower permission). Each part carries a regression test, and each test fails with its fix removed.

## What changed

**Backup manifest location** (`adminBackup.js`, `backupService.js`, `BackupConfiguration.tsx`)
- `backup_manifest_path` and `backup_manifest_format` decide where the manifest file is written, so they now join the Super-Admin-only destination settings (`DESTINATION_SETTING_RE`).
- `backup_manifest_format` must be `json` or `yaml` (400 otherwise). The service also only uses those two values, and reduces the manifest file name to its basename, for both the local and the S3 manifest. The S3 path's unreachable `xml` branch is gone.
- The backup form's mirror of the restricted keys includes the two manifest keys, so a non-super admin's save leaves them out like the destination fields.

**Generic settings writers** (`adminSettings.js`)
- `PUT /api/admin/settings/{general,seo,analytics,security}` answer **400 `BACKUP_SETTINGS_ELSEWHERE`** (with the key list) for `backup_*` / `database_backup_*`. Those keys belong to `/admin/backup/config` and `/admin/database-backup/config`, which apply their own rules. The generic reads still return them.
- `analytics_tracker_provider`, `analytics_umami_url`, `analytics_rybbit_url` and `analytics_custom_head_html` are now protected keys requiring **`settings.integrations`** on every generic writer. They decide which script the tracker proxy serves from the app's own origin. Protection uses the existing change-detection, so an unchanged round-trip still saves; a provider with no row yet compares as the Analytics tab derives it (from the legacy `analytics_umami_enabled`), so a first save of an untouched tab does not 403.

**Tracker script in the admin UI** (`analytics.service.ts`)
- The Umami/Rybbit script now gets the same treatment the custom head HTML already has: it is not loaded while an admin route is shown, and entering the admin UI after it ran reloads into a clean document.

**SSO email linking** (`oidcService.js`, `adminAuth.js`, `auth.js`, `picpeakImportService.js`, `AdminLoginPage.tsx`, `en.json` / `de.json`)
- First-login linking by email matched `admin_users.email`, which is never verified and which an admin can change on their own profile. New column **`admin_users.email_link_eligible`** records whether the email was set by a trusted flow; email linking now requires it.
- A profile email change by anyone but a super_admin clears it. A super_admin saving a user's email in user management (changed or not) sets it — that is how linking is re-enabled.
- If the only matching row is not eligible, the callback redirects with **`sso_error=email_unverified`** (new en/de message) instead of falling through to JIT provisioning and colliding on the unique email.
- A `.picpeak` restore keeps the operator's `email_link_eligible`, like the other auth columns in `PRESERVED_AUTH_FIELDS`.

**Super Admin accounts** (`userManagementService.js`)
- Only a super_admin may change a super_admin account through `PUT /api/admin/users/:id` (`users.edit`) and `POST /:id/deactivate`, `POST /:id/activate`, `DELETE /:id` (`users.delete`) — the same line invitations, role assignment and password resets already hold.
- `PUT`'s `is_active` change applies the self-deactivation and last-super-admin guards the deactivate route has, and reads `'false'` as false (it was stored as active on SQLite).

**Workflow emails** (`workflows/actions.js`, `workflows/approvals.js`)
- `send_email` and the approval-gate email no longer take `attachments` from node config or run vars (run vars can come from a test-run payload). Attachments are file paths the mailer reads from disk; actions that attach documents (e.g. `escalate_to_collections`) build that list themselves from stored document paths and are unchanged.

## Migration 227

`227_admin_email_link_eligible.js` adds `admin_users.email_link_eligible boolean NOT NULL DEFAULT true`, `hasColumn`-guarded both ways. Existing rows get `true`, so SSO linking keeps working on upgrade. Every insert path (wizard, invite, CLI, JIT, restore) leaves it at the default. Reads go through `hasColumnCached` + `formatBoolean`. **227, not 226:** #1539 is open with `226_customer_groups.js`.

## Decisions

- **Existing installs stay eligible.** An email an admin already self-changed before this migration can't be told apart, so it stays linkable. The alternative (default false) would break SSO linking for every existing admin.
- **`analytics_umami_enabled` stays unprotected.** It only switches between tracker URLs someone with `settings.integrations` set, and the tab re-derives it on every save; protecting it would 403 ordinary saves.
- **Backup keys are refused loudly on the generic writers**, not dropped, in line with how protected keys already 403 instead of silently doing nothing. No frontend, wizard or other route writes them there.

## Behaviour changes on upgrade

- A custom role with `settings.edit` but not `settings.integrations` gets 403 when it *changes* the tracker provider/URLs or custom head HTML. Presets are unaffected: super_admin and solo_photographer hold `settings.integrations`, the built-in `admin` role never had `settings.edit`.
- An admin whose email was changed on their own profile (or by a non-super admin) is not linked by SSO email matching until a super_admin saves that email.
- A non-super admin can no longer set the backup manifest path/format.

## Testing

- **New/extended suites:** `backupRestoreSuperAdminOnly`, `adminSettingsProtectedKeys`, `adminUsersSuperAdminTarget` (new), `oidcSso`, `workflowSendEmailAttachments` (new), `workflowApprovalEmailAttachments` (new), `src/routes/__tests__/adminAuth` (mock sequence updated for the extra lookup), frontend `analytics.trackerProxy`.
- Each fix was reverted in turn and its test re-run: every one fails without the fix.
- Full backend Jest locally (SQLite): 418 suites pass. `storageBackend` and `imageProcessor.storage` fail because they need MinIO; they fail identically on `main` (checked by running them there).
- Frontend: analytics + BackupConfiguration Vitest pass; `npm run build` passes; eslint clean on changed files (the one remaining warning in `AdminLoginPage.tsx` is on an untouched line).
- Migration 227 run on SQLite; Postgres not run locally — the column is a plain `boolean NOT NULL DEFAULT true` read via `formatBoolean`.
- Review depth: backend read in full; frontend at diff level.

## Notes for review

- **Size gate:** the diff is ~720 lines (about half tests), over the `bypass-size-gate` threshold — that's the review-size policy, not a code failure. The five parts are separate commits if you'd rather review them one by one.
- **Screenshots:** the only UI-visible changes are the new SSO error toast text and two backup-form fields that are no longer saved for non-super admins; no layout changes. Happy to add screenshots if useful.
- **Locales:** `email_unverified` added in `en` and `de`, matching the other `ssoErrors` keys, which only exist in those two.
- **Follow-up, not in this PR:** the last-super-admin guard is count-then-write in four places (demote, deactivate, delete, and now PUT), so two concurrent requests could still pass it; only `oidcService.syncAdminRole` does it under a row lock. Moving all four onto one locked helper is a separate change.


the-luap:
Reviewed at `340d0b4`. Backend read at source level throughout; frontend at diff level plus the analytics suite run locally. No blockers — the five boundaries hold under a bypass sweep. Two concerns worth an explicit answer before merge.

## 🔴 Blockers

None.

## 🟡 Concerns

### 1. `analytics_umami_enabled` is the master switch, not a selector

`adminSettings.js` leaves `analytics_umami_enabled` out of `TRACKER_CODE_KEYS` on the rationale that "it only selects between tracker URLs someone with `settings.integrations` already set". The code says otherwise.

`publicSettings.js:173-176` gates the URL and website id on that same flag:

```js
umami_enabled: settingsObject.analytics_umami_enabled === true || ... === 'true',
umami_url: (... === true || ... === 'true') ? (settingsObject.analytics_umami_url || null) : null,
umami_website_id: (... === true || ... === 'true') ? (settingsObject.analytics_umami_website_id || null) : null,
```

and `App.tsx:150-154` ORs it into the provider check:

```js
if (
  (provider === 'umami' || settings?.umami_enabled)
  && settings?.umami_url && settings?.umami_website_id
) {
```

So the flag is the on/off switch for the whole Umami path. `PUT /analytics` is `requirePermission('settings.edit')`, which means a `settings.edit`-only holder can re-enable same-origin tracker script injection after a `settings.integrations` holder set `analytics_tracker_provider = 'none'`. They can't choose a *new* URL, but "previously vetted" isn't the same as "still safe" — a decommissioned tracker domain can be re-registered.

That's a hole in the boundary this PR is establishing. Either protect the flag too, or drop the `|| settings?.umami_enabled` when an explicit provider row exists. If it's deliberate, say so in the comment — the current rationale doesn't match the behaviour.

### 2. Gallery → admin now reloads the page for every Umami install

`analytics.service.ts:233-241` — entering an admin route after the script ran triggers a full `window.location.reload()`. No infinite loop: `trackerScriptInjected` resets on load, `initialize` skips injection on an admin path, and `handleRouteChange` returns early. I verified that, and the last commit pins it with a test.

The behaviour change is that this used to apply only to the rare custom-head case and now applies to the common Umami/Rybbit setup: a viewer who browses a gallery and then opens the admin UI in the same tab gets a hard reload. Correct, but worth an explicit ack rather than being discovered in production.

### 3. `email_unverified` has no self-service recovery

`oidcService.js:653-663` throws rather than falling through, and only a super_admin can re-set the flag (`userManagementService.js:281-289`). If the locked-out admin *is* the last super_admin, the only way back is SQL. The error string points at "ask a Super Admin", which is the right guidance for every other case. Worth a line in the docs for the last-super-admin case.

## 🟢 Nits

- `backupService.js:1055` uses `path.basename(manifestFileName)` for the temp file but `:1059` builds `manifestKey` from the raw name. Not exploitable — the name is built internally at `:1184-1186` from the backup id and the now-coerced format — but the PR body claims the basename reduction covers "both the local and the S3 manifest", and it covers the local file only. Hoist `const safeName = path.basename(manifestFileName)` and use it for both.
- `BackupConfiguration.tsx:55` — the `manifest_path$|manifest_format$` alternatives are dead: `BackupFormData` (`:19-43`) has no manifest fields and `grep backup_manifest frontend/src` is empty. Harmless as a mirror of the backend rule, just noting it changes nothing today.

## Screenshots

UI-touching PRs need a screenshot attached — that's a checklist item here, not a nit. The surfaces are small (the new SSO error toast, and the two backup fields that stop being saved for non-super admins), so the toast alone is enough. Please add it.

- **Migration 227** — `hasColumn`-guarded both ways, additive, sane `down()`, `NOT NULL DEFAULT true`. Correctly numbered around PR 1539's 226.
- **`DESTINATION_SETTING_RE`** — ran the regex against 11 key shapes rather than reading it. The `$` anchors inside the alternation behave: both manifest keys restricted, `backup_manifest_path_extra` / `backup_manifest_formatx` not over-matched, `backup_schedule` and `backup_retention_days` correctly left on `backup.create`.
- **The manifest fix closes a read hole, not just a write one** — `backup_manifest_path` is a root for the restore file browser (`adminRestore.js:419`, `:812`, `restoreService.js:41`). A `backup.create` holder could point it at an arbitrary directory and browse it. Good catch making it super-admin-only.
- **No second write path** for `backup_manifest_*`: only `adminBackup.js` PUT `/config`. `adminDatabaseBackup.js:57-67` and the six other settings writers are explicit allowlists, so the new 400 on `/general`, `/security`, `/analytics`, `/seo` is sufficient coverage.
- **No client breakage** from `BACKUP_SETTINGS_ELSEWHERE` — nothing in `settings.service.ts` or the wizard sends a `backup_*` key to a generic writer.
- **Legacy `xml` rows** can't mis-serialize: all three reads coerce (`backupService.js:1047`, `:1056`, `:1165`), and `loadManifest` still parses existing `.xml` files by extension + content sniff.
- **`email_link_eligible` completeness** — all six `admin_users` inserts land on the `DEFAULT true` trusted-flow set, and the JIT row carries `external_subject` so the linking query excludes it regardless. Both email-UPDATE paths are covered; no third path to `admin_users.email` exists. `resetAdminPassword` doesn't touch email and is super-admin gated at the route.
- **`hasColumnCached` not in a transaction** at any of the three new call sites — the SQLite deadlock doesn't apply here.
- **`req.admin.roleName`** is really populated (`middleware/auth.js:66`) and `adminUsers.js:177-181` passes it through, so the super-admin check isn't comparing against `undefined`.
- **Profile email casing** — `adminAuth.js:47` lowercases before the `:76` comparison, so both sides match. No missed clear from case drift.
- **Workflow attachments** — the strip is scoped to `emailData.attachments` on those two paths. `escalate_to_collections` (`actions.js:103-123`) builds its own list from stored paths and calls `queueEmail` directly. No regression.
- **Tests are real** — I reverted the super-admin guard and `assertMayManageSuperAdminTarget` and re-ran: 4 tests fail. Your "each fails without its fix" claim holds on the slice I spot-checked. All 7 backend suites (51 tests) and the frontend analytics suite pass locally.

## Verdict

Give me a decision on concern 1 — protect the flag, drop the OR, or state that re-enabling a previously vetted URL is acceptable — and ack 2. The nits are optional. Everything else is cleared from my side.

the-luap:
Following up on concern 1 with a concrete recommendation, since I dug into whether the stated obstacle is real.

**Protect the flag.** The rationale for leaving `analytics_umami_enabled` open — the Analytics tab re-derives it on every save, so protecting it would 403 ordinary saves — is actually the reason protecting it is nearly free.

`useSettingsState.ts:488` is the only writer in the frontend:

```js
settingsData.analytics_umami_enabled = analyticsSettings.tracker_provider === 'umami';
```

It's a pure function of the provider dropdown; there's no independent control for it anywhere in the UI. So:

- Through the UI it can only change when `tracker_provider` changes, and that key is already `settings.integrations`-gated. The 403 would land on exactly the request that's already refused — no additional legitimate save breaks.
- A `settings.edit`-only admin saving the tab without touching the provider sends an unchanged derived value, which `collectUnauthorizedProtectedKeys` deletes before the permission check bites.

The hole doesn't need the UI at all: `PUT /analytics` with `analytics_umami_enabled: true` re-enables the proxied script directly, bypassing the provider gate. Given `publicSettings.js:174-175` gates `umami_url` and `umami_website_id` on that same flag, it's the master switch, not a selector.

**One caveat if you take this.** Adding the key to `TRACKER_CODE_KEYS` alone introduces a false 403: on a legacy install with no `analytics_umami_enabled` row, a `settings.edit` admin saving with provider `none` sends derived `false`, and the comparison is `String(null ?? '')` → `''` against `'false'`, which reads as a change. So it's two edits, mirroring the pattern you already built for the provider:

```js
const TRACKER_CODE_KEYS = new Set([
  'analytics_tracker_provider',
+ 'analytics_umami_enabled',
  'analytics_umami_url',
  'analytics_rybbit_url',
  'analytics_custom_head_html',
]);

const effectiveMissingSetting = async (key) => {
  if (key === 'analytics_tracker_provider') { ... }
+ if (key === 'analytics_umami_enabled') return false;
  return null;
};
```

I'd skip the other option I floated (dropping `|| settings?.umami_enabled` from `App.tsx:152`). That fixes the symptom at the render layer and would silently turn analytics off for legacy installs that have the flag but no provider row. This PR enforces these boundaries at the permission layer; the fix belongs there too.

Luca-Timo:
Thanks — all three answered, pushed as `fc95e5e4` (three commits on top of `340d0b4`). Body updated with a **Review round 2** section.

## 1. Protected the flag

Took your recommendation verbatim, including the `effectiveMissingSetting` half — you're right that the "it would 403 ordinary saves" rationale was the argument *for* protecting it, not against. `useSettingsState.ts:488` is the only writer and it's a pure function of the provider dropdown, so the 403 can only land on a request that's already refused.

I confirmed both citations before acting on them: `publicSettings.js:173-176` gates `umami_url`, `umami_website_id` and `umami_share_url` on the flag, and `App.tsx:150-154` ORs it in. And the direct `PUT /analytics {analytics_umami_enabled: true}` path is exactly as you describe — no UI needed.

I did **not** touch `App.tsx:152`, for your reason: it would turn analytics off for legacy installs that have the flag and no provider row. The rationale comment now states what the code does instead of the old reading.

New test: *"settings.edit role is 403d when it re-enables the tracker through the legacy flag"*. Took the key back out of `TRACKER_CODE_KEYS` and re-ran — that one test fails, the other eight pass.

## 2. Acked

The gallery → admin reload now applies to ordinary Umami/Rybbit installs, not just the rare custom-head case: one full reload for a viewer who browses a gallery and then opens the admin UI in the same tab. Deliberate — it's what keeps a tracker script out of an admin session, and it's the treatment the custom head HTML already had. Written into *Behaviour changes on upgrade* so it isn't discovered in production.

## 3. Documented — and it doesn't need SQL

Checked the last-Super-Admin case and there is a supported way back, so I documented that rather than a database edit. `SIMPLE_SETUP.md` troubleshooting, next to the admin-password reset:

- normally, any Super Admin re-saves the address in **Settings → Users** (saving it unchanged is enough);
- if the refused account is the *only* Super Admin **and** `oidc_disable_local_login` is on, `OIDC_BREAK_GLASS=true` re-opens the password route (`oidcService.isLocalLoginDisabled`), they sign in locally and confirm their own email — a super_admin editing their own row hits the same `email_link_eligible = true` branch — then the variable comes back out.

A pointer to both sits at the `OIDC_EMAIL_UNVERIFIED` throw.

## Nits

- `manifestKey` now uses the same `safeName` as the temp file. You're right the body over-claimed; that's now true as written.
- `BackupConfiguration.tsx:55` — left the dead alternatives in on purpose, so the mirror stays a complete copy of `DESTINATION_SETTING_RE` rather than a subset someone has to re-derive if the form ever gains those fields. Noted in the body so it doesn't read as an oversight.

## Screenshot

Attached in the next comment (the new SSO error toast on `/admin/login?sso_error=email_unverified`). The two backup fields don't change visually — they're simply not sent for a non-super admin — so there's nothing to show there.

Re-ran `adminSettingsProtectedKeys`, `backupRestoreSuperAdminOnly`, `adminUsersSuperAdminTarget` and `oidcSso`: 47 tests pass. eslint clean on the four changed files. No frontend file changed in this round.

Luca-Timo:
01-sso-email-unverified-toast

the-luap:
Re-verified `340d0b4..fc95e5e4` (three commits) and the screenshot. Concerns 1 and 2 and the nit are closed. Concern 3 is not, and part of that is on me: the recovery path both of us described does not exist in the UI.

## Concerns 1 and 2, nit — closed

| Item | Where | |
|---|---|---|
| Legacy flag protected | `adminSettings.js:117` in `TRACKER_CODE_KEYS`, so it falls under the `settings.integrations` rule | ✅ |
| No false 403 on a legacy install | `adminSettings.js:134` returns `false` for a missing row; `collectUnauthorizedProtectedKeys` compares `String(stored)` to `String(incoming)`, so a derived `false`, a stored `'true'` string vs. a boolean `true`, and an untouched tab all read as unchanged and are dropped | ✅ |
| No second writer of the flag | `grep analytics_umami_enabled backend/src`: only `adminSettings.js` writes; `publicSettings.js`, `analyticsTrackerProxy.js` and `trackers/index.js` read. The four generic writers (`/general` `:1544`, `/security` `:1736`, `/analytics` `:1779`, `/seo` `:1840`) all pass through the check | ✅ |
| `App.tsx` left alone | Correct, for the legacy-install reason | ✅ |
| Rationale comment matches the code now | `adminSettings.js:110-114` | ✅ |
| Reload on gallery → admin acked and written into *Behaviour changes on upgrade* | PR body | ✅ |
| S3 manifest key uses the basename too | `backupService.js:1055` `safeName`, used at `:1056` and `:1060` | ✅ |
| Dead alternatives in `BackupConfiguration.tsx:55` kept on purpose | Fine, and noted in the body | ✅ |

Both halves of the flag fix are pinned, I reverted them separately:
- without the `TRACKER_CODE_KEYS` entry → *"settings.edit role is 403d when it re-enables the tracker through the legacy flag"* fails, 8 pass
- without the `effectiveMissingSetting` line → *"settings.edit role can save the Analytics tab unchanged on an install with no tracker settings yet"* fails, 8 pass

So the false-403 caveat is covered by an existing test, not just by reasoning. Your four suites: 47 tests pass locally (32 + 15 in `integration/oidcSso`), eslint clean on the three changed source files.

One edge I looked at and am fine with: an install where the provider row says `umami` but the flag row is missing or `false` (only reachable by writing the provider through the API without the flag). There the tab derives `true`, the effective value is `false`, and a `settings.edit`-only save of the Analytics tab 403s naming the key. That is the right answer: on that install Umami is actually off (`publicSettings.js:174-176` nulls the URL), so the save would switch it on.

## 🟡 Concern 3 — the documented way back is not reachable from the UI

The new docs (`SIMPLE_SETUP.md:478`), the code comment at the `OIDC_EMAIL_UNVERIFIED` throw, and the toast in your screenshot all give the same instruction: a Super Admin opens user management and **saves the admin's email**.

The backend half is real: `PUT /admin/users/:id` accepts an optional `email` (`adminUsers.js:171`), and `userManagementService.js:283-289` sets `email_link_eligible = true` when the actor is a super_admin, unchanged value included.

The frontend never sends it:

- `userManagement.service.ts:103-106` — `UpdateUserData` is `{ roleId?, isActive? }`
- `userManagement.service.ts:190-197` — the payload is built from exactly those two fields
- `UserManagementPage.tsx:441` — the only edit call is `updateUser(id, { roleId })`; the one email input on that page belongs to the invite modal

So today there is no screen on which anyone can "save the email". That makes both documented cases wrong, not only the last-Super-Admin one: the break-glass login works (`oidcService.js:176-177`), but once signed in there is still nothing to click. The only real path is a hand-made `PUT` with `email` in the body. I pointed you at `userManagementService.js:281-289` in round 1 as "only a super_admin can re-set the flag" without checking whether the UI reaches it, so this gap is shared.

Why it matters more after this PR than before: an admin who changed their own email on the profile page used to link by email (the hole you are closing). After it they get a toast telling them to ask for something a Super Admin cannot do.

What I'd take, smallest first:

1. A row action on the Users page, visible to a Super Admin only when the row is not link-eligible: **"Confirm email for SSO"**, sending `PUT /admin/users/:id { email: <current> }`. No email editing form needed, and it says what it does. That needs `email_link_eligible` (or a derived boolean) in the user list payload, the string in en/de like the rest of the `ssoErrors` block, and a screenshot.
2. If you would rather keep this PR backend-only: change the three texts to say what is true today (the address has to be re-saved through the API by a Super Admin, with the request shown in `SIMPLE_SETUP.md`), and open a follow-up for the button. I'd accept that, but it is a poor fit for the people who run this, so I prefer 1.

Small wording point either way: the page is **Users** in the sidebar (`AdminSidebar.tsx:86`, `/admin/users`), not under Settings.

## CI

`backend` failed on `fc95e5e4` with one test: `photoEngagementCounters.test.js:256`, *"skipped archive entries do not count (missing source file)"*, expected 1, received 0 after `settle()`. Not yours: the PR touches nothing under downloads or engagement, the test passes 3/3 locally on your head, and the re-run I triggered is green. It is a timing flake in that suite and I'll track it separately. `bypass-size-gate` is red by design (800 lines, needs a maintainer review).

## Screenshot

Checked. The toast renders and is readable. It is also the clearest illustration of the concern above, since its last sentence is the instruction that cannot be followed.

## Verified vs. trusted

Directly inspected: the three commits at their final file:line, the comparison logic in `collectUnauthorizedProtectedKeys`, all writers and readers of the flag, the break-glass gate, the email branch in `userManagementService`, the frontend user-update path end to end, both revert checks, the four suites, the CI log. Trusted at description level: the updated PR body sections other than the two I quote. Not exercised: a real OIDC provider run of the refused-then-confirmed flow.

## Verdict

The five permission boundaries hold, and everything from round 1 except concern 3 is closed. Make the recovery instruction true — option 1 or option 2 — and I'm comfortable merging.

Luca-Timo:
@the-luap rebased onto `main` (`7aa4c0cc`) — the conflict with #1571 in `userManagementService.js` is resolved, PR is mergeable again.

Resolved in favour of #1571:

- **`is_active` on `PUT /users/:id` is out of this PR.** #1571 rejects the field on that route outright, before any partial write, which is stronger than the guards this PR was adding to that path — so they are gone, together with their three tests. `adminUserStatus.test.js` covers that behaviour now. What stays is the super-admin-*target* check on `PUT`: who may edit a super_admin's profile at all, which #1571 does not cover.
- **Guard order on deactivate/activate/delete:** `assertMayManageSuperAdminTarget` now runs *after* the self and last-super-admin guards, so `deactivateAdminUser(lastSuper, someoneElse)` still throws *Cannot deactivate the last Super Admin* — the message your new test pins. Nothing is written before the check either way, so the boundary is unchanged.
- Migration stays **227** — `main` has 226 and 228–230, so the slot is still free.

Re-ran locally after the rebase: your `adminUserStatus`, `apiForcedPasswordReset`, `oidcEgress`, `umamiAdapter`, `rybbitAdapter` suites plus all of this PR's own suites and the nine user-management/role suites — 137 tests pass; frontend `npm run build` clean. The PR description has the same summary at the bottom.

`bypass-size-gate` is the known size-policy failure, not a code failure — the five parts are separate commits if you'd rather take them one by one. Good to merge from my side once CI is green.

the-luap:
Re-verified the rebase onto `main` (`7aa4c0cc`), head `4891286`. `git range-diff` against the last reviewed head `fc95e5e4`: 12 of the 15 commits are identical, 3 changed in the conflict resolution (`7239d4af`, `df6d76f5`, `7f57f3d7`), and there are no new commits.

## Rebase — clean

| Item | Where | |
|---|---|---|
| `is_active` dropped from `PUT /users/:id` in favour of #1571 | `main` rejects the field in `updateAdminUser` before the user is even loaded (`userManagementService.js:244-246`), so nothing is written from a mixed update. The guards this PR used to add there would be dead code. Dropping them with their three tests is right | ✅ |
| Super-admin-*target* check on `PUT` kept | Runs after the lookup, before any write: a `users.edit` holder still gets 403 on a super_admin's email or username | ✅ |
| Guard order on deactivate / activate / delete | Self and last-super-admin invariants → `assertMayManageSuperAdminTarget` → write. Nothing is written before the check, and #1571's *"Cannot deactivate the last Super Admin"* message survives for every caller | ✅ |
| `email_link_eligible` branch | Unchanged by the rebase; it uses `db`, not a transaction, so `hasColumnCached` is safe there | ✅ |
| Migration 227 | Still free: `main` has 226 and 228–230 | ✅ |

Suites at `4891286`, run locally on SQLite: `oidcSso`, `adminSettingsProtectedKeys`, `adminUsersSuperAdminTarget`, `backupRestoreSuperAdminOnly`, both workflow attachment suites, `adminAuth`, 49 tests, plus #1571's `adminUserStatus`, 8 tests. All green. CI: everything passes except `bypass-size-gate`, which is the known size policy, not code.

## 🟡 Concern 3 — still open

Nothing in the rebase touches it, and the docs commit is byte-identical (`fc95e5e4` = `48912866` in the range-diff). At `4891286`:

- `userManagement.service.ts:103-106` — `UpdateUserData` is still `{ roleId?, isActive? }`, and `:192` sends only `role_id`;
- `UserManagementPage.tsx:441` — the only edit call is `updateUser(id, { roleId })`;
- no frontend code reads or shows `email_link_eligible`;
- `SIMPLE_SETUP.md:485` and the toast still tell a Super Admin to *save the email* on the Users page, which has no control for it.

So the instruction the user sees after an `OIDC_EMAIL_UNVERIFIED` refusal still can't be followed without a hand-made `PUT`. The two options from the last round stand; I still prefer the first:

1. a **"Confirm email for SSO"** row action on **Users**, for Super Admins, on rows that aren't link-eligible, sending `PUT /admin/users/:id { email: <current> }` — plus the eligibility boolean in the list payload, en/de strings, and a screenshot;
2. or keep the PR backend-only and make the three texts say what is true today (re-save through the API, with the request in `SIMPLE_SETUP.md`), and open a follow-up for the button.

## 🟢 Nit

`userManagementService.js`, the comment above the super-admin-target check still says a `users.edit` holder could *"rewrite a super_admin's email or deactivate them"*. After #1571 the deactivate half can't happen on this route. Trim it to the profile.

## Verified vs. trusted

Directly inspected: the range-diff of all 15 commits, the three changed commits at their final file:line, `main`'s `is_active` refusal, the guard order in all three account-status functions, the frontend user-update path, the docs text, CI. Run locally: the eight suites above. Trusted at description level: your note that `npm run build` is clean. Not exercised: a real OIDC provider.

## Verdict

The rebase is correct and all five permission boundaries hold. Close concern 3 — option 1 or option 2 — and I'm comfortable merging.

Luca-Timo:
@the-luap concern 3 closed with **option 1**, one commit: `e271a45`. You were right that it's the better fit — the toast now names something the Super Admin can actually click.

**The action.** On **Users**, a row whose email is not link-eligible is marked *Email not confirmed for SSO* and carries **Confirm email for SSO**, which `PUT`s that row's own address back. That is exactly the trusted-flow save `userManagementService.js:300-305` already treated as a confirmation — the backend rule is unchanged, it just became reachable. `updateUser` sends `email` now; it only ever sent `role_id`.

**The field.** `getAllAdminUsers` / `getAdminUserById` select `email_link_eligible` behind `hasColumnCached`, and `transformUser` emits `emailLinkEligible` **only for a super_admin caller**. Marker, action and payload draw the same line, so a delegated `users.view` clerk isn't told which rows an IdP assertion could land on. A pre-227 backend sends no key, which the page reads as eligible and offers nothing on.

**Two things I'd flag as decisions rather than bury:**

1. **The marker is Super-Admin-only too**, not just the button. An admin who cannot confirm gains nothing from a warning about SSO trust — but if you'd rather every `users.view` holder saw the state, that's a one-line change and I'll make it.
2. **Confirming no longer rewrites the address.** Found while writing the dialog copy: the route lowercases what it receives (`normalizeEmail` keeps `all_lowercase`), so re-saving `Mixed.Case@Example.com` used to edit the stored value — and the dialog's "the address itself is not changed" would have been false for exactly the accounts this targets. `updateAdminUser` now compares without case and writes `email` only when it differs.

**Your nit is in:** the comment above the super-admin-target check no longer claims a `users.edit` holder could deactivate through `PUT` — since #1571 that route refuses the field outright.

**Verification.** 5 page tests, 3 service tests, 5 added to `adminUsersSuperAdminTarget` (list round trip; field withheld from a non-super caller; a `users.edit` holder re-saving an address leaving `false` at `false`; a mixed-case address surviving confirmation; the pre-227 shape through `transformUser`). Each fix reverted in turn — the payload field, the `email` mapping, the `isSuperAdmin` gate, the payload gate, the no-rewrite compare — each fails exactly one test and leaves the rest green. 76 backend tests over the ten user-management suites, 8 frontend, eslint clean, `npm run build` clean.

Also **exercised end to end**, not only unit-tested: fresh SQLite instance migrated to 230, one super_admin plus a non-eligible admin. The click sent `PUT /api/admin/users/2 {"email":"mara@example.com"}`, `email_link_eligible` flipped to `1`, the marker went away. Screenshots from that run — the row, the dialog, the confirmed state — in the next comment.

`bypass-size-gate` is still red by design.

Luca-Timo:
02-users-confirm-email-for-sso
03-confirm-email-dialog
04-email-confirmed

the-luap:
Re-reviewed the new commit `e271a45` on top of the last verified head `4891286` (`git log 4891286..e271a45`: one commit, 12 files, +429/−49). Concern 3 is closed in the shape I asked for, and your two decisions are the right ones. The "no rewrite" change is what exposes the blocker below, though, so it needs one more pass.

## 🔴 Blockers

### 1. A confirmed admin whose stored email has capitals is never linked, and SSO creates a second admin instead
`userManagementService.js:299`, `oidcService.js:616` / `:668`. After this commit, confirming a mixed-case address leaves it as stored. `resolveAdminFromClaims` lowercases the claim (`:567`), then compares exactly: `.where('email', email)`, both in the link query (`:616`) and in the `OIDC_EMAIL_UNVERIFIED` query (`:668`). Both engines compare case-sensitively. So for `Mara.Owner@Example.com`:

- **Confirm:** "Confirm email for SSO" sets `email_link_eligible = 1`, and the marker disappears.
- **Link query:** the SSO login with that verified email misses the row.
- **Refusal query:** misses it too.
- **Result:** with `oidc_autoprovision` on, the login is **JIT-provisioned as a second admin** with the default role, and the original row stays unlinked. With it off, the login gets `not_provisioned`.

Before this commit, re-saving lowercased the address, so the link query matched. The case-sensitive lookup is older than this commit, but this commit removed the accidental workaround and ships the button for exactly these accounts. The most likely one is the bootstrap Super Admin: `001_init.js:30` stores `ADMIN_EMAIL` verbatim.

Reproduced on `e271a45` by adding this to `oidcSso.test.js`:

```js
await db('admin_users').insert({ username: 'mixed-case-admin', email: 'Mara.Owner@Example.com',
  password_hash: await bcrypt.hash('MixedPass123', 4), role_id: adminRole.id, is_active: 1,
  auth_provider: 'local', email_link_eligible: 1, created_at: new Date(), updated_at: new Date() });
idp.setNextUser({ sub: 'sub-mara', email: 'Mara.Owner@Example.com', email_verified: true });
const res = await ssoRoundTrip();
// → location /admin/dashboard, and admin_users now has TWO rows:
//   { id: 1, email: 'Mara.Owner@Example.com', external_subject: null,       role_id: 2 }
//   { id: 2, email: 'mara.owner@example.com', external_subject: 'sub-mara', role_id: 4 }
```

**Fix:** keep "no rewrite" and make the two OIDC lookups case-insensitive: `.whereRaw('LOWER(email) = ?', [email])` at `:616` and `:668`. The claim side is already lowercased. Then extend the test above so it expects exactly one row, linked, with the original role. A second test should cover the refusal: a mixed-case row that isn't eligible must get `sso_error=email_unverified`, not a JIT duplicate. Today it gets the duplicate, and that bug was in the PR before this commit too.

## 🟡 Concerns

1. **Confirmation can trust an address the Super Admin never saw** (`userManagementService.js:299-302` and `:360`). `sameAddress` is computed from the row read at the top of `updateAdminUser`. On a match, the final `update` writes only `email_link_eligible = true`, with no `email` and no condition on it. The target admin can change their own address through `PUT /admin/profile` (`adminAuth.js:75-83`, which correctly sets eligibility to `false`). If that lands between the read and the write, the confirm marks the **new**, unapproved address as link-eligible. Before this commit the approved address was written together with the flag, so the race couldn't happen. The window is small, but this flag exists precisely to stop a self-typed address from being trusted. **Fix:** when confirming, make the write conditional, e.g. `update(...).where('id', id).where('email', user.email)`. On 0 rows, answer 409 ("the address changed, reload"). A test can stub the profile write in between.
2. **Email-in-use check is case-sensitive** (`userManagementService.js:289`). `where('email', updates.email)` with the lowercased input misses an existing `Mara@Example.com`, so a second admin could be given `mara@example.com`. This is older than your change, and it's the same class as blocker 1, so use the same `LOWER(email)` comparison. It isn't blocking alone.

## 🟢 Nits

- The dialog copy "The address itself is not changed" is true now. Keep it once blocker 1 is fixed the way I suggest; if you'd rather lowercase on confirm instead, the copy has to change.
- Your decision 1, the marker for Super Admins only: agreed. A delegated `users.view` holder gains nothing from it, and it leaks which rows an IdP assertion could land on.

## ✅ Verified

- **Directly inspected at `e271a45`:**
  - the backend diff in full: `transformUser`'s `ssoEligibility` gate on all three routes, and the `hasColumnCached` column selection in `getAllAdminUsers` / `getAdminUserById`
  - `updateAdminUser`'s new `sameAddress` branch and its eligibility writes
  - `resolveAdminFromClaims` `:564-680`
  - the profile route's eligibility reset
  - the frontend action: `isSuperAdmin && emailLinkEligible === false` gates marker and button, and it sends `updateUser(id, { email })`
  - en/de strings
  - the `SIMPLE_SETUP.md` text
- **Reproduced locally:** blocker 1, with the test above on SQLite.
- **Second opinion:** an external review of `4891286..e271a45` independently flagged blocker 1 and concern 1. I checked both against the code before including them.
- **Trusted at description level:** your end-to-end click-through and screenshots (they look right, and the `PUT` payload matches the code), and the revert-each-fix test discipline.
- **CI:** `backend` failed on `photoEngagementCounters.test.js:196` (`download-all increments every downloadable photo`, a timing assertion after `settle()` in a 30s suite), and the job was then cancelled at its time limit. This PR doesn't touch downloads, and `main` is green, so I re-ran it as a flake; the re-run is in progress. `bypass-size-gate` is the known size policy.

## Verdict

Close blocker 1 (case-insensitive OIDC lookups plus the two tests) and concern 1 (conditional confirm write), and I'm comfortable merging. Concern 2 can ride along or go to a follow-up, your call.

Luca-Timo:
@the-luap blocker 1 and concern 1 closed in `8878ea7`, and I took concern 2 along — with a bit more reach than you asked for, so flag it if that's too much in one PR.

**Blocker 1.** Fixed the way you suggested: both OIDC lookups compare `LOWER(email)` (`oidcService.js:616`, `:672`), "no rewrite" stays, and the dialog copy stays true. Your repro is in the suite as two tests — the eligible mixed-case row links onto the *existing* row (one row afterwards, original role, address untouched, `external_subject` stamped), and the non-eligible one gets `sso_error=email_unverified` instead of the JIT duplicate. Both fail if I put either lookup back.

**The same rule everywhere it rests.** Concern 2 alone would have left the invariant half-held: an invite could still create the second case-variant row the OIDC fix depends on not existing. So `LOWER(email)` now also covers the user-management uniqueness check, the profile route's (`adminAuth.js:62`), and both on the invitation path (`createInvitation`, `acceptInvitation`). `admin_invitations.email` is deliberately left alone — two case-variant invites can both be pending, and whichever is accepted second is refused at the `admin_users` check. Noted in *Behaviour changes on upgrade*: an install that already has two case-variant admin rows now gets a 409 on the next save of either, which is the ambiguity surfacing rather than being silently resolved.

**Concern 1.** `updateAdminUser` remembers the address it read and makes the write conditional on it, 409 on 0 rows. The confirm mutation passes the server message through to the toast, so the admin reads *"The email address changed while this was open — reload and try again"* rather than the generic failure. The test queues a competing profile-style write from a `db.on('query')` hook when the service issues its read — same shape as the retention race test — and asserts 409, the flag still `false`, and the address still the one the target typed.

**Also ran both suites on PostgreSQL 15**, since the whole fix is about engine-level case comparison: `adminUsersSuperAdminTarget` 15/15 on pg and SQLite, `oidcSso` 17/17 on SQLite and 16/17 on pg. The one pg failure is *"stores the client secret encrypted and survives a config round-trip"*, and it fails identically on the pre-commit tree under pg — I checked by stashing and re-running, not by reasoning. Not this branch's, and CI only runs SQLite, so it's invisible there; say the word if you'd like it in an issue.

90 backend tests over the twelve user-management/SSO suites, 8 frontend, eslint and `npm run build` clean. Nothing changed visually in this commit — the screenshots from round 3 still hold.

Luca-Timo:
CI on `8878ea7` is green — everything passes except `bypass-size-gate`.

One thing worth recording, since `backend` showed **cancelled** twice before that and it isn't the flake from last round: both attempts hit `timeout-minutes: 20` with no failing test, 90 of 444 suites done. The runner was simply slow — from the *first* suites, before any code of this branch runs: `workflowEngine` 69.9s vs 9.9s on `e271a45`, `customerDocuments` 70.3s vs 14.7s, uniformly ~7x across unrelated suites. Locally on this commit the four suites CI showed at 36–42s each finish in 6.4s combined. The third attempt passed in **7m07s**, in line with `main` (7.5–8.5 min).

So: nothing to fix, but the job has now been cancelled or flaked on three consecutive rounds, and at ~8 min against a 20-minute cap there is not much headroom when a runner is having a bad day. Happy to open an issue for the cap, or for the `photoEngagementCounters` timing assertion from last round, if you want either tracked.

the-luap:
Re-verified `8878ea7` on top of `e271a45` (`git log e271a45..8878ea7`: one commit, 7 files, +142/−11). Blocker 1 and concern 1 are closed at the backend in the shape I asked for, and taking concern 2 across the invitation and profile paths is the right call: the OIDC fix only holds if no second case-variant row can be created. One frontend regression left, and it sits in the path this commit added.

## 🔴 Blockers

### 1. The 409 toast shows "Request failed with status code 409", not the server message
`UserManagementPage.tsx:485`: `errorMessage: (error: Error) => error.message || t(...)`. `useMutationWithToast.ts:62-68` calls a **function** `errorMessage` with the raw error and skips `extractServerError` entirely. Only the string form reads `response.data.error` first. `config/api.ts` doesn't rewrite `error.message` in its response interceptor, so `error.message` is axios's own text. For the race this commit guards, the admin reads `Request failed with status code 409`, not *"The email address changed while this was open — reload and try again"*. Every other failure (500, network) now shows the raw axios string too, where it used to show the translated `confirmEmailError`. The commit message's "which the dialog's toast passes through" doesn't hold.

The invite mutation at `:425` has the same shape, so that's where it was copied from. It's older than this PR, but it hides every server message on invite too.

**Fix:** use the string form, which already does server-message-first:
```tsx
errorMessage: t('userManagement.confirmEmailError', 'Failed to confirm the email'),
```
(and the same at `:425` with `t('userManagement.invitationError')`). Add a page test that rejects `updateUser` with `{ response: { status: 409, data: { error: '…changed while this was open…' } } }` and asserts the toast text.

## 🟡 Concerns

1. **Ambiguous case-variant rows link onto an arbitrary one** (`oidcService.js:619-625`). `byEmailQuery.first()` has no `orderBy`, and `LOWER(email) = ?` can now match more than one row on an install that already has `Mara@x.com` and `mara@x.com` (no case-insensitive unique index exists). Before this commit the exact-case row won deterministically. Now the IdP identity, and its mapped role, lands on whichever row the engine returns first, possibly the one with the higher role. You already treat pre-existing case-variant pairs as an ambiguity to surface (the 409 on the next save). The SSO path should do the same: fetch up to 2 rows and, on more than one, refuse with `OIDC_EMAIL_UNVERIFIED` (or a dedicated code) rather than pick. Rare on real installs, but it's the one place the new comparison widens what a login can land on.
2. **The PostgreSQL failure in `oidcSso` ("stores the client secret encrypted and survives a config round-trip").** Yes, please open an issue. CI runs SQLite only, so a pg-only failure in the SSO config path won't surface anywhere else.

## 🟢 Nits

- `createInvitation`'s pending-invite check (`userManagementService.js:36`) is still exact-case on `admin_invitations`. Your reasoning (the second acceptance is refused at the `admin_users` check) holds. A one-line comment there would save the next reader from "fixing" it.

## ✅ Verified

- **Directly inspected at `8878ea7`:**
  - Both OIDC lookups (`:620`, `:672`) compare `LOWER(email)` against the already-lowercased claim (`:567`).
  - The profile route's conflict check (`adminAuth.js:65`) is case-insensitive, and its own write is lowercased at `:48`.
  - `createInvitation` / `acceptInvitation` / `updateAdminUser` uniqueness checks use `LOWER(email)`.
  - Last round's concern 1: `confirmedAddress` is set only on the same-address branch, the write is `where('id').where('email', confirmedAddress)`, and 0 rows gives a `ConflictError` (409). `user.email` can't be null on that branch because `sameAddress` requires a match against the validated input.
  - No other `admin_users` email lookup is left exact-case (grepped `backend/src`).
- **Trusted at description level:** the new tests and your revert-each-fix discipline, and the pg runs (15/15 and 16/17 with the pre-existing failure).
- **CI:** green on `8878ea7` apart from `bypass-size-gate`, the known size policy. Your read of the two cancelled attempts matches mine. Neither had a failing test: every suite ran 6–8× slower from the first ones on, while other branches' runs at the same time finished in about 8 minutes. The full suite also passes locally on this head: 439 suites, 4556 tests, 150 s. Yes to both issues: the 20-minute cap has too little headroom, and the `photoEngagementCounters` timing assertion is worth tracking.

## Verdict

Close blocker 1 (string-form `errorMessage` plus the toast test) and I'm comfortable merging. Concern 1 can ride along or go to a follow-up, your call.

Luca-Timo:
@the-luap blocker 1 and concern 1 closed in `a955aa7`, and the three issues are filed.

**Blocker 1.** Confirmed at `useMutationWithToast.ts:62-68` — the function form skips `extractServerError` entirely, so the 409 arrived as axios's own sentence and every other failure lost its translated fallback too. Both mutations pass the string now, the invite one included. Two tests: a rejection carrying `{ response: { status: 409, data: { error } } }` must toast the server's sentence, and a bare `Error` must toast `confirmEmailError`. Both fail if I put the function form back. My "passes through" claim last round was wrong — I wrote the code, saw 409 handling on both ends and never checked what the hook did in between.

**Concern 1, taken rather than deferred.** Both OIDC lookups read two rows and refuse on more than one. I gave it a dedicated **`sso_error=email_ambiguous`** rather than reusing `email_unverified`: the fix there is to resolve the pair, and sending that admin to "Confirm email for SSO" would be the same wrong instruction I spent two rounds making right. New en/de string, and it is in the login page's known-key list at `AdminLoginPage.tsx:71` — otherwise it would have been silently downgraded to the generic `idp` message, which is the sort of dead fallback you'd have caught next round. Test: two eligible rows differing only in case → `email_ambiguous`, both still unlinked.

**Nit** taken — the pending-invitation check now says in a comment why it stays exact-case.

**Issues.** PicPeak/picpeak#1615, PicPeak/picpeak#1616, PicPeak/picpeak#1617. One correction worth having up front on #1615: **the secret is not exposed.** I read the row on Postgres before filing — `app_settings.setting_value` is a `json` column, `pg` returns it already parsed, and the test's `JSON.parse` then chokes on the bare ciphertext. The stored value is the `iv.tag.payload` ciphertext on both engines, so the encryption the test guards does hold; only its parse step is engine-specific. Worth fixing anyway, since any test reading `app_settings` directly has the same trap.

**Verification:** 90 backend tests over eleven suites on SQLite, `oidcSso` 17/18 on PostgreSQL 15 (the one failure is #1615's, on an unmodified tree too), 10 frontend tests, eslint clean, `npm run build` clean. Nothing visual changed — the round-3 screenshots still hold.

the-luap:
Re-verified `a955aa7` on top of `8878ea7` (one commit, 9 files, +96/−11). Blocker 1 and concern 1 are closed, the nit is in, and the three issues are filed. Giving the ambiguous-address case its own `email_ambiguous` code was the right call: sending that admin to "Confirm email for SSO" would have been the wrong instruction.

## Blockers — closed

| | Where | Status |
|---|---|---|
| 1. 409 toast showed axios text | `UserManagementPage.tsx:427`, `:488`: both mutations now pass the string `errorMessage`, so `useMutationWithToast` reads `response.data.error` first and falls back to the translated text | ✅ |
| Tests | `ssoConfirmEmailAction.test.tsx`: the 409 with `data.error` toasts the server's sentence, and a bare `Error` toasts `confirmEmailError` | ✅ |

## Concerns — closed

| | Where | Status |
|---|---|---|
| 1. Ambiguous case-variant rows | `oidcService.js:630-636`: the link query reads `limit(2)` and refuses on more than one row with `OIDC_EMAIL_AMBIGUOUS`. The refusal query (`:682-692`) gets the same treatment, so a pair of unconfirmed rows also gets `email_ambiguous`, not `email_unverified`. The code is mapped in `auth.js:1116`, is on the login page's known-key list (`AdminLoginPage.tsx:71-72`), and has en/de strings | ✅ |
| 2. pg-only `oidcSso` failure | Filed as issue 1615, with the correction that the stored value is ciphertext on both engines and only the test's `JSON.parse` depends on the engine. Agreed, thanks for checking the row rather than assuming | ✅ |
| Nit: exact-case pending-invite check | `userManagementService.js:34-37`, comment added | ✅ |

One behaviour I checked and agree with: a pair where one row is confirmed (`email_link_eligible`) and the other isn't still links, to the confirmed row, because the link query only reads eligible rows. That isn't the arbitrary pick the concern was about. The choice is deterministic and rests on a Super Admin's confirmation.

## 🟢 Nits (non-blocking)

- The new ambiguity test covers two **eligible** rows. The second refusal path (two **unconfirmed** case variants at `:682-692`, giving `email_ambiguous`) has no test of its own. It's a small addition next to the existing one, whenever convenient.

## ✅ Verified

- **Directly inspected:** the full `a955aa7` diff, the lookup flow in `resolveAdminFromClaims` (eligible link, then the refusal query, then JIT provisioning), the hook's string-vs-function branch in `useMutationWithToast.ts`, and issues 1615–1617.
- **Run locally on `a955aa7` (SQLite):** `oidcSso` + `adminUsersSuperAdminTarget` 33/33, `ssoConfirmEmailAction` 7/7.
- **Trusted at description level:** the revert-each-fix checks and the PostgreSQL run (17/18, with the one failure being issue 1615's).
- **CI:** green on `a955aa7`: `backend` 8m43s, `e2e-smoke` passed. `bypass-size-gate` is the known size policy.

## Verdict

From my side, merge cleared.

PR #1542, opened 2026-09-19
fix(gallery): stop lightbox pinch-zoom flickering in touchscreen Safari

the-luap:
## Problem

Pinch-zooming a photo in the lightbox flickers in Safari on touchscreens (iPad / iPhone).

## Root causes

All in `frontend/src/components/gallery/PhotoLightbox.tsx`.

1. **A CSS transition retargeted on every touchmove.** The image had `transition: isDragging ? 'none' : 'transform 0.2s'`, and `isDragging` is only set by a single-finger pan on an already zoomed image. A pinch from 100% therefore ran with the 0.2s transition on, and each touchmove started a new one. Measured in WebKit (below): the rendered scale does not follow the fingers at all, then jumps once the target stops changing. It was also inconsistent — one finger landing first on a zoomed image set `isDragging`, so some pinches had the transition and some did not.
2. **No layer of its own.** The image lives inside the 300%-wide carousel track (`will-change: transform`), so every scale step repainted that whole layer.
3. **The first finger of a pinch dragged the carousel.** Two fingers never land in the same event; finger one put the track into `dragging`, finger two triggered a 280ms spring-back underneath the zoom.
4. **`handleTrackTransitionEnd` accepted bubbled events.** It checked `propertyName === 'transform'` but not the target, so the image's own `transitionend` ended the track's spring early and the track snapped.
5. **Native page zoom could run alongside ours.** `touch-action: none` was only on the track container. iOS Safari ignores `user-scalable=no`, so a pinch with one finger on the toolbar or a button zoomed the page while the handlers zoomed the image.

## Fix

- `transition: none` and `will-change: transform` while `isDragging || isPinching`; both go back when the fingers lift, so Safari repaints sharp once at the final scale and the toolbar buttons / wheel zoom keep their easing.
- Pinch baseline moved from state to a ref (`pinchRef`), re-baselined when a third touch (palm) lifts so the surviving pair does not jump. Previously that case killed the pinch.
- `SWIPE_SLOP_PX = 8` before a single finger moves the track. The commit threshold still uses the raw finger travel.
- `e.target !== e.currentTarget` guard in `handleTrackTransitionEnd`.
- `touch-action: pan-x pan-y` on the lightbox root (feedback panel still scrolls; pinch and double-tap zoom are off) plus a `gesturestart` `preventDefault`, registered only when `navigator.maxTouchPoints > 0` so a trackpad pinch in desktop Safari behaves as before.

Desktop mouse pan, wheel zoom and the zoom buttons are unchanged.

## Verification

**Vitest** — new `PhotoLightbox.pinchZoom.test.tsx`, 7 tests driving real touch events through the component. 5 of them fail against `main`; the other two pin the new ref-based pinch (touchmove chaining, third-finger re-baseline). `src/components/gallery`: 28 files / 196 tests green. ESLint: no new warnings. `tsc`: clean for the touched files.

**Real WebKit (iOS 26.4 simulator, iPad Pro 11", Safari via safaridriver)** — same scripted pinch (finger 1 lands and drifts 6px, finger 2 lands, 45 touchmoves at one per frame from 1x to 3x), sampling `getComputedStyle` every frame, `main` build vs this branch:

| per pinch frame | `main` | this branch |
|---|---|---|
| image `transition` | `transform 0.2s` | `none` |
| rendered scale vs. target scale, max gap | **2.0** | **0.0** |
| rendered scale vs. target scale, mean gap | 0.83 | 0.0 |
| track offset while finger 1 is alone | 0 / -2 / -4 px | 0 / 0 / 0 px |
| track offset during the pinch, max | 4.5 px (spring-back) | 0 px |

On `main` the rendered scale stays at 1.0 for ~30 frames while the target climbs to 2.95, then animates to 3 after the target clamps.

**Not verified:** safaridriver collapses two touch pointers into one, so the pinch above is dispatched as `TouchEvent`s from script. That exercises WebKit's style/transition/compositing path but not the native gesture recogniser, so item 5 (native page zoom fighting ours) is covered by the unit test and by reading `touch-action` back in Safari, not by an observed native pinch. Needs a pass on a physical iPad/iPhone before merge.

## Screenshot

The change is motion-only — there is no static before/after to show. A screen recording from a device is the meaningful artefact here.


the-luap (commented):
## 🔴 Blockers

None.

## 🟡 Concerns

None material.

## ✅ Verified

- No XSS: touch math + CSS (`touch-action`, `willChange`) only.
- Pinch/swipe state clears on `touchEnd` / `touchCancel`; `transitionend` gated to track so image transitions cannot advance the carousel.
- Close/Escape still work — no stuck-overlay path found.
- `gesturestart` `preventDefault` is touch-device gated (intentional Safari native-zoom block).
- Vitest coverage for the pinch path.

UI change — a short phone/Safari clip or screenshot would help confirm the flicker is gone.

## Verdict

Merge-cleared.

the-luap:
Recording added, as asked. iPhone 17 Pro simulator, iOS Safari, the 546-photo gallery. Main on the left, this branch on the right; red dots are the two touch points.



Mid-pinch still: main at 231 % with the photo still at full size, this branch at 223 % already scaled.



Separate clips: [main](https://raw.githubusercontent.com/PicPeak/picpeak/screenshots/pr-1542/shots/1542-pinch-main.gif) · [this branch](https://raw.githubusercontent.com/PicPeak/picpeak/screenshots/pr-1542/shots/1542-pinch-pr.gif)

Measured per frame over a 100-frame pinch: on main the image's computed scale stayed at 1.0 for the first 60 frames while the zoom counter climbed, then jumped to 3.0. On this branch it matched the target on every frame.

Caveat: safaridriver collapses two touch sources into one, so the pinch is synthetic `TouchEvent`s dispatched from the page. That exercises WebKit's style, transition and compositing path, the part this PR changes, but not the native gesture recogniser (page zoom, `gesturestart`). A pinch on a physical device is still worth one look before merge.

PR #1543, opened 2026-09-19
fix(transfers): keep files picked in the create dialog

Luca-Timo:
In the admin **New transfer** dialog, files picked with **Choose files** sometimes never showed up in the list, and **Create transfer** stayed disabled. The dialog showed no error, and no request reached the backend. On my instance it happened reliably once a title was typed: an empty or one-character title worked, but `LB` or `LBM Logos` didn't. The file type and name didn't matter.

**Cause:** the input's `onChange` passed its live `FileList` into `addFilesToList`, whose `setFiles` updater calls `Array.from(list)`. The handler then clears the input (`e.target.value = ''`) straight away. Clearing empties that same `FileList` object in the browser, and React doesn't always run an updater at `setState` time; after earlier re-renders (typing a title) it runs it on the next render. By then the list is empty, so nothing is added.

## What changed

**Frontend**
- `frontend/src/pages/admin/transfers/TransfersPage.tsx`: `addFilesToList` copies the files with `Array.from` before calling `setFiles`, so clearing the input can't empty them. This matches what the other file inputs already do (`TransferUploadPage`, `PhotoUpload`, `UserPhotoUpload`, and this page's own "add files" button on an existing transfer).
- `frontend/src/pages/admin/transfers/__tests__/TransfersPage.createFilePicker.test.tsx` (new): opens the dialog, optionally types a title (`""`, `L`, `LB`, `LBM Logos`), picks two files through an input whose `files` list is emptied when the value is cleared (as a browser does), and checks that both files are listed and **Create transfer** is enabled.

There are no backend, route, migration, i18n or settings changes.

I also searched the frontend for the same pattern: every other `target.files` / `dataTransfer.files` handler already copies the files or takes `files[0]` before any state update. This was the only instance.

## Testing

- **The new test, with and without the fix (run locally):** with the fix, 4/4 pass. With `TransfersPage.tsx` reverted to `main`, 3 fail (`L`, `LB`, `LBM Logos`) and the empty-title case passes. That's the same shape as the bug.
- **Full frontend suite:** `npx vitest run`, 129 files / 835 tests passing.
- **Build and lint:** `npm run build` and `eslint src/pages/admin/transfers/` are clean.
- **The browser behaviour, checked by hand:** in Chromium, a `FileList` reference taken from an input has length 0 after `input.value = ''`.
- **On prod, before the fix:** reproduced in Safari. Titles `""`, `t` and `L` worked, while `LB`, `Lb`, `LBMLogos` and `LBM Logos` didn't.
- **Not verified yet:** the fix running in a real browser. The screenshots below are still to be added.

EDIT: ## Screenshots
No screenshots necessary, only visible change is the way the "create transfer" button behaves. Now a transfer without files can be created to add files later.


the-luap:
Reviewed at `407dcacf`.

## 🔴 Blockers

None.

## 🟡 Concerns

1. **Screenshots.** The body still carries the two `<!-- before / after -->` placeholders. UI PRs need them before merge — you already list it as open, so this is just the reminder.

## ✅ Verified

Directly inspected:

- **Diagnosis is right.** `TransfersPage.tsx:285` hands the live `e.target.files` to `addFilesToList` and clears the input in the same handler. The old updater only called `Array.from(list)` when React got round to running it, which after a prior re-render is the next render — by then the list is empty.
- **Fix** at `TransfersPage.tsx:194-197` copies before `setFiles`. `addFilesToList` has exactly one caller, so nothing else depends on the old shape.
- **The test has teeth.** Ran it in a worktree: 4/4 with the fix; with `main`'s `TransfersPage.tsx` swapped back in, `L`, `LB` and `LBM Logos` fail and the empty title passes. Same shape as your report.
- **"Only instance" holds.** Went through all 27 `target.files` / `dataTransfer.files` handlers under `frontend/src`: every other one takes `files[0]` or copies first, `TransferUploadPage.tsx:75` included.

## Verdict

Add the screenshots and from my side, merge cleared.

Luca-Timo:
@the-luap sorry my mistake.

No screenshots necessary, only visible change is the way the "create transfer" button behaves. Now a transfer without files can be created to add files later. (Edit 1)

the-luap:
Re-checked at `407dcacf` — no new commits since the review above, so everything verified there still stands.

## On the screenshot reasoning

Two corrections, because the stated reason doesn't match the diff.

**The create button's behaviour is not what this PR changes.** The `disabled` expression at `TransfersPage.tsx:362-365` is byte-identical to `main`:

```tsx
disabled={
  (picked.length === 0 && files.length === 0 && !allowUploads)
  || (deliveryMethod === 'email' && parsedEmails.length === 0)
}
```

**And a transfer without files could already be created before this PR** — that is the `allowUploads` branch of the same expression, i.e. the "give the client an upload link" checkbox (`:234`). Nothing in these four lines touches it.

What does change is the thing the PR is named after: the picked files stay in the list. The button going from stuck-disabled to enabled is a *consequence* of that, since `allowUploads` defaults to `false` (`:161`) — with the files dropped and no event photos picked, the admin couldn't create the transfer at all.

## On the screenshots themselves — you're right, just not for that reason

I tried to produce the before/after pair myself and **could not make `main` misbehave under automation**:

| Engine | FileList liveness after `input.value = ''` | Files kept on `main` |
|---|---|---|
| Chromium 141 | same object, `length` 0 → premise holds | **yes** (3 of 3) |
| WebKit 26.0 | same object, `length` 0 → premise holds | **yes** (3 of 3) |

Title typed first in both, files assigned through a real `DataTransfer` and a dispatched `change`. So the mechanism you diagnosed is real — the held `FileList` genuinely empties in place — but React evaluates the updater eagerly under a scripted pick, before the clear, and the files survive. Your unit test reproduces it because its fake list is emptied by the `value` setter synchronously; a native pick in Safari, which is what you had, is not scriptable here.

So: no before/after pair. Dropping that requirement.

## What is still worth having

The PR body lists *"**Not verified yet:** the fix running in a real browser"*. That one I did, on `pr-1543` in WebKit — title typed, three files picked, all three listed, **Transfer erstellen** enabled:



Please fold that line in the body into "verified" and replace the screenshot placeholders with this shot, or your own from Safari.

## Verdict

Merge cleared. The fix is right whatever the scheduling does — copying before `setState` is what every other file handler in the tree already does, and it costs nothing.

PR #1545, opened 2026-09-19
fix(contracts): close the signing gaps found auditing issue 1446

the-luap:
Slice 1 of the implementation plan on issue 1446 — the part that is defects in shipped behaviour rather than missing features. Everything here is on code that went out with #1464.

### 1. A co-signer saw the account holder's email address and company

`sessionView` handed every verified signer the contract's `recipient` block (`publicView.js:79-83`), so a second signer received the address their co-signer's login codes go to — one signer's authentication data reaching another, which is exactly what the issue rules out.

The display name stays (it is printed in the contract they are reading); the email and company are withheld from anyone who is not the account holder. The holder is identified by email hash — the same match `portalSigningAccess` already makes, since `contract_signers` carries no `customer_account_id`.

### 2. A lost response invited a blind resubmission

With no HTTP status the sign form said *"Check your connection and try again"* and did not re-read the session (`ContractResponsePage.tsx:692`). The server side was already correct — the idempotency key lives in `sessionStorage` and a replay returns `{replayed:true}` — but the signer was never told whether the first attempt landed, on the one request in the product that must not be repeated blind.

Now the failure asks the server. If the signature landed, the signed result appears. If it did not, the page offers **Check again** and **Send my signature again**, the second under the same idempotency key. The typed name and the chosen mode survive a reload; a drawn signature stays in memory and never reaches browser storage. Declining and the admin's counter-signature get the same posture instead of a bare retry.

### 3. Erasing a customer left the signing surface untouched

`eraseCustomer` anonymised `customer_accounts` and stopped there: a live signing link on an unsigned contract kept working, and the customer's name and address stayed inside the frozen `rendered_content` — with no written rule saying which of that was deliberate.

The rule now lives in `backend/src/services/contract/erasure.js`:

- **no signature yet** (draft or sent) → cancelled; every invitation, session and action token revoked; the signers' `*_enc` columns cleared; the customer's name and address taken out of the snapshot;
- **at least one signature** → kept whole as the contractual record, same stance erasure already takes for contract-linked documents, with only the live access revoked.

`rendered_content_sha256` is deliberately **not** recomputed — it is the hash a signer would have been bound to. Migration 228 adds `contracts.rendered_content_redacted_at`, so a later integrity check reads the mismatch as a redaction rather than tampering. Two residues are documented in the module rather than silently handled: the event log keeps the signer labels it already carries (editing one breaks the chain for every event after it).

### 4. The evidence key was invisible to operators

`PICPEAK_EVIDENCE_KEY` was in no `.env.example`, neither compose file and no documentation. With no key set, one is generated silently into `<storage>/business-docs/keys/evidence.key` — inside the backup path, but then the only copy, and losing it makes every signer's name, address and IP unreadable.

It is documented in all four places now, the server logs a warning at boot while the key comes from that file, and System Health counts stored evidence **per key id across all five encrypted columns**. The old check read the key id off the newest `name_enc` row alone, so one fresh row reported "all fine" over rows it could not read.

### 5. An admin's paper copy completed a multi-signer contract with no statement of what it carried

The customer's own upload is refused outright on a multi-signer contract (`MULTIPLE_SIGNERS`); the admin's was not. The capability stays — it is how an offline signature reaches the record at all — but it is now the explicit offline-evidence workflow the issue asks for: the dialog lists the customer signers who have neither signed nor declined, the admin ticks each one the paper copy bears, and the server refuses with `SIGNERS_NOT_COVERED` unless all of them are covered. The `wet_upload` event payload records `{ coversSignerIds, uploadedByAdminId, fileSha256 }` — ids only, no names or addresses.

### 6. The signing certificate had no download route

It only ever left the server as an email attachment, so a lost email was a lost evidence record. `GET /api/admin/contracts/:id/certificate` and `GET /api/customer/contracts/:id/certificate` serve it, `attachment` + `nosniff`, scoped exactly like the contract PDF routes. The portal list carries `hasCertificate` so the button is only offered where it works.

**One deviation from the plan, on purpose:** it also called for `…/session/certificate` for a signer session and a certificate button on `SignedResult`. Neither can work. `countersign` calls `signers.revokeAccess` when it seals the contract, so every session is dead by the time a certificate exists; and at `SignedResult` time the contract is at most `signed_by_customer`, which has no certificate yet. Signers get it by email and through the portal.

## Tests

| Area | Where |
|---|---|
| co-signer payload carries no `@`, holder's does | `contractSigningV2.test.js` |
| paper coverage: none / partial / foreign id / all; ids-only payload; no confirmation once everyone signed | `contractSigningV2.test.js` |
| certificate: 404 before, byte-identical to the stored artifact after | `contractSigningV2.test.js` |
| erasure: cancelled + redacted + hash unchanged; links dead; signed contract kept whole; log names both sets | `contractErasure.test.js` (new) |
| evidence key: counted per column, older rows behind a readable newest row, never the key itself | `adminSystemHealthEvidenceKey.test.js` (new) |
| portal certificate: owner only, `CERTIFICATE_MISSING`, other customer and draft 404 | `customerPortalDocuments.test.js` |
| lost response: signature that landed, and one that did not; same idempotency key on resend; draft keeps the typed name and not the image | `ContractResponsePage.signingV2.test.tsx` |

Every new test was checked against the code before the change and fails there. Backend suites are green on SQLite **and** PostgreSQL 15. Full backend run: 419 suites, 4329 tests. Frontend: 128 files, 834 tests. `build:check` adds no new TS errors (5 pre-existing, all in untouched files).

## Still to come

Screenshots (light, dark, 390px) for the four UI surfaces — the uncertain-signature panel, the paper-upload dialog, and the two certificate buttons. Kept as a draft until they are attached.

Relates to issue 1446.


the-luap:
## Screenshots

### The uncertain-signature panel

A signature whose response never arrived. The plain "Sign contract" button is replaced by the two deliberate actions — the entered name and mode survive, and the copy says why sending again cannot sign twice.

| Light | Dark |
|---|---|
|  |  |

390px:



### The paper-upload dialog

Both customer signers listed, neither ticked, upload disabled until each one is confirmed.

| Light | Dark |
|---|---|
|  |  |

### The certificate download

Admin, on a completed contract — next to the existing PDF downloads:

| Light | Dark |
|---|---|
|  |  |

Customer portal — offered only where `hasCertificate` is true, so it never leads to a 404:

| Desktop | 390px |
|---|---|
|  |  |

Captured against the dev rig on the real code path: a two-signer contract out for signature, and a completed one whose certificate was issued by `issueCertificate`.

the-luap (commented):
## 🔴 Blockers

None.

## 🟡 Concerns

1. **Wet upload after partial e-sign** (`signingV2.js` / `signatures.js`) — paper can overwrite `signed_pdf_path` and flip `fully_signed` without marking already-e-signed slots. Dialog copy says paper becomes authoritative; the attestation understates vs discarded e-stamps. Prefer refusing wet upload when any customer row is already `signed`, or requiring attestation for all customer slots.

2. **Erasure retention** (`erasure.js`) — signed/partly-signed contracts keep encrypted PII and event `actorLabel`s; only links/sessions die. Documented tradeoff — confirm it matches the privacy/retention doc before calling GDPR done.

3. **Evidence key rotation** — health scan detects mismatch; no re-encrypt path. Operators still need a manual recovery story.

4. **Certificate vs evidence permissions** — `GET …/certificate` is `contracts.view` (emails on the cert); IP/UA stay behind `contracts.manage`. Consistent with overview, but view-only admins get co-signer emails via PDF.

## 🟢 Nits

- Portal session/idempotency/draft share one scope — a stale `/contract/signing` bookmark can reopen the last session until expiry.
- Decline path has no idempotency key (unlike sign).

## ✅ Verified

- Cross-signer substitution blocked: session hashed to `signer_id`; conditional `status: 'invited'` on sign.
- Co-signer PII stripped in `sessionView` for non-account-holders; invite pre-OTP is issuer + masked email only.
- Evidence key: AES-GCM at rest; boot refuse near-miss keys; System Health returns metadata only; no default secret in compose/.env.example.
- Migration 228: `hasTable`/`hasColumn` up; guarded `down()`.
- Double-submit: idempotency replay + frontend uncertain UI.
- Feature flag `contracts` on public signing router; customer portal cert scoped by `customer_account_id`.

## Verdict

No blockers. Comfortable merging after a clear answer on wet-upload-after-partial-e-sign (concern 1) and an explicit OK on signed-contract retention under erasure (concern 2).

the-luap:
Answers and fixes for all four concerns. Commits: `1dd6a9c7`, `08c0caba`, `2d4bda08`.

## Concerns

### 1. Paper upload after a partial e-sign — closed in `1dd6a9c7`
`signingV2.js:754` `electronicSignaturePresent()`. `assertPaperCoversSigners` (`:775`) now refuses with `409 ELECTRONIC_SIGNATURE_PRESENT` as soon as any customer signer row is `signed`. The coverage endpoint (`adminContracts.js:730`) returns the flag, so the dialog explains the refusal instead of offering an upload that will fail. `ContractDetailPage.tsx:190` maps the code if it comes back anyway.

The refusal also covers the case where every customer has e-signed and only the counter-signature is open. The old test "a paper copy needs no confirmation once every signer has signed" allowed that deliberately. It is replaced, because the paper would still overwrite the e-signatures; the counter-signature is given in the browser.

Known residual: a signer can still e-sign between the check and the write on a multi-signer contract, where status stays `sent` until the last signature. Only v2 contracts are covered.

| Light | Dark |
|---|---|
|  |  |

390px:

| Light | Dark |
|---|---|
|  |  |

Captured with this branch's frontend against the rig. The rig backend runs `main`, which has no `paper-signature-coverage` route yet, so that one response is intercepted in Playwright (`electronicSignaturePresent: true`); everything else is live.

### 2. Erasure retention — confirmed, tightened in `2d4bda08`, documented
Signed contracts are kept whole on erasure; that is intended (statutory retention of the signed contract and its evidence). Documenting it showed `plan()` kept more than that: contracts that were already `declined` or `cancelled` kept all signer PII (name, email, IP, user agent, decline reason) even though nobody signed.

`erasure.js` now decides on the existing "any signature" test only. Every unsigned contract is redacted the same way draft/sent ones were: signer PII cleared, `customer_name` / `customer_address` blanked (hash not recomputed, `rendered_content_redacted_at` set), links, sessions and tokens revoked. Only draft/sent are moved to `cancelled`; terminal statuses stay. The hash-chained event-log labels stay, for the reason in the header comment. The erasure activity log records the redacted-without-status-change contracts separately. New cases in `contractErasure.test.js`: declined and cancelled without a signature (PII gone, status and chain intact), and signed-then-cancelled stays whole.

The retention documentation `erasure.js` referred to did not exist; it is PicPeak/docs PR 31 ("What erasing a customer keeps" on the contracts page), linked from the code in `08c0caba`.

### 3. Evidence-key rotation
No re-encrypt path in code; decryption accepts the current key only, so recovery is restoring the previous value (`PICPEAK_EVIDENCE_KEY` or `storage/business-docs/keys/evidence.key`, env wins). `PICPEAK_EVIDENCE_KEY` was not documented at all; the same docs PR adds it to the environment-variables page with that recovery note. A key list / re-encrypt tool stays a follow-up.

### 4. Certificate under `contracts.view`
Deliberate. The contract overview already shows co-signer emails to view-only admins, so the certificate exposes nothing new; IP and user agent stay behind `contracts.manage`.

## Nits

- Decline without an idempotency key: skipped. A successful decline revokes every session, so a retry is refused before it could match a key, and the status lock stops a second notification. Adding a key would mean letting a revoked session through.
- Shared portal scope for a stale `/contract/signing` bookmark: left as is; it ends with the session.

Tests: 18 suites / 195 backend, 29 vitest; the new backend and dialog tests fail before their fixes.

PR #1547, opened 2026-09-19
test(contracts): pin the signing isolation and permission boundaries

the-luap:
Slice 2 of the implementation plan on issue 1446. **Stacked on #1545** — base is `feat/1446-signing-hardening`, so the diff here is only this slice.

`contractSigningV2.test.js` signs its happy paths as a `super_admin` with one customer, which left the refusals around it unpinned. This adds them at route level.

## What is now pinned

| Boundary | Test |
|---|---|
| A verified session reaches only its own contract's attachments | a real attachment on a *different* contract → 404 |
| A signer's session signs their own slot and nobody else's | Ben signs typing *Anna's* name; `customer-2` is signed, `customer-1` still `invited`, contract still `sent` |
| A revoked link cannot ask for a code | resend revokes the first link → `410 SIGNING_LINK_REVOKED`, and **no** `code_sent` appended to the chain |
| Portal access on another customer's v2 contract | `signing-access` and `/pdf` → 404 **even though this customer is listed as a signer on it**, and no session is minted |
| The evidence view is behind `contracts.manage` | a role with only `contracts.view` → 403 while it still reads the contract itself; with the permission → 200 **and** one `contract_signing_evidence_viewed` row naming the contract |

Two items from the plan were already covered upstream and are not duplicated: a second use of the same OTP (asserted in "a link shows nothing about the customer until the emailed code is entered") and a session used after the contract is sealed.

These five are regression pins on behaviour that is already correct — a test-hardening slice cannot honestly claim to fail before the change. The sixth needed a fix.

## The one real gap: a key replayed with different details

`sign()` answered a replay like this:

```js
if (idempotencyKey && signer.idempotency_key === idempotencyKey) {
  return { status: contract.status, signedAt: signer.signed_at, replayed: true };
}
```

— without looking at what the replay actually asked for. The key is generated once per link and resent after a lost response, so a key arriving with a **different name or mode** is not the request that succeeded, and answering "done" reports a signature nobody made that way.

It now answers `409 IDEMPOTENCY_KEY_REUSED` when the submitted name or mode differs from the one recorded (the drawn image is deliberately not compared — a legitimate retry re-renders the canvas). The signing page treats the code like `ALREADY_SIGNED`: the signature *is* on record, so it re-reads the session and shows it rather than raising an error.

Test: same key twice → `replayed: true`; same key with another name → 409; same key with `drawn` instead of `typed` → 409; the row still holds exactly one signature, typed, under the original name.

## Checks

Green on SQLite **and** PostgreSQL 15. Full backend run: 419 suites, 4335 tests. Frontend: 9 tests in the signing page file, full suite green. No new TS errors.

No UI surface changes beyond the error-code branch, which is covered by a frontend test rather than a screenshot.

Relates to issue 1446.


the-luap (commented):
## 🔴 Blockers

None.

## 🟡 Concerns

None material. The idempotency-key reuse → `409 IDEMPOTENCY_KEY_REUSED` when the same key is resent with a different name/mode is a real fix, not test theatre — UI treating that as already-signed is correct.

## ✅ Verified

Tests actually pin boundaries:
- foreign attachment via own session → 404
- parallel signer cannot fill another slot
- revoked link → 410 + no new `code_sent`
- portal `signing-access` for foreign contract even when listed as signer → 404 + no session mint
- evidence needs `contracts.manage` + audit log row

## Verdict

Merge-cleared (onto #1545).

the-luap:
Merged the updated #1545 twice (`f3ec9b63`, `4ca1d354`) — the paper-upload refusal and the erasure tightening — no conflicts. 18 suites / 200 tests pass. No findings of its own on this one.

PR #1548, opened 2026-09-19
feat(contracts): freeze what a contract costs, not only what it says

the-luap:
Slice 1 of the implementation plan on issue 1445. Branches off `main`, independent of the 1446 stack (#1545 / #1547) — but **1446 slice 3 depends on this**: until it lands, binding a signature to "everything" is hollow, because the hash does not cover the price.

## The gap

`buildContentSnapshot` froze title, intro, outro, placeholders and clauses. The **quote line items the contract prints** were not in it — the `quote_line_items_table` block re-read `quote_line_items` on every render (`renderContext.js:251-272`).

A sent contract opens its stored PDF, so nothing a customer ever saw could change. But `rendered_content_sha256`, the hash `contract_signers.content_sha256` records and a signature is bound to, did not cover the commercial terms. Editing the source quote after sending left that hash untouched while changing what a re-render produced.

## Snapshot format 2

```
format: 2,
…clauses, placeholders, title, intro, outro as before…
quote: {
  number, currency,
  lineItems: [ …the counted rows, in order… ],
  totals: { netMinor, vatRatePercent, vatMinor, shippingMinor, grossMinor }
}
```

Three decisions worth stating:

- **Totals are read from the `quotes` row, not recomputed.** Those are the amounts the quote itself states and the customer accepted. Re-deriving them here would let a later change to, say, the rounding setting disagree with the document that went out. (The plan sketched a `taxLines[]` array; a quote carries a single `vat_rate`, so this is one rate, named.)
- **Every number is coerced on the way in.** `*_minor` are bigints and `quantity` / `discount_percent` decimals: PostgreSQL returns those as strings and SQLite as numbers. Freezing them raw would give one contract two different content hashes on the two engines — and that hash is what a signature is bound to. There is a test asserting each field's `typeof`.
- **Line items keep their DB column names.** `buildRenderContext` hands snapshot rows to exactly the mapper it already uses for live rows, so a frozen contract and a live one go down one code path.

`quote` is frozen whenever the contract has a source quote, whether or not the line-items block is included: what is frozen is the commercial basis, and which blocks print it is the renderer's question.

**Format 1 keeps the live read.** A contract sent before this had its terms frozen nowhere and they cannot be reconstructed; its stored PDF is what opens anyway.

## What now shows the frozen figures

- **The PDF** draws a totals row under the line table. It printed a column of amounts and never named the sum being signed for. `drawContractTotals` is separate from `drawTotals` on purpose: that one is pinned to a fixed offset from the page bottom and shares geometry with the payment block beneath it, neither of which exists in a contract, where the table sits between clauses. Same column arithmetic, same labels, same money formatting. `RENDERER_VERSION` → `3`. `CONTRACT_SIGNATURE_LAYOUT` is untouched — the signature page is always its own final page.
- **The signing page** gains the line items and totals it never had, so "what you see is what you sign" reaches the price. From the snapshot only: showing a signer live figures that differ from the document they are signing would be worse than showing none. A format-1 contract shows nothing, and there is a test for that.

## Audit gaps

1. **Every generated PDF is logged.** `persistContractPdf` is the single choke point for every contract PDF this app writes, so it is where `contract_document_generated` now records the contract, the `generated_documents` id, the kind, the PDF checksum, the content hash and the template and renderer versions. Best-effort — a failed log line must never cost a contract its PDF. The send passes its content hash in explicitly, because `completeSend` only writes it to the row *after* the PDF exists.
2. **`contract_text_sections` and `contract_attachment_inclusions` are audited tables now**, and their five writers go through `auditedInsert` / `auditedDelete`. Both cascade from `contracts`, so `accountingHistoryReferences` declares them — `accountingHistoryReferences.test.js` catches that omission on its own, which is a good guard.
3. **A template draft save records which parts it rewrote** — `changed: ['name', 'items', 'items:3->4', 'attachments']`. Keys and counts, never clause text.

## Two fixes found on the way

- `listForDocument` never returned `manifest`, so nothing could read which attachments a PDF was made of, in what order, or their checksums — including the separately delivered ones, which are bound into nothing else. The admin documents card now lists them.
- A manifest recorded **no signature page** whenever every attachment was delivered separately: `insertBeforeLastPage` returns early with `lastPageIndex: null` when it has nothing to merge. The placed slots know the page, so they are the fallback.

## Tests

New `contractFrozenSnapshot.test.js`, 7 tests — the frozen shape and its coercion, editing the source quote after a send (hash, stored snapshot, on-disk PDF and render context all unchanged), the format-1 fallback, a contract with no quote, the signing page with and without frozen terms, the manifest plus its log entry, and both newly audited tables landing in the contract's change history. All seven fail against the code before the change.

One thing the test deliberately does **not** assert: byte equality of two `renderContractPdfBuffer` calls. Every PDFKit render writes a fresh random `/ID`, so two renders of one unchanged contract differ by construction; `pdfService.theme.test.js` pins byte stability at the renderer, where `generatedAt` is fixed. The test pins the render *context* instead, which is where the snapshot-versus-live decision is actually made.

Green on SQLite **and** PostgreSQL 15. Full backend run: 418 suites, 4323 tests. Frontend: 128 files, 831 tests. `build:check` adds no new TS errors (5 pre-existing, untouched files).

## Still to come

Screenshots (light, dark, 390px) for the two UI surfaces — the price table on the signing page and the attachment manifest in the admin documents card. Draft until they are attached.

Relates to issue 1445.


the-luap:
## Screenshots

### The price on the signing page

The figures a signer now sees before signing — from the frozen snapshot, so they are the ones the content hash covers.

Note what is **not** there: the quote's optional "Fotobuch 30x30", which nobody selected. `countedLineItems` leaves it out of the snapshot, so it is not part of what is signed. The arithmetic below it is the quote's own: 3,200 + 900 + (120 × 4.50 − 10%) = 4,586 net, 8.1% VAT = 371.47, total 4,957.47.

| Light | Dark |
|---|---|
|  |  |

390px:



### The attachment manifest

Under "Generated documents" — what the PDF was made of, which was recorded at send and until now unreadable through any API. The attachment here is delivered as a **separate file**, so its checksum in the manifest is the only record of what went out.

| Light | Dark |
|---|---|
|  |  |

Captured against the dev rig: a contract created from a real accepted quote and sent through `sendContract`, so the snapshot, the manifest and the PDF are the ones this code produces.

the-luap (commented):
## 🔴 Blockers

### 1. Send-time PDF never draws frozen totals
`backend/src/services/contract/sending.js` renders the unsigned PDF via `buildRenderContext` **before** `buildContentSnapshot` runs. On a draft, `buildRenderContext` (`renderContext.js`) sets line items from the live quote but leaves `quoteTotals` null — totals are only filled from a format-2 snapshot.

`pdfService.js` (~2241) only calls `drawContractTotals` when `ctx.quoteTotals` is set. Effect: the stored unsigned/signed PDF can omit net/VAT/gross, while the signing HTML and post-send rematerialize show totals from the snapshot. That undercuts the claim that the document being signed names the sum.

**Fix:** build (or pass) `buildQuoteSnapshot` into the send render context **before** `renderContractWithSlots`, then freeze that same object in `completeSend`.

## 🟡 Concerns

1. `rendered_content_sha256` still does not cover attachment hashes — those live on the document manifest only. Incomplete for “everything bound into the signature” until a later slice.
2. Money via `ensureInt` → JS number — fine for typical CHF minors; not BigInt-safe above `Number.MAX_SAFE_INTEGER` (#596 class).
3. Admin rematerialize after freeze can disagree with the stored unsigned PDF until blocker 1 is fixed (does not overwrite `pdf_path`).

## 🟢 Nits

- `ContractPrice` hides the whole block when `lineItems.length === 0` even if totals exist.
- Screenshots still pending if this touches the customer/admin contract UI enough to need one.

## ✅ Verified

- Post-send mutation of freeze refused for non-draft; format-2 re-render uses snapshot only; integration test covers quote rewrite after send.
- Public commercial view from snapshot only — no live quote, no admin cost/margin fields.
- Contract edit race guarded by `lock_version` in `completeSend`.

## Verdict

Close blocker 1 before merge. Concern 1 is expected follow-up for full signature binding.

the-luap:
Blocker closed in `746c6026`.

## Blocker — closed

**Send-time PDF without totals.** Confirmed as described: `sendContract` rendered from the bare draft, `buildRenderContext` reads the live line items and no totals on a draft, and `pdfService.js:2245` draws totals only when `ctx.quoteTotals` is set.

`sending.js:55-86` now builds the content snapshot once, renders from `{ ...contract, rendered_content: renderedContent }`, and `completeSend` freezes that same serialised string. The PDF is drawn from the object that is frozen, so the two cannot disagree. Nothing is written to the draft before `completeSend`, so a send that fails at the attachment check still leaves no snapshot behind.

New case in `contractFrozenSnapshot.test.js` spies on `renderContractWithSlots` during a send and asserts it received the frozen totals, line items and quote number. It failed with `quoteTotals: null` before the fix.

## Concerns

1. Attachment hashes outside `rendered_content_sha256` — agreed, follow-up for full signature binding.
2. Money as JS numbers: values above `MAX_SAFE_INTEGER` are silently imprecise, not rejected (`'9007199254740993'` → `9007199254740992`). The quote routes validate `isInt()` with no upper bound and the quote maths (`quoteService.js:148-208`) is already plain numbers, so the precision is lost at quote creation; the snapshot only carries it. Belongs in the quote layer, not this PR.
3. Admin re-render vs stored PDF — ✅ closed by the blocker fix; both read the same snapshot.

## Nits

- `ContractPrice` — ✅ renders whenever `commercial` is set; only the line table depends on there being lines. New case in `ContractResponsePage.signingV2.test.tsx`. The PDF still omits totals when no line was counted (`pdfService.js:2203-2205`), so that edge case differs between page and PDF.
- Screenshots of the signing price and the manifest are in the comment above.

Contract suites: 108 integration + 94 service/route/pdf tests pass; ContractResponsePage 10/10.

PR #1549, opened 2026-09-19
fix(documents): close the gaps the issue 1444 audit found in what shipped

the-luap:
Slice 1 of the implementation plan on issue 1444 — the part that is correctness on the document slice that shipped with #1464, not new features. Branches off `main`, independent of the other stacks.

## 1a — The sandbox CSP did not survive nginx

`sendPdfAttachment` sets `Content-Security-Policy: default-src 'none'; sandbox` on every document download: the bytes came from outside the building and the browser must not interpret them. `frontend/nginx.conf` has a **server-level** `proxy_hide_header Content-Security-Policy` so nginx can be the single source of the site CSP — which stripped that header too, and sent the site policy in its place. `attachment` and `nosniff` still held; the isolation did not.

A regex location for the two download paths now lets the backend's value through. Inside a location, `proxy_hide_header` and `add_header` **replace** the server-level lists rather than adding to them, which is why every other security header is re-listed there — leaving one out would silently drop it for these responses.

Verified against real nginx with a stub upstream, not by reading the config:

| Path | CSP headers | Value |
|---|---|---|
| `/api/customer/documents/7/download` | 1 | `default-src 'none'; sandbox` |
| `/api/admin/customers/3/documents/7/download` | 1 | `default-src 'none'; sandbox` |
| `/api/customer/documents` | 1 | the site CSP, unchanged |

`X-Frame-Options`, `X-Content-Type-Options`, `Referrer-Policy` and `Permissions-Policy` all still present on the download path.

## 1b — A registered scanner did not gate admin uploads

```js
const status = verdict === 'clean' || uploaderType === 'admin' ? 'clean' : 'pending';
```

An admin upload was stored `clean` whatever the scanner said. With a scanner that was down, timed out or answered `pending`, it went through unscanned and shareable — the one case where a scanner would have been doing something.

The admin now vouches for an upload only when **no** scanner is registered (which is the shipped state, and why that rule exists at all); with one registered, its verdict decides for both sides. Sharing a document that is not `clean` was already refused with `DOCUMENT_NOT_CLEAN` — the plan asked for that guard and it turned out to be there.

## 1c — Deleting was not contract-aware

`markErasedForCustomer` keeps a contract-linked document as evidence, unshared and renamed. `softDelete` had no such rule, and the retention sweep then purged the bytes of any deleted row — the silent destruction of a contractual record the issue rules out.

Deleting a contract-linked document is now refused with `409 DOCUMENT_CONTRACT_LINKED` until it is unlinked (the existing `PATCH` with `contractId: null`), and the sweep skips contract-linked rows in **both** of its transitions, whatever state they are in. That makes the deliberate path the only path.

## 1d — PDF checking was shallow here and deep elsewhere

Customer uploads got a 20-byte magic check plus a string search for `/Encrypt` in the last MB. Contract attachments — which come from an *admin* — were parsed in a worker with heap, time, RSS and decompression budgets and refused for JavaScript, launch actions, embedded files and XFA. The path that takes files from outside the building was the less careful one.

The customer path now uses the same inspector, with two deliberate differences:

- **Forms stay allowed.** Customers upload *signed* contracts, and a signed PDF is an AcroForm with a signature field. `pdfInspect` refuses actions that run, submit or import (`/SubmitForm`, `/ResetForm`, `/ImportData`) but not the presence of a form, so this needed no change — there is a test pinning it, because it is exactly what a tightening of `pdfInspect` would break.
- **The original bytes are stored**, not the `normalised` re-serialisation an attachment keeps. Rewriting the file would break the byte ranges a digital signature covers, and these documents are evidence. There is a test asserting the stored file is byte-identical to the upload.

New refusals map to `PDF_ACTIVE_CONTENT`, `PDF_TOO_COMPLEX` and `PDF_TOO_MANY_PAGES`, each with portal copy in en + de.

## Test fixtures

The suite's PDFs were hand-written stubs — `%PDF-1.4\n1 0 obj << /Type /Catalog >> endobj\ntrailer...` — that start with the right bytes and parse as nothing. Enough for the check they were written against; not for one that actually loads the document. They are real pdf-lib documents now, the same `makePdf` shape `pdfValidation.test.js` and `contractAttachments.test.js` use.

The file's boot hook also gets a longer budget. 120s is not enough to run the core migration set against a containerised PostgreSQL — the boot alone takes ~110s there — and a timeout in `beforeAll` fails every test in the file at once, which is what it did. With 300s the suite passes on Postgres; before this change it could not be run against it at all.

## Checks

30 tests in `customerDocuments.test.js`, green on SQLite **and** PostgreSQL 15. Three of the four new tests fail against the code before the change; the two that don't ("a form still uploads", "with no scanner the admin still vouches") are guards on invariants this change must not break, which is the point of them. Full backend run: 417 suites, 4321 tests. Frontend: 128 files, 831 tests.

## Still to come

No visual surface changed beyond three new upload error messages, so there is no screenshot to attach — but this is the base of the 1444 stack, so it stays a draft until slice 11 is on top of it.

Relates to issue 1444.


the-luap (commented):
## 🔴 Blockers

None.

## 🟡 Concerns

### 1. Retention soft-delete UPDATE is not contract-atomic
`customerDocumentRetentionService.js` selects rejected rows with `.whereNull('contract_id')`, then soft-deletes via `whereIn(id)` **without** re-asserting `contract_id IS NULL`. A concurrent `updateLinks` that attaches a contract between select and update can soft-delete a still-linked row (API `softDelete` cannot). Bytes stay protected by the purge filter; you still get the hidden linked state the 409 was meant to forbid.

**Fix:** add `.whereNull('contract_id')` on the update (and optionally on the purge stamp).

### 2. Admin `share=true` + scanner-pending can stamp `shared_at` early
Create can set `shared_at` while status is still `pending`; `setShared` would 409. Customer list still requires `clean`, so no IDOR — state is just inconsistent until review. Gate `shared_at` on `status === 'clean'`.

## 🟢 Nits

- Download nginx regex requires exact path (no trailing `/`); miss falls through and nginx strips backend CSP again.
- Download location re-adds `X-Frame-Options: SAMEORIGIN` (not `DENY`); mitigated by `attachment` + sandbox CSP.

## ✅ Verified

- Cross-customer ownership unchanged on get/download/share/delete paths; wrong-tenant → 404.
- nginx download location omits CSP from `proxy_hide_header` so backend `default-src 'none'; sandbox` passes.
- Contract-linked delete refused via `assertNotContractLinked`; no hard-delete API.
- Upload path still multer temp → content-addressed key + `assertPathInside` / `validatePdf`.

## Verdict

No blockers. Take concern 1 if you want the contract invariant airtight under concurrency — otherwise merge-cleared from this review.

the-luap:
Both concerns and both nits closed in `9b3e4968`.

## Concerns — closed

| | Where | |
|---|---|---|
| 1. Retention soft-delete not contract-atomic | `customerDocumentRetentionService.js:58` — the update re-asserts `whereNull('contract_id')` and `whereNull('deleted_at')` | ✅ |
| 1b. Purge stamp | `customerDocumentsService.js:606-620` `purgeFiles` — a stamp-side check alone would come after the bytes were already gone, so the order is inverted: claim `purged_at` first, only while `contract_id IS NULL`, then delete the bytes; a failed delete releases the claim so the next sweep retries | ✅ |
| 2. `share=true` on a scanner-pending upload | `customerDocumentsService.js:429` — `shared_at` only when `status === 'clean'`. The share intent on a pending upload is not remembered; the admin shares after review. Remembering it would need a column | ✅ |

## Nits — closed

- `nginx.conf:139` — download location matches with an optional trailing slash ✅
- `nginx.conf:159` — `X-Frame-Options "DENY"` on that location; the frontend fetches these as blobs, nothing embeds them ✅

`nginx -t` clean on `nginx:1.30-alpine`. Three new cases in `customerDocuments.test.js` — a contract link landing between the sweep's select and update, a stale snapshot handed to `purgeFiles`, a share on a scanner-pending admin upload — all fail before the change. 63/63 across `customerDocuments`, `customerPortalDocuments`, `maintenance.spaShell`.

PR #1550, opened 2026-09-19
test(documents): pin the cross-customer boundaries on every document route

the-luap:
Slice 11 of the implementation plan on issue 1444. **Stacked on #1549** — base is `fix/1444-document-hardening`, so the diff here is only this slice.

The suite pinned list and download isolation and *one* admin mutation (`review`) addressed through the wrong customer's id. Every other admin route reaches the same document by id, so each one needed checking on its own — that is what an object-level authorization matrix is for.

## What is now pinned

| Boundary | Test |
|---|---|
| Every admin mutation through the wrong customer | `share`, `unshare`, `download`, `DELETE` and `PATCH` → 404, **and the row is byte-for-byte unchanged afterwards** — a 404 that still wrote is not a refusal |
| An upload linked to another customer's contract | 400, no row, no temp file left behind |
| A customer naming a project | `projectId` is admin-only on the way in, so it is ignored rather than trusted — the document is created with `project_id` null |
| A deactivated customer | shut out of list, upload and download |
| CSRF on the multipart upload | cross-site `Sec-Fetch-Site`, and a foreign `Origin`, both refused before the route |

## The CSRF gap in the harness

The portal upload is a cookie-authenticated `multipart/form-data` POST — the exact shape a cross-site page can send from a plain HTML form with no preflight, and therefore the exact thing the gate exists for. The document suite builds its apps with `buildRouteApp`, which mounts no CSRF, so the gate had never been exercised on this route at all. It gets its own file, following the `csrfChunkedUpload.test.js` pattern: parsers, then the real gate, then a stub handler, so what is being asserted is whether the request reaches a handler.

## A dark-mode bug the state tests found

```js
const STATUS_STYLE = {
  clean: 'bg-green-100 text-green-800',
  pending: 'bg-amber-100 text-amber-800',
  rejected: 'bg-red-100 text-red-800',
};
```

Light values only. The status chip is the one thing the documents page exists to communicate — whether the file you sent has been accepted — and in dark mode all three were unreadable. They now carry the same dark pairs the signing page's signer chips use, and there is a test asserting every chip has a `dark:` class, which fails without the fix.

The new upload refusals from slice 1 (`PDF_ACTIVE_CONTENT`, `PDF_TOO_COMPLEX`, `PDF_TOO_MANY_PAGES`) are pinned to name the file and say what to do, rather than falling back to "could not be uploaded".

## Not done from this slice

The end-to-end customer-portal flow test (login → upload → admin accepts → customer downloads). It needs the isolated e2e stack rather than the jest harness, and is worth its own pass rather than a rushed one at the end of this branch.

## Checks

33 backend tests in `customerDocuments.test.js` plus 3 CSRF tests, green on SQLite **and** PostgreSQL 15. Full backend run: 418 suites, 4327 tests. Frontend: 128 files, 833 tests.

The dark-mode fix is a colour change on an existing chip, so a screenshot is coming with the rest of the 1444 stack before review.

Relates to issue 1444.


the-luap:
## Screenshots — and a correction the screenshots forced

The first attempt at the dark-mode chip fix added Tailwind `dark:` variants. The screenshot showed them **still rendering their light values** on the portal's dark ground, which is how I learned that `dark:` is inert on the customer surface: it needs the `dark` class the admin shell toggles on `<html>`, while the portal themes through CSS tokens — something `CustomerLayout` documents in a comment I had not read carefully enough.

The chips now derive both colours from the surface they sit on (`color-mix`), so one rule reads on either ground. Pushed as a follow-up commit on this branch.

| Portal documents — light | Portal documents — dark |
|---|---|
|  |  |

390px (dark):



All three states are shown: **Rejected** with its reason, **Available** with a download, and **Awaiting review** with the explanation. The light capture was taken with the rig's branding palette temporarily switched to light and restored afterwards, since the rig's own palette is dark.

## A wider finding, not fixed here

**40 other `dark:` occurrences across five customer-portal files are inert for the same reason** — including the status chips on the contracts page (`Fully signed`, `Sent`), visible in the 1446 portal screenshot. Fixing them all is a bigger, purely visual change than belongs in a test-hardening branch, and it deserves its own before/after pass. Happy to open an issue for it.

## A pre-existing smoke-test failure, also not mine

The pre-push hook's E2E smoke suite fails two Branding tests on this rig:

```
strict mode violation: getByRole('button', { name: /Save Changes|…/i }) resolved to 2 elements:
  1) 'Save Changes'   2) 'Save changes'  (disabled)
```

That suite runs against the **built** frontend image on :7100, which predates every frontend change in this stack (image created 2026-09-15), so it cannot be exercising this code. Flagging it rather than silently skipping past it.

the-luap:
## Both open items from this PR are now closed

**The E2E flow is written** — pushed to this branch as `97b83a7e`. Login → upload → the review gate holds → admin marks it clean → the customer downloads it as an attachment. It runs on the isolated e2e stack (`scripts/e2e.sh`), alongside the existing gallery-handoff test in the same spec.

Two things the first run taught me, both now comments in the spec:
- `toHaveURL(/\/customer\//)` also matches `/customer/login`, so a failed sign-in sailed straight past it and failed later somewhere vaguer. It waits for a path that is *not* the login page now.
- The portal may be in **German**: a fresh invite leaves `preferred_language` unset. The assertions take either language, and the filename locator is `exact` because the success toast names the file too.

The upload is a real, parseable PDF built in the spec rather than a `%PDF-` stub — the content check this branch's parent added loads the document, and correctly refuses a stub.

And I checked it fails for the right reason rather than assuming: with `vouched` forced true in the running backend (the review gate removed), the spec fails on the download button being present before review, not on something incidental.

**The wider dark-mode gap is fixed** in **#1552**, stacked on this branch: all 34 inert `dark:` occurrences plus the contracts-page chips that never had them, through the same token mechanism, verified in a browser on both a dark and a light branding palette.

**The smoke-suite failure** I flagged as pre-existing turned out to be a real defect and is fixed in **#1551**: `/admin/branding` had two save buttons with the same accessible name — byte-identical in German. The whole local smoke suite passes again, 13 of 13.

the-luap (commented):
## 🔴 Blockers

None.

## 🟡 Concerns

1. Deactivated-customer **download** allows `[401,403,404]` against another customer’s pending id — 404 is also the healthy cross-customer answer, so that one assertion is soft. List/upload checks are solid.
2. Stacked on #1549 — inherits the retention soft-delete race unless that lands with the `whereNull('contract_id')` fix.

## ✅ Verified

- Admin wrong-customer matrix expects 404 **and** asserts DB fields unchanged (not status-only).
- Foreign `contractId` upload → 400 + no row + empty temp; ignored `projectId` → null.
- CSRF test exercises the real middleware with a stub route — useful, not cosmetic.

## Verdict

Merge-cleared (onto #1549). Real isolation pins.

the-luap:
Rebuilt on the updated #1549 (merge `ad2e7298`), and the soft assertion is fixed in `1f7ce24d`.

## Concerns — closed

| | Where | |
|---|---|---|
| 1. Deactivated-customer download accepted `[401,403,404]` | `customerDocuments.test.js:315` — now uses customer B's **own** document, marked clean, and first proves B downloads it while active (200). After deactivation it expects exactly `401` / `CUSTOMER_NOT_FOUND`. The list and upload checks in the same test are pinned to the same answer, since one auth gate refuses all three | ✅ |
| 2. Inherited retention race | closed at the source in #1549 `9b3e4968`, merged here | ✅ |

45/45 backend, `customerDocuments.test.tsx` 7/7.

PR #1551, opened 2026-09-19
fix(branding): name the second save button on the Branding page

the-luap:
Found while running the pre-push E2E smoke suite during the issue-1444 work.

## The defect

`/admin/branding` renders two save buttons:

| Where | Key | English | German |
|---|---|---|---|
| Page header | `branding.saveChanges` | `Save Changes` | `Änderungen speichern` |
| Customer dashboard header card | `settings.customerSurface.save` | `Save changes` | `Änderungen speichern` |

In German they were **byte-identical**; in English they differed only in the case of one letter. A screen reader reads the page as having two buttons with the same name and no way to tell which one saves what — and the buttons do genuinely different things (the card writes `/admin/settings/customer-surface`, the header writes the branding settings).

The smoke suite hit the same wall from the other side. Specs 06 and 07 have both been failing on the local rig since that card was added:

```
strict mode violation: getByRole('button', { name: /Save Changes|Änderungen speichern/i })
resolved to 2 elements:
  1) <button …>Save Changes</button>
  2) <button disabled …>Save changes</button>
```

## The fix

The card's button now names what it saves — **"Save header settings"** / **"Kopfzeilen-Einstellungen speichern"** — rather than teaching the test selector to guess which of two identically-named buttons it wants. One line of component change plus the two locale strings.

## Verification

Rebuilt the frontend image from this branch and ran the local smoke suite against it:

```
13 passed (21.8s)
  ✓ 06-theme-save-without-live-preview.spec.ts › …persists after Save + reload…
  ✓ 07-branding-default-on-create-event.spec.ts › Branding-default theme is applied…
```

Both previously failing specs pass, and nothing else moved. Frontend unit suite: 128 files, 831 tests.

No screenshot: the only visual change is the text on a button that is disabled unless the card is dirty, and its new label is quoted above.


the-luap (commented):
## 🔴 Blockers

None.

## 🟡 Concerns

Only `en`/`de` strings updated; other locales keep the old `settings.customerSurface.save` value, so the duplicate accessible-name issue can linger there until translated. Key itself is unchanged — no collision.

## ✅ Verified

Surgical label fix on the second Branding save button. No logic/auth changes.

UI change — a before/after screenshot on the Branding page would help.

## Verdict

Merge-cleared.

the-luap:
Both review items closed.

**Other locales** — `2a8748d8`: `settings.customerSurface.save` changed in es, fr, nl, pt, ru and sl as well. Checked case-insensitively against `branding.saveChanges` in all eight locales, no collision left. `es` had no `customerSurface` block and was showing the English fallback; it gets the one key.

| | First button | Second button |
|---|---|---|
| es | Guardar cambios | Guardar ajustes del encabezado |
| fr | Enregistrer les modifications | Enregistrer les paramètres de l'en-tête |
| nl | Wijzigingen opslaan | Kopregelinstellingen opslaan |
| pt | Salvar Alterações | Salvar configurações do cabeçalho |
| ru | Сохранить изменения | Сохранить настройки шапки |
| sl | Shrani spremembe | Shrani nastavitve glave |

**Screenshots** — the two buttons sit far apart, so each image stacks the page-header row and the header card from the same page.

| Before (main) | After |
|---|---|
|  |  |
|  |  |

German was the byte-identical case ("Änderungen speichern" twice). The card button renders disabled because nothing is unsaved.

PR #1552, opened 2026-09-19
fix(portal): colour every customer-surface status through the theme tokens

the-luap:
Stacked on **#1550** (base `test/1444-document-isolation-matrix`), because that PR fixed the documents page's chips and this finishes the job everywhere else.

## The defect, and why it hid so well

Tailwind's `dark:` modifier **does not fire on the customer surface**. `darkMode: 'class'` matches a `dark` class on an ancestor, which only the admin shell (`AdminDarkModeContext`) and the public gallery (`usePublicDarkMode`) set. The portal renders the photographer's branding palette through CSS tokens instead — `CustomerLayout.tsx:120-124` says so in a comment.

So all 34 `dark:` classes under `pages/customer` were dead, and the **light** half of each pair is what showed on a dark portal. The contracts page never even tried: its chips were fixed light colours, so a customer on a dark portal read the status of their own contract as dark text on a pale chip.

This is invisible to unit tests — a class-string assertion passes on a chip nobody can read. It is exactly how the first attempt in #1550 looked fixed when it was not.

## Why not just add the `dark` class

Putting `dark` on the portal's root would make all 34 variants fire, and break what they sit inside: `.dark` also re-declares `--color-surface`, `--color-text` and the rest with PicPeak's generic palette (index.css:65). Every operator running custom branding would lose their colours below that element. Rejected.

## The fix

A `hue-*` class names the meaning; `.status-chip` washes that hue over whatever surface it lands on, and `.text-status` pulls it toward that surface's own text colour:

```css
.status-chip {
  background-color: color-mix(in srgb, var(--chip-hue) 18%, var(--color-surface, #ffffff));
  color:            color-mix(in srgb, var(--chip-hue) 72%, var(--color-text, #171717));
}
```

One rule, readable on a white portal and on a near-black one, following the operator's palette rather than replacing it. Six hues: success, warning, danger, info, storno, neutral — plus `.row-muted` and `.hover-surface` for the two backgrounds that had the same problem.

Converted across nine files: status chips on bills, quotes, contracts and documents; the storno amount and lineage lines; outstanding and late-fee amounts; error text and required-field markers; row hover and the cancelled-row background.

## Verified in a browser, on both palettes

Not asserted in jsdom — that is what made the first attempt look right.

| Invoices — dark palette | Invoices — light palette |
|---|---|
|  |  |

All six hues in one view: storno purple (chip, amount and lineage line), neutral grey on the cancelled row with its dimmed background, danger red with the amber late fee, success green, and the accent links.

| Contracts — dark | Documents — dark |
|---|---|
|  |  |

| Quotes — dark | Quotes — light |
|---|---|
|  |  |

The light captures were taken with the rig's branding palette temporarily switched and restored afterwards; the rig's own palette is dark.

## The guard

`portalStatusColours.test.tsx` is a source check, not a render check, because that is the shape of this mistake — it cannot be seen in jsdom. It fails on any `dark:` variant or fixed `bg-/text-{green,red,amber,blue,purple}-NNN` under `pages/customer`, so the next one shows up in CI instead of in a screenshot weeks later.

Frontend suite: 129 files, 835 tests.

Relates to issue 1444.


the-luap (commented):
## 🔴 Blockers

None.

## 🟡 Concerns

None. No XSS via status: chip classes come from fixed maps (`hue-*`), labels via i18n; names/rejection reasons are React text nodes.

## ✅ Verified

Theme coverage for portal pages under `pages/customer`; source lint bans leftover `dark:` and fixed `bg|text-(green|red|…)-N`. Residual fixed greys / `text-accent` look intentional.

UI change — screenshot on the PR would help review (dark + light status chips).

## Verdict

Merge-cleared (onto #1550).

the-luap:
Merged the updated #1550 (`f8daee45`), no conflicts; 45/45 backend, `src/pages/customer` 16/16.

On the screenshot request: they are in the PR body — invoices (dark and light palette, all six hues), contracts, documents and quotes, captured against the rig with the branding palette switched for the light shots.

PR #1554, opened 2026-09-20
fix(security): close the IPv6 rate-limit bypass and stop answering scanner probes with the SPA shell

the-luap:
## Summary

Two small hardening fixes that came out of analysing how self-hosted instances hold up against botnet scanning. No UI change.

### 1. IPv6 clients could rotate around both rate limiters

`rateLimitService.js` keyed the general `/api` limiter and the auth limiter on the raw `req.ip`. A single IPv6 host is routinely handed a whole /64 and can use every address in it, so it could present a fresh "client" per request and never spend a budget. The 5-failed-logins limit and the 300-per-window `/api` budget were effectively unlimited for anyone on IPv6. The per-account lockout in `authSecurity.js` does not catch it either, because it counts per `(identifier, ip)` pair.

`backend/src/utils/rateLimitKey.js`:
- IPv6 → keyed on the /64 (`2001:db8:1:2::/64`)
- IPv4 → unchanged
- IPv4-mapped (`::ffff:203.0.113.7`, dotted or hex) → the plain IPv4 address, so one client cannot hold two buckets
- unparseable → raw value, as before

/64 and not wider on purpose: mobile carriers give each handset its own /64, and a /56 mask would make unrelated gallery guests share one budget. An attacker with a delegated /56 still gets 256 buckets. Bounded, not zero.

### 2. Scanner probes were answered with `200` and the SPA shell

`/wp-login.php`, `/cgi-bin/luci`, `/phpmyadmin/` and similar fell through to the history fallback: `try_files … /index.html` in `frontend/nginx.conf`, and the Express `app.get('*')` fallback the single-container image uses. PicPeak has no route shaped like these. The 200 tells a scanner something is there, and it blinds log-based tools an operator runs in front. CrowdSec's `http-probing` and fail2ban's 404 jails count 404s (relevant to the new docs page on bots and scanners).

Both layers now return 404 for:
- paths ending in `.php`, `.php5`/`.php7`, `.phtml`, `.asp(x)`, `.jsp(x)`, `.cgi`
- `/wp-admin`, `/wp-content`, `/wp-includes`, `/phpmyadmin`, `/cgi-bin` (as a full path segment)

Deliberately narrow: no bare words, so a CMS page called `/vendor` or `/php` still renders, and nothing a download could be named. `^~` locations (`/uploads`, `/s/`, the analytics tracker proxy) still outrank the new regex blocks; the blocks sit ahead of the static-asset regex because nginx takes the first matching regex.

## Verification

- **Red first.** `rateLimitIpv6Key.test.js` drives the real limiters through `X-Forwarded-For`. On the old key, 2 of its tests failed: six failed logins from six addresses in one /64 all returned 401, and the general budget never tripped. Green after the change.
- `__tests__/middleware` + `adminSettingsRateLimit`: 19 suites, 166 tests pass.
- **nginx, real container** (`nginx:1.30-alpine` with `frontend/nginx.conf`): `nginx -t` OK. `/wp-login.php`, `/index.PHP`, `/cgi-bin/luci`, `/wp-content/plugins/x.js`, `/phpMyAdmin/`, `/default.aspx` → 404 with the security headers still present. `/admin`, `/gallery/<slug>`, `/vendor`, `/impressum`, `/wp-administration` → 200 shell. `/assets/app.js` → 200 JS. `/.env` → 403 as before.
- **Express fallback, real app** (`server.js` loaded with `SERVE_FRONTEND=true` on a throwaway SQLite DB): `/wp-login.php`, `/cgi-bin/luci`, `/phpMyAdmin/` → 404; `/setup`, `/impressum`, `/vendor` → 200 shell.
- `tests/nginx/test_request_logging.py` passes.
- ESLint clean on all new and changed files except `server.js`, whose 25 existing findings are unchanged (same count on `origin/main`).
- Full backend suite: 418 suites pass, 1 fails. `__tests__/utils/streamResponse.test.js` failed once under full-suite load and passes 7/7 in isolation; it does not touch anything in this diff.

## Not in this PR

- The ~14 route-local limiters (`publicQuotes`, `publicTransfer`, `publicContracts`, …) still use express-rate-limit's default key, which is the raw `req.ip`. They guard high-entropy token links, so the exposure is smaller, but they should adopt `rateLimitKey` in a follow-up.
- `nginx/nginx.conf*` and `frontend/nginx.dev.conf` are untouched: the first two are operator-side proxy examples that forward to the bundled nginx, which now answers 404.
- No stable twin yet.


the-luap (commented):
## 🔴 Blockers

None.

## 🟡 Concerns

1. **Incomplete limiter coverage** — auth + global API budgets use `rateLimitKey`; many public limiters still key on raw `req.ip` (`publicQuotes`, `publicContracts`, transfers, payment-check, newsletter, `feedbackRateLimit.js`). Same IPv6 per-address bypass remains there. Follow-up: reuse `rateLimitKey` on those instances.

2. **Trailing-slash false negative** — `scannerPaths.js` / nginx regex is `$`-anchored, so `/wp-login.php/` still gets the SPA. Allow optional trailing slash.

3. **Root-only directory list** — nested probes (`/wordpress/wp-admin/`) still get the shell. Deliberately narrow; document as known residual.

4. **Bracketed XFF** (`[2001:db8::1]`) fails `net.isIPv6` and skips /64 collapse — only material if `TRUST_PROXY=true`. Strip `[]` before parse.

## 🟢 Nits

- Deprecated IPv4-compatible `::203.0.113.7` keys as an IPv6 /64, not the dotted IPv4; mapped `::ffff:` is handled correctly.
- Rebase onto current `main` before merge (release commits landed).

## ✅ Verified

- /64 collapse; leading zeros + case fold; zone id stripped; IPv4-mapped → same IPv4 key; uses `req.ip` only (aligned with trust-proxy defaults).
- Both `createRateLimiter` and `createAuthRateLimiter` wired; integration tests drive 429 across addresses in one /64.
- Scanner probes skip SPA via `next()` → Express 404; gallery slugs cannot look like `.php`.
- nginx scanner regexes sit before static-asset regex; CSP unchanged; normal routes unaffected.

## Verdict

Merge-cleared on the stated fix. Worth a follow-up for other IP-keyed limiters and the trailing-slash FN.

the-luap:
Rebased onto current `main` (no conflicts) and all four concerns plus the nit closed. Three new commits on top: `a151b0ba`, `c30d58ad`, `0a0b4d91`.

## Concerns — closed

| | Where | |
|---|---|---|
| 1. Limiters still on raw `req.ip` | `c30d58ad` — `keyGenerator: rateLimitKey` on publicQuotes, publicContracts, publicContractSigning, publicDocumentVerification, publicPaymentCheck, publicTransfer, publicTransferUpload, publicNewsletter and analyticsTrackerProxy (express-rate-limit 6.11.2 falls back to raw `req.ip` without one); adminNewsletters falls back to `rateLimitKey(req)` without an admin id; the in-memory counters too — `feedbackRateLimit.js:228` (strictRateLimit), `galleryGuests.js:61/226/274`, `publicTokenGuards.js:80` (bad-token lockout) | ✅ |
| 2. Trailing-slash false negative | `0a0b4d91` — `scannerPaths.js:19` and `nginx.conf:92` accept an optional `/` | ✅ |
| 3. Nested probes | documented next to the list as a deliberate residual (`/wordpress/wp-admin/` stays unmatched) | ✅ |
| 4. Bracketed XFF | `a151b0ba` — `rateLimitKey.js:53-67` strips `[…]` and `[…]:port` before parsing | ✅ |

## Nit — closed

- IPv4-compatible `::203.0.113.7` (and the long form) folds to the dotted IPv4. Only the dotted spelling, so `::1` stays loopback ✅

## Left out on purpose — tracked in #1564

`generateGuestIdentifier` (feedback) and `createClientFingerprint` (secure images) still hash the raw IP, because both hashes double as an identity — guest dedup for likes/ratings, and the secure-image token binding. Collapsing them to the /64 would merge a household into one guest and loosen the token binding. #1564 splits bucket key from identity for both.

## Tests

New `__tests__/middleware/rateLimitKeyCoverage.test.js` asserts every express-rate-limit instance sets a keyGenerator, and drives strictRateLimit and the token lockout across one /64; `rateLimitIpv6Key.test.js` and `scannerPaths.test.js` extended. Every new test fails with the fix stashed (6 / 12 / 4). Full backend run: 421 suites, 4400 tests, 0 failures. `nginx -t` clean on `nginx:1.30-alpine`.

PR #1557, opened 2026-09-20
fix(video): let the admin regenerate button repair video thumbnails

the-luap:
## Summary

Follow-up to #1417. That PR taught `ensureThumbnail` to rebuild a video's poster frame, but the admin **Regenerate thumbnails** button never reached it: the route filtered videos out before the call (`adminThumbnails.js`, "Skip videos, matching /regenerate-previews"). Opening a gallery repaired broken video tiles lazily; pressing the button did nothing for them.

This removes the filter on `/regenerate` and gives videos their own path through the loop. `/regenerate-previews` keeps its filter, the preview tier is image-only.

## Why videos are not just "un-filtered"

In the review of #1417 I suggested letting the admin route opt out of the size bound and otherwise treating videos like stills. Reading `videoProcessor.generateVideoThumbnail` changed that: the poster is rendered at a **fixed 300x300** and does not depend on the thumbnail settings this endpoint exists to apply. The loop force-rebuilds every row (`thumbnail_path: null`, `force: true`), so for videos that would mean:

- every press re-reads every video, a full object download each on S3, to write back identical bytes
- on S3, any video over `VIDEO_THUMBNAIL_MAX_SOURCE_BYTES` that got a real poster at upload time (the upload path is unbounded, the file is local) would have it **replaced by the placeholder**

So, for a video row:

| | Stills (unchanged) | Videos (new) |
|---|---|---|
| `thumbnail_path` | nulled | passed through |
| `force` | `true` | off, so a valid poster is skipped by `ensureThumbnail`'s own check |
| Size bound | n/a | lifted: `{ boundVideoSource: false }` |
| Tier deletion | yes | no, videos never take the tier path |

The bound is lifted because it guards the guest request path. This is an admin-initiated background job that walks one video at a time, and bounded it could never give a large S3 video with a missing thumbnail anything but the placeholder. The option threads `ensureThumbnail` → `regenerateThumbnail` → `regenerateVideoThumbnail` and defaults to bounded, so every existing caller behaves as before.

A legacy row with no `media_type` is recognised by `mime_type`, same inline predicate the gallery routes use. These rows were never caught by the old filter and went to Sharp, one error each.

## Testing

- `adminThumbnails.regenerate.test.js`: the old "leaves videos alone" test is replaced by five. A video is passed through with its `thumbnail_path`, no `force`, `boundVideoSource: false`, no tier deletion; a `mime_type`-only row takes the same path; stills still get `force` and a nulled path; a video that returns `null` and one that throws are both counted and the batch continues (`1 success, 2 errors`); `/regenerate-previews` still excludes videos. Three of the five fail on `main`.
- `imageProcessor.videoThumbnail.test.js`: a 4 GB S3 video with `boundVideoSource: false` is downloaded and gets the real poster, `stat` is never called.
- Full backend suite: 418 suites, 4330 tests pass. ESLint clean on all four files.
- Not exercised: a real S3 backend, and a real ffmpeg run through the route (the route test mocks `imageProcessor`, as it did before).

## Known limits

- A row that already holds a **placeholder** (from the bounded guest path, or from an unreadable file) counts as a valid thumbnail and is skipped. The button repairs missing and broken thumbnails; it does not upgrade placeholders. Telling the two apart needs a marker on the row, which is out of scope here.
- A non-forced call can join a guest-triggered regeneration already in flight for the same video and adopt its bounded result. Same outcome as the guest path, and only in that race.

Backend only, no UI change, so no screenshot. Needs a stable twin once merged; `stable` has the same filter.


the-luap (commented):
## 🔴 Blockers

None.

## 🟡 Concerns

`boundVideoSource: false` lets the admin regenerate job download unbounded S3 objects. Guest path still defaults bound. Mitigated by `adminAuth` + `photos.edit` / `mayStartRegeneration`, job mutex, sequential loop, and video `force: false` (skips healthy posters). Intentional admin-cost tradeoff — call out in the PR if not already.

## ✅ Verified

- No new path/FFmpeg surface: sources still via `resolvePhotoStorageKey` / `resolvePhotoFilePath` on DB rows.
- Videos are repaired, not force-rebuilt; Sharp no longer fed containers.
- Authz unchanged (`photos.edit`).

## Verdict

Merge-cleared.

the-luap:
On the `boundVideoSource: false` concern: already called out in the PR body ("The bound is lifted because it guards the guest request path…", and the size-bound row of the behaviour table), with the option defaulting to bounded so every other caller is unchanged. No code change.

PR #1559, opened 2026-09-20
fix(video): let the admin regenerate button repair video thumbnails (stable)

the-luap:
## Summary

Stable twin of #1557. Clean cherry-pick of its single commit onto `stable` after #1556 landed, no conflicts, nothing dropped and nothing added.

`stable` has the same filter on `/regenerate` ("Skip videos, matching /regenerate-previews"), so the admin **Regenerate thumbnails** button does nothing for the video rows #1556 made repairable.

Same behaviour as the main twin, for a video row: `thumbnail_path` passed through and no `force` (a healthy poster is skipped, because posters render at a fixed 300x300 and do not depend on the thumbnail settings), size bound lifted via `boundVideoSource: false`, no tier deletion. Stills and `/regenerate-previews` are unchanged. The full reasoning is in #1557.

## Merge order

**Do not merge before #1557.** A stable twin must not carry something its main twin does not. If #1557 changes in review, this branch gets the same change before it merges.

## Verification, on this branch

- `adminThumbnails.regenerate`, `imageProcessor.videoThumbnail`, `imageProcessor.singleFlight`, and the two `adminThumbnails` route suites: 5 suites, 53 tests pass.
- Revert check: putting `adminThumbnails.js` and `imageProcessor.js` back to `origin/stable` fails 7 of the tests in the first two suites.
- ESLint clean on both source files.
- Not exercised: a real S3 backend, and a real ffmpeg run through the route.


the-luap (commented):
## 🔴 Blockers

None.

## 🟡 Concerns

Same as #1557: `boundVideoSource: false` is an intentional admin-cost tradeoff on large S3 videos (guest path still bound).

## ✅ Verified

Stable backport of the main fix. Authz unchanged; path resolution still from DB rows; videos repaired not force-rebuilt.

## Verdict

Merge-cleared (stable).

the-luap:
Same answer as #1557: the lifted bound is the intentional admin-cost tradeoff, described in the #1557 body and referenced from this one. No change.

